PR #15, opened 2025-08-01
FROM ryaneggz/8-add-langchain-mcp-adapter-example TO hwchase17/deepagents@master

ryaneggz:
Closes https://github.com/hwchase17/deepagents/issues/8

PR #17, opened 2025-08-02
Mcp adapters

Cam10001110101:


hwchase17:
I don't think all this is needed - i think we can just do something like #15 ?

hwchase17:
closing in favor of other pr. thank you for looking into this though! If you think any changes are needed - please open them against main!

PR #29, opened 2025-08-07
feat: add checkpointer support to deep agent creation

prasrahul:
- Updated create_deep_agent to accept an optional checkpointer parameter.
- Modified _create_task_tool to pass checkpointer to create_react_agent.

### DIFF
--- a/src/deepagents/sub_agent.py
+++ b/src/deepagents/sub_agent.py
@@ -18,9 +18,9 @@ class SubAgent(TypedDict):
     tools: NotRequired[list[str]]
 
 
-def _create_task_tool(tools, instructions, subagents: list[SubAgent], model, state_schema, checkpointer):
+def _create_task_tool(tools, instructions, subagents: list[SubAgent], model, state_schema):
     agents = {
-        "general-purpose": create_react_agent(model, prompt=instructions, tools=tools, checkpointer=checkpointer)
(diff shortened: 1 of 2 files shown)


prasrahul:
@hwchase17 check this out! 

this is blocking us for going live with the deepagents. 
We want to contribute more going further to make deepagents production ready and easily implement with all the core features of langgraph.

src/deepagents/sub_agent.py
```diff
@@ -18,9 +18,9 @@ class SubAgent(TypedDict):
     tools: NotRequired[list[str]]
 
 
-def _create_task_tool(tools, instructions, subagents: list[SubAgent], model, state_schema):
+def _create_task_tool(tools, instructions, subagents: list[SubAgent], model, state_schema, checkpointer):
```
hwchase17: what does adding a checkpointer here do? they are being used in a stateless way right? is this just for durable execution inside an agent?

prasrahul: yes, it is not required for the sub-agents. 
We are using deepagent as one of the agent in our multi-agent setup and need the agent to persist the history for multi-turn interaction. so our usecase is not stateless. 
checkpointer is required for deep-agent but not required for sub-agents. 

updated with the latest commit @hwchase17

hwchase17:
thank you!

PR #34, opened 2025-08-12
refactor: Improve efficiency of Tavily client usage in examples

doobidoo:
This pull request addresses a performance inefficiency in the example code located in README.md and examples/research/research_agent.py.

## Changes

The TavilyClient was previously being instantiated inside the internet_search tool function. This resulted in a new client object being created on every single tool call, which is inefficient.
This commit refactors the code to initialize the TavilyClient once at the module level. The single client instance is then reused across all subsequent calls to the internet_search tool.

## Benefits

- **Performance**: Reduces the overhead of creating new client objects repeatedly in an agentic loop.
- **Best Practices**: Aligns the example code with the best practice for handling API clients, providing a better template for users of the library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ from typing import Literal
 from tavily import TavilyClient
 from deepagents import create_deep_agent
 
+tavily_client = TavilyClient(api_key=os.environ["TAVILY_API_KEY"])
 
 # Search tool to use to do research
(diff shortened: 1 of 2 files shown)


hwchase17 (approved):

hwchase17:
thank you!

PR #35, opened 2025-08-12
Update prompts.py

Akshu24Tech:


hwchase17:
left comments on other pr (which looks the same?)

PR #37, opened 2025-08-12
docs: port to unified documentation

Palashio:
This PR modifies the README to direct users to the unified documentation under langchain labs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,14 +7,12 @@ a **planning tool**, **sub agents**, access to a **file system**, and a **detail
 
 <img src="deep_agents.png" alt="deep agent" width="600"/>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,240 +1,20 @@
...
-)
-
-# Invoke the agent
-result = agent.invoke({"messages": [{"role": "user", "content": "what is langgraph?"}]})
-```
-
-See [examples/research/research_agent.py](examples/research/research_agent.py) for a more complex example.
-
-The agent created with `create_deep_agent` is just a LangGraph graph - so you can interact with it (streaming, human-in-the-loop, memory, studio)
-in the same way you would any LangGraph agent.
```
hwchase17, line 72: can we leave this as a "quckstart"? I think having that in the readme makes it way easier to get started

PR #38, opened 2025-08-13
Eat/subagent model selection 33

emanueleielo:
**Benefits**  
- Allow different models/settings per sub-agent.  
- Control cost/latency by routing tasks to cheaper/faster models.  
- Mix providers (Anthropic/OpenAI/local) in one agent.  
- Backward compatible: defaults to main model if not set.  

**Changes**  
- Added optional `model`, `model_provider`, `max_tokens`, `temperature`, `api_key` to `SubAgent`.  
- Build logic: use sub-agent config if present, else fallback to parent.  
- Updated example: critique uses `claude-3-5-haiku-20241022` (temp 0), others `claude-sonnet-4-20250514`.  
- Added tests for override, fallback, mixed providers.  
- Updated README with schema and usage example.  

**Closes:** hwchase17/deepagents#33

### DIFF
--- a/examples/research/research_agent.py
+++ b/examples/research/research_agent.py
@@ -65,12 +65,6 @@ critique_sub_agent = {
     "name": "critique-agent",
     "description": "Used to critique the final report. Give this agent some infomration about how you want it to critique the report.",
     "prompt": sub_critique_prompt,
-    # Use a lightweight model specialized for fast critique, with deterministic output
-    "model_settings": {
-        "model": "anthropic:claude-3-5-haiku-20241022",
-        "temperature": 0,
-        "max_tokens": 8192
-    }
 }
 
 
--- a/examples/research/research_agent.py
+++ b/examples/research/research_agent.py
@@ -36,7 +36,7 @@ research_sub_agent = {
     "name": "research-agent",
     "description": "Used to research more in depth questions. Only give this researcher one topic at a time. Do not pass multiple sub questions to this researcher. Instead, you should break down a large topic into the necessary components, and then call multiple research agents in parallel, one for each sub question.",
     "prompt": sub_research_prompt,
-    "tools": ["internet_search"]
+    "tools": ["internet_search"],
 }
 
 sub_critique_prompt = """You are a dedicated editor. You are being tasked to critique a report.
@@ -65,6 +65,10 @@ critique_sub_agent = {
     "name": "critique-agent",
     "description": "Used to critique the final report. Give this agent some infomration about how you want it to critique the report.",
     "prompt": sub_critique_prompt,
+    # Use a lightweight model specialized for fast critique, with deterministic output
+    "model": "claude-3-5-haiku-20241022",
+    "model_provider": "anthropic",
+    "temperature": 0,
 }
 
 
--- a/examples/research/research_agent.py
+++ b/examples/research/research_agent.py
@@ -66,9 +66,11 @@ critique_sub_agent = {
     "description": "Used to critique the final report. Give this agent some infomration about how you want it to critique the report.",
     "prompt": sub_critique_prompt,
     # Use a lightweight model specialized for fast critique, with deterministic output
-    "model": "claude-3-5-haiku-20241022",
-    "model_provider": "anthropic",
-    "temperature": 0,
+    "model_settings": {
+        "model": "anthropic:claude-3-5-haiku-20241022",
+        "temperature": 0,
+        "max_tokens": 8192
+    }
 }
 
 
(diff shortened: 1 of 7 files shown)


src/deepagents/graph.py
```diff
@@ -8,6 +7,8 @@
 
 from langgraph.prebuilt import create_react_agent
 
+from deepagents.model import get_default_model
```
hwchase17: weird reformating

src/deepagents/graph.py
```diff
@@ -53,6 +56,8 @@ def create_deep_agent(
     built_in_tools = [write_todos, write_file, read_file, ls, edit_file]
     if model is None:
         model = get_default_model()
+    elif isinstance(model, str):
```
hwchase17: why is this needed? i think just passing model string to create_react_agent is good enough?

src/deepagents/sub_agent.py
```diff
@@ -5,17 +5,23 @@
...
 
 class SubAgent(TypedDict):
     name: str
     description: str
     prompt: str
     tools: NotRequired[list[str]]
+    # Optional per-subagent model configuration
+    model: NotRequired[str]
+    model_provider: NotRequired[str]
+    max_tokens: NotRequired[int]
```
hwchase17: what about adding a model settings dictionary? makes it easier to add more parameters?

hwchase17 (commented):
i like this pr!

big comments:
- i think model_settings as a dictionary may be better? makes it easier to pass in even more, without blowing up the SubAgent class? (could be typed or untyped)
- dont need model provider, can just be specified as part of model (https://python.langchain.com/api_reference/langchain/chat_models/langchain.chat_models.base.init_chat_model.html)

emanueleielo:
> i like this pr!
> 
> big comments:
> 
> * i think model_settings as a dictionary may be better? makes it easier to pass in even more, without blowing up the SubAgent class? (could be typed or untyped)
> * dont need model provider, can just be specified as part of model (https://python.langchain.com/api_reference/langchain/chat_models/langchain.chat_models.base.init_chat_model.html)

thanks, yes that's a good idea i'll implement it

src/deepagents/graph.py
emanueleielo: i agree just forgot it

src/deepagents/sub_agent.py
emanueleielo: Switched to `model_settings` dict

emanueleielo:
### Summary of updates:
- Introduced `model_settings` dict for sub-agents.
- Dropped `model_provider`; provider is specified via `model`.
- Updated Model Resolution Logic
- Updated Examples
- Updated Documentation

@hwchase17

src/deepagents/model.py
```diff
@@ -1,5 +1,15 @@
-from langchain_anthropic import ChatAnthropic
+from langchain.chat_models import init_chat_model
+from typing import Optional
 
 
-def get_default_model():
-    return ChatAnthropic(model_name="claude-sonnet-4-20250514", max_tokens=64000)
+def get_default_model(
+    model: str = 'anthropic:claude-sonnet-4-20250514',
+    max_tokens: int = 8192,
```
hwchase17: this changes the default behavior - it used to be 64000 (that is intentional)

emanueleielo: i understand the max_tokens, about the `get_default_model` if we left the `ChatAnthropic` it will be not flexible (the orchestrator agent), is this intentional because the other models doesn't manage the **write_todos/sub-agents** correctly?

hwchase17: you can pass in a model currently though right? so unless im misunderstanding, it IS flexible.

examples/research/research_agent.py
```diff
@@ -65,6 +65,12 @@ def internet_search(
     "name": "critique-agent",
     "description": "Used to critique the final report. Give this agent some infomration about how you want it to critique the report.",
     "prompt": sub_critique_prompt,
+    # Use a lightweight model specialized for fast critique, with deterministic output
+    "model_settings": {
+        "model": "anthropic:claude-3-5-haiku-20241022",
+        "temperature": 0,
+        "max_tokens": 8192
+    }
```
hwchase17:
```suggestion
```

hwchase17 (approved):

src/deepagents/sub_agent.py
```diff
@@ -2,9 +2,10 @@
 from deepagents.state import DeepAgentState
 from langgraph.prebuilt import create_react_agent
 from langchain_core.tools import BaseTool
-from typing_extensions import TypedDict
+from typing_extensions import TypedDict, Any
 from langchain_core.tools import tool, InjectedToolCallId
 from langchain_core.messages import ToolMessage
+from langchain.chat_models import init_chat_model
 from typing import Annotated, NotRequired
```
hwchase17:
```suggestion
from typing_extensions import TypedDict
from langchain_core.tools import tool, InjectedToolCallId
from langchain_core.messages import ToolMessage
from langchain.chat_models import init_chat_model
from typing import Annotated, NotRequired, Any
```

PR #43, opened 2025-08-15
Using .ainvoke() to support async tools (e.g. tools retrieved via langchain-mcp-adapters) at subagent level

tnzk:
I've realized that subagents don't handle async tools properly as they use non-async `.invoke` inside. This becomes problematic when you use langchain-mcp-adapters which [doesn't support sync tools](https://github.com/langchain-ai/langchain-mcp-adapters/issues/158#issuecomment-2894629847) (always returns async ones). What makes this issue particularly tricky is that the main agent works fine with async tools when you use `.ainvoke` so the problem only shows up when the main agent delegates a task involving async tools to one of subagents. 

This PR fixes the issue by updating the subagents to call `.ainvoke` instead. [Here's the reproduction](https://github.com/tnzk/deepagents-async-subagents-repro/blob/main/main.py).


hwchase17 (approved):
thanks!

PR #58, opened 2025-08-19
feat: interrupt config 

Palashio:
This PR does two things:
- passes down the `state_schema` to the general purpose agent. currently, the state schema was only being passed to subagents and not the general purpose agent
- adds interrupt config with usage as seen below 
- creates ability to specify human interrupt config's separately for each tool call 

```
  # Using custom configs
  agent = create_deep_agent(
      tools=[...],
      instructions="...",
      interrupt_config={
          "custom_tool": HumanInterruptConfig(
              allow_ignore=True,
              allow_respond=False,
              allow_edit=True,
              allow_accept=True,
          )
      }
  )

``

### DIFF
--- a/src/deepagents/interrupt.py
+++ b/src/deepagents/interrupt.py
@@ -46,7 +46,7 @@ def create_interrupt_hook(
 
         # If no interrupts needed, return early
         if not interrupt_tool_calls:
-            return state
+            return
 
         approved_tool_calls = auto_approved_tool_calls.copy()
 

### CODE REVIEW

File: src/deepagents/graph.py [hunk 1]
```diff
@@ -67,11 +72,18 @@ def create_deep_agent(
         state_schema
     )
     all_tools = built_in_tools + list(tools) + [task_tool]
+    
+    # Create post model hook if interrupt config is provided
+    post_model_hook = None
```
hwchase17: should check that both are not specified (or, if both are specified, still create both)

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,82 @@
+"""Interrupt configuration functionality for deep agents."""
+
+from typing import Dict, Any, List
+from langgraph.types import interrupt
+
+
+class InterruptConfig:
```
hwchase17: i think we should use the interrupt schemas here: https://github.com/langchain-ai/agent-inbox?tab=readme-ov-file#interrupts

i think we have them in prebuilts here:https://langchain-ai.github.io/langgraph/reference/agents/#langgraph.prebuilt.interrupt.HumanInterruptConfig

Palashio: i used the `HumanInterruptConfig` for the schema parameters, and then `HumanResponse`. I can't use them fully because i need to provide the tool strings that they should act on at the top level.

src/deepagents/graph.py
```diff
@@ -5,13 +5,15 @@
...
 from langchain_core.language_models import LanguageModelLike
+from deepagents.interrupt import InterruptConfig, create_interrupt_hook
 from langgraph.types import Checkpointer
-
 from langgraph.prebuilt import create_react_agent
 
 StateSchema = TypeVar("StateSchema", bound=DeepAgentState)
 StateSchemaType = Type[StateSchema]
 
+
```
hwchase17: stray new lines (intentional?)

src/deepagents/graph.py
```diff
@@ -67,11 +80,20 @@ def create_deep_agent(
         state_schema
     )
     all_tools = built_in_tools + list(tools) + [task_tool]
+    
+    # Should never be the case that both are specified
+    selected_post_model_hook = None
+    if post_model_hook is not None:
+        selected_post_model_hook = post_model_hook
+    elif interrupt_config:
```
hwchase17: nit: good practice to try to end with a `else` so reader can understand fully
lets maybe put the raising of the error here so logic is all together (easier to read/understand)

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,122 @@
...
+        last_message = messages[-1]
+
+        if not hasattr(last_message, "tool_calls") or not last_message.tool_calls:
+            return state
+
+        approved_tool_calls = []
+
+        # Check each tool call for approval
+        for tool_call in last_message.tool_calls:
+            tool_name = tool_call.get("name", "")
```
hwchase17: when would these not exist?

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,122 @@
...
+        allow_respond=True,
+        allow_edit=True,
+        allow_accept=True,
+    ),
+}
+
+
+def create_interrupt_hook(
+    tool_configs: ToolInterruptConfig,
+    message_prefix: str = "Tool execution requires approval",
```
hwchase17, line 16: no way for user to specify this currently FYI (eg its not exposed in create_deep_agent)

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,122 @@
...
+                    "action_request": ActionRequest(
+                        action=tool_name,
+                        args=tool_args,
+                    ),
+                    "config": tool_config,
+                    "description": description,
+                }
+
+                interrupt_dict: Dict[str, HumanInterrupt] = {tool_name: request}
+                responses: List[HumanResponse] = interrupt([request])
```
hwchase17: i think this interrupts them sequentially. is this intentional or should be do in paralel?

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,122 @@
...
+                request: HumanInterrupt = {
+                    "action_request": ActionRequest(
+                        action=tool_name,
+                        args=tool_args,
+                    ),
+                    "config": tool_config,
+                    "description": description,
+                }
+
+                interrupt_dict: Dict[str, HumanInterrupt] = {tool_name: request}
```
hwchase17: i dont think this is used anywhere

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,122 @@
...
+                responses: List[HumanResponse] = interrupt([request])
+                if not responses:
```
hwchase17: when would this happen?

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,122 @@
...
+                elif response["type"] == "edit":
+                    edited: ActionRequest = response["args"] 
+                    # Replace args with edited args
+                    new_tool_call = {
+                        "name": tool_name,
+                        "args": edited["args"],
+                        "id": tool_call.get("id", ""),
+                    }
+                    approved_tool_calls.append(new_tool_call)
+                else:
```
hwchase17: i actually forget what the other types are - what are the other types?

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,122 @@
...
+                    continue
+            else:
+                approved_tool_calls.append(tool_call)
+
+        if len(approved_tool_calls) != len(last_message.tool_calls):
```
hwchase17: if its the same length but weve edited tool calls, we dont need to create new_message because... we are updating tool calls in place?

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,122 @@
...
+            new_message = type(last_message)(
```
hwchase17: can we just do

new_message.tool_calls = ...

src/deepagents/sub_agent.py
```diff
@@ -20,7 +20,7 @@ class SubAgent(TypedDict):
 
 def _create_task_tool(tools, instructions, subagents: list[SubAgent], model, state_schema):
     agents = {
-        "general-purpose": create_react_agent(model, prompt=instructions, tools=tools)
+        "general-purpose": create_react_agent(model, prompt=instructions, tools=tools, state_schema=state_schema)
```
hwchase17: i had actually not done this because if we need to do this i think we alrso probably want to update other keys? eg if sub agent changes a key, we need to propogate that back up?

hwchase17: should be sep pr

Palashio: sounds good will remove

src/deepagents/interrupt.py
Palashio: added

src/deepagents/interrupt.py
Palashio: was thinking edge cases, but think you're right in that there aren't any where this would trigger

src/deepagents/interrupt.py
Palashio: accept/edit/ignore, ignore is handled by the continue statement which skips adding the tool call to all approved ones

src/deepagents/interrupt.py
Palashio: i switched it to parallel, lmk if this isn't what you mean

src/deepagents/interrupt.py
hwchase17: what would it look like to make that part of the config? is that better? pro would be only one new argument to top level create agent

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,95 @@
...
+                approved_tool_calls.append(tool_call)
+            elif response["type"] == "edit":
+                edited: ActionRequest = response["args"]
+                new_tool_call = {
+                    "name": tool_call["name"],
+                    "args": edited["args"],
+                    "id": tool_call["id"],
+                }
+                approved_tool_calls.append(new_tool_call)
+            # For "ignore" and other types, we skip adding the tool call
```
hwchase17: what are other types?

i would maybe list out the other types and throw error if unknown type - in the case we add a new type, would catch that and not silently do something we dont expect

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,95 @@
...
+
+        last_message.tool_calls = approved_tool_calls
+
+        return state
```
hwchase17: i think more intuitive to return `{"messages": [last_message]}` (really clear that we are updating this)

src/deepagents/interrupt.py
Palashio: we'd have to create a new type, maybe `ToolConfig` that encapsulates the `ToolInterruptConfig` + `message_prefix`. And then users would specify if through there. So con would be that we wouldn't use the standard schemas. i think for now maybe we keep it (expose both) and then if people want more granular control (new message prefix for each) then we switch it over.

src/deepagents/interrupt.py
Palashio: accept/edit seems to be the only one's to handle after looking deeper, found this here: [https://langchain-ai.github.io/langgraph/how-tos/human_in_the_loop/add-human-in-the-loop/#review-tool-calls](https://langchain-ai.github.io/langgraph/how-tos/human_in_the_loop/add-human-in-the-loop/#review-tool-calls). modified this to an error.

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,96 @@
...
+    Args:
+        tool_configs: Dict mapping tool names to HumanInterruptConfig objects
+        message_prefix: Optional message prefix for interrupt descriptions
+    """
+    
+    def interrupt_hook(state: Dict[str, Any]) -> Dict[str, Any]:
+        """Post model hook that checks for tool calls and triggers interrupts if needed."""
+        messages = state.get("messages", [])
+        if not messages:
+            return state
```
hwchase17:
```suggestion
            return
```

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,96 @@
...
+    def interrupt_hook(state: Dict[str, Any]) -> Dict[str, Any]:
+        """Post model hook that checks for tool calls and triggers interrupts if needed."""
+        messages = state.get("messages", [])
+        if not messages:
+            return
+
+        last_message = messages[-1]
+
+        if not hasattr(last_message, "tool_calls") or not last_message.tool_calls:
+            return state
```
hwchase17:
```suggestion
            return
```

src/deepagents/interrupt.py
```diff
@@ -0,0 +1,96 @@
...
+        for tool_call in last_message.tool_calls:
+            tool_name = tool_call["name"]
+            if tool_name in tool_configs:
+                interrupt_tool_calls.append(tool_call)
+            else:
+                auto_approved_tool_calls.append(tool_call)
+
+        # If no interrupts needed, return early
+        if not interrupt_tool_calls:
+            return state
```
hwchase17:
```suggestion
            return
```

hwchase17 (approved):

PR #62, opened 2025-08-20
Accept Model Instances Instead of Dictionary Settings

riadzeitounn:
## The Problem

Right now, there's a weird inconsistency between how the main agent and subagents handle models:

- The main agent uses proper model instances like `ChatAnthropic(model_name="claude-sonnet-4", max_tokens=64000)`
- But subagents have to use dictionary configs like `{"model": "anthropic", "model_name": "claude-haiku"}`

This is pretty annoying because:

1. **It's inconsistent** - why should main agents and subagents work differently?
2. **Code duplication** - you have to repeat the same model settings everywhere
3. **It's just not intuitive**

## My Solution

I updated the `SubAgent` TypedDict to accept actual model instances instead of dictionaries.

### Before (the old way):

```python
subagents = [
    {
        "name": "researcher",
        "description": "Does research tasks", 
        "prompt": "You are a researcher...",
        "model_settings": {"model": "anthropic", "model_name": "claude-haiku-3"}  # Ugh, dictionaries
    }
]
```

### After (much cleaner):

```python

# Create your models once
model1 = ChatAnthropic(model_name="claude-haiku-3", max_tokens=4000)
model2 = ChatAnthropic(model_name="claude-sonnet-4", max_tokens=64000)

subagents = [
    {
        "name": "researcher", 
        "description": "Does research tasks",
        "prompt": "You are a researcher...",
        "model": model1  # Clean and reusable!
    },
    {
        "name": "writer",
        "description": "Does writing tasks", 
        "prompt": "You are a writer...",
        "model": model2  # Different model for different needs
    }
]
```

## What I Changed

1. Updated `SubAgent` TypedDict in `src/deepagents/sub_agent.py` to use `model: NotRequired[LanguageModelLike]` instead of `model_settings: NotRequired[dict[str, Any]]`
2. Simplified the model resolution logic in `_create_task_tool` - no more dictionary unpacking
3. Updated the documentation


riadzeitounn:
I agree, I pushed a version of the code to support both methods

PR #76, opened 2025-08-27
Fixed a typo for creating the subagents example code in the README file

armaanirani:
The section 'Creating a custom deep agent' had a typo and inconsistency in the naming. 
When defining the `research_subagent` in the example Python code for creating subagents had an extra '_' between 'sub' and 'agent', which was inconsistent with the name when adding it to the list of subagents right after that.


hwchase17 (approved):
thank you!

PR #79, opened 2025-08-27
Harrison/improve interrupt

hwchase17:


README.md
```diff
@@ -253,38 +253,111 @@ By default, deep agents come with five built-in tools:
...
 ```
 
-When a tool call requires approval, the agent will pause and wait for human input before proceeding. The message shown to users will include your custom prefix (or "Tool execution requires approval" by default) followed by the tool name and arguments. Multiple tool calls are processed in parallel, allowing you to review and approve multiple operations at once.
+#### Approve
+
+To "approve" a tool call means the agent will execute the tool call as is.
+
+This flow shows how to approve a tool call (assuming the tool requiring approval is called):
+
+```python
```
Palashio, line 311: these examples are just a bit confusing, took me a second to understand what was going on?

Palashio (approved):
looks good to me, just one small question.

i didn't test in the researcher example though

PR #80, opened 2025-08-28
Add support for custom graph subagents

Asya02:
### The Problem
Currently, subagents can only be react agents made using `create_react_agent`. This is inconvenient because often the subagent that the main agent wants to delegate a task to may be more efficient with a different architecture, with a custom graph. For example, a simple RAG
image

And it would be good to make such custom graphs specifically as separate subagents, and not transfer them as a tool to the main agent, so as not to transfer the entire message history to them, but only transfer the task to them.

### My Solution
I updated the SubAgent TypedDict to use a `graph` field, so that in case of a custom subagent architecture, the compiled graph can be passed to this field.
Now subagents can be declared with a `graph` field
```
sub_agent = {
    "name": "research-agent",
    "description": "Used to research more in depth questions. Only give this researcher one topic at a time. Do not pass multiple sub questions to this researcher. Instead, you should break down a large topic into the necessary components, and then call multiple research agents in parallel, one for each sub question.",
    "prompt": sub_research_prompt,
    "tools": ["internet_search"],
    "graph": <your_compiled_graph>
}
```
The `graph` field is optional. You can specify subagents in the same way as before:
```
sub_agent = {
    "name": "research-agent",
    "description": "Used to research more in depth questions. Only give this researcher one topic at a time. Do not pass multiple sub questions to this researcher. Instead, you should break down a large topic into the necessary components, and then call multiple research agents in parallel, one for each sub question.",
    "prompt": sub_research_prompt,
    "tools": ["internet_search"],
}
```
Then the old approach will be applied, where the subagent will be formed as before via `create_react_agent`.

### DIFF
--- a/src/deepagents/sub_agent.py
+++ b/src/deepagents/sub_agent.py
@@ -1,16 +1,16 @@
+from typing import Annotated, Any, NotRequired, Union
+
+from langchain.chat_models import init_chat_model
+from langchain_core.language_models import LanguageModelLike
+from langchain_core.messages import ToolMessage
+from langchain_core.runnables import Runnable
+from langchain_core.tools import BaseTool, InjectedToolCallId, tool
+from langgraph.prebuilt import InjectedState, create_react_agent
+from langgraph.types import Command
+from typing_extensions import TypedDict
+
 from deepagents.prompts import TASK_DESCRIPTION_PREFIX, TASK_DESCRIPTION_SUFFIX
 from deepagents.state import DeepAgentState
-from langgraph.prebuilt import create_react_agent
-from langchain_core.tools import BaseTool
-from typing_extensions import TypedDict
-from langchain_core.tools import tool, InjectedToolCallId
-from langchain_core.messages import ToolMessage
-from langchain_core.language_models import LanguageModelLike
-from langchain.chat_models import init_chat_model
-from typing import Annotated, NotRequired, Any, Union
-from langgraph.types import Command
-
-from langgraph.prebuilt import InjectedState
 
 
 class SubAgent(TypedDict):
@@ -20,6 +20,7 @@ class SubAgent(TypedDict):
     tools: NotRequired[list[str]]
     # Optional per-subagent model: can be either a model instance OR dict settings
     model: NotRequired[Union[LanguageModelLike, dict[str, Any]]]
+    graph: NotRequired[Runnable]
 
 
 def _create_task_tool(tools, instructions, subagents: list[SubAgent], model, state_schema):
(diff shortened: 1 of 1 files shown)


hwchase17:
let just make `subagents: list[SubAgent]` -> `subagents: list[Union[SubAgent, CompliedGraph]]`

Asya02:
> let just make `subagents: list[SubAgent]` -> `subagents: list[Union[SubAgent, CompliedGraph]]`

Thank you. It sounds interesting, I also initially thought of making a similar implementation. 

But I don't understand how to correctly add a name and description for the subagent, which are important for the main agent to understand what tasks should be given to this subagent (function `_create_task_tool`). It seems that if you simply add attributes to the compiled graph
```
compiled_graph.name = "SomeAgent"
compiled_graph.description = "SomeDescription"
```
, then this will differ little from the original dictionary that described the subagent, but a little less intuitive.
I would be grateful for help in this matter and am ready to correct this PR

riadzeitounn:
How will the communication happen between the deep agent and the compiled graph?

riadzeitounn:
> > let just make `subagents: list[SubAgent]` -> `subagents: list[Union[SubAgent, CompliedGraph]]`
> 
> Thank you. It sounds interesting, I also initially thought of making a similar implementation.
> 
> But I don't understand how to correctly add a name and description for the subagent, which are important for the main agent to understand what tasks should be given to this subagent (function `_create_task_tool`). It seems that if you simply add attributes to the compiled graph
> 
> ```
> compiled_graph.name = "SomeAgent"
> compiled_graph.description = "SomeDescription"
> ```
> 
> , then this will differ little from the original dictionary that described the subagent, but a little less intuitive. I would be grateful for help in this matter and am ready to correct this PR

if you pass the same json as your example, the description and name will pass into the _get_subagent_description and will be added for the deep agent:

sub_agent = {
    "name": "research-agent",
    "description": "Used to research more in depth questions. Only give this researcher one topic at a time. Do not pass multiple sub questions to this researcher. Instead, you should break down a large topic into the necessary components, and then call multiple research agents in parallel, one for each sub question.",
    "graph": <your_compiled_graph>
}

Asya02:
> How will the communication happen between the deep agent and the compiled graph?

Communication between the deep agent and the subagent, represented as a compiled graph in my solution, occurs in a similar way as in the original solution. The deep agent has a task tool, whose description lists all the subagents to which a specific task can be assigned. The names and descriptions of the subagents are taken from the sub_agent dictionary as before. The difference is that the subagent itself is not created using create_react_agent, but is used already compiled.

hwchase17:
how about this? https://github.com/langchain-ai/deepagents/pull/94

Asya02:
> how about this? #94


Thank you, this solution completely covers my cases

PR #86, opened 2025-09-01
Feature/regex search tool

hiroishda:
Added a regex search tool to the built-in tools.

This enables better context management when the agent is initialized with the files state variable prepopulated.
For example, you can prepopulate a deep agent with a large filesystem in the state, allowing it to operate over the entire filesystem more precisely.

Also added a demo that shows how to regex tool is called


hwchase17:
someting like this is great, but it should be be something that is similar to what you can do on a real filesytem (so that you can do it on a real filesystem). so like "grep"

hiroishda:
makes sense, I'll raise a new PR with the grep-like tool

hiroishda:
Closing to change the implementation of file system search to grep-like tool

PR #92, opened 2025-09-04
Support `create_react_agent`'s `pre_model_hook` arg

TaroVard:
Regarding the type - I chose to stick to the original `RunnableLike` type of `create_react_agent` even though we used the partial `Callable` for `post_model_hook` - I do think these passthrough args should stick to the original type but if you guys prefer simplicity and uniformness that's fine

### DIFF
--- a/src/deepagents/graph.py
+++ b/src/deepagents/graph.py
@@ -3,6 +3,7 @@ from deepagents.model import get_default_model
 from deepagents.tools import write_todos, write_file, read_file, ls, edit_file
 from deepagents.state import DeepAgentState
 from typing import Sequence, Union, Callable, Any, TypeVar, Type, Optional
+from langchain_core.runnables.base import RunnableLike
 from langchain_core.tools import BaseTool, tool
 from langchain_core.language_models import LanguageModelLike
 from deepagents.interrupt import create_interrupt_hook, ToolInterruptConfig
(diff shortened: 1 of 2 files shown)


hwchase17:
ooc - what are you using pre_model_hook for?

TaroVard:
My specific use case is actually a workaround for something more basic that I haven't been able to achieve 😅 
TL;DR I use it to globally trim tool results beyond my acceptable limit 

My non-deepagents version had a custom `ToolNode` that did that, which is probably the more idiomatic way of doing it
however when using deepagents passing around ToolNode instead of the list of tools quickly became a problem with all the subagent routing and general method signature mismatches along the way (even when I'm currently running my own fork due to other limitations)

With custom ToolNode out of the question I resulted to my second-best option, which is definitely the cleanest in terms of added code (simplified version):

```
def trim_tool_response_hook(state: AgentState) -> dict[str, Any]:
    max_chars = 16000 * 4
    messages = state.get("messages", [])

    for message in [
        message
        for message in messages
        if isinstance(message, ToolMessage)
        and not message.additional_kwargs.get("checked_for_trimming")
    ]:
        text = message.text()
        if len(text) > max_chars:
            trimmed_content = text[:max_chars]

            warning_prefix = (
                "**TRIMMED RESULT** (Response was too large)\n"
                "--- PARTIAL CONTENT ---\n"
            )
            message.content = (
                warning_prefix + trimmed_content + "\n--- END OF PARTIAL CONTENT ---\n"
            )

            message.additional_kwargs["trimmed"] = True
        message.additional_kwargs["checked_for_trimming"] = True

    return {"messages": messages}
```

lpb1:
+1, we need this feature too.

nhuang-lc:
Hey all, pre-model hook is now implemented through Middleware in LangChain v1 ([see docs](https://docs.langchain.com/oss/python/langchain/middleware#state-extension)) 

Deepagents now supports custom middleware extensions, so you can implement pre-model hooks and use this in deepagents

PR #93, opened 2025-09-07
Harrison/allow filtering of main agent tools

hwchase17:


PSig714:
There's also an issue with subagent misalignment if they are unaware of eachothers existence. They must have an interconnected prompt structure where all subagents are aware of eachother along with the parent deepagent.

hwchase17:
> There's also an issue with subagent misalignment if they are unaware of eachothers existence. They must have an interconnected prompt structure where all subagents are aware of eachother along with the parent deepagent.

no i dont think this is true. if the main agent is aware of all subagents, it should be the one (and only one) to do the orchestration. if you want more connectedness - that is just a different archticture entierely

PR #94, opened 2025-09-07
Harrison/custom sub agent

hwchase17:


riadzeitounn:
Thats a great solution, I would suggest adding the config to the ainvoke of the subagent/Custom Graph too

fmdelgado:
what about having a SUBAGENT_RECURSION_LIMIT too?

PR #102, opened 2025-09-11
Allow passing response format into deep agent

willatlawford:
Langgraph's create_react_agent supports passing a response format argument. [Docs here](https://langchain-ai.github.io/langgraph/agents/agents/#6-configure-structured-output)

When passed, output will be formatted to match the given schema and returned in the 'structured_response' state key. 

I am using deepagents for agentic data extraction, so this is a useful addition.


sydney-runkle:
Closing as stale, now supported!

Issue #99, opened 2025-09-11
Unable to Pass Config or Use astream with Subagent Custom Graph


riadzeitounn:
Hey,
There are some features that would be helpful in the subagent graph and would make it more dynamic and work as any custom langgraph graph:

1- Config Not Passing to Subagent Graph
Unable to pass a custom config to the subagent's graph as we pass to any graph.

2- Can't Use astream with Subagent Graph
Unable to use astream() with the subagent graph — only ainvoke() works. Ideally, we'd like to stream results from the subagent as we do with the main graph.

#107

hwchase17:
thanks for raising. why would you want to pass a custom config to the sub agents graph?

streaming from sub graphs should be possible already with: https://docs.langchain.com/oss/python/langgraph/streaming#stream-subgraph-outputs

if this doesnt work... we should fix it (possilbe that being inside a tool is causing it to mess up?)

riadzeitounn:
Hey, I didn't try this streaming method, thanks for pointing it out.
For the config, imagine having a custom graph as a subagent, and the graph depends on a value in config to change its route. In addition, some tools inside subagents, would use config to do specific stuff on the run.

riadzeitounn:
This is now possible with new Middleware architecture. Thanks

PR #107, opened 2025-09-12
Added Config to Subagents and allowed streaming

riadzeitounn:
This was related to issue #99


riadzeitounn:
You can pass the stream_mode in different ways, I used config, but I am open to a better solution.

PR #111, opened 2025-09-16
Improve default deepagents performance

nhuang-lc:
Addresses the following issues

- Overkill usage of To-Dos for planning when tasks are simple
- Unclear when something should be a Task vs just direct tool-calling


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [project]
 name = "deepagents"
-version = "0.0.5"
+version = "0.0.6rc1"
```
hwchase17: lets not merge this

PR #118, opened 2025-09-19
WIP: Migrate to langchain v1

nhuang-lc:
**Features**
- Expose PlanningMiddleware, FilesystemMiddleware, and SubagentMiddleware as three independent middlewares that can be added to an agent. 
- create_deep_agent and async_create_deep_agent still create prebuilts with these middlewares
- Use create_agent instead of create_react_agent
- Migrate post_model_hook to Middleware, users can supply their own middleware for the agent to use

**TODO**
- Tools still take strings (subset of main agent tools)
- Outstanding bug with InjectedToolState being investigated in built-in tools
- Extended state_schemas are not supported via a state_schema variable, users will need to specify their own middleware.
- Fix and test interrupts and HITL support


src/deepagents/builder.py
```diff
@@ -35,14 +35,20 @@ def build_agent(config: Optional[dict] = None):
         else:
             config = {}
         config_fields = {
-            k: v for k, v in config.items() if k in ["instructions", "subagents"]
+            k: v for k, v in config.items() if k in ["instructions", "subagents", "tools"]
```
nhuang-lc: Need to add this - previously we were ignoring tools so tools passed in the config didn't do anything

README.md
```diff
@@ -24,20 +24,20 @@ pip install deepagents
...
 from typing import Literal
-
+from langchain_core.tools import tool
 from tavily import TavilyClient
 from deepagents import create_deep_agent
 
 tavily_client = TavilyClient(api_key=os.environ["TAVILY_API_KEY"])
 
 # Search tool to use to do research
+@tool(description="Use this tool to run a web search")
```
hwchase17: do we need this?

README.md
```diff
@@ -128,6 +130,7 @@ research_subagent = {
     "name": "research-agent",
     "description": "Used to research more in depth questions",
     "prompt": sub_research_prompt,
+    "tools": [internet_search]
```
hwchase17, line 132: can we keep making this optional, and if key not provided then pass through all tools

README.md
```diff
@@ -230,6 +221,46 @@ agent = create_deep_agent(
 )
 ```
 
+## Deep Agents Middleware
```
hwchase17: this should either be documented in more detail (what exactly each one does, API, etc) or not at all

I would not document to start, we can add in after and do another release on that

examples/research/research_agent.py
```diff
@@ -2,22 +2,21 @@
...
-
-from deepagents import create_deep_agent, SubAgent
+from deepagents import create_deep_agent
 
 # It's best practice to initialize the client once and reuse it.
 tavily_client = TavilyClient(api_key=os.environ["TAVILY_API_KEY"])
 
-
 # Search tool to use to do research
+@tool(description="Use this tool to run a web search")
```
hwchase17: why?

examples/research/research_agent.py
```diff
@@ -37,7 +36,8 @@ def internet_search(
     "name": "research-agent",
     "description": "Used to research more in depth questions. Only give this researcher one topic at a time. Do not pass multiple sub questions to this researcher. Instead, you should break down a large topic into the necessary components, and then call multiple research agents in parallel, one for each sub question.",
     "prompt": sub_research_prompt,
-    "tools": ["internet_search"],
+    "tools": [internet_search],
+    "middleware": []
```
hwchase17: is this needed?

examples/research/research_agent.py
```diff
@@ -66,6 +66,7 @@ def internet_search(
     "name": "critique-agent",
     "description": "Used to critique the final report. Give this agent some information about how you want it to critique the report.",
     "prompt": sub_critique_prompt,
+    "middleware": []
```
hwchase17: needed?

src/deepagents/builder.py
```diff
@@ -70,14 +76,20 @@ def build_agent(config: Optional[dict] = None):
...
         }
         config = AgentConfig(**config_fields)
+        subagents_with_tools = [
+            {
+                **subagent,
```
hwchase17: hm i dont love this... i think we need to figure this out... is a bit annoying to pass all tools in...

Could we get away with passing in a function that takes a tool name and returns the tool itself?

how would we use this in OAP?

src/deepagents/graph.py
```diff
@@ -1,130 +1,86 @@
...
-            if not isinstance(tool_, BaseTool):
-                tool_ = tool(tool_)
-            if tool_.name in main_agent_tools:
-                passed_in_tools.append(tool_)
-    else:
-        passed_in_tools = list(tools)
-    all_tools = built_in_tools + passed_in_tools + [task_tool]
-
-    return create_react_agent(
+    elif isinstance(model, str):
```
hwchase17: can we not just pass this through? do places down stream not handle str?

src/deepagents/graph.py
```diff
@@ -1,130 +1,86 @@
...
+    #     deepagent_middleware.append(AnthropicPromptCachingMiddleware(ttl="5m"))
+
+    return create_agent(
         model,
-        prompt=prompt,
-        tools=all_tools,
-        state_schema=state_schema,
-        post_model_hook=selected_post_model_hook,
-        config_schema=config_schema,
+        prompt=instructions + "\n\n" + BASE_AGENT_PROMPT,
```
hwchase17, line 57: is BASE_AGENT_PROMPT even needed?

src/deepagents/tools.py
```diff
@@ -29,15 +29,15 @@ def write_todos(
...
+def ls(state: Annotated[dict, InjectedState]) -> list[str]:
     """List all files"""
     return list(state.get("files", {}).keys())
 
 
 @tool(description=READ_FILE_TOOL_DESCRIPTION)
 def read_file(
     file_path: str,
-    state: Annotated[DeepAgentState, InjectedState],
+    state: Annotated[AgentState, InjectedState],
```
hwchase17: dont we want this typed better so it knows of files?

src/deepagents/builder.py
nhuang-lc: Resolving per slack convo - removing builder

README.md
nhuang-lc: Needs to be BaseTool to be passed to create_agent. Are you suggesting that if not tool we wrap it for them?

src/deepagents/graph.py
nhuang-lc: We need to check if the model is Anthropic or not to choose whether or not to add prompt caching, I can remove this for now but we either need to add this back later, or I talked to sydney about updating the middleware to handle for folks

src/deepagents/graph.py
nhuang-lc: Personally I think it's helpful for framing so we don't just launch into the tool definitions without any preamble

examples/research/research_agent.py
nhuang-lc: Same as above ^^

examples/research/research_agent.py
nhuang-lc: Not required, I aded just to show that users CAN pass in more custom middleware

examples/research/research_agent.py
nhuang-lc: Same as above ^^

README.md
nhuang-lc: Does not need to be basetool, reverted this change

README.md
```diff
@@ -24,7 +24,7 @@ pip install deepagents
 ```python
 import os
 from typing import Literal
-
+from langchain_core.tools import tool
```
hwchase17: not needed

examples/research/research_agent.py
```diff
@@ -2,14 +2,13 @@
 from typing import Literal
 
 from tavily import TavilyClient
+from langchain_core.tools import tool
```
hwchase17: not neded

hwchase17 (approved):

Issue #105, opened 2025-09-12
subagent create


arkonchen:
Would it be possible to place the creation of the sub agent in the task function instead of _create_task_tool?

nhuang-lc:
Can you elaborate on why this would be helpful for you?

arkonchen:
For example, there are some mcp tools that I don't want to load at the beginning of the process for creating a sub agent, I want to save some resources by loading them when I want to call them.

phanijapps:
can't this be done via subgraph with middleware modify_model_request?

nhuang-lc:
Yep, that's one workaround @phanijapps!

Another workaround would be to create a custom subagent graph (that lazily loads tools to your liking) and then to point to that graph with a [CompiledSubAgent](https://docs.langchain.com/oss/python/deepagents/subagents#compiledsubagent). This is relatively new so it likely wasn't possible when you opened this issue.

Personally I don't think this is that much of an issue (though your specific use case may be different) because in production graph initialization happens once. It's preferable to load all of the tools once at startup, than to load the tools over and over when the subagents are called.

PR #123, opened 2025-09-21
Move sub-agent creation from _create_task_tool to inside task function

ummtushar:
This was related to issue #105 

This change moves the creation of the agents dictionary from the tool creation time to inside the task function, making it dynamic per invocation. This allows for potential future flexibility but may impact performance.


ummtushar:
can someone review this change?

vtrivedy:
hey thanks for the PR, we'll leave the semantics as they are but have new capabilities in subagents released soon!

PR #125, opened 2025-09-23
Update prompts.py slight grammatical error

roombawulf:
just adding a missing space


nhuang-lc:
Thanks for opening, appreciate the eye for detail! I fixed this as part of a larger PR

PR #126, opened 2025-09-23
chore(deepagents): use new decorator syntax for simple middlewares

sydney-runkle:
This is w/ a local version of `langchain` v1, can cut a release if we like this API.

This seems to be working w/ the `chat-langchain` application :)


sydney-runkle:
Closing, I think we just want to use the subclassing API for now.

PR #130, opened 2025-09-24
Turn on anthropic prompt caching

nhuang-lc:
Also reduce number of tokens before summarization (150k + 64k was exceeding 200k limits)


nhuang-lc:
Added in https://github.com/langchain-ai/deepagents/pull/135

PR #157, opened 2025-10-08
Add long-term memory support through Store

nhuang-lc:
- Adds longterm memory support for store
- Adds configurability for system prompt and tool descriptions for the FilesystemMiddleware
- Uses different prompts for filesystem tools based on whether or not we have long-term memory access

### DIFF
--- a/src/deepagents/middleware.py
+++ b/src/deepagents/middleware.py
@@ -7,7 +7,7 @@ from langchain_core.tools import BaseTool, tool, InjectedToolCallId
 from langchain_core.messages import ToolMessage
 from langchain.chat_models import init_chat_model
 from langgraph.types import Command
-from langgraph.runtime import get_runtime, Runtime
+from langgraph.runtime import Runtime
 from langchain.tools.tool_node import InjectedState
 from typing import Annotated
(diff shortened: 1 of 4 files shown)


src/deepagents/graph.py
```diff
@@ -20,14 +21,18 @@ def agent_builder(
     subagents: Optional[list[SubAgent | CustomSubAgent]] = None,
     context_schema: Optional[Type[Any]] = None,
     checkpointer: Optional[Checkpointer] = None,
+    store: Optional[BaseStore] = None,
+    use_longterm_memory: bool = False,
```
Ry-DS, line 25: Why have this Boolean when you can just decide whether to use the store based off if it's none or not?

nhuang-lc: People may add a Store for use in tool calls to store other things, not related to long-term memory.

Ry-DS: This should be mentioned in the doc comment on line 97, not clear otherwise

nhuang-lc:
Closing this PR in favor of a separate one that will include these updates to long-term mem

PR #201, opened 2025-10-23
Change API to accept a single memory_backend, and fix all tests

nhuang-lc:


src/deepagents/memory/backends/composite.py
```diff
@@ -190,7 +190,7 @@ def grep(
         pattern: str,
         path: str = "/",
         include: Optional[str] = None,
-        output_mode: str = "files_with_matches",
```
nhuang-lc, line 193: We should support all modes in composite

src/deepagents/memory/backends/state.py
```diff
@@ -200,7 +200,7 @@ def grep(
         pattern: str,
         path: str = "/",
         include: Optional[str] = None,
-        output_mode: str = "files_with_matches",
```
nhuang-lc, line 203: Same here, need to support more output modes

src/deepagents/memory/backends/store.py
```diff
@@ -126,7 +173,7 @@ def ls(self, prefix: Optional[str] = None, runtime: Optional["ToolRuntime"] = No
         namespace = self._get_namespace()
         
         # Search store with optional prefix filter
-        items = store.search(namespace, filter={"prefix": prefix} if prefix else None)
+        items = self._search_store_paginated(store, namespace, filter={"prefix": prefix} if prefix else None)
```
nhuang-lc, line 176: Use paginated search

src/deepagents/memory/backends/utils.py
```diff
@@ -94,6 +95,7 @@ def update_file_data(file_data: dict[str, Any], content: str) -> dict[str, Any]:
         Updated FileData dict
     """
     lines = content.split("\n") if isinstance(content, str) else content
+    lines = [line[i:i+MAX_LINE_LENGTH] for line in lines for i in range(0, len(line) or 1, MAX_LINE_LENGTH)]
```
nhuang-lc, line 98: Ensures that we only update files with lines of MAX_LINE_LENGTH

src/deepagents/memory/protocol.py
```diff
@@ -151,13 +151,14 @@ def grep(
         """
         ...
     
-    def glob(self, pattern: str, runtime: Optional["ToolRuntime"] = None) -> list[str]:
+    def glob(self, pattern: str, path: str = "/", runtime: Optional["ToolRuntime"] = None) -> list[str]:
```
nhuang-lc, line 154: We need to accept path parameter here

src/deepagents/middleware/filesystem.py
```diff
@@ -538,49 +537,26 @@ class FilesystemMiddleware(AgentMiddleware):
...
         Args:
-            backend: Optional backend for file storage. If provided, uses it directly.
-            long_term_memory: Optional backend for /memories/ files. If provided, creates CompositeBackend.
+            memory_backend: Backend for file storage. Defaults to StateBackend if not provided.
             system_prompt: Optional custom system prompt override.
             custom_tool_descriptions: Optional custom tool descriptions override.
             tool_token_limit_before_evict: Optional token limit before evicting a tool result to the filesystem.
         """
         self.tool_token_limit_before_evict = tool_token_limit_before_evict
-        # Handle shortcut for default longterm memory
```
nhuang-lc, line 557: Removed this complicated logic

PR #208, opened 2025-10-27
experimental backend

hwchase17:


README.md
```diff
@@ -384,15 +393,24 @@ Context engineering is one of the main challenges in building effective agents.
 ```python
 from deepagents.middleware.filesystem import FilesystemMiddleware
+from deepagents.backends import (
+    StateBackendProvider,
+    CompositeStateBackendProvider,
```
nickhuang25: Maybe CompositeStateBackendProvider was from an earlier iteration? I don't think exists anymore 

Should replace with `CompositeBackend` or `build_composite_state_backend`

PR #209, opened 2025-10-27
simplify backend protocol

eyurtsev:
Simplify backend protocol

### DIFF
--- a/src/deepagents/backends/protocol.py
+++ b/src/deepagents/backends/protocol.py
@@ -20,23 +20,21 @@ class WriteResult:
     Attributes:
         error: Error message on failure, None on success.
         path: Absolute path of written file, None on failure.
-        content: Written file content, may be None.
         files_update: State update dict for checkpoint backends, None for external storage.
             Checkpoint backends populate this with {file_path: file_data} for LangGraph state.
             External backends set None (already persisted to disk/S3/database/etc).
 
     Examples:
         >>> # Checkpoint storage
-        >>> WriteResult(path="/f.txt", content="hi", files_update={"/f.txt": {...}})
+        >>> WriteResult(path="/f.txt", files_update={"/f.txt": {...}})
         >>> # External storage
-        >>> WriteResult(path="/f.txt", content="hi", files_update=None)
+        >>> WriteResult(path="/f.txt", files_update=None)
         >>> # Error
         >>> WriteResult(error="File exists")
     """
 
     error: str | None = None
     path: str | None = None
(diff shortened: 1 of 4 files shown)


src/deepagents/backends/protocol.py
```diff
@@ -1,218 +1,200 @@
...
+        >>> # Checkpoint storage
+        >>> EditResult(path="/f.txt", files_update={"/f.txt": {...}}, occurrences=1)
+        >>> # External storage
+        >>> EditResult(path="/f.txt", files_update=None, occurrences=2)
+        >>> # Error
+        >>> EditResult(error="File not found")
+    """
+
+    error: str | None = None
+    path: str | None = None
```
eyurtsev: We can either standardize what the backends return (e.g., path, occurrences) this allows us to standardize formatting also.

and/oR we allow the backends to ship a free form `msg`

src/deepagents/backends/composite.py
```diff
@@ -1,36 +1,45 @@
...
+    """
 
-class _CompositeBackend:
-    
     def __init__(
         self,
-        default: BackendProtocol | StateBackend,
-        routes: dict[str, BackendProtocol],
+        default: Backend,
+        routes: dict[str, Backend],
```
eyurtsev, line 20: Why is routes a dictionary? Why so general?

src/deepagents/backends/protocol.py
```diff
@@ -1,218 +1,200 @@
...
-@runtime_checkable
-class StateBackendProvider(Protocol):
 
-    def get_backend(self, runtime: ToolRuntime) -> StateBackendProtocol:
-        """Get the backend."""
-        ...
\ No newline at end of file
+# Backend factory function types
+BackendProvider = Callable[[ToolRuntime], Backend]
+AsyncBackendProvider = Callable[[ToolRuntime], Awaitable[Backend]]
```
eyurtsev, line 200: We'll need async support at some point for all operations + the factory itself. Would suggest **not** adding it until we're ready to add full async support

src/deepagents/graph.py
```diff
@@ -48,7 +48,7 @@ def create_deep_agent(
     context_schema: type[Any] | None = None,
     checkpointer: Checkpointer | None = None,
     store: BaseStore | None = None,
-    backend: BackendProtocol | None = None,
+    backend: BackendProvider | None = None,
```
eyurtsev, line 51: This should be Backend

PR #216, opened 2025-10-28
Update ls behavior

nhuang-lc:
- ls now lists files and directories at the provided path
- Raises error if path passed to `root_dir` in FilesystemBackend is not an absolute path

### DIFF
--- a/tests/backends/test_filesystem_backend.py
+++ b/tests/backends/test_filesystem_backend.py
@@ -24,6 +24,7 @@ def test_filesystem_backend_normal_mode(tmp_path: Path):
     paths = {i["path"] for i in infos}
     assert str(f1) in paths  # File in root should be listed
     assert str(f2) not in paths  # File in subdirectory should NOT be listed
+    assert (str(root) + "/dir/") in paths  # Directory should be listed
 
     # read, edit, write
     txt = be.read(str(f1))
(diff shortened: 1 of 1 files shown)


src/deepagents/backends/filesystem.py
```diff
@@ -51,6 +51,9 @@ def __init__(
                      all file paths will be resolved relative to this directory.
                      If not provided, uses the current working directory.
         """
+        if root_dir and not root_dir.startswith("/"):
```
hwchase17: can use pathlib more here. eg `if root_dir and not Path(root_dir).is_absolute()`

also can combine with logic on line 57

tests/backends/test_filesystem_backend.py
```diff
@@ -19,10 +19,11 @@ def test_filesystem_backend_normal_mode(tmp_path: Path):
 
     be = FilesystemBackend(root_dir=str(root), virtual_mode=False)
 
-    # ls_info absolute path
+    # ls_info absolute path - should only list files in root, not subdirectories
     infos = be.ls_info(str(root))
     paths = {i["path"] for i in infos}
-    assert str(f1) in paths and str(f2) in paths
+    assert str(f1) in paths  # File in root should be listed
```
hwchase17, line 25: can we test that root/"dir" is listed?

tests/backends/test_filesystem_backend.py
```diff
@@ -50,10 +51,11 @@ def test_filesystem_backend_virtual_mode(tmp_path: Path):
 
     be = FilesystemBackend(root_dir=str(root), virtual_mode=True)
 
-    # ls_info from virtual root
+    # ls_info from virtual root - should only list files in root, not subdirectories
     infos = be.ls_info("/")
     paths = {i["path"] for i in infos}
-    assert "/a.txt" in paths and "/dir/b.md" in paths
+    assert "/a.txt" in paths  # File in root should be listed
+    assert "/dir/b.md" not in paths  # File in subdirectory should NOT be listed
```
hwchase17, line 59: can we test that "dir" is listed?

hwchase17 (approved):

PR #222, opened 2025-10-29
Large Tool Interceptor should write to backend

nhuang-lc:
The large tool interceptor used to always write to state, now we should write to the provided backend

### DIFF
--- a/src/deepagents/middleware/filesystem.py
+++ b/src/deepagents/middleware/filesystem.py
@@ -28,6 +28,7 @@ from deepagents.backends.utils import (
     format_content_with_line_numbers,
     format_grep_matches,
     truncate_if_too_long,
+    sanitize_tool_call_id,
 )
 
 EMPTY_CONTENT_WARNING = "System reminder: File exists but has empty contents"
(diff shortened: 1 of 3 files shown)


src/deepagents/middleware/filesystem.py
```diff
@@ -568,54 +567,69 @@ async def awrap_model_call(
...
+    def _process_large_message(
+        self,
+        message: ToolMessage,
+        resolved_backend: BackendProtocol,
+    ) -> tuple[ToolMessage, dict[str, FileData] | None]:
+        content = message.content
+        if not isinstance(content, str) or len(content) <= 4 * self.tool_token_limit_before_evict:
+            return message, None
+
+        file_path = f"/large_tool_results/{message.tool_call_id}"
```
eyurtsev: could you sanitize the tool call id by replacing catching `..` and `/` and `\` and replacing them with something less threatening?

src/deepagents/middleware/filesystem.py
```diff
@@ -568,54 +567,69 @@ async def awrap_model_call(
...
-                            tool_call_id=tool_result.tool_call_id,
-                        )
-                    ],
-                    "files": {file_path: file_data},
-                }
-                return Command(update=state_update)
+            if not (self.tool_token_limit_before_evict and
+                    len(tool_result.content) > 4 * self.tool_token_limit_before_evict):
+                return tool_result
+            resolved_backend = _get_backend(self.backend, runtime)
```
eyurtsev: nit: `_get_backend` should be a method in this case since it's relying on private knowledge (self.backend)

src/deepagents/middleware/filesystem.py
eyurtsev: Also how do we handle separators?

nhuang-lc: Will replace them with _ when sanitizing tool call id

PR #224, opened 2025-10-30
Add deepagents cli scaffolding

eyurtsev:
Add deep-agents cli scaffolding

### DIFF
--- a/libs/deepagents-cli/pyproject.toml
+++ b/libs/deepagents-cli/pyproject.toml
@@ -22,7 +22,17 @@ deepagents-cli = "deepagents_cli:cli_main"
 [tool.uv.sources]
 deepagents = { workspace = true }
 
-[project.optional-dependencies]
+[dependency-groups]
+test = [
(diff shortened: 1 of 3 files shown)


libs/deepagents-cli/pyproject.toml
```diff
@@ -0,0 +1,97 @@
+[project]
+name = "deepagents-cli"
+version = "0.0.1"
+description = "Deepagents CLI"
+readme = "README.md"
+license = { text = "MIT" }
+requires-python = ">=3.11,<4.0"
+dependencies = [
+  "deepagents",
```
hwchase17, line 9: version lower bound?

libs/deepagents-cli/pyproject.toml
```diff
@@ -0,0 +1,97 @@
+[project]
+name = "deepagents-cli"
+version = "0.0.1"
```
hwchase17, line 3: ive already published some versions here, would need to b 0.0.4 or 0.1

PR #244, opened 2025-10-30
fix ui height overflow, new ascii banner, fix model line

casparb:
- fixed screen height by adding reserve_space_for_menu=0
- moved console.clear() before create_model() to prevent double banner
- moved model info display from create_model() to simple_cli() and positioned it before working directory line

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/main.py
+++ b/libs/deepagents-cli/deepagents_cli/main.py
@@ -95,9 +95,7 @@ def parse_args():
 
 async def simple_cli(agent, assistant_id: str | None, session_state, baseline_tokens: int = 0):
     """Main CLI loop."""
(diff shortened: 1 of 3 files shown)


libs/deepagents-cli/deepagents_cli/main.py
```diff
@@ -174,6 +173,10 @@ async def main(assistant_id: str, session_state):
     """Main entry point."""
     # Create the model (checks API keys)
     model = create_model()
+    session_state.model = model
```
eyurtsev, line 176: Mutable assignment here?

FWIW something related to this change crashes the app

eyurtsev: I'll try to resolve now

PR #246, opened 2025-10-30
fix(cli): Abort functionality, autocomplete, and fixed memory prompts

vtrivedy:
Fixes terminal state corruption after agent execution.

## Changes
- Flush stdout/stderr after task completion
- Reset ANSI attributes to clean state
- Prevents Rich console from corrupting prompt_toolkit

## Issue
After agent execution, the terminal state could become desynchronized causing prompt rendering issues.

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/execution.py
+++ b/libs/deepagents-cli/deepagents_cli/execution.py
@@ -363,6 +363,10 @@ async def execute_task(
                                     console.print()
                                     decisions.append({"type": "approve"})
 
+                                # Flush console output before resuming to prevent terminal interference
+                                sys.stdout.flush()
+                                sys.stderr.flush()
+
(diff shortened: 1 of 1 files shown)


libs/deepagents-cli/deepagents_cli/input.py
```diff
@@ -20,53 +21,65 @@
...
+                p = Path(completed_path).expanduser()
+                if not p.is_absolute():
+                    p = Path.cwd() / p
+
+                # If it's a directory and doesn't already end with /
+                if p.is_dir() and not completion.text.endswith("/"):
+                    # Modify the completion to include trailing /
                     yield Completion(
-                        text=completion.text,
+                        text=completion.text + "/",
```
eyurtsev, line 48: this will only work on posix

PR #268, opened 2025-11-03
refactor(graph): apply dependency inversion to default model creation

zhyg:
- Change get_default_model() return type from ChatAnthropic to BaseChatModel
- Move ChatAnthropic import from module level to function scope
- Remove direct dependency on concrete implementation at module boundary


eyurtsev:
We generally discourage refactors as contributions. The best way to contribute is to ask for features and explain the use cases (not via implementation)

PR #274, opened 2025-11-04
fix: improve tool rejection handling and add double Ctrl+C exit

vtrivedy:
Remove background thread hack in rejection handling that caused race conditions. Use LangGraph Command to resume and update agent state synchronously when tools are rejected
Add stern system prompt instructions to prevent agent chatter after rejections
Implement double Ctrl+C to exit with auto-hiding toolbar message to exit session. Show both auto-accept status and quit warning in combined toolbar
Fixes issues with:

Agent state should not be handled by background thread, could cause inconsistencies in state
Normal ctrl+c exit pattern

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -211,3 +211,8 @@ __marimo__/
 
 #claude
 .claude
+
+# planning files
+IMPLEMENTATION_GUIDE.md
+pr_251_edits.md
+improvements.md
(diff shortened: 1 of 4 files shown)


vtrivedy:
Fix unnecessary files and add small UI changes

PR #278, opened 2025-11-04
fix (cli): autocomplete for @ and / commands through directories, bash mode in TUI

vtrivedy:
- autocomplete through directories, works on backspace
- overhaul search for substrings to use regex and FilePathComplete to find valid directories
- fixed terminal spacing at bottom so room for menu
- removed bash menu autocomplete because unnecessary, but added bash mode when user types !

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/input.py
+++ b/libs/deepagents-cli/deepagents_cli/input.py
@@ -46,17 +46,25 @@ class FilePathCompleter(Completer):
 
         path_fragment = m.group("path")
 
-        # Create temporary document for just the path fragment
-        temp_doc = Document(text=path_fragment, cursor_position=len(path_fragment))
+        # Unescape the path for PathCompleter (it doesn't understand escape sequences)
+        unescaped_fragment = path_fragment.replace("\\ ", " ")
+
+        # Strip trailing backslash if present (user is in the process of typing an escape)
+        if unescaped_fragment.endswith("\\"):
+            unescaped_fragment = unescaped_fragment[:-1]
+
+        # Create temporary document for the unescaped path fragment
+        temp_doc = Document(text=unescaped_fragment, cursor_position=len(unescaped_fragment))
 
         # Get completions from PathCompleter and use its start_position
         # PathCompleter returns suffix text with start_position=0 (insert at cursor)
         for comp in self.path_completer.get_completions(temp_doc, complete_event):
             # Add trailing / for directories so users can continue navigating
-            completed_path = Path(path_fragment + comp.text).expanduser()
-            completion_text = comp.text
-            if completed_path.is_dir() and not completion_text.endswith(os.sep):
-                completion_text += os.sep
+            completed_path = Path(unescaped_fragment + comp.text).expanduser()
+            # Re-escape spaces in the completion text for the command line
+            completion_text = comp.text.replace(" ", "\\ ")
+            if completed_path.is_dir() and not completion_text.endswith('/'):
+                completion_text += '/'
 
             yield Completion(
                 text=completion_text,


libs/deepagents-cli/deepagents_cli/input.py
```diff
@@ -9,118 +9,85 @@
...
 from prompt_toolkit.enums import EditingMode
 from prompt_toolkit.formatted_text import HTML
 from prompt_toolkit.key_binding import KeyBindings
 
-from .config import COLORS, COMMANDS, COMMON_BASH_COMMANDS, SessionState, console
+from .config import COLORS, COMMANDS, SessionState, console
+
+
+# Regex patterns for context-aware completion
+AT_MENTION_RE = re.compile(r'@(?P<path>[A-Za-z0-9._~/-]*)$')
```
eyurtsev: OK for now -- though it'll fail for whitespace (which is allowed in file names)

libs/deepagents-cli/deepagents_cli/input.py
```diff
@@ -150,15 +117,34 @@ def parse_file_mentions(text: str) -> tuple[str, list[Path]]:
...
+
+        # Check if we're in BASH mode (input starts with !)
+        try:
+            session = session_ref.get('session')
+            if session and hasattr(session, 'default_buffer'):
+                current_text = session.default_buffer.text
+                if current_text.startswith("!"):
+                    parts.append(("bg:#ff1493 fg:#ffffff bold", " BASH MODE "))
+                    parts.append(("", " | "))
+        except:
```
eyurtsev: This is a broad except, could you replace it with a specific except?

libs/deepagents-cli/deepagents_cli/input.py
```diff
@@ -150,15 +117,34 @@ def parse_file_mentions(text: str) -> tuple[str, list[Path]]:
     return text, files
 
 
-def get_bottom_toolbar(session_state: SessionState):
-    """Return toolbar function that shows auto-approve status."""
+def get_bottom_toolbar(session_state: SessionState, session_ref: dict):
+    """Return toolbar function that shows auto-approve status and BASH MODE."""
 
     def toolbar():
```
eyurtsev: annotate the return type if you can, especially when it's easy. `list[tuple[str, str]]`

libs/deepagents-cli/deepagents_cli/input.py
```diff
@@ -150,15 +117,34 @@ def parse_file_mentions(text: str) -> tuple[str, list[Path]]:
...
+        parts = []
+
+        # Check if we're in BASH mode (input starts with !)
+        try:
+            session = session_ref.get('session')
+            if session and hasattr(session, 'default_buffer'):
```
eyurtsev: Why is `hasattr` required?

libs/deepagents-cli/deepagents_cli/input.py
```diff
@@ -9,118 +9,85 @@
...
+        # Create temporary document for just the path fragment
+        temp_doc = Document(text=path_fragment, cursor_position=len(path_fragment))
+
+        # Get completions from PathCompleter and use its start_position
+        # PathCompleter returns suffix text with start_position=0 (insert at cursor)
+        for comp in self.path_completer.get_completions(temp_doc, complete_event):
+            # Add trailing / for directories so users can continue navigating
+            completed_path = Path(path_fragment + comp.text).expanduser()
+            completion_text = comp.text
+            if completed_path.is_dir() and not completion_text.endswith('/'):
```
eyurtsev: How is path delimiter handled on windows if we're using `endswith`? is `/` getting mapped to `\`?

libs/deepagents-cli/deepagents_cli/input.py
vtrivedy: removed, defauly_buffer always in PromptSession so unnecessary check.  every session has a default_buffer

libs/deepagents-cli/deepagents_cli/input.py
vtrivedy: done

libs/deepagents-cli/deepagents_cli/input.py
vtrivedy: done still pass

libs/deepagents-cli/deepagents_cli/input.py
vtrivedy: / works on windows as well afaik.

libs/deepagents-cli/deepagents_cli/input.py
vtrivedy: adjusted regex to support spaces 
"""
> @hi\ there\ yo.md

● I see the file...
"""

eyurtsev (approved):

PR #281, opened 2025-11-04
feat(acp): acp adapter (wip)

eyurtsev:
* Partial implementation of an ACP adapter
* Not expected to be complete for a while


mdrxy:
I think this is stale? https://github.com/langchain-ai/deepagents/tree/master/libs/acp

PR #283, opened 2025-11-04
feat: add provider selection and persistence

tode-rl:
Add ability to switch between OpenAI and Anthropic models during a session with the /model slash command. Model choice is persisted in config.json and restored in future sessions.

Changes:
- config.py: Add SessionState.preferred_provider field, refactor create_model() with force_provider parameter, add helper functions (_create_openai_model, _create_anthropic_model, _show_api_key_error), add load_agent_config() and save_agent_config() for persistence
- commands.py: Add /model command handler that returns dict signal for agent recreation, change return type to str | bool | dict
- main.py: Update simple_cli() and main() to return dict | None for special actions, add model switching loop in cli_main() that recreates agent on provider change, load config at startup and save on model switch
- README.md: Document /model command, add Model Switching section, update Agent Storage to mention config.json, update module documentation

Model switching recreates the agent and clears conversation history while preserving memories and custom instructions.


libs/deepagents-cli/deepagents_cli/config.py
```diff
@@ -98,9 +102,55 @@ def get_default_coding_instructions() -> str:
     return default_prompt_path.read_text()
 
 
-def create_model():
+def _create_openai_model():
+    """Create an OpenAI model instance."""
+    model_name = os.environ.get("OPENAI_MODEL", "gpt-5-mini")
```
eyurtsev: I don't want a lot of different env variables for specifying model preferences

Would `DEEPAGENTS_MODEL=`  work?

tode-rl:
I thought `/provider` made a bit more sense, `/model` can come later with local configs

vtrivedy:
hey thanks for the PR, we'll be using checkpointers for persistence instead, should be out soon!

PR #284, opened 2025-11-04
feat: add option to start cli in interactive mode with arg as prompt

tode-rl:
Add ability to pass a prompt as a command-line argument that executes on entering interactive mode.

Usage:
  deepagents "create hello.py"
  deepagents --agent mybot "review the last commit"
  deepagents --auto-approve "add a new unit test"

Changes:
- parse_args(): Detect if first positional arg is a prompt vs subcommand
- simple_cli(): Accept and execute initial_prompt parameter
- main(): Pass initial_prompt through
- cli_main(): Extract prompt from args and pass to main()
- ui.py: Update help text with [PROMPT] usage and example
- README.md: Document command-line prompt execution feature

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/README.md
+++ b/libs/deepagents-cli/deepagents_cli/README.md
@@ -28,7 +28,9 @@ cli/
   - `cli_main()` - Console script entry point (called when you run `deepagents`)
   - `main()` - Async main function that orchestrates agent creation and CLI
   - `simple_cli()` - Main interactive loop handling user input
+    - Executes optional initial prompt before entering interactive mode
   - `parse_args()` - Command-line argument parsing
+    - Accepts optional positional `prompt` argument for piping prompts
   - `check_cli_dependencies()` - Validates required packages are installed
 
 ### `config.py` - Configuration & Constants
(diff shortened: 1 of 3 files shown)


eyurtsev (approved):

vtrivedy:
looks good with added flags, but instead of entering interactive mode it should exit after it's done

tode-rl:
> looks good with added flags, but instead of entering interactive mode it should exit after it's done

I think it can do that but I believe it should have a separate flag for this behaviour. My thought is that a larger task would benefit from having the context in the cli and the ability to do follow-ups

for a usecase like `deepagents --auto-approve "Write a new unit test case"` some flag like `--script` could exit from the loop on completion? (I've seen `exec`, and `-p` used in other clis to toggle this behaviour)

tode-rl:
@vtrivedy Any thoughts on https://github.com/langchain-ai/deepagents/pull/284#issuecomment-3487811596? I can do in a follow-up if you're ok with adding another arg in the cli. I don't mind the cli exiting after complete but would like some control over it

vtrivedy:
agree on separate flag, we're leaning towards having the "exec" style only right now (shell exits) to bootstrap the run. if you want to make a new PR we can review it today

mdrxy:
Implemented in #755

PR #298, opened 2025-11-05
fix (cli-ui): remove brittle/dead code for summarization tracking

vtrivedy:
Now let LangChain’s SummarizationMiddleware recap flow through as a normal assistant message with no special UI.

Future suggestion: because the middleware emits a plain HumanMessage with no metadata to key off, any change to the hard-coded prefix will silently break UI customizations. We should _build_new_messages to tag that message additional_kwargs={"is_summary": True} which would allow prompt/prefix changes to be decoupled.

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/execution.py
+++ b/libs/deepagents-cli/deepagents_cli/execution.py
@@ -311,7 +311,7 @@ def execute_task(
                     message, metadata = data
 
                     if isinstance(message, HumanMessage):
-                        content = str(message.content).strip()
+                        content = message.text()
                         if content:
                             flush_text_buffer(final=True)
                             if spinner_active:


libs/deepagents-cli/deepagents_cli/execution.py
```diff
@@ -345,6 +310,21 @@ def flush_summary_buffer() -> None:
 
                     message, metadata = data
 
+                    if isinstance(message, HumanMessage):
+                        content = str(message.content).strip()
```
eyurtsev: I believe we can use the .text() attribute as of langchain 1 :)

eyurtsev: it'll handle situations in which the human message has multi-modal inputs (it'll only project the text contents)

vtrivedy: changed and nice :)

eyurtsev (approved):

PR #300, opened 2025-11-05
ctrl-c protection, add buffer to avoid accidentally exiting thread

vtrivedy:
- add time buffer on ctrl+c on prompt input box to avoid accidental exit, tells user can press twice to exit

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/input.py
+++ b/libs/deepagents-cli/deepagents_cli/input.py
@@ -159,7 +159,7 @@ def get_bottom_toolbar(
         parts.append((base_class, base_msg))
 
         # Show exit confirmation hint if active
-        hint_until = getattr(session_state, "exit_hint_until", None)
+        hint_until = session_state.exit_hint_until
         if hint_until is not None:
             now = time.monotonic()
             if now < hint_until:


libs/deepagents-cli/deepagents_cli/input.py
```diff
@@ -154,6 +158,16 @@ def toolbar() -> list[tuple[str, str]]:
 
         parts.append((base_class, base_msg))
 
+        # Show exit confirmation hint if active
+        hint_until = getattr(session_state, "exit_hint_until", None)
```
eyurtsev: `getattr` usage here is likely masking a bug in the code or incorrect initialization?

vtrivedy: removed, it's actually already initialized so we don't need to check

libs/deepagents-cli/deepagents_cli/input.py
```diff
@@ -168,6 +182,44 @@ def create_prompt_session(assistant_id: str, session_state: SessionState) -> Pro
...
+            app.exit(exception=KeyboardInterrupt())
+            return
+
+        session_state.exit_hint_until = now + EXIT_CONFIRM_WINDOW
+
+        handle = session_state.exit_hint_handle
+        if handle:
+            handle.cancel()
+
+        loop = asyncio.get_running_loop()
```
eyurtsev, line 207: Is there a way to do this without fetching the asyncio loop? I haven't read the code carefully, but this looks suspicious (looks like we're mixing sync and async code

eyurtsev (approved):

PR #314, opened 2025-11-06
Add simple sandbox execute

eyurtsev:
Add protocol extension to support shell execution

Daytona

```python
from daytona import CreateSandboxFromSnapshotParams, Daytona, DaytonaConfig
from daytona import Daytona
from deepagents_cli.integrations.daytona import DaytonaBackend

client = Daytona()

daytona_sandbox = client.create(
    params=CreateSandboxFromSnapshotParams(
        auto_stop_interval="15",
        auto_delete_interval="15",
    )
)

backend = DaytonaBackend(daytona_sandbox)
```

Runloop

```python
import os

client = Runloop(
    bearer_token=os.environ.get("RUNLOOP_API_KEY"),  # This is the default and can be omitted
)

devbox_view = client.devboxes.create_and_await_running()
print(devbox_view.id)

backend = RunloopBackend(devbox_id=devbox_view.id, client=client)
```

### DIFF
--- a/libs/deepagents/backends/sandbox.py
+++ b/libs/deepagents/backends/sandbox.py
@@ -6,8 +6,8 @@ only need to implement the execute() method.
 """
 
 from __future__ import annotations
-import json
 
+import json
 from abc import ABC, abstractmethod
 
 from deepagents.backends.protocol import (


libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,324 @@
...
+        output = result.output.rstrip()
+        exit_code = result.exit_code
+
+        if exit_code != 0 or "Error: File not found" in output:
+            return f"Error: File '{file_path}' not found"
+
+        return output
+
+    def write(
+        self,
```
vtrivedy, line 193: string replacing could be an issue.  should we use shlex for shell string ops?  like shlex.quote()

libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,324 @@
...
+from deepagents.backends.protocol import (
+    EditResult,
+    ExecuteResponse,
+    FileInfo,
+    GrepMatch,
+    SandboxBackendProtocol,
+    WriteResult,
+)
+
+_GLOB_COMMAND_TEMPLATE = """python3 -c "
```
vtrivedy, line 22: will hard coded python3 break across sandbox providers>

libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,324 @@
...
+        """Structured listing with file metadata."""
+        files = self.execute(f"ls -la '{path}' 2>/dev/null || true")
+
+        # Parse ls output - this is a simple implementation
+        # You might want to use find for more structured output
+        result: list[FileInfo] = []
+        for line in files.output.strip().split("\n"):
+            if not line or line.startswith("total"):
+                continue
+            parts = line.split()
```
vtrivedy, line 156: would break for files with spaces, "this file.txt"

libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,324 @@
...
+            file_path=file_path,
+            offset=offset,
+            limit=limit
+        )
+        result = self.execute(cmd)
+
+        output = result.output.rstrip()
+        exit_code = result.exit_code
+
+        if exit_code != 0 or "Error: File not found" in output:
```
vtrivedy, line 187: vary of string search because false positives

libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,324 @@
...
+
+        # Build grep command to get structured output
+        grep_opts = "-rHn"  # recursive, with filename, with line number
+
+        # Add glob pattern if specified
+        glob_pattern = ""
+        if glob:
+            glob_pattern = f"--include='{glob}'"
+
+        # Escape pattern for shell
```
vtrivedy, line 273: another string escape, can choose to standardize around shlex or another os utility

libs/deepagents/backends/sandbox.py
eyurtsev: Yeah it might break if some providers don't have python3 built in, but I think let's ship this for now and can revisit once it doesn't work. 

Providers don't have to subclass anyway

libs/deepagents/backends/sandbox.py
eyurtsev: Good catch!

libs/deepagents/backends/sandbox.py
eyurtsev: We can do json encoding on the payloads and do something a bit more structured. What do you think?

libs/deepagents/backends/sandbox.py
eyurtsev: Yeah we should use shlex here.

We'll need integration tests as well to go over all of these edge cases for each provider since I suspect there will be other issues that pop up!

eyurtsev:
Split into these PRs:

* https://github.com/langchain-ai/deepagents/pull/319
* https://github.com/langchain-ai/deepagents/pull/320

eyurtsev:
Closing this PR and will apply review comments on top of 319

PR #316, opened 2025-11-06
Fix mcp tools async 23b73

kilingzhang:


examples/research/requirements.txt
```diff
@@ -1,3 +1,6 @@
 -e ../../
 langgraph-cli[inmem]
 tavily-python
+tavily
+langchain-openai>=1.0.0,<2.0.0
+git+https://github.com/kilingzhang/langchain-mcp-adapters.git
```
eyurtsev, line 6: Why is this here?

libs/deepagents/middleware/tool_exception_handler.py
```diff
@@ -0,0 +1,92 @@
+"""Middleware to handle tool exceptions gracefully."""
```
eyurtsev, line 1: @kilingzhang -- we'd want to add this kind of middleware to langchain directly rather than deepagents. 

It would need to be done a bit differently and require parameterizastion by tool name and exception raised within it.

We don't want a blank catch of all exceptions since exceptions arise due to different reasons and the strategy for handling them is different.

For example, if it's a network connectivity issue, we want to retry the tool directly without using an LLM to do retry logic!

eyurtsev (changes requested):
* Middleware should go into langchain.
* OK to update example here, but requirements shouldn't use a different version langchain-mcp-adapters

PR #315, opened 2025-11-06
Add skills and dual-scope memory to deepagents CLI

rlancemartin:
# Add Skills System and Dual-Scope Memory Support

This updates the memory directory structure and adds `skills` to deepagent-cli based on [design doc](https://www.notion.so/Deepagents-CLI-Directory-Structure-2a7808527b1780c99ca7cf688e73b124?source=copy_link).

 ## Overview

 Three major enhancements:
1. **Skills System** - Per-agent skills 
2. **Global and local memory** - Automatic global + project-scoped `agent.md` loading
3. **Removed virtualization** - Use real filesystem paths 
 
### 1. Skills System

- Added global skills for any agent configuration
- Instructions for skill usage in README
- Added example skills
```
~/.deepagents/<agent_name>/
 ├── agent.md              # Global personality/style
 └── skills/                    # Per-agent skills
     ├── web-research/
     │   └── SKILL.md
     └── langgraph-docs/
         └── SKILL.md
```

### 2. Global and local memory

- In addition to global `agent.md` (shown above), we also will read from project
- All `agent.md` content injected into system prompt 
 ```
 my-project/
   ├── .git/
   └── .deepagents/
       └── agent.md
 ```
- Also improved rules for writing to either local or project `agent.md` file
```
 | Scope | Path | Purpose | Examples |
 |-------|------|---------|----------|
 | **Global** | `~/.deepagents/{agent}/agent.md` | Personality, style, universal behavior |
 Tone, coding style, general preferences |
 | **Project** | `[project]/.deepagents/agent.md` | Project-specific instructions |
 Architecture, conventions, testing |
```

### 3. Removed virtualization

1. Skills Need Real Paths: Bash scripts in skills (like arxiv_search.py) need to execute w/ absolute paths
2. Simpler Architecture: Direct filesystem access is more straightforward than path translation
3. Better Error Messages: Path errors are clearer without virtual→real translation
4. Consistent Across Modes: Both local and sandbox modes use routes={} for consistency

```python
  # OLD (virtualized):
  backend = CompositeBackend(
      default=FilesystemBackend(),
      routes={"/memories/": long_term_backend}  # Virtual path
  )
```

```
  # NEW (direct filesystem):
  backend = CompositeBackend(
      default=FilesystemBackend(),
      routes={}  # No virtualization
  )
```

Also added <env> tag and explicit path handling guidance to the system prompt. 

### Testing

run the CLI
```
[...]

PR #318, opened 2025-11-06
fix(cli): handle multiple concurrent interrupts in HITL workflow

rlancemartin:
## Summary

Fixes `RuntimeError` when parallel subagents create multiple pending interrupts in the human-in-the-loop (HITL) workflow.

## Problem

I was testing multiple sub-agents, which are used in the [web-research skill](https://github.com/langchain-ai/deepagents/pull/315). When spawning multiple subagents in parallel that require HITL approval (e.g., 3 research subagents all calling `web_search`), LangGraph raises:
```
RuntimeError: When there are multiple pending interrupts, you must specify the interrupt id when resuming.
```

The previous code only handled one interrupt at a time and passed a simple response dict. LangGraph requires responses for **all** pending interrupts to be mapped by their interrupt IDs when multiple are present.

## Solution

**Changes to `execution.py`:**

1. **Collect all pending interrupts** - Changed from tracking a single `hitl_request` to a dict `pending_interrupts` keyed by interrupt ID
2. **Handle each interrupt** - Loop through all interrupts and collect responses (auto-approve or manual prompt)
3. **Build proper resume format**:
   - Single interrupt: pass response directly `{"decisions": [...]}`
   - Multiple interrupts: pass dict `{interrupt_id_1: {"decisions": [...]}, interrupt_id_2: {...}}`

## Test Plan

Tested all w/ `web-research skill`.

- [x] Test with single subagent (HITL) - should work as before
- [x] Test with 3 parallel subagents + `--auto-approve` - should auto-approve all
- [x] Test with 3 parallel subagents + manual HITL - should prompt for each and resume correctly

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/execution.py
+++ b/libs/deepagents-cli/deepagents_cli/execution.py
@@ -545,7 +545,6 @@ async def execute_task(
                             raise
 
                 # Build response for all pending interrupts
-                all_responses: dict[str, HITLResponse] = {}
                 any_rejected = False
 
                 for interrupt_id, hitl_request in pending_interrupts.items():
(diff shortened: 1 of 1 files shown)


eyurtsev:
Pushed a rebase on top of master -- taking over the PR @rlancemartin

PR #320, opened 2025-11-06
feature: add sandbox integrations

eyurtsev:
Add protocol extension to support shell execution

Daytona

```python
from daytona import CreateSandboxFromSnapshotParams, Daytona, DaytonaConfig
from daytona import Daytona
from deepagents_cli.integrations.daytona import DaytonaBackend

client = Daytona()

daytona_sandbox = client.create(
    params=CreateSandboxFromSnapshotParams(
        auto_stop_interval="15",
        auto_delete_interval="15",
    )
)

backend = DaytonaBackend(daytona_sandbox)
```

Runloop

```python
import os

client = Runloop(
    bearer_token=os.environ.get("RUNLOOP_API_KEY"),  # This is the default and can be omitted
)

devbox_view = client.devboxes.create_and_await_running()
print(devbox_view.id)

backend = RunloopBackend(devbox_id=devbox_view.id, client=client)
```

### DIFF
--- a/libs/deepagents/backends/sandbox.py
+++ b/libs/deepagents/backends/sandbox.py
@@ -8,6 +8,8 @@ only need to implement the execute() method.
 from __future__ import annotations
 
 import json
+import shlex
+import base64
 from abc import ABC, abstractmethod
 
 from deepagents.backends.protocol import (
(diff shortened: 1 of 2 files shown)


libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,313 @@
+"""Base sandbox implementation with execute() as the only abstract method.
```
vtrivedy, line 1: general design decision of "everything is execute()" way of fs ops + bash in sandbox?  vs Provider specific file APIs (ex: from modal)

app = modal.App.lookup("sandbox-fs-demo", create_if_missing=True)
sb = modal.Sandbox.create(app=app)

with sb.open("test.txt", "w") as f:
    f.write("Hello World\n")

f = sb.open("test.txt", "rb")
print(f.read())
f.close()

libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,336 @@
...
+        # Create JSON payload with all parameters
+        payload = json.dumps({"pattern": pattern, "path": path})
+
+        # Python code that reads from stdin and performs glob
+        python_code = """import sys, json, glob, os
+data = json.load(sys.stdin)
+pattern = data['pattern']
+path = data['path']
+
+os.chdir(path)
```
vtrivedy, line 300: changing dir here doesn't seem right, shouldn't affect state of cwd

libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,336 @@
...
+    print(f'{line_num:6d}\\t{line_content}')"""
+
+        # Execute with JSON payload via stdin
+        cmd = f"echo {shlex.quote(payload)} | python3 -c {shlex.quote(python_code)} 2>&1"
+        result = self.execute(cmd)
+
+        output = result.output.rstrip()
+        exit_code = result.exit_code
+
+        if exit_code != 0 or "Error: File not found" in output:
```
vtrivedy, line 118: not sure we want to string match for this?  seems brittle

libs/deepagents/backends/sandbox.py
```diff
@@ -0,0 +1,336 @@
...
+                filename = " ".join(parts[8:])
+                is_dir = parts[0].startswith("d")
+                result.append(
+                    {
+                        "path": filename,
+                        "is_dir": is_dir,
+                    }
+                )
+
+        return result
```
vtrivedy: better off using os.scandir(), the parsing logic [:8] seems like it would break and os package has lots of good primitives here

libs/deepagents/backends/sandbox.py
eyurtsev: good catch

PR #332, opened 2025-11-07
fix(ci): use --test group only  to install test dependencies 

rlancemartin:
Resolve repeated test failures in https://github.com/langchain-ai/deepagents/pull/315 where responses (needed by test_tools.py) wasn't being installed during CI, causing:

```
ModuleNotFoundError: No module named 'responses'
```

image

Even though responses>=0.25.0 was listed in the test dependency group, the --group test --dev command wasn't successfully installing it in the CI environment. By using --all-groups, we ensure that every dependency group (including test) is fully resolved and installed, making the test suite work correctly.

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: "📦 Install Test Dependencies"
         shell: bash
-        run: uv sync --all-groups
+        run: uv sync --group test
 
       - name: "🧪 Run Unit Tests"
         shell: bash


.github/workflows/_test.yml
```diff
@@ -43,7 +43,7 @@ jobs:
 
       - name: "📦 Install Test Dependencies"
         shell: bash
-        run: uv sync --group test --dev
+        run: uv sync --all-groups
```
eyurtsev: The PR description isn't correct -- group test is the test dependencies.

I think CI is failing b/c the repsonses isn't in the test group probably?

Let's change this to just `uv sync --group test` so we don't have dep issues later if dev starts to include things a jupyter notbeook

eyurtsev:
Adjusted to just the test group and merging

PR #343, opened 2025-11-10
fix: Don't error when "subagent" does not exist. And add test for truncation logic

nhuang-lc:
- Don't raise an error when an incorrect subagent is used, just tell the agent
- Add test for truncation logic


vtrivedy:
could you run the lint and make format for the tests

PR #355, opened 2025-11-12
add summary_threshold for setting compact length

aotsukiqx:
Add threshold parameter for context compaction length in custom model usage scenarios


eyurtsev:
cc @ccurme

PR #380, opened 2025-11-12
feat(cli): Parallel Web Search integration

NormallyGaussian:
## Summary

Adds Parallel web search as an alternative to Tavily and exposes both as separate tools.

## Design Decision: Separate Tools

We expose `tavily_search()` and `parallel_search()` as **separate tools** rather than a unified interface because:

**Different Input Signatures:**
- Tavily: `query: str` - single search query
- Parallel: `queries: list[str]`, `objective: str | None` - multi-query with optional goal

**Different Output Formats:**
- Tavily:  Includes`content` field (string), relevance `score` and `query`
- Parallel: Includes `excerpts` field (array of markdown strings), `publish_date`, and `search_id`

**Different Use Cases:**
- Tavily: Single-query search with topic filtering ("general", "news", "finance")
- Parallel: Multi-query research optimized for LLM synthesis

Normalizing into one interface would lose functionality or create a confusing API.

## Implementation
- Conditionally registered based on API keys
- Tavily still takes priority when both keys are set
- Separate approval configs for each tool

## Testing

Added some tests; they all pass. I also tested manually with the Parallel API key enabled.

image

## Breaking Changes

None. Existing `web_search()` maintained as backward-compatible wrapper which still uses Tavily.


libs/deepagents-cli/deepagents_cli/main.py
```diff
@@ -135,7 +146,7 @@ async def simple_cli(
     console.print(DEEP_AGENTS_ASCII, style=f"bold {COLORS['primary']}")
     console.print()
 
-    if backend and isinstance(backend, SandboxBackendProtocol):
+    if backend and isinstance(backend, SandboxBackendProtocol) and hasattr(backend, "id"):
```
ghost: I think this is needed after another recent change to get the CLI to work

eyurtsev:
```suggestion
    if backend and isinstance(backend, SandboxBackendProtocol):
```

eyurtsev:
@NormallyGaussian we're holding off on additional integrations for now until we expose a proper way to configure the integrations. I'll keep the PR open and will ping here once ready

ghost:
> @NormallyGaussian we're holding off on additional integrations for now until we expose a proper way to configure the integrations. I'll keep the PR open and will ping here once ready

@eyurtsev I wanted to check in. Would now be a good time to reengage? Thanks!

mdrxy:
Hey @NormallyGaussian — thank you for your patience here, and sorry for leaving this in limbo for so long. We know Eugene mentioned we'd follow up once we had a better integration configuration story, but we've since reassessed and have decided to close this PR.

Some additional context: the CLI is likely entering maintenance mode, and going forward we'd want to have better scaffolding in place for how integrations get added before taking on new ones. This isn't a reflection of the quality of the work — the implementation is solid and the use case makes sense — it just doesn't fit where we're currently headed.

If you're open to it, we'd encourage you to open an issue to track interest in Parallel Search as an integration. That gives us a better signal on community demand, and if we see enough interest we'll revisit it as the project evolves.

Thanks again for the contribution and for checking back in — we appreciate it.

PR #393, opened 2025-11-13
Add custom base url from env

clem844:
Add custom base url for deepagents-cli

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/config.py
+++ b/libs/deepagents-cli/deepagents_cli/config.py
@@ -96,11 +96,19 @@ def create_model():
         from langchain_openai import ChatOpenAI
 
         model_name = os.environ.get("OPENAI_MODEL", "gpt-5-mini")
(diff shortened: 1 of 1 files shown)


flingjie:
> I don't want to introduce more env variables for this. We'll add a configuration file

In deepagents-cli, the `create_model` method was implemented in place of `init_chat_model` from langchain.chat_models. Could you share the key differences in the approach, and how do you see this change potentially improving flexibility, scalability, or performance in the future?

Dishwasha:
> I don't want to introduce more env variables for this. We'll add a configuration file

Will this be per-agent?  Would like to be able to use different models and/or providers for each agent.

Dishwasha:
@clem844 FYI looks like ChatOpenAI already supports OPENAI_BASE_URL and I have verified this works with deepagents-cli.

vtrivedy:
closing as fix supported above + avoiding more env vars

PR #396, opened 2025-11-13
feature(cli); plumb through dev server (experimental)

eyurtsev:
```bash
uv run deepagents-cli dev --allow-blocking
```


eyurtsev:
We'll need to re-open against new CLI when pick this up

PR #407, opened 2025-11-14
fix: fetch resumable shell from LC

sydney-runkle:
* use resumable shell middleware from langchain
* fix imports to not use private LC constructs

### DIFF
--- a/libs/deepagents/deepagents/middleware/resumable_shell.py
+++ b/libs/deepagents/deepagents/middleware/resumable_shell.py
@@ -17,4 +17,3 @@ class ResumableShellToolMiddleware(ShellToolMiddleware):
     actually active. This keeps behaviour identical for uninterrupted runs while
     allowing HITL pauses to succeed.
     """
-


libs/deepagents/deepagents/middleware/resumable_shell.py
```diff
@@ -1,20 +1,6 @@
...
-    _SessionResources,
-)
-from langchain.agents.middleware.types import AgentState
-from langchain.tools.tool_node import ToolCallRequest
-from langchain_core.messages import ToolMessage
-from langgraph.types import Command
+from langchain.agents.middleware.shell_tool import ShellToolMiddleware
 
 
 class ResumableShellToolMiddleware(ShellToolMiddleware):
```
sydney-runkle, line 6: kept around for backwards compat

eyurtsev:
I changed the implementation a bit just deleted from deepagents since it's not used in deepagents

PR #428, opened 2025-11-18
fix(subagents): pass runtime config to subagent invocations

rlancemartin:
Fixes LangSmith UUID v7 deprecation warning by ensuring subagents inherit the parent agent's runtime configuration (including thread_id and tracing context) instead of generating new trace IDs with uuid.uuid4().

Changes:
- Pass runtime.config to both subagent.invoke() and subagent.ainvoke()
- Ensures proper config inheritance for LangSmith tracing
- Prevents "LangSmith now uses UUID v7" deprecation warning


sydney-runkle:
I think this is probably not the right fix, seems like a deeper issue

sydney-runkle:
Done now

PR #429, opened 2025-11-19
Remove unnecessary dependencies from `deepagents` module

loganrosen:
The `deepagents` module currently has three dependencies that are not used by it but rather by the `deepagents-cli` module: `daytona`, `runloop-api-client`, and `tavily`. Since `deepagents-cli` already separately depends on these libraries, I've removed the dependencies from the `deepagents` module accordingly.


eyurtsev:
good catch thank you

eyurtsev:
Note to self: i verified the uv.lock files were unchanged after rerunning uv sync

PR #436, opened 2025-11-20
fix: resolve Pydantic serialization warning for context in SubAgent (issue #430)

### LINKED ISSUE #430: Pydantic Serialization Warning When Using Context Parameter with SubAgent Invocation
## Description

When using the `invoke()` method of a `CompiledStateGraph` created by `deepagents.create_deep_agent()` with a non-None `context` parameter, Pydantic emits a serialization warning **only when the agent invokes a subagent** (calls the `task` tool). Direct responses without subagent invocation do not trigger this warning.

## Environment

- **deepagents**: 0.2.7
- **langgraph**: 1.0.3
- **pydantic**: 2.12.4
- **Python**: 3.13

## Steps to Reproduce

```python
from dataclasses import dataclass
from typing import Optional
from langgraph._internal._typing import DataclassLike

@dataclass
class Context(DataclassLike):
    """Agent context configuration"""
    key: Optional[str] = None

# Create agent with context_schema
agent = create_deep_agent(
    model="gpt-4o",  # or any OpenAI model
    context_schema=Context,
    system_prompt="Conduct research and write a polished report.",
)

# Test 1: Simple greeting that gets answered directly (no warning)
print("Test 1: Direct response")
result1 = agent.invoke(
    {"messages": [{"role": "user", "content": "hi"}]},
    context=Context(key="12312")
)
print(f"Messages count: {len(result1['messages'])}\n")

# Test 2: Chinese greeting that triggers subagent call (warning appears!)
print("Test 2: Triggers subagent")
result2 = agent.invoke(
    {"messages": [{"role": "user", "content": "你好"}]},
    context=Context(key="12312")
)
print(f"Messages count: {len(result2['messages'])}")
```

## Actual Behavior

**Test 1 (Direct response)**: No warning, 2 messages in result
```
Test 1: Direct response
Messages count: 2
```

**Test 2 (Subagent invocation)**: Warning appears, 4 messages in result (includes tool call)
```
Test 2: Triggers subagent
Messages count: 4
/path/to/pydantic/main.py:464: UserWarning: Pydantic serializer warnings:
  PydanticSerializationUnexpectedValue(Expected `none` - serialized value may not be as expected [field_name='context', input_value=Context(key='12312'), input_type=Context])
  return self.__pydantic_serializer__.to_python(
```

## Expected Behavior

No warning should be emitted when passing a valid `Context` instance that matches the `context_schema` type, regardless of whether a subagent is invoked or not.

## Analysis

The warning only appears when:
1. The `context` parameter is provided with a non-None value
2. The agent decides to invoke the `task` tool (spawning a subagent)

When the agent responds directly without invoking subagents, no warning occurs.

This suggests the issue is related to how state (including `context`) is serialized and passed during subagent invocation, specifically in the `_return_command_with_state_update` function in `deepagents/middleware/subagents.py`.

The warning message indicates Pydantic expects the `context` field to be `None` (based on the default value), even though `Optional[Context]` explicitly allows non-None values. This appears to be a type definition or serialization configuration issue when creating the internal state schema.

## Impact

- **Functionality**: The code works correctly despite the warning
- **User Experience**: Confusing warning messages pollute logs
- **Debugging**: Makes it harder to identify real issues
- **Production**: Can cause concern in production environments with strict warning policies

## Additional Notes

- The warning is non-deterministic from a user's perspective - it depends on the agent's internal decision to use subagents
- Simple queries that the agent can answer directly don't trigger the warning
- More complex queries or certain input patterns (like Chinese greetings) that cause subagent invocation will consistently trigger the warning

## Related Issues

This may be related to other context handling issues in LangGraph 1.0, such as #6342 which discusses problems with using `context` and `config` together.

sinboun:
This PR fixes a PydanticSerializationUnexpectedValue warning observed when running subagents.

Issue: The runtime argument in task and atask functions lacked the generic type annotation [ContextT]. This caused the Pydantic serializer to expect None for the context field, resulting in a warning when a valid Context object was passed.

Changes:

Imported ContextT from langchain.tools.tool_node.

Updated type hints from ToolRuntime to ToolRuntime[ContextT].

Error Log Resolved:

PydanticSerializationUnexpectedValue(Expected none - serialized value may not be as expected [field_name='context', input_value=Context(...)])

## Linked Issue  
Closes #430


eyurtsev (approved):

sinboun (commented):
good

sinboun:
This is ready to merge!  @eyurtsev

sydney-runkle:
Thanks for the PR. I'm going to fix at a lower level in langchain instead.

warrenseine:
Hey @sydney-runkle, is there any PR I could watch to follow the progress? I've bumped all my LangGraph stack and still observe the same warning, so I assume this is not yet resolved?

warrenseine:
Hello @sydney-runkle, just raising it again in case this slipped through! I've confirmed that this PR does fix the warning locally, so maybe it's worth merging as is even though a deeper fix can be implemented?

PR #456, opened 2025-11-22
Harrison/add gemini support

hwchase17:


libs/deepagents-cli/deepagents_cli/main.py
```diff
@@ -373,6 +365,11 @@ async def main(
 
 def cli_main() -> None:
     """Entry point for console script."""
+    # Fix for gRPC fork issue on macOS
+    # https://github.com/grpc/grpc/issues/37642
+    if sys.platform == "darwin":
```
eyurtsev, line 370: This should be be happening in config.py -- we want all configuration logic to take place there since there are multiple entrypoints

eyurtsev: Feel free to merge as is i can move

PR #478, opened 2025-11-24
Harrison/fix diffing

hwchase17:


libs/deepagents-cli/deepagents_cli/file_ops.py
```diff
@@ -278,7 +292,20 @@ def update_args(self, tool_call_id: str, args: dict[str, Any]) -> None:
...
+                        responses = self.backend.download_files([path_str])
+                        if (
+                            responses
+                            and responses[0].content is not None
+                            and responses[0].error is None
+                        ):
+                            record.before_content = responses[0].content.decode("utf-8")
+                        else:
+                            record.before_content = ""
+                    except Exception:
```
eyurtsev, line 306: we can just let it crash, no need to catch broad exceptions

PR #483, opened 2025-11-24
test make format

hwchase17:
this is ai generated

what i want to do:

- control what dir its run in
- control interrupts


eyurtsev:
Run with `make test_integration`

PR #484, opened 2025-11-24
feat(deepagents): use context window fractions for SummarizationMiddleware when available

ccurme:


libs/deepagents/deepagents/graph.py
```diff
@@ -98,6 +98,18 @@ def create_deep_agent(
     if model is None:
         model = get_default_model()
 
+    if (
+        model.profile is not None
+        and isinstance(model.profile, dict)
+        and "max_input_tokens" in model.profile
+        and isinstance(model.profile["max_input_tokens"], int)
+    ):
+        trigger = ("fraction", 0.85)
```
ccurme, line 107: 0.85 was chosen to keep the threshold token count (170000) unchanged for the default model.

PR #499, opened 2025-11-26
fix: improve write_file tool description for better LLM clarity

### LINKED ISSUE #416: write_file tool => no clear description on content
I've found that at least for Claude Haiku 4.5, the model will call the `write_file` tool many times in a row without providing the `content` parameter (3-8 times incorrectly before finally providing `content`). This issue might be alleviated by refining the tool description, currently in which `content` is never explicitly described as the content of the file:

```python
WRITE_FILE_TOOL_DESCRIPTION = """Writes to a new file in the filesystem.

Usage:
- The file_path parameter must be an absolute path, not a relative path
- The content parameter must be a string
- The write_file tool will create the a new file.
- Prefer to edit existing files over creating new ones when possible."""
```

While the purpose of the `content` parameter might seem obvious, it could be interpreted by a naive LLM as "content type" parameter and thus not seen as a truly necessary parameter. 

Also, the LLM might be reluctant to write out such a long parameter value to a tool; in my case, the content is >60k tokens. Anecdotally, the reluctancy to provide the `content` parameter seems to increase as the content size increases (e.g., 8-10 failed tool calls if the `content` token count is >100k tokens).

Ghazi-raad:
## Description
Improves the `write_file` tool description to explicitly clarify what the `content` parameter represents.

## Problem
The current description is ambiguous, leading to LLMs (especially Claude models) failing 3-10 times before correctly providing the content parameter. Multiple community members have reported this issue, particularly with larger content sizes (>60k tokens).

Related community reports:
- @quinlanjager: "I see this a lot too with Claude Sonnet 4.5"
- @WenRichard: "I meet many times when I use claude-4.5"

## Solution
- Explicitly states that `content` contains the complete file contents
- Adds clarifying language: "(this is the actual text/code/data that will be written to the file)"
- Provides concrete examples showing proper usage
- Clarifies overwrite behavior
- Fixes typo: "create the a new file" → "create a new file"

## Impact
- Reduces failed tool call attempts
- Improves user experience and reduces API costs
- Fully backward compatible (documentation-only change)
- Particularly helps with large file creation (>60k tokens)

## Testing
- Existing tests pass
- Backward compatible (documentation-only change)
- No breaking changes
- No functional code changes

Fixes #416

### DIFF
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -201,7 +201,7 @@ Usage:
 - If a file already exists at the path, it will be overwritten
 - Prefer to edit existing files over creating new ones when possible
 
-Example:
+Examples:
 - To create a Python file: write_file(file_path="/app/main.py", content="def hello():\\n    print('Hello, World!')")
 - To create a text file: write_file(file_path="/data/notes.txt", content="Meeting notes from today...")"""
 


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -196,9 +196,14 @@ class FilesystemState(AgentState):
 
 Usage:
 - The file_path parameter must be an absolute path, not a relative path
-- The content parameter must be a string
-- The write_file tool will create the a new file.
-- Prefer to edit existing files over creating new ones when possible."""
+- The content parameter contains the complete file contents as a string (this is the actual text/code/data that will be written to the file)
+- The write_file tool will create a new file at the specified path
+- If a file already exists at the path, it will be overwritten
```
eyurtsev, line 201: at the moment it's not overwritten. it errors (i.e., overwrites are not allowed)

williamchen433:
I have tried this solution in 'claude-sonnet-4-5-20250929', But it does't work.

vtrivedy:
closing, overwrites not allowed in w/ write_file

PR #503, opened 2025-11-26
re-implement base sandbox without relying on python3

eyurtsev:
(not intended for merging) -- just testing whether we should do this. terminal-bench containers don't have python3 so issues w/ some of the file ops


eyurtsev:
The python3 implementation is pretty reasonable and python is fairly easy to install in sandboxes. Users have always the option to provide their own implementation.

PR #508, opened 2025-11-27
chore: update OpenAi default model to GPT-5.2

Chesars:
## Summary
  - Update default OpenAI model from GPT-5-mini to GPT-5.2

  ## Notes
  - Users can use the other models via environment variables:
    - `ANTHROPIC_MODEL=claude-opus-4-5-20251101`
    - `OPENAI_MODEL=gpt-5-mini`

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/config.py
+++ b/libs/deepagents-cli/deepagents_cli/config.py
@@ -374,7 +374,7 @@ def create_model() -> BaseChatModel:
     if settings.has_openai:
         from langchain_openai import ChatOpenAI
 
-        model_name = os.environ.get("OPENAI_MODEL", "gpt-5.1")
+        model_name = os.environ.get("OPENAI_MODEL", "gpt-5.2")
         console.print(f"[dim]Using OpenAI model: {model_name}[/dim]")
         return ChatOpenAI(
             model=model_name,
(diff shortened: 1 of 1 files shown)


libs/deepagents-cli/deepagents_cli/config.py
```diff
@@ -374,15 +374,15 @@ def create_model() -> BaseChatModel:
...
+        model_name = os.environ.get("OPENAI_MODEL", "gpt-5.1")
         console.print(f"[dim]Using OpenAI model: {model_name}[/dim]")
         return ChatOpenAI(
             model=model_name,
         )
     if settings.has_anthropic:
         from langchain_anthropic import ChatAnthropic
 
-        model_name = os.environ.get("ANTHROPIC_MODEL", "claude-sonnet-4-5-20250929")
+        model_name = os.environ.get("ANTHROPIC_MODEL", "claude-opus-4-5-20251101")
```
mdrxy: I don't think we want this as it is heavy and more expensive

Chesars: Opus 4.5 on SWE-bench Verified with Live-SWE-agent is ~26% more expensive per task than Sonnet 4.5 for a ~5% relative higher resolve rate, so let’s keep Sonnet 4.5 as the default.

sydney-runkle:
We don't want to implement this change at the moment.

PR #509, opened 2025-11-27
A concept for adding subagent support to deepagents-cli

Dishwasha:


vtrivedy:
thanks for the PR, subagents are supported via the task tool in the cli

PR #520, opened 2025-12-01
fix(deepagents): lower threshold for summarization

ccurme:
We use `count_tokens_approximately` to measure context lengths in SummarizationMiddleware.

This appears to under-estimate Anthropic token counts by 22%.

Assuming Claude Sonnet 4.5, we are triggering summarization at 170,000 tokens, but this is actually more like 207,000 Anthropic tokens, which is above its max input size of 200,000.

Here we compensate by adjusting the trigger down to 70%, or 140,000 tokens as measured by `count_tokens_approximately`, which is about the desired 170,000 tokens.


ccurme:
Closing in favor of https://github.com/langchain-ai/langchain/pull/34161

PR #524, opened 2025-12-02
feat: allow overriding default middleware configuration in create_deep_agent

saiprasaad2002:
- Add override_default_middleware and custom_middleware parameters to create_deep_agent
- Preserve existing default middleware behavior when override is disabled
- Support fully custom middleware stacks (including custom models/params) while still wiring interrupt_on via HumanInTheLoopMiddleware

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -43,6 +43,8 @@ def create_deep_agent(
     *,
     system_prompt: str | None = None,
     middleware: Sequence[AgentMiddleware] = (),
+    override_default_middleware: bool = False,
+    custom_middleware: Sequence[AgentMiddleware] | None = None,
     subagents: list[SubAgent | CompiledSubAgent] | None = None,
     response_format: ResponseFormat | None = None,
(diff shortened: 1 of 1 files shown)


saiprasaad2002:
But what if I need to change the model for my middleware instead of keeping the same from my deep agent. Maybe I can prefer the summarization middleware to get gpt-5-mini with my deepagent having o3.

I was just exploring this repo for some days and kindly apologize me if i'm wrong here as this is my very first PR in my professional career. Still at the learning process..

vtrivedy:
> But what if I need to change the model for my middleware instead of keeping the same from my deep agent. Maybe I can prefer the summarization middleware to get gpt-5-mini with my deepagent having o3.
> 
> I was just exploring this repo for some days and kindly apologize me if i'm wrong here as this is my very first PR in my professional career. Still at the learning process..

this is supported in the summarization middleware via the model param:
```
SummarizationMiddleware(
            model="gpt-4o-mini",
            trigger=("tokens", 4000),
            keep=("messages", 20),
        ),
```

closing, as mentioned above create_agent can be used if you want to support these customizations

PR #555, opened 2025-12-10
feat: add ability to paste images in input

casparb:
add ability to paste images from clipboard into deepagents cli
- detect images from clipboard when user pastes (mac only)
- convert to base64 to send to llm
- display as [image 1], [image 2] etc 'tags' instead of raw base64 (I want to do the same thing for large code blocks that are pasted so they don't fill the view)
- backspace deletes entire image tag in one keypress and fully removes it from input
- supports pngpaste which is fast with an osascript fallback which is built in

and of course, adds unit tests for image utilities

to use pngpaste instead of the much slower osascript, users should run `brew install pngpaste`

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/image_utils.py
+++ b/libs/deepagents-cli/deepagents_cli/image_utils.py
@@ -56,7 +56,7 @@ def _get_macos_clipboard_image() -> ImageData | None:
     # Try pngpaste first (fast if installed)
     try:
         result = subprocess.run(
-            ["pngpaste", "-"],  # Output to stdout
+            ["pngpaste", "-"],
             capture_output=True,
             check=False,
             timeout=2,
@@ -68,7 +68,7 @@ def _get_macos_clipboard_image() -> ImageData | None:
                 base64_data = base64.b64encode(result.stdout).decode("utf-8")
                 return ImageData(
                     base64_data=base64_data,
-                    format="png",
+                    format="png",  # 'pngpaste -' always outputs PNG
                     placeholder="[image]",
                 )
             except Exception:


libs/deepagents-cli/deepagents_cli/image_utils.py
```diff
@@ -0,0 +1,209 @@
...
+            timeout=2,
+        )
+        if result.returncode == 0 and result.stdout:
+            # Successfully got PNG data
+            try:
+                Image.open(io.BytesIO(result.stdout))  # Validate it's a real image
+                base64_data = base64.b64encode(result.stdout).decode("utf-8")
+                return ImageData(
+                    base64_data=base64_data,
+                    format="png",
```
eyurtsev: How do you know it's a png? (does `pngpaste` reutnrs png always?`) (maybe add a sentence to explain what it does)

libs/deepagents-cli/deepagents_cli/execution.py
```diff
@@ -36,6 +37,34 @@
 _HITL_REQUEST_ADAPTER = TypeAdapter(HITLRequest)
 
 
+def _display_user_message_with_images(text: str) -> None:
+    """Display user message with image placeholders colored in magenta.
+
+    Args:
+        text: User message text potentially containing [image] or [image N] placeholders
+    """
+    import re
```
eyurtsev: Could you move these to global scope

libs/deepagents-cli/deepagents_cli/image_utils.py
casparb: When no output file / format is specified (or stdout) pngpaste always outputs PNG. added a comment.

Issue #559, opened 2025-12-10
read_file in CLI will exceed line offset often - this shouldn't error


nhuang-lc:
Instead of throwing error we should just return all of the lines

<img width="421" height="157" alt="Image" src="https://github.com/user-attachments/assets/64e9d2cb-ef7c-4bf4-8b2b-23e12f5af005" />

CrazyWillBear:
I'm working on this right now, will make a pull request.

eyurtsev:
fix merged

PR #567, opened 2025-12-14
Handle read_file offset exceeding file length by returning all lines (issue #559)

CrazyWillBear:
This is meant to address [issue #559](https://github.com/langchain-ai/deepagents/issues/559) and does so by updating the read_file behavior so that if the offset is larger than the number of lines in the file, it returns all lines starting from the first line instead of throwing an error. This is as-described in the referenced issue.

This is my first contribution to an open source project, so I appreciate any and all feedback.


CrazyWillBear:
Apologies, replaced with PR #568 with correct changes.

PR #588, opened 2025-12-18
relaxed version check for daytona to resolve daytona>=0.113.0's obsto…

Jaiaid:
Both daytona>=0.113.0 and deepagents==0.2.8 requires obstore package but daytona>=0.113.0 requires obstore>=0.7.0, <0.8.0 which is not available in Python 3.14.2 (has obstore==0.8.2)

To solve this we relax daytona's version number

Tested in Ubuntu 20.04, Python 3.14.2

### DIFF
--- a/libs/deepagents-cli/pyproject.toml
+++ b/libs/deepagents-cli/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
   "langchain-openai>=0.1.0",
   "tavily-python",
   "python-dotenv",
-  "daytona>=0.113.0",
+  "daytona",
   "modal>=0.65.0",
   "markdownify>=0.13.0",
   "langchain>=1.0.7",


eyurtsev:
This does not seem to resolve the issue: 

```
Resolved 146 packages in 1.68s
error: Distribution `obstore==0.7.3 @ registry+https://pypi.org/simple` can't be installed because it doesn't have a source distribution or wheel for the current platform

hint: You're using CPython 3.14 (`cp314`), but `obstore` (v0.7.3) only has wheels with the following Python ABI tags: `cp311`, `cp312`, `cp313`
```

eyurtsev:
I filed an issue w/ daytona to unblock

eyurtsev:
OK daytona released new version of client so this is no longer necessary.

PR #592, opened 2025-12-19
Comply with Anthropic Agent Skills spec

vtrivedy:
### Summary
Update skills system to comply with Anthropic Agent Skills specification.

### Problem
- Skills didn't enforce the Agent Skills spec (allowed invalid names, no length limits)
- Optional fields not parsed (license, compatibility, metadata, allowed-tools)
- System prompt used markdown instead of recommended XML format for Claude

### Changes

pyproject.toml: Added pyyaml>=6.0

load.py:
- Warn on invalid names but still load (backwards compatible)
- Added optional fields to SkillMetadata (license, compatibility, metadata, allowed_tools)
- Switched to yaml.safe_load() for proper nested YAML parsing

commands.py:
- Strict validation on create (reject invalid names)
- Updated template with optional field examples

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/skills/middleware.py
+++ b/libs/deepagents-cli/deepagents_cli/skills/middleware.py
@@ -52,40 +52,50 @@ class SkillsStateUpdate(TypedDict):
     """List of loaded skill metadata (name, description, path)."""
 
 
-# Skills System Documentation (per Agent Skills spec: https://agentskills.io/specification)
+# Skills System Documentation
 SKILLS_SYSTEM_PROMPT = """
 
 ## Skills System
 
 You have access to a skills library that provides specialized capabilities and domain knowledge.
 
+{skills_locations}
+
+**Available Skills:**
+
 {skills_list}
 
 **How to Use Skills (Progressive Disclosure):**
 
-Skills follow a progressive disclosure pattern - you see their name and description above,
-but read full instructions only when needed:
+Skills follow a **progressive disclosure** pattern - you know they exist (name + description above), but you only read the full instructions when needed:
 
-1. **Recognize when a skill applies**: Check if the user's task matches any skill description
-2. **Read full instructions**: Use `read_file` with the `<location>` path from the skill list
-3. **Follow the instructions**: SKILL.md contains workflows, best practices, and examples
-4. **Access supporting files**: Skills may include scripts, configs, or docs in the same dir
+1. **Recognize when a skill applies**: Check if the user's task matches any skill's description
+2. **Read the skill's full instructions**: The skill list above shows the exact path to use with read_file
+3. **Follow the skill's instructions**: SKILL.md contains step-by-step workflows, best practices, and examples
+4. **Access supporting files**: Skills may include Python scripts, configs, or reference docs - use absolute paths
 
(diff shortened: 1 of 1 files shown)


libs/deepagents-cli/deepagents_cli/skills/load.py
```diff
@@ -23,31 +23,53 @@
...
+    license: NotRequired[str | None]
+    """License name or reference to bundled license file."""
+
+    compatibility: NotRequired[str | None]
+    """Environment requirements (max 500 chars)."""
+
+    metadata: NotRequired[dict[str, str] | None]
+    """Arbitrary key-value mapping for additional metadata."""
+
+    allowed_tools: NotRequired[str | None]
```
eyurtsev, line 70: `allowed_tools: NotRequired[list[str]]`

libs/deepagents-cli/deepagents_cli/skills/load.py
```diff
@@ -23,31 +23,53 @@
...
     """Source of the skill ('user' or 'project')."""
 
+    # Optional fields per Agent Skills spec
+    license: NotRequired[str | None]
+    """License name or reference to bundled license file."""
+
+    compatibility: NotRequired[str | None]
+    """Environment requirements (max 500 chars)."""
+
+    metadata: NotRequired[dict[str, str] | None]
```
eyurtsev, line 67: metadata: NotRequired[dict[str, str]]

libs/deepagents-cli/deepagents_cli/skills/load.py
```diff
@@ -23,31 +23,53 @@
...
+    """Description of what the skill does (max 1024 chars)."""
 
     path: str
     """Path to the SKILL.md file."""
 
     source: str
     """Source of the skill ('user' or 'project')."""
 
+    # Optional fields per Agent Skills spec
+    license: NotRequired[str | None]
```
eyurtsev, line 61: If you use a NotRequired, the type shouldn't be an optional, it's instead a Maybe type.

i.e., you don't want Maybe[Optional[T]], but either Maybe[T] or Optional[T]

libs/deepagents-cli/deepagents_cli/skills/load.py
```diff
@@ -113,32 +164,65 @@ def _parse_skill_metadata(skill_md_path: Path, source: str) -> SkillMetadata | N
...
+            return None
+
+        if not isinstance(frontmatter_data, dict):
+            logger.warning("Skipping %s: frontmatter is not a mapping", skill_md_path)
+            return None
 
         # Validate required fields
-        if "name" not in metadata or "description" not in metadata:
+        name = frontmatter_data.get("name")
+        description = frontmatter_data.get("description")
```
eyurtsev, line 185: Is description required or could we relax this and rely on name only for the agent to know what it's doing?

vtrivedy: required in spec

libs/deepagents-cli/deepagents_cli/skills/load.py
vtrivedy: current follows spec

PR #593, opened 2025-12-19
Tool selector middleware

limaolin2017:
- Adds a new ToolSelectorMiddleware that selects a Top‑K subset of tools per model call while always keeping DeepAgents
    built‑ins. This reduces tool schema tokens without breaking “self‑rescue” capability.
  - Supports fallback selection, unindexed tool preservation, and an explicit empty‑tools mode via config.
  - Includes unit tests (now in tests/unit_tests/test_middleware.py) that exercise filtering behavior with
    create_deep_agent + a fake model.

  Key files

  - libs/deepagents/deepagents/middleware/tool_selector.py
  - libs/deepagents/deepagents/middleware/__init__.py
  - libs/deepagents/tests/unit_tests/test_middleware.py

  Testing

  - uv run pytest -q libs/deepagents/tests/unit_tests/test_middleware.py::TestToolSelectorMiddleware


eyurtsev (changes requested):
Thanks for the contribution, but I don't think we want to add this to the package. You can include this in your own code right now as custom middleware.

Issue #527, opened 2025-12-02
InvalidUpdateError When Calling Multiple Sub Agents in Parallel


dpresonate:
I am tying to use a deepagent to call multiple instances of a sub-agent in parallel. The sub agent is built with langchain's `create_agent` with a Pydantic Structured Response output. All of the sub agents run in parallel like they are supposed to, but when they all finish I get the following error. 
```
Error: At key 'structured_response': Can receive only one value per step. Use an Annotated key to handle multiple values.
For troubleshooting, visit: https://docs.langchain.com/oss/python/langgraph/errors/INVALID_CONCURRENT_GRAPH_UPDATE
Traceback (most recent call last):
  File "/Users/daniel.pinson/IdeaProjects/resonate-ai-agent/src/deep_agent_cli.py", line 133, in main
    await stream_response(graph, user_input, config_params, langfuse_client, session_id)
  File "/Users/daniel.pinson/IdeaProjects/resonate-ai-agent/src/deep_agent_cli.py", line 73, in stream_response
    await _do_stream()
  File "/Users/daniel.pinson/IdeaProjects/resonate-ai-agent/src/deep_agent_cli.py", line 48, in _do_stream
    async for chunk, _metadata in graph.astream(state, config=config, stream_mode="messages"):
  File "/Users/daniel.pinson/IdeaProjects/resonate-ai-agent/.venv/lib/python3.12/site-packages/langgraph/pregel/main.py", line 2971, in astream
    loop.after_tick()
  File "/Users/daniel.pinson/IdeaProjects/resonate-ai-agent/.venv/lib/python3.12/site-packages/langgraph/pregel/_loop.py", line 542, in after_tick
    self.updated_channels = apply_writes(
                            ^^^^^^^^^^^^^
  File "/Users/daniel.pinson/IdeaProjects/resonate-ai-agent/.venv/lib/python3.12/site-packages/langgraph/pregel/_algo.py", line 294, in apply_writes
    if channels[chan].update(vals) and next_version is not None:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/daniel.pinson/IdeaProjects/resonate-ai-agent/.venv/lib/python3.12/site-packages/langgraph/channels/last_value.py", line 64, in update
    raise InvalidUpdateError(msg)
langgraph.errors.InvalidUpdateError: At key 'structured_response': Can receive only one value per step. Use an Annotated key to handle multiple values.
[...]

PR #603, opened 2025-12-23
fix(deepagents): exclude structured response from state update

eyurtsev:
This change will resolve issue: https://github.com/langchain-ai/deepagents/issues/527

Confirmed that regardless of chosen strategy information about the structured response is already present in the last AIMessage from the sub-agent. So we're OK with the current solution.


eyurtsev:
Should resolve https://github.com/langchain-ai/deepagents/issues/527

PR #612, opened 2025-12-24
Improve create_deep_agent() to support SystemMessage type

### LINKED ISSUE #597: Support SystemMessage type for create_deep_agent(system_prompt=...) (align with langchain.agents.create_agent)
#### Summary

`langchain.agents.create_agent` now supports:

```python
system_prompt: str | SystemMessage | None = None
```

However, `deepagents.create_deep_agent` still exposes:

```python
system_prompt: str | None = None
```

This means DeepAgents cannot directly consume a `SystemMessage` (or a pipeline that already models system prompts as `SystemMessage`), even though the underlying `create_agent` does support it.

It would be very helpful if `create_deep_agent` could accept both `str` **and** `SystemMessage`, and internally adapt to the existing DeepAgents prompt composition logic.

---

#### Motivation

* In many codebases, system prompts are already modeled as **structured objects** (`SystemMessage` or custom prompt objects) rather than plain strings.
* `create_agent` now allows passing `SystemMessage` directly, which makes it easy to reuse these higher-level models when building agents.
* DeepAgents is positioned as a higher-level **Agent Harness** on top of LangGraph/LangChain. It would be natural if it did not “force” callers to downcast back to `str` while the lower-level `create_agent` already supports `SystemMessage`.

Concretely, in our project:

* We maintain our own `SystemPrompt` / `SystemMessage` abstraction (with sections like *role*, *constraints*, *style*, etc.).
* For `create_agent`, we can now pass a `SystemMessage` directly.
* For `create_deep_agent`, we currently need a separate `to_legacy_system_prompt_str()` adapter solely to satisfy the `str` type, even though downstream everything ends up as messages anyway.

This makes DeepAgents feel “less typed” than the underlying `create_agent`, and it complicates sharing a common prompt model between DeepAgents and non-DeepAgents agents.

---

#### Current behavior

```python
from langchain_core.messages import SystemMessage

sys_msg = SystemMessage(content="You are a deep research assistant...")

# ❌ Type mismatch: create_deep_agent only accepts `str | None`
agent = create_deep_agent(
    model="openai:gpt-4o",
    tools=[...],
    system_prompt=sys_msg,   # not allowed by the type signature
)
```

We instead have to do:

```python
agent = create_deep_agent(
    model="openai:gpt-4o",
    tools=[...],
    system_prompt=sys_msg.content,  # manual flattening
)
```

or:

```python
system_prompt = my_system_prompt.to_legacy_system_prompt_str()

agent = create_deep_agent(
    model="openai:gpt-4o",
    tools=[...],
    system_prompt=system_prompt,
)
```

---

#### Desired behavior

Ideally, `create_deep_agent` would have a signature like:

```python
from langchain_core.messages import SystemMessage

def create_deep_agent(
    *,
    ...
) -> CompiledStateGraph[...]:
    ...
```

Semantics:

* If `system_prompt` is a `str`, keep the current behavior (prepend to the base DeepAgents system prompt).
* If `system_prompt` is a `SystemMessage`, internally convert it to a string (or to a single system message block) following the existing DeepAgents composition logic.
* If `None`, keep the current default behavior (use the built-in base prompt).

This keeps full backwards compatibility while allowing higher-level code to pass `SystemMessage` directly, just like `create_agent` does.

---

#### Alternatives considered

* **Keep a separate `to_legacy_system_prompt_str()` adapter on the user side**

  This works today, but:

  * It duplicates logic that `create_agent` already encapsulates for `SystemMessage`.
  * It forces DeepAgents users to “downgrade” to `str` even if their internal representation is already `SystemMessage`.

* **Bypass `create_deep_agent` and manually compose middlewares + `create_agent`**

  Technically possible, but it defeats the purpose of DeepAgents as a higher-level harness:

  * We would have to reimplement parts of DeepAgents behavior (planning, filesystem, subagents, self-management) just to support `SystemMessage` at the top level.
  * This increases complexity and fragmentation across projects.

---

#### Backwards compatibility

* The change is **fully backward compatible** if implemented as `str | SystemMessage | None`:

  * Existing code passing `str` continues to work.
  * New code can opt in to pass `SystemMessage`.
* Internally, DeepAgents can still normalize everything to a single system prompt string if that’s what the current infrastructure expects.

---

#### Additional context

* `langchain.agents.create_agent` already supports `SystemMessage` for `system_prompt`, so aligning DeepAgents with this behavior would make the ecosystem **more consistent** and easier to reason about.
* Many of the emerging patterns around **provider-specific system prompts** (e.g. Anthropic multi-block system prompts, cache control, etc.) are modeled in user code as `SystemMessage` or higher-level prompt objects. Being able to pass these directly into DeepAgents would reduce friction and boilerplate.

Thanks for considering this! Happy to adjust the proposal if there’s a better place in the API to expose `SystemMessage` support (e.g. a separate `system_message` parameter, a builder pattern, etc.).

RussellLuo:
Closes #597.

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -150,7 +150,7 @@ def create_deep_agent(
     # Convert system_prompt to SystemMessage if needed
     system_message: SystemMessage = SystemMessage(content=BASE_AGENT_PROMPT)
     if system_prompt is not None:
-        content = system_prompt.content if isinstance(system_prompt, SystemMessage) else system_prompt
+        content = system_prompt.text if isinstance(system_prompt, SystemMessage) else system_prompt
         system_message = SystemMessage(content=content + "\n\n" + BASE_AGENT_PROMPT)
 
     return create_agent(


alario-tang:
Thanks for the update — aligning `deepagents.create_deep_agent(system_prompt=...)` with `langchain.agents.create_agent` and accepting `SystemMessage` is definitely the right direction.

One concern: the current implementation assumes `SystemMessage.content` is always a `str` and does string concatenation (`content + "\n\n" + BASE_AGENT_PROMPT`). In our case we pass `SystemMessage(content=[...blocks...])` (i.e., block/list content), and this will either:

1. **TypeError** at runtime when attempting `list + str`, or
2. Force users to flatten structured system content into a plain string.

Also, by creating a new `SystemMessage(content=...)`, you may inadvertently **drop message metadata** (`additional_kwargs`, name/id, provider-specific fields, etc.) that can matter for some integrations.

**Suggestion:** handle both string and block/list content and preserve the original `SystemMessage` fields when possible. For example (rough sketch):

```python
from langchain_core.messages import SystemMessage

def _merge_system_prompt(system_prompt: str | SystemMessage | None) -> SystemMessage:
    base = BASE_AGENT_PROMPT

    if system_prompt is None:
        return SystemMessage(content=base)

    if isinstance(system_prompt, SystemMessage):
        content = system_prompt.content
        if isinstance(content, str):
            new_content = f"{content}\n\n{base}"
        elif isinstance(content, list):
            # Append base as an additional text block (block schema may vary by provider)
            new_content = content + [{"type": "text", "text": base}]
        else:
            # Fallback: coerce to string
            new_content = f"{str(content)}\n\n{base}"

        # Preserve other fields on the original SystemMessage
        return system_prompt.model_copy(update={"content": new_content})

    # system_prompt is a str
    return SystemMessage(content=f"{system_prompt}\n\n{base}")
```

It would also be great to add tests for:

* `system_prompt` as `str`
* `system_prompt` as `SystemMessage(content="...")`
* `system_prompt` as `SystemMessage(content=[...blocks...])` (ensuring it doesn’t crash and preserves blocks + metadata)

Happy to help tailor this to the exact “block” schema you’re targeting (Anthropic-style blocks vs other formats), but at minimum we need to avoid assuming `content` is always a string.

RussellLuo:
@alario-tang Thanks for the suggestions!

I didn't notice that [SystemMessage](https://reference.langchain.com/python/langchain/messages/#langchain.messages.SystemMessage)'s content could take more than just a string:

```python
SystemMessage(content: str | list[str | dict], **kwargs: Any)

SystemMessage(
    content: str | list[str | dict] | None = None,
    content_blocks: list[ContentBlock] | None = None,
    **kwargs: Any,
)

SystemMessage(
    content: str | list[str | dict] | None = None,
    content_blocks: list[ContentBlock] | None = None,
    **kwargs: Any,
)
```

I found that the metadata fields in `SystemMessage` actually come from [BaseMessage](https://reference.langchain.com/python/langchain_core/language_models/#langchain_core.messages.BaseMessage.content). But honestly, it feels like all those metadata fields, are mostly tailored for `AIMessage` and `HumanMessage` (i.e. the message created by the provider/model):

```python
class BaseMessage(Serializable):
    """Base abstract message class.

    Messages are the inputs and outputs of a chat model.

    Examples include [`HumanMessage`][langchain.messages.HumanMessage],
    [`AIMessage`][langchain.messages.AIMessage], and
    [`SystemMessage`][langchain.messages.SystemMessage].
    """

    content: str | list[str | dict]
    """The contents of the message."""

    additional_kwargs: dict = Field(default_factory=dict)
    """Reserved for additional payload data associated with the message.

    For example, for a message from an AI, this could include tool calls as
    encoded by the model provider.

    """

    response_metadata: dict = Field(default_factory=dict)
    """Examples: response headers, logprobs, token counts, model name."""

    type: str
    """The type of the message. Must be a string that is unique to the message type.

    The purpose of this field is to allow for easy identification of the message type
    when deserializing messages.

    """

    name: str | None = None
    """An optional name for the message.

    This can be used to provide a human-readable name for the message.

    Usage of this field is optional, and whether it's used or not is up to the
    model implementation.

    """

    id: str | None = Field(default=None, coerce_numbers_to_str=True)
    """An optional unique identifier for the message.

    This should ideally be provided by the provider/model which created the message.

    """

    model_config = ConfigDict(
        extra="allow",
    )
```

So I tweaked the PR a bit, using `SystemMessage.text` to grab the text content (kinda like how [ModelRequest.system_prompt](https://github.com/langchain-ai/langchain/blob/557eddfd51f7801bb54af773b19dcad4df0aee4b/libs/langchain_v1/langchain/agents/middleware/types.py#L157) works).

For the test cases, @ps06756 has kindly added them in another PR (#617), so I'm thinking we could team up to get this done.

alario-tang:
Thanks for the tweak! Using `SystemMessage.text` avoids the type issue, but it also flattens `SystemMessage(content=[...blocks...])` into a single string, which drops the original block boundaries.

In our usage, those boundaries are not accidental formatting — they’re part of the prompt’s structure. Many chat model adapters (and some providers) treat system content as a sequence of blocks/segments, and the segmentation can affect how the model “chunks” and prioritizes instructions. When everything is collapsed into one string:

- The model no longer sees distinct segments; it sees one continuous instruction.
- Separation cues that were explicit in the block structure become implicit (at best) and may be interpreted differently.
- Ordering and grouping effects can change: what was intended as “two separate directives” can become a single blended directive, which can alter compliance or emphasis.
- Even if you insert `\n\n` between pieces, that’s only a textual hint; it’s not the same as preserving structured boundaries that the adapter/provider might pass through as separate blocks.

So the main concern is: `.text` makes it robust against type errors, but it converts a structured system prompt into an unstructured one, and that loss of block boundaries can lead to different LLM behavior compared to keeping the blocks intact.

RussellLuo:
@alario-tang Thanks for the breakdown! Guess you're right, I've noticed Claude Code's system prompt is structured in multiple blocks too.

If crafting system prompts is such a big deal and users wanna have fine control, then personally, I think the simplest way is to just use whatever system prompt the user provides without tacking on the `BASE_AGENT_PROMPT`. Kinda like how it's handled in [`langchain.agents.create_agent()`](https://github.com/langchain-ai/langchain/blob/557eddfd51f7801bb54af773b19dcad4df0aee4b/libs/langchain_v1/langchain/agents/factory.py#L688-L694):

```python
# Convert system_prompt to SystemMessage if needed
system_message: SystemMessage | None = None
if system_prompt is not None:
    if isinstance(system_prompt, SystemMessage):
        system_message = system_prompt
    else:
        system_message = SystemMessage(content=system_prompt)
```

I'd like to hear the official take (like from @eyurtsev). If this is the suggested approach, I can tweak my PR.

aaron-seq:
## Code Review

Thank you for working on this PR to add SystemMessage support to `create_deep_agent()`. This is an important enhancement that aligns DeepAgents with the upstream LangChain API. However, there are several concerns with the current implementation that need to be addressed.

### Critical Issues

**1. Data Loss with SystemMessage.text Property**

The current implementation uses `system_prompt.text` which flattens structured content. As alario-tang correctly points out, this causes:

- Loss of block boundaries when `SystemMessage.content` is a list of blocks (e.g., Anthropic-style multi-block system prompts)
- Potential loss of metadata fields (`additional_kwargs`, `name`, `id`, `response_metadata`)
- Inability to preserve provider-specific fields like cache control or structured content

When users pass `SystemMessage(content=[{"type": "text", "text": "...", "cache_control": {...}}])`, the `.text` property converts this to a plain string, losing all structure.

**2. Type Error Risk**

Line 153 attempts string concatenation:
```python
content = system_prompt.text if isinstance(system_prompt, SystemMessage) else system_prompt
system_message = SystemMessage(content=content + "\n\n" + BASE_AGENT_PROMPT)
```

If `system_prompt.text` returns anything other than a string (edge case), this will fail at runtime.

### Recommendations

**Option 1: Preserve SystemMessage Structure (Recommended)**

Follow alario-tang's suggestion to handle both string and structured content:

```python
def _merge_system_prompt(system_prompt: str | SystemMessage | None) -> SystemMessage:
    base = BASE_AGENT_PROMPT
    if system_prompt is None:
        return SystemMessage(content=base)
    
    if isinstance(system_prompt, SystemMessage):
        content = system_prompt.content
        if isinstance(content, str):
            new_content = f"{content}\n\n{base}"
        elif isinstance(content, list):
            # Append base as additional text block
            new_content = content + [{"type": "text", "text": base}]
        else:
            # Fallback: coerce to string
            new_content = f"{str(content)}\n\n{base}"
        # Preserve metadata
        return system_prompt.model_copy(update={"content": new_content})
    
    # system_prompt is str
    return SystemMessage(content=f"{system_prompt}\n\n{base}")
```

This approach:
- Handles string content (backward compatible)
- Preserves block structure for list content
- Maintains all SystemMessage metadata via `model_copy`
- Follows the pattern suggested by maintainer alario-tang

**Option 2: Simpler Alignment with langchain.agents.create_agent**

As RussellLuo mentioned, `langchain.agents.create_agent` doesn't merge system prompts with a base prompt. Consider whether DeepAgents should:
- Let users handle the full system prompt themselves (including `BASE_AGENT_PROMPT`)
- Only concatenate when `system_prompt` is a string
- Keep SystemMessage objects as-is without modification

This would match the pattern in langchain and require users to include `BASE_AGENT_PROMPT` in their SystemMessage if needed.

### Additional Considerations

**Type Signature**: The type signature change is correct:
```python
```

**Backward Compatibility**: Maintained for string inputs, which is good.

**Testing**: The related PR #617 adds unit tests, which should cover:
- `str` input
- `SystemMessage(content="...")` input  
- `SystemMessage(content=[...blocks...])` input
- Metadata preservation

### Alignment with Issue #597

The issue explicitly requests supporting SystemMessage to avoid forcing users to downcast to string. The current implementation technically accepts SystemMessage, but by using `.text`, it defeats the purpose since structured content is flattened.

### Recommendation Summary

I suggest implementing Option 1 to properly handle structured SystemMessage content while preserving metadata. This addresses the concerns raised by alario-tang and fully satisfies the requirements in issue #597. The implementation should delegate the merging logic to a helper function and use `model_copy` to preserve SystemMessage fields.

Would appreciate hearing from @eyurtsev or other maintainers on the preferred approach for handling BASE_AGENT_PROMPT merging with SystemMessage objects.

sydney-runkle:
We need some more robust testing here + support across all of the other middlewares. Going to implement via https://github.com/langchain-ai/deepagents/pull/803.

Thanks for the initiative here! Happy to help get this feature across the line.

PR #638, opened 2025-12-30
feat(cli): Add Session Persistence to CLI

vtrivedy:
# Session Persistence via LangGraph Checkpoints

Conversations are persisted using LangGraph's built-in SQLite checkpointing. Each thread is identified by an 8-character hex ID (e.g., `a1b2c3d4`).

**Database:** `~/.deepagents/sessions.db`

LangGraph manages the schema automatically:
* `checkpoints` table stores conversation state
* `metadata` column (JSON) contains `agent_name` and `updated_at`

## Starting a New Session
```
deepagents --agent mybot
```
1. New 8-char thread ID generated
2. LangGraph stores checkpoints with metadata: `{"agent_name": "mybot", "updated_at": "..."}`

## Listing Threads

```
deepagents threads list
```
Shows a table with Thread ID, Agent, and Last Used timestamp.

## Resuming
```
deepagents --resume                    # Resume most recent
deepagents --resume a1b2c3d4           # Resume specific thread
```

## The `/clear` Command

1. Generates a new thread ID
2. Old thread remains in DB and is resumable
3. New thread starts with fresh state

### DIFF
--- a/libs/deepagents-cli/uv.lock
+++ b/libs/deepagents-cli/uv.lock
@@ -702,7 +702,7 @@ dependencies = [
 requires-dist = [
     { name = "langchain", specifier = ">=1.1.0,<2.0.0" },
     { name = "langchain-anthropic", specifier = ">=1.2.0,<2.0.0" },
-    { name = "langchain-core", specifier = ">=1.1.0,<2.0.0" },
+    { name = "langchain-core", specifier = ">=1.2.5,<2.0.0" },
     { name = "langchain-google-genai" },
     { name = "wcmatch" },
 ]
@@ -1434,7 +1434,7 @@ wheels = [
 
 [[package]]
 name = "langchain-core"
-version = "1.1.0"
+version = "1.2.6"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "jsonpatch" },
@@ -1444,10 +1444,11 @@ dependencies = [
     { name = "pyyaml" },
     { name = "tenacity" },
     { name = "typing-extensions" },
+    { name = "uuid-utils" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/1e/17/67c1cc2ace919e2b02dd9d783154d7fb3f1495a4ef835d9cd163b7855ac2/langchain_core-1.1.0.tar.gz", hash = "sha256:2b76a82d427922c8bc51c08404af4fc2a29e9f161dfe2297cb05091e810201e7", size = 781995, upload-time = "2025-11-21T21:01:26.958Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/b9/ce/ba5ed5ea6df22965b2893c2ed28ebb456204962723d408904c4acfa5e942/langchain_core-1.2.6.tar.gz", hash = "sha256:b4e7841dd7f8690375aa07c54739178dc2c635147d475e0c2955bf82a1afa498", size = 833343, upload-time = "2026-01-02T21:35:44.749Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/71/1e/e129fc471a2d2a7b3804480a937b5ab9319cab9f4142624fcb115f925501/langchain_core-1.1.0-py3-none-any.whl", hash = "sha256:2c9f27dadc6d21ed4aa46506a37a56e6a7e2d2f9141922dc5c251ba921822ee6", size = 473752, upload-time = "2025-11-21T21:01:25.841Z" },
+    { url = "https://files.pythonhosted.org/packages/6f/40/0655892c245d8fbe6bca6d673ab5927e5c3ab7be143de40b52289a0663bc/langchain_core-1.2.6-py3-none-any.whl", hash = "sha256:aa6ed954b4b1f4504937fe75fdf674317027e9a91ba7a97558b0de3dc8004e34", size = 489096, upload-time = "2026-01-02T21:35:43.391Z" },
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


libs/deepagents-cli/deepagents_cli/commands.py
```diff
@@ -1,35 +1,52 @@
...
     if cmd in ["quit", "exit", "q"]:
         return "exit"
 
     if cmd == "clear":
-        # Reset agent conversation state
-        agent.checkpointer = InMemorySaver()
+        new_thread_id = None
+        if session_state is not None:
+            new_thread_id = generate_thread_id()
+            session_state.thread_id = new_thread_id
```
eyurtsev: This code is doing mutation in place -- which we should avoid as much as possible

libs/deepagents-cli/deepagents_cli/execution.py
```diff
@@ -218,9 +218,16 @@ async def execute_task(
     else:
         message_content = final_input
 
+    from datetime import UTC, datetime
```
eyurtsev: import should be global

libs/deepagents-cli/deepagents_cli/config.py
```diff
@@ -375,14 +374,33 @@ def ensure_project_skills_dir(self) -> Path | None:
...
-    """Holds mutable session state (auto-approve mode, etc)."""
+    """Holds mutable session state (auto-approve mode, thread info, etc)."""
+
+    def __init__(
+        self,
+        auto_approve: bool = False,
+        no_splash: bool = False,
+        *,
+        thread_id: str | None = None,
+        agent_name: str = "agent",
```
eyurtsev: Is there a place where we can look up the default value of "agent"

libs/deepagents-cli/deepagents_cli/main.py
```diff
@@ -311,6 +347,7 @@ async def _run_agent_session(
     sandbox_backend=None,
     sandbox_type: str | None = None,
     setup_script_path: str | None = None,
+    checkpointer=None,
```
eyurtsev: type information missing

libs/deepagents-cli/deepagents_cli/main.py
```diff
@@ -449,9 +500,65 @@ def cli_main() -> None:
             reset_agent(args.agent, args.source_agent)
         elif args.command == "skills":
             execute_skills_command(args)
+        elif args.command == "threads":
+            if args.threads_command == "list":
+                asyncio.run(
```
eyurtsev: can we match the rest of the pattern here?

This should do something like `execute_threads_command(threads)` (see execute_skills_command)

If not, is there anyway to factor out the details correctly? The output from these two branches should be `session_state`


```python
elif args.command == "threads":
   session_state = asyncio.run(...)
else:
  session_state = create_..
```

libs/deepagents-cli/deepagents_cli/main.py
```diff
@@ -27,6 +27,15 @@
     create_sandbox,
     get_default_working_dir,
 )
+from deepagents_cli.sessions import (
```
eyurtsev, line 37: This is a large surface area to expose in `main` Can the details stay in `sessions`? (e.g., take a look at `execute_skills_command` and `setup_skills_parser`)

libs/deepagents-cli/pyproject.toml
```diff
@@ -19,6 +19,8 @@ dependencies = [
   "langchain>=1.0.7",
   "runloop-api-client>=0.69.0",
   "pillow>=10.0.0",
+  "aiosqlite>=0.20.0,<0.22.0",
```
eyurtsev, line 22: could we remove the upper bound from aiosqlite

eyurtsev (commented):
Could we add extended unit tests for this functionality?

libs/deepagents-cli/pyproject.toml
vtrivedy: needed for a bug

libs/deepagents-cli/deepagents_cli/main.py
vtrivedy: yeah this is better

libs/deepagents-cli/deepagents_cli/commands.py
vtrivedy: Fixed - now returns `{"clear": True, "thread_id": new_thread_id}` and the caller in main.py handles the state update.

libs/deepagents-cli/deepagents_cli/config.py
vtrivedy: Added DEFAULT_AGENT_NAME = "agent" constant and updated

libs/deepagents-cli/deepagents_cli/main.py
vtrivedy: Added BaseCheckpointSaver | None type hint.

libs/deepagents-cli/deepagents_cli/main.py
vtrivedy: Moved parser setup to setup_threads_parser() in sessions module. Reduced imports to only what's needed

libs/deepagents-cli/deepagents_cli/sessions.py
```diff
@@ -0,0 +1,254 @@
...
+    db_dir.mkdir(parents=True, exist_ok=True)
+    return db_dir / "sessions.db"
+
+
+def generate_thread_id() -> str:
+    """Generate a new 8-char hex thread ID."""
+    return uuid.uuid4().hex[:8]
+
+
+async def list_threads(
```
eyurtsev, line 44: I suspect a bunch of these can be replaced w/ the checkpointer API.

For example, 

```    def delete_thread(self, thread_id: str) -> None:
        """Delete all checkpoints and writes associated with a thread ID.

        Args:
            thread_id: The thread ID to delete.

        Returns:
            None
        """
        try:
            # check if we are in the main thread, only bg threads can block
            # we don't check in other methods to avoid the overhead
            if asyncio.get_running_loop() is self.loop:
                raise asyncio.InvalidStateError(
                    "Synchronous calls to AsyncSqliteSaver are only allowed from a "
                    "different thread. From the main thread, use the async interface. "
                    "For example, use `checkpointer.alist(...)` or `await "
                    "graph.ainvoke(...)`."
                )
        except RuntimeError:
            pass
        return asyncio.run_coroutine_threadsafe(
            self.adelete_thread(thread_id), self.loop
        ).result()
```

delete_thread already exists

vtrivedy:
used this but differently in new CLI

PR #644, opened 2026-01-02
Create devcontainer.json

DOUGLASDAVIS08161978:


eyurtsev:
Thanks, the core maintainers don't use devcontainers and we've had issues in the past keeping these up to date

PR #646, opened 2026-01-03
feature(deepagents): add memory

vtrivedy:
## Summary

**MemoryMiddleware** (new)
- Loads persistent context from `AGENTS.md` files following the [agents.md](https://agents.md) spec
- Always-on injection into system prompt (unlike skills which are on-demand)
- Supports multiple sources (e.g., user-level + project-level)
- Content wrapped in XML tags (`<user_memory>`, `<project_memory>`)

### Memory (always-on context)

```python
memory = MemoryMiddleware(
    backend=FilesystemBackend(root_dir="/"),
    sources=[
        {"path": "~/.deepagents/AGENTS.md", "name": "user"},
        {"path": "./.deepagents/AGENTS.md", "name": "project"},
    ],
)

- Sources are loaded in order, combined into prompt
- Each wrapped in XML: <user_memory>, <project_memory>
- Missing files skipped silently
- File format: plain markdown

### DIFF
--- a/libs/deepagents/tests/unit_tests/middleware/test_memory_middleware.py
+++ b/libs/deepagents/tests/unit_tests/middleware/test_memory_middleware.py
@@ -6,6 +6,7 @@ and temporary directories with the FilesystemBackend in normal (non-virtual) mod
 
 from datetime import UTC, datetime
 from pathlib import Path
+from types import SimpleNamespace
 
 from langchain.agents import create_agent
 from langchain_core.messages import AIMessage, HumanMessage
@@ -35,6 +36,23 @@ def make_memory_content(title: str, content: str) -> str:
 
 {content}
 """
+
+
+def create_store_memory_item(content: str) -> dict:
+    """Create a memory item in StoreBackend FileData format.
+
+    Args:
+        content: Memory content string
+
+    Returns:
+        Dict with content (as list of lines), created_at, and modified_at
+    """
+    timestamp = datetime.now(UTC).isoformat()
+    return {
+        "content": content.split("\n"),
+        "created_at": timestamp,
+        "modified_at": timestamp,
+    }
 
 
 def test_format_memory_locations_empty() -> None:
@@ -543,8 +561,6 @@ def test_memory_middleware_with_state_backend_factory() -> None:
     assert middleware.sources[0] == "/memory/AGENTS.md"
 
     # Create a mock Runtime (simplified for testing)
-    from types import SimpleNamespace
-
     state = {"messages": [], "files": {}}
     runtime = SimpleNamespace(
         context=None,
@@ -570,8 +586,6 @@ def test_memory_middleware_with_store_backend_factory() -> None:
     assert callable(middleware._backend)
 
     # Create a mock Runtime with store
-    from types import SimpleNamespace
-
     store = InMemoryStore()
     state = {"messages": []}
     runtime = SimpleNamespace(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: libs/deepagents/tests/integration_tests/test_memory_integration.py [hunk 1]
```diff
@@ -0,0 +1,492 @@
+"""Comprehensive integration tests for MemoryMiddleware.
```
eyurtsev, line 1: When possible would be better to remove integration tests and instead think of them as extended unit tests.

I'd prefer thinking of unit tests as anything that we can run deterministically and doesn't require external network connections to hit an API.

If we're testing against some backends (e.g., file system) -- we can do that as a unit test.

libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -0,0 +1,321 @@
...
+
+
+class MemoryState(AgentState):
+    """State schema for MemoryMiddleware.
+
+    Attributes:
+        memory_contents: Dict mapping source names to their loaded content.
+    """
+
+    memory_contents: NotRequired[dict[str, str]]
```
eyurtsev: Could rely on empty dict to mark the fact that there are no memories?

```suggestion
    memory_contents: dict[str, str]
```

libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -0,0 +1,321 @@
...
+            # download_files returns a list of FileDownloadResponse objects
+            for response in results:
+                if response.path == path and response.content is not None:
+                    # Content may be bytes or string
+                    if isinstance(response.content, bytes):
+                        return response.content.decode("utf-8")
+                    if isinstance(response.content, list):
+                        return "\n".join(response.content)
+                    return str(response.content)
+        except Exception as e:
```
eyurtsev: Could we let it crash?

libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -0,0 +1,321 @@
...
+        memory_locations = self._format_memory_locations()
+        memory_contents = self._format_memory_contents(contents)
+
+        memory_section = self.system_prompt_template.format(
+            memory_locations=memory_locations,
+            memory_contents=memory_contents,
+        )
+
+        if request.system_prompt:
+            system_prompt = memory_section + "\n\n" + request.system_prompt
```
eyurtsev, line 341: should we be relying on system_message here?

libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -0,0 +1,321 @@
...
+            runtime: Runtime context for factory functions.
+
+        Returns:
+            Resolved backend instance.
+        """
+        if callable(self._backend):
+            return self._backend(runtime)
+        return self._backend
+
+    def _format_memory_locations(self) -> str:
```
eyurtsev, line 166: (nit): I'd probably inline this code into _format_memory_contents or into modifyRequest directly

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -0,0 +1,628 @@
...
+    license: str | None
+    """License name or reference to bundled license file."""
+
+    compatibility: str | None
+    """Environment requirements (max 500 chars)."""
+
+    metadata: dict[str, str]
+    """Arbitrary key-value mapping for additional metadata."""
+
+    allowed_tools: list[str]
```
eyurtsev, line 147: We don't do anything with allowed_tools atm right?

libs/deepagents/examples/skills_example.py
```diff
@@ -0,0 +1,244 @@
...
+                print(f"  Registry: {skill['registry']}")
+                if skill.get("allowed_tools"):
+                    print(f"  Allowed Tools: {skill['allowed_tools']}")
+
+        # Show what the system prompt would look like
+        print("\n" + "=" * 60)
+        print("SYSTEM PROMPT INJECTION")
+        print("=" * 60)
+
+        skills_locations = middleware._format_skills_locations()
```
eyurtsev: The demo is relying on private functions?

libs/deepagents/examples/memory_demo.py
```diff
@@ -0,0 +1,217 @@
+"""Demo showing exactly how MemoryMiddleware works.
```
eyurtsev: How do we expect users to use this? I would've expected an example file to give me code that I can copy and paste without running -- this looks more like a file that I would run to get some instruction about how to use this stuff?

libs/deepagents/examples/skills_example.py
```diff
@@ -0,0 +1,244 @@
+"""Example demonstrating the Skills Middleware.
```
eyurtsev: Is this meant to be run like a script?

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -0,0 +1,628 @@
...
+
+import yaml
+
+if TYPE_CHECKING:
+    from deepagents.backends.protocol import BackendProtocol
+
+from collections.abc import Awaitable, Callable
+
+# Type alias for backend or factory function
+BackendOrFactory = "BackendProtocol | Callable[[Runtime], BackendProtocol]"
```
eyurtsev: type alias should be after the imports

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -0,0 +1,628 @@
...
+        # Construct SKILL.md path using PurePosixPath for safe, standardized path operations
+        skill_dir = PurePosixPath(skill_dir_path)
+        skill_md_path = str(skill_dir / "SKILL.md")
+        skill_md_paths.append((skill_dir_path, skill_md_path))
+
+    # Batch download all SKILL.md files
+    paths_to_download = [skill_md_path for _, skill_md_path in skill_md_paths]
+    try:
+        responses = backend.download_files(paths_to_download)
+    except Exception as e:
```
eyurtsev: We should remove the try except here and let the code throw the exception

libs/deepagents/deepagents/middleware/__init__.py
```diff
@@ -1,11 +1,17 @@
...
 from deepagents.middleware.filesystem import FilesystemMiddleware
+from deepagents.middleware.memory import MemoryMiddleware, MemorySource
+from deepagents.middleware.skills import SkillRegistry, SkillsMiddleware
 from deepagents.middleware.subagents import CompiledSubAgent, SubAgent, SubAgentMiddleware
 
 __all__ = [
     "CompiledSubAgent",
     "FilesystemMiddleware",
+    "MemoryMiddleware",
+    "MemorySource",
```
eyurtsev: `MemorySource` and `SkillRegistry` have different suffixes now, but behave in the same way

libs/deepagents/examples/memory_demo.py
vtrivedy: let's remove all these in examples, was useful to test, users won't use - how to use will be in our docs

libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -0,0 +1,379 @@
...
+    AgentMiddleware,
+    AgentState,
+    ModelRequest,
+    ModelResponse,
+    PrivateStateAttr,
+)
+from langchain.tools import ToolRuntime
+from langgraph.runtime import Runtime
+
+logger = logging.getLogger(__name__)
```
eyurtsev, line 72: remove logger

vtrivedy:
all good to merge

vtrivedy:
will do a follow-up on fixing up/removing/adding integration tests
unit tests are good

PR #678, opened 2026-01-08
fix(deepagents): throw a clear exception message when 'messages' key missing from the output of the subagent

hwchase17:
Adds validation to throw a clear error when a CompiledSubAgent with a custom StateGraph is missing the required 'messages' key in its state schema. Previously, this resulted in an unhelpful KeyError.

### DIFF
--- a/libs/deepagents/tests/unit_tests/test_subagents.py
+++ b/libs/deepagents/tests/unit_tests/test_subagents.py
@@ -5,11 +5,15 @@ are invoked, how they return results, and how state is managed between parent
 and child agents.
 """
 
+from typing import TypedDict
+
+import pytest
 from langchain.agents import create_agent
 from langchain.agents.middleware import TodoListMiddleware
 from langchain.agents.structured_output import ToolStrategy
 from langchain_core.messages import AIMessage, HumanMessage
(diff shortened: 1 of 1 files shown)


khteh:
```
  File "/usr/src/Python/rag-agent/.venv/lib/python3.13/site-packages/deepagents/middleware/subagents.py", line 367, in atask
    return _return_command_with_state_update(result, runtime.tool_call_id)
  File "/usr/src/Python/rag-agent/.venv/lib/python3.13/site-packages/deepagents/middleware/subagents.py", line 320, in _return_command_with_state_update
    "messages": [ToolMessage(result["messages"][-1].text, tool_call_id=tool_call_id)],
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'text'
During task with name 'tools' and id 'c7f36538-8a5f-c1d7-4754-ebfbedabd5ee'
```

khteh:
OK. I have switched to self-hosted `Ollama` `gpt-oss` LLM model and it works perfectly.
```
messages: Annotated[Sequence[BaseMessage], add_messages]
```
does indeed solve the `AttributeError: 'str' object has no attribute 'text'` problem and the PR is valid.

PR #684, opened 2026-01-08
feat(sdk): expose `state_schema` parameter in `create_deep_agent`

lanadon:
Add support for passing a custom state_schema to create_deep_agent,
which is forwarded to the underlying langchain create_agent function.
This allows users to define custom state fields for their agents.

- Add state_schema parameter to create_deep_agent signature
- Pass state_schema through to create_agent call
- Add unit test to verify custom state fields appear in stream_channels

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -52,6 +52,7 @@ def create_deep_agent(
     skills: list[str] | None = None,
     memory: list[str] | None = None,
     response_format: ResponseFormat | None = None,
+    state_schema: type[Any] | None = None,
     context_schema: type[Any] | None = None,
     checkpointer: Checkpointer | None = None,
     store: BaseStore | None = None,
@@ -117,6 +118,7 @@ def create_deep_agent(
 
             Memory is loaded at agent startup and added into the system prompt.
         response_format: A structured output response format to use for the agent.
+        state_schema: Optional custom state schema for the agent (must extend AgentState).
         context_schema: The schema of the deep agent.
         checkpointer: Optional `Checkpointer` for persisting agent state between runs.
         store: Optional store for persistent storage (required if backend uses `StoreBackend`).
@@ -232,6 +234,7 @@ def create_deep_agent(
         tools=tools,
         middleware=deepagent_middleware,
         response_format=response_format,
+        state_schema=state_schema,
         context_schema=context_schema,
         checkpointer=checkpointer,
         store=store,
--- a/libs/deepagents/tests/unit_tests/test_end_to_end.py
+++ b/libs/deepagents/tests/unit_tests/test_end_to_end.py
@@ -6,7 +6,7 @@ from typing import Any
 from unittest.mock import patch
 
 import pytest
-from langchain.agents.middleware import AgentMiddleware
+from langchain.agents.middleware import AgentMiddleware, AgentState
 from langchain.agents.middleware.types import ModelRequest, ModelResponse
 from langchain.tools import ToolRuntime
 from langchain_core.language_models import LanguageModelInput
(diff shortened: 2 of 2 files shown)


eyurtsev:
Thanks! We'll discuss internally whether to add or keep as is.

We did not want to add it to `create_agent` originally and only added it to make **migration** from create_react_agent (the old version) to create_agent (current version) easier.

The reason we did not want to add it is that most state extension schema is associated with either tools that modify the state OR with nodes that modify the state both of which can be specified with middleware.

lanadon:
Thanks for the context @eyurtsev! I understand the concern about keeping the API surface minimal.

Let me explain our use case more concretely. We have read-only context that needs to be available throughout the agent's execution - things like organization identifiers, user identifiers, and other tenant-scoping information. These are not modified during execution - they're configuration/context that tools and middleware need to read to do their job (e.g., fetching data scoped to the right tenant, logging with proper context).


### The problem with using config.metadata
Currently, we pass these via config.metadata and then patch them into state in a middleware's before_model hook. This works, but has some downsides:
- No type safety: metadata is dict[str, Any], so we lose all typing. We have to manually parse UUIDs from strings with try/except blocks and handle type coercion ourselves.
- No validation: There's no schema validation that required fields are present or have the correct types at invocation time.
- The patching feels redundant: We define a proper typed state schema on our middleware, but then have to manually copy values from untyped metadata into typed state. This indirection adds complexity and potential for bugs.

### What we'd ideally want
With state_schema exposed, we could:
1. Define our context fields with proper types in the schema
2. Pass them directly in the graph input
3. Have them available in state throughout execution - typed and validated from the start

### Question
If exposing state_schema isn't the right approach, what's the recommended pattern for passing typed, read-only context that needs to be available throughout an agent's execution? Is there a cleaner way than metadata → middleware patching?

sydney-runkle:
Curious - what information are you storing in state? Are you using it in tools? Could use a bit of context here to decide if this is a compelling enough use case for another arg.

lanadon:
Thanks for the review @sydney-runkle! I think you will find most of the context you are looking for in my previous message answering eyurtsev.
Basically I am looking for a strongly typed way to inject read-only context that needs to be available throughout the agent's execution - things like organization identifiers, user identifiers, and other tenant-scoping information. Before migrating my agents to deep-agents I was using custom LangGraph agents and having a custom state was the perfect way to do so.

This read-only context is then used in tools, middlewares and subagents.

I guess this could be a common pattern if merged.

If not I would be keen on having your feedback on the question at the end of my previous message :)

sydney-runkle:
> Basically I am looking for a strongly typed way to inject read-only context 

@lanadon, can you use runtime context for that?

```py
agent = create_agent(..., context_schema=MyContext)

agent.invoke({'messages': [...]}, context=MyContext())
```

You could use a pydantic `BaseModel` for your context schema if you'd like to enforce validation?

sydney-runkle:
Closing as not planned for now, happy to revisit if we get a more compelling reason as to why this would be helpful

rhlarora84:
Users currently using `create_agent` which support `state_schema` may want an easy path towards migration to `create_deep_agent`. The middleware framework allows you to mutate the state and then leverage it for tools/pre processing of messages/post processing etc. It would be great to carry over the same concepts to `create_deep_agent` for that reason, and continue to maintain consistency with what langchain released just few months back.

coder-xiaotian:
I also need state_schema, because the context won't be persisted to the database

akhil189:
Hello @lanadon were you able to find a workaround for this? If yes, could you please share how did you achieve that?

akhil189:
@eyurtsev @sydney-runkle 

I have a workflow with 3 tools (SQL tool, Retriever Tool, Report Tool).

Retriever Tool - retrieves context from a `vectorstore` (`List[Document]`)
SQL Tool - creates a valid SQL query
Report Tool - accepts a SQL query and creates a csv report

Based on user query, the agent has to call one or more tools. The workflow is specified in skills. I need my tools to 
1. Store  the context in agent's state and
2. Return a `ToolMessage` to the main agent 

I need them in agent's state because:
1. I need the context for highlighting (grounding) on the UI and also logging.
2. If there a filepath, I want to provide a download option on the UI.

```python
from langchain.tools import ToolRuntime, tool
from langgraph.types import Command
from langchain.agents import AgentState, create_agent
from typing import Annotated, List
from langchain_core.documents import Document


@tool
def generate_report(self, sql_query: str, runtime: ToolRuntime) -> Command:
    """Takes in a SQL query and generates a csv file with postprocessing

    Args:
        sql_query: A valid SQL query
        runtime: LangChain ToolRuntime providing the tool_call_id for ToolMessage.

    Returns:
        Command updating the agent state with the saved document and a ToolMessage.
    """

    # <Logic to create the csv file>
    path = ...

    return Command(
        update={
            "reports": [path],
            "messages": [
                ToolMessage(
                    content=f"report successfully generated",
                    tool_call_id=runtime.tool_call_id,
                )
            ],
        }
    )

@tool
def context_retriever(query: str, runtime: ToolRuntime) -> Command:
    """Retrieves context from a vectorstore

    Args:
        query: The query to find relevant context.
    """

    # logic to retrieve Documents from a vectorstore
    docs = ...
    formatted = "\n".join([doc.content for doc in docs])

    return Command(
        update={
            "agent_context": docs,
            "messages": [
                ToolMessage(
                    content=formatted,
                    tool_call_id=runtime.tool_call_id,
                )
            ],
        }
    )

class CustomState(AgentState):
    """Extended state that tracks context alongside messages."""

    agent_context: List[Document] # docs from vectorstore
    reports = List[str] # filepaths

### with create_agent()
agent = create_agent(
    model=...,
    system_prompt=...,
    tools=[...],
    state_schema=CustomState
)

### with deep_agent()
agent = create_deep_agent(
    model=...,
    system_prompt=...,
    tools=[...],
    state_schema=CustomState # not accepted
)
```

Could you please guide me on how to achieve this with `create_deep_agent()`?

PR #685, opened 2026-01-08
feat: update memory prompt to be more barebones

nhuang-lc:
- Updates the memory prompt to be more barebones
- Puts sources and content together (instead of all sources, then all content)

### DIFF
--- a/libs/deepagents/deepagents/middleware/memory.py
+++ b/libs/deepagents/deepagents/middleware/memory.py
@@ -150,6 +150,7 @@ MEMORY_SYSTEM_PROMPT = """<agent_memory>
 </memory_guidelines>
 """
 
+
 class MemoryMiddleware(AgentMiddleware):
     """Middleware for loading agent memory from AGENTS.md files.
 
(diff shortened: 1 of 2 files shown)


libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -89,21 +89,65 @@ class MemoryStateUpdate(TypedDict):
...
-- Memory content above provides project-specific context and instructions
-- Follow any guidelines, conventions, or patterns described in memory
-- Memory is read-only during this session (loaded at startup)
-- If you need to update memory, use the appropriate file editing tools
+MEMORY_SYSTEM_PROMPT = """<agent_memory>
+{agent_memory}
+</agent_memory>
+
+<memory_guidelines>
+    The above <agent_memory> was loaded in from files in your filesystem. As you learn from your interactions with the user, you can save new knowledge by calling the `edit_file` tool.
```
vtrivedy, line 97: Good call out on Edit File tool we want it to use this and not bash if enabled right?

vtrivedy (approved):

PR #691, opened 2026-01-09
Port SDK Memory to CLI

vtrivedy:
## Summary
- Port CLI from custom `AgentMemoryMiddleware` to SDK's `MemoryMiddleware`
- Standardize on `AGENTS.md` filename (per [agents.md](https://agents.md/) spec)
- Fix default prompt that was causing agents to overcheck memory in protocol

## Changes
**CLI:**
- Replace `AgentMemoryMiddleware` with SDK's `MemoryMiddleware`
- Update memory paths to use `AGENTS.md` instead of `agent.md`
- Remove aggressive "Memory-First Protocol" from default prompt that instructed agents to run `ls /memories/` at session start
- Delete dead code: `token_utils.py` (unused), `agent_memory.py` (replaced), stale benchmark test

**SDK:**
- Remove debug `print("HI")` statement from memory middleware

## Test plan
- [x] all 139 unit tests pass


libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -342,7 +342,6 @@ async def abefore_agent(self, state: MemoryState, runtime: Runtime, config: Runn
         if "memory_contents" in state:
             return None
 
-        print("HI")
```
eyurtsev, line 345: oh my why is this here

libs/deepagents-cli/deepagents_cli/agent.py
```diff
@@ -388,17 +387,27 @@ def create_cli_agent(
 
     # Add memory middleware
     if enable_memory:
-        agent_middleware.append(AgentMemoryMiddleware(settings=settings, assistant_id=assistant_id))
+        memory_sources = [str(settings.get_user_agent_md_path(assistant_id))]
```
eyurtsev, line 390: We should figure out what to do in sandbox mode for memory and skills. In some sense it's OK given that there's HIL, but users may not realize that this is the configuration, so we may need to configure w/ a different default?

eyurtsev (approved):

PR #696, opened 2026-01-09
feat(sdk): MCP Progressive Disclosure

### LINKED ISSUE #480: Support iterative discovery of MCP servers and tools
Read tool definitions on-demand from the filesystem, rather than reading them all up-front.

The agent can then write code to interact with MCP servers / tools.

https://www.anthropic.com/engineering/code-execution-with-mcp

rkbadhan:
Closes #480

This PR introduces MCP (Model Context Protocol) progressive disclosure middleware to deepagents, achieving lower token reduction for MCP tool integration through lazy schema loading and filesystem-based tool discovery. This is inspired by the blog from cursor - `https://cursor.com/blog/dynamic-context-discovery`

Instead of loading all MCP tool schemas into the system prompt upfront (which consumes 5K-20K tokens for 10 tools), this implementation stores tool metadata in a virtual filesystem structure and lets agents discover tools on-demand using existing filesystem operations (`ls`, `read_file`). Only the schemas needed for the current task are loaded into context.

---

### 1. New MCPMiddleware (`libs/deepagents/deepagents/middleware/mcp.py`)

**Architecture:**

- Implements `AgentMiddleware` interface for seamless integration with deepagents middleware stack
- Manages lifecycle of MCP server connections via `langchain-mcp-adapters`
- Syncs tool metadata to virtual filesystem on initialization
- Injects discovery instructions into system prompt instead of full tool schemas
- Provides `mcp_invoke` tool for executing discovered MCP tools

**Filesystem Structure:**

```
/.mcp/
  ├── brave-search/
  │   ├── search.json       # Tool metadata with schema
  │   └── suggest.json
  ├── filesystem/
  │   ├── read_file.json
  │   └── write_file.json
  └── github/
      └── create_issue.json
```

**Discovery Workflow:**

1. Agent runs `ls /.mcp/` → sees available servers
2. Agent runs `ls /.mcp/brave-search/` → sees available tools
3. Agent runs `read_file /.mcp/brave-search/search.json` → loads full schema
4. Agent invokes `mcp_invoke(tool_name="search", arguments={...})` → executes tool

**Token Economics:**

| Approach | Tokens | Details |
|----------|--------|---------|
| Traditional | ~10,000 | 10 tools × 1,000 tokens avg |
| Progressive disclosure | ~1,200 | ~200 tokens (discovery) + ~1,000 tokens (1-2 tools used) |
| **Savings** | **~8,800** | **~47% reduction** |

### 2. Integration with `create_deep_agent` (`libs/deepagents/deepagents/graph.py`)

**New Parameters:**

```python
mcp_servers: list[dict[str, Any]] | None  # MCP server configurations
mcp_root: str = "/tmp/.mcp"              # Root directory for metadata storage
```

**Configuration Example:**

```python
agent = create_deep_agent(
    mcp_servers=[
        {
            "name": "brave-search",
            "command": "npx",
            "args": ["-y", "@modelcontextprotocol/server-brave-search"],
            "env": {"BRAVE_API_KEY": os.getenv("BRAVE_API_KEY")}
        }
    ],
    mcp_root="/var/lib/deepagents/.mcp"  # Optional custom path
)
```

**Middleware Ordering:**
MCPMiddleware is inserted before FilesystemMiddleware to ensure MCP metadata is available to filesystem operations.

### 3. Comprehensive Testing

**Unit Tests** (`test_mcp_middleware.py` - 722 lines):

- Server configuration validation
- Lifecycle management (connect, close, context manager)
- Tool schema extraction and metadata sync
- System prompt injection
- Server/tool name extraction logic
- Error handling for missing dependencies

**Integration Tests** (`test_mcp_integration.py` - 492 lines):

- End-to-end agent creation with MCP servers
- Filesystem metadata structure verification
- Integration with other middleware (skills, memory)
- Async invocation patterns
- Tool invocation with mock MCP clients
- Backward compatibility verification

**Test Coverage:**

- All error paths (invalid configs, duplicate names, missing dependencies)
- Async and sync code paths
- Integration with full middleware stack
- Metadata persistence across sessions

### 4. Documentation & Examples

**Example Script** (`examples/mcp_tools_optimization.py` - 359 lines):

Demonstrates:
- Basic MCP usage with single server
- Multiple MCP servers (brave-search, filesystem, github)
- Custom MCP root configuration
- Integration with skills and memory middleware
- Detailed explanation of token reduction mechanics

---

## Technical Design Decisions

**Virtual Filesystem Approach:** Uses `FilesystemBackend(virtual_mode=True)` for MCP metadata to avoid polluting user's actual filesystem while maintaining discoverability.

**Lazy Connection:** MCP client connection deferred to first `before_agent` call to support agent creation without immediate network calls.

**Graceful Degradation:** Connection failures logged but don't block agent initialization; tools become unavailable rather than causing crashes.

**Server Name Prefixing:** Tool names follow `{server}_{tool}` convention from MCP adapters; middleware strips prefixes for clean discovery.

**No Schema Duplication:** Tool schemas stored once in metadata files, loaded on-demand rather than duplicated in system prompt.

---

## Dependencies

**Required:**
- `langchain-mcp-adapters` - MCP protocol client implementation

**Installation:**

```bash
pip install langchain-mcp-adapters
```

---

## Backward Compatibility

✅ **Fully backward compatible:**

- `mcp_servers` parameter is optional (defaults to `None`)
- Existing agents without MCP continue working unchanged
- All existing middleware (skills, memory, filesystem) compatible
- Standard `tools` parameter works alongside `mcp_servers`

---

## Performance Characteristics

**Initialization:**
- Sync on startup: One-time metadata sync (~100ms for 10 tools)
- Can disable with `sync_on_startup=False` for faster startup

**Runtime:**
- First tool discovery: 2-3 filesystem operations (~10ms)
- Subsequent invocations: Cached in agent context
- Tool execution: Same latency as traditional MCP integration

**Memory:**
- Metadata files: ~1-2KB per tool
- Virtual filesystem overhead: Negligible (in-memory)

---

## Security Considerations

**Sandboxed Metadata:** Virtual filesystem isolates MCP metadata from user files

**Command Injection:** Server configs validated; command/args passed directly to subprocess without shell interpretation

**Environment Isolation:** Each server gets isolated environment variables

**Error Sanitization:** Tool execution errors sanitized before returning to agent

---

## Files Changed

| File | Change | Lines |
|------|--------|-------|
| `libs/deepagents/deepagents/middleware/mcp.py` | New | +321 |
| `examples/mcp_tools_optimization.py` | New | +359 |
| `libs/deepagents/tests/unit_tests/middleware/test_mcp_middleware.py` | New | +722 |
| `libs/deepagents/tests/integration_tests/test_mcp_integration.py` | New | +492 |
| `libs/deepagents/deepagents/graph.py` | Modified | +24 |
| `libs/deepagents/deepagents/middleware/__init__.py` | Modified | +5 |
| **Total** | | **+1,923** |

---

## Testing Instructions

Run unit tests:
```bash
pytest libs/deepagents/tests/unit_tests/middleware/test_mcp_middleware.py -v
```

Run integration tests:
```bash
pytest libs/deepagents/tests/integration_tests/test_mcp_integration.py -v
```

Run example (requires `BRAVE_API_KEY`):
```bash
export BRAVE_API_KEY="your-key"
python examples/mcp_tools_optimization.py
```

---

## Migration Guide

**Before (Traditional MCP):**

```python
# All tool schemas loaded into prompt
from langchain_mcp_adapters import MultiServerMCPClient

client = MultiServerMCPClient(servers)
tools = client.get_tools()  # ~10,000 tokens
agent = create_deep_agent(tools=tools)
```

**After (Progressive Disclosure):**

```python
# Tools discovered on-demand
agent = create_deep_agent(
    mcp_servers=[
        {"name": "brave-search", "command": "npx", "args": [...]}
    ]
)
# Only ~200 tokens in system prompt
```

---

## Future Enhancements

Potential follow-ups (not in this PR):

- Metadata caching across agent sessions
- Tool usage analytics for optimization
- Dynamic server addition/removal
- Resource-based tool filtering
- MCP server health monitoring


libs/deepagents/deepagents/middleware/mcp.py
```diff
@@ -0,0 +1,321 @@
...
+        mcp_root: str = "/tmp/.mcp",
+        sync_on_startup: bool = True,
+    ) -> None:
+        self._validate_servers(servers)
+        self.servers = servers
+        self.mcp_root = mcp_root
+        self.sync_on_startup = sync_on_startup
+        self._mcp_client: Any | None = None
+
+        from deepagents.backends.filesystem import FilesystemBackend
```
eyurtsev: nit: should be global

libs/deepagents/tests/integration_tests/test_mcp_integration.py
```diff
@@ -0,0 +1,492 @@
+"""Integration tests for MCP middleware with create_deep_agent.
```
eyurtsev: We won't need the end to end integration tests -- for mcp it's easy to launch an mcp server as a subprocess from within unit tests, makes it very nice to be able to tests against server configurations etc

libs/deepagents/tests/unit_tests/middleware/test_mcp_middleware.py
```diff
@@ -0,0 +1,722 @@
+"""Unit tests for MCP middleware.
```
eyurtsev: The unit tests can be written in the style of langchain-mcp-adapters. Using a sub-process call to launch a local mcp server using a context manager.

libs/deepagents/deepagents/middleware/mcp.py
```diff
@@ -0,0 +1,321 @@
+"""MCP (Model Context Protocol) middleware for progressive tool disclosure."""
```
eyurtsev, line 1: Hey thanks for the contribution! 

Code looks good, but there's a few big changes / design problems that would be addressed before we could consider merging:

1. Server usage of MCP is virtually never via stdio transport (wrong paradigm for web deployments unless folks are using web-sockets). We'd likely need to be looking at http transport as the default. I'm OK adding stdio later as an option, but that's not the case that makes sense for the SDK by default.
2. deepagents uses `backends` for any file operations, so the middleware needs to work against the backend protocol not against custom state.
3. http can be stateful and stateful MCP relies on MCP-Session-ID, the session ID would need to be tied to the thread id? (although I am not sure what would be expected in the event of sub-agents)

eyurtsev (changes requested):
Hey thanks for the contribution!

Code looks good, but there's a few big changes / design problems that would be addressed before we could consider merging:

1. Server usage of MCP is virtually never via stdio transport (wrong paradigm for web deployments unless folks are using web-sockets). We'd likely need to be looking at http transport as the default. I'm OK adding stdio later as an option, but that's not the case that makes sense for the SDK by default.
2. deepagents uses backends for any file operations, so the middleware needs to work against the backend protocol not against custom state.
3. http can be stateful and stateful MCP relies on MCP-Session-ID, the session ID would need to be tied to the thread id? (although I am not sure what would be expected in the event of sub-agents)

libs/deepagents/deepagents/middleware/mcp.py
rkbadhan: Hi @eyurtsev Thanks for the review. I have made changes to 1 and 2. 

For 3 - here is what I am thinking - 
1. Most of the MCP servers are stateless by default like it covers 90% of them.
2. For remaining scenarios - thread id  is a suitable and proven approach 
3. Given subagents are being executed based on agent's conversation context, they are not independent. they can inherit the agent's session. If parallel sub-agents cause state conflicts on a stateful MCP server, MCP server should handle this concurrent access.
4. Solution could be - for stateless MCP servers, I'll add a session_mode="shared" option that uses a single connection across all threads, avoiding the overhead of per-thread connections when it's unnecessary.

Does this align with your thinking, or do you see any other edge cases that are not covered?

eyurtsev: > Most of the MCP servers are stateless by default like it covers 90% of them

You're right... maybe even more than that. To be honest, I'd be OK handling stateless http servers, and extending support after the fact for stateful ones.

---

For (2), the implementation is currently not coding against a backend interface, but is instead assuming we're working with files in the langgraph checkpointer (i.e., StateBackend). The goal is to really have it work against an arbitrary backend.

rkbadhan: Hi @eyurtsev I have tried to take care of the (2) as per your suggestion. please have a look. 

For (3) - 
1. Fixed resource leak - The close() method now properly closes the MCP client when the last reference is removed. Added _cleanup_client() method that handles both async (aclose()) and sync (close()) cleanup methods.
2. Added clear documentation - Updated the class docstring to explicitly state this middleware is designed for stateless servers (90%+ of use cases) and that it shares a single HTTP client across all threads. 
3. compute_cache_key() function can be enhanced for per-thread logic when/if stateful server support is needed.

I didn't add a session_mode parameter since we're not implementing per-thread support yet. Please let me know if you'd like any other changes.

rkbadhan:
Hi @eyurtsev is there anything more needed in this PR?

zimo888:
Hi, is it possible to support Skill, so that the scripts within Skill.md can be replaced with MCP calls? In this way, progressive loading of MCP can be implemented.

rkbadhan:
Hi @eyurtsev  and @mdrxy is there any thing more needed on this PR? Please let me know

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #697, opened 2026-01-09
feat: add agent name tag to `task` tool calls

sydney-runkle:
This adds a `subagent:{subagent_name}` tag whenever the `task` tool is called.

before

Screenshot 2026-01-09 at 2 04 25 PM

after

Screenshot 2026-01-09 at 2 04 57 PM

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -356,7 +356,7 @@ def _create_task_tool(
         if rt := ls.get_current_run_tree():
             if not rt.tags:
                 rt.tags = []
-            rt.tags.insert(0, f"subagent_name:{subagent_type}")
+            rt.tags.insert(0, f"subagent:{subagent_type}")
         result = subagent.invoke(subagent_state, runtime.config)
         if not runtime.tool_call_id:
             value_error_msg = "Tool call ID is required for subagent invocation"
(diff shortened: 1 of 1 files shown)


vtrivedy:
can we get a unit testing for this under libs/deepagents/tests/unit_tests

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -352,6 +353,10 @@ def task(
             allowed_types = ", ".join([f"`{k}`" for k in subagent_graphs])
             return f"We cannot invoke subagent {subagent_type} because it does not exist, the only allowed types are {allowed_types}"
         subagent, subagent_state = _validate_and_prepare_state(subagent_type, description, runtime)
+        if rt := ls.get_current_run_tree():
```
vtrivedy, line 356: AFAICT this won't propagate to the tools and other actions right? should we pass through config so they propagate?  so can filter subagent:general-purpose and see those calls.

sydney-runkle: was of the opinion that we shouldn't propagate here (cleaner, if you get to the tool you can inherently see the subagent logic), but could be convinced

sydney-runkle:
> can we get a unit testing for this under libs/deepagents/tests/unit_tests

pretty difficult to do this w/o mocking the tracing stuff :/

vtrivedy:
closing, we're handling this another way :)

PR #706, opened 2026-01-10
cli-token-tracking-fixes

vtrivedy:
## Summary
  - Fix `/clear` not resetting token count - now resets tokens and clears status message
  - Fix token count not updating after regular text responses (was only updating after tool calls)
  - Fix token display showing stale values - now uses `total_tokens` (input + output) for accurate context size
  - Fix tokens disappearing during streaming - hide during stream, restore after completion or interrupt
  - Suppress Pydantic V1/Python 3.14 deprecation warning on startup

  ## Details
  - Moved token extraction before `content_blocks` check so we capture usage from all message types
  - Changed from `input_tokens` to `total_tokens` which represents actual context size after response
  - Added `hide()` and `show()` methods to token tracker for streaming state management
  - Token display now restored in cleanup to handle early cancellation edge cases

  ## Test plan
  - [x] Unit tests added for `TextualTokenTracker` (5 tests passing)

  ## Known limitation
  Very quick interrupts (before any streaming data arrives) may occasionally leave token display blank until next message.

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/app.py
+++ b/libs/deepagents-cli/deepagents_cli/app.py
@@ -40,9 +40,7 @@ if TYPE_CHECKING:
 class TextualTokenTracker:
     """Token tracker that updates the status bar."""
 
-    def __init__(
-        self, update_callback: callable, hide_callback: callable | None = None
-    ) -> None:
+    def __init__(self, update_callback: callable, hide_callback: callable | None = None) -> None:
         """Initialize with callbacks to update the display."""
         self._update_callback = update_callback
         self._hide_callback = hide_callback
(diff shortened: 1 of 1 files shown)


libs/deepagents-cli/deepagents_cli/main.py
```diff
@@ -1,6 +1,12 @@
 """Main entry point and CLI loop for deepagents."""
 # ruff: noqa: T201
 
+# Suppress deprecation warnings from langchain_core (e.g., Pydantic V1 on Python 3.14+)
+# ruff: noqa: E402
+import warnings
+
+warnings.filterwarnings("ignore", module="langchain_core._api.deprecation")
```
eyurtsev, line 8: nice thank you!

PR #707, opened 2026-01-10
fix(cli,harbor): clipboard copy on message widgets across terminals

vtrivedy:
- Fix text selection not copying to clipboard when highlighting text in the CLI
  - Add `pyperclip` dependency for reliable clipboard access on macOS Terminal (OSC52 doesn't work there)
  - Enable `ALLOW_SELECT` on message widgets so text can be selected

Note: confirm we're ok with pyperclip dependency

### DIFF
--- a/libs/harbor/uv.lock
+++ b/libs/harbor/uv.lock
@@ -576,6 +576,7 @@ dependencies = [
     { name = "modal" },
     { name = "pillow" },
     { name = "prompt-toolkit" },
+    { name = "pyperclip" },
     { name = "python-dotenv" },
     { name = "pyyaml" },
     { name = "requests" },
@@ -598,6 +599,7 @@ requires-dist = [
     { name = "modal", specifier = ">=0.65.0" },
     { name = "pillow", specifier = ">=10.0.0" },
     { name = "prompt-toolkit", specifier = ">=3.0.52" },
+    { name = "pyperclip", specifier = ">=1.8.0" },
     { name = "python-dotenv" },
     { name = "pyyaml", specifier = ">=6.0" },
     { name = "requests" },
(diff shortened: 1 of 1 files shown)


vtrivedy:
@eyurtsev quick review on supporting copy across terminals with fallbacks + adding pyperclip as dependency

libs/deepagents-cli/deepagents_cli/widgets/messages.py
```diff
@@ -48,13 +47,9 @@ def __init__(self, content: str, **kwargs: Any) -> None:
             content: The message content
             **kwargs: Additional arguments passed to parent
         """
-        super().__init__(**kwargs)
+        super().__init__(f"[bold cyan]>[/bold cyan] {content}", **kwargs)
```
eyurtsev, line 50: We potentially have a security issue here... could you check what kind of escaping rules we need to apply to `content` to avoid malicious actor from generating links using textual format

eyurtsev: This doesn't actually matter for User Message that much (since it's user controlled)...

unless 
1. this raw content strategy is used in either tool message or assistant message or system message (if we're showing system message which we don't right now)
2. there's something that can produce a user message that's not a user

libs/deepagents-cli/deepagents_cli/widgets/messages.py
eyurtsev, line 243: OK basically any `f` string on untrusted inputs for Static, we might need to do escape rules if i remember correctly. I'd double check against textual syntax

eyurtsev: This is from memory... against i may be wrong, but we should check what kind of stuff textual markup is capable of (it may be possible to have it link to **code** in the existing textual app)

mdrxy:
@vtrivedy @eyurtsev is this still an issue?

PR #711, opened 2026-01-11
Fix Windows path handling in middleware

johnpccd:
This PR includes fixes for Windows path handling in the middleware components:

- Normalize Windows paths in skills middleware for cross-platform compatibility
- Enhance memory and skills middleware for cross-platform compatibility with virtual paths support

### DIFF
--- a/libs/deepagents/tests/unit_tests/middleware/test_validate_path.py
+++ b/libs/deepagents/tests/unit_tests/middleware/test_validate_path.py
@@ -2,6 +2,7 @@
 
 import pytest
 
+from deepagents.backends.filesystem import FilesystemBackend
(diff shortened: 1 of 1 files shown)


eyurtsev:
Hi @johnpccd we currently do not support windows for the CLI. Your best bet for now is https://learn.microsoft.com/en-us/windows/wsl/install

johnpccd:
Hi @eyurtsev,
If you could be so kind to tell me what are the other areas that need modification to support Windows, I could also support with that.
With this PR, I am able to run the cli on Windows and use skills and file system tools.

sydney-runkle:
In the short term we're not planning on adding windows support, going to close for now.

Issue #705, opened 2026-01-10
Interrupt Really Quickly After Message leaves "Interrupted" in Status Bar


vtrivedy:
Edge case where interrupt comes before any streaming of tokens, "Interrupted" stays in bottom bar.  Minor issue

tadeodonegana:
Fixed in #712

PR #712, opened 2026-01-11
fix(cli): reset status to Ready after early interrupt

tadeodonegana:
Previously, when an interrupt occurred before any tokens were streamed, the status bar would get stuck showing 'Interrupted' instead of resetting to 'Ready'. This was because the exception handlers for CancelledError and KeyboardInterrupt returned early without updating the status.

Added adapter._update_status('Ready') before return in both exception handlers to ensure the UI recovers properly from interrupts.

This fix the following issue: #705 

Before:
Screenshot 2026-01-11 at 12 35 51 PM

After:
Screenshot 2026-01-11 at 12 48 06 PM


mdrxy:
We've removed this status bar - thank you for the contribution, though!

PR #713, opened 2026-01-11
feat: improve create_deep_agent() to support SystemMessage type

### LINKED ISSUE #597: Support SystemMessage type for create_deep_agent(system_prompt=...) (align with langchain.agents.create_agent)

ps06756:
## Summary

- Add `_merge_system_prompt()` helper that properly handles `SystemMessage` inputs
- Preserve block structure when `SystemMessage.content` is a list
- Preserve metadata (`name`, `additional_kwargs`, etc.) via `model_copy()`
- Add unit tests for all cases

This addresses the concerns raised in #612 about data loss when using `.text` property:
- No more flattening of structured content
- Block boundaries preserved for list content
- Metadata preserved via `model_copy()`
- Backward compatible with string inputs

Closes #597.

## Test plan

- [x] Unit tests added for `_merge_system_prompt()` function
- [x] All 212 unit tests pass

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -11,6 +11,7 @@ from langchain.agents.structured_output import ResponseFormat
 from langchain_anthropic import ChatAnthropic
 from langchain_anthropic.middleware import AnthropicPromptCachingMiddleware
 from langchain_core.language_models import BaseChatModel
+from langchain_core.messages import SystemMessage
 from langchain_core.tools import BaseTool
 from langgraph.cache.base import BaseCache
 from langgraph.graph.state import CompiledStateGraph
@@ -23,6 +24,25 @@ from deepagents.middleware.patch_tool_calls import PatchToolCallsMiddleware
 from deepagents.middleware.subagents import CompiledSubAgent, SubAgent, SubAgentMiddleware
 
 BASE_AGENT_PROMPT = "In order to complete the objective that the user asks of you, you have access to a number of standard tools."
+
+
+def _merge_system_prompt(system_prompt: str | SystemMessage | None) -> SystemMessage:
+    """Merge a system prompt with BASE_AGENT_PROMPT, preserving SystemMessage metadata."""
+    if system_prompt is None:
+        return SystemMessage(content=BASE_AGENT_PROMPT)
+
+    if isinstance(system_prompt, str):
+        return SystemMessage(content=f"{system_prompt}\n\n{BASE_AGENT_PROMPT}")
+
+    content = system_prompt.content
+    if isinstance(content, str):
+        new_content = f"{content}\n\n{BASE_AGENT_PROMPT}"
+    elif isinstance(content, list):
+        new_content = content + [{"type": "text", "text": BASE_AGENT_PROMPT}]
+    else:
+        new_content = f"{content}\n\n{BASE_AGENT_PROMPT}"
+
+    return system_prompt.model_copy(update={"content": new_content})
 
 
 def get_default_model() -> ChatAnthropic:
@@ -41,7 +61,7 @@ def create_deep_agent(
     model: str | BaseChatModel | None = None,
     tools: Sequence[BaseTool | Callable | dict[str, Any]] | None = None,
     *,
-    system_prompt: str | None = None,
+    system_prompt: str | SystemMessage | None = None,
     middleware: Sequence[AgentMiddleware] = (),
     subagents: list[SubAgent | CompiledSubAgent] | None = None,
     response_format: ResponseFormat | None = None,
@@ -66,8 +86,7 @@ def create_deep_agent(
     Args:
         model: The model to use. Defaults to Claude Sonnet 4.
         tools: The tools the agent should have access to.
-        system_prompt: The additional instructions the agent should have. Will go in
-            the system prompt.
+        system_prompt: Additional instructions for the agent (str or SystemMessage).
         middleware: Additional middleware to apply after standard middleware.
         subagents: The subagents to use. Each subagent should be a dictionary with the
             following keys:
@@ -148,7 +167,7 @@ def create_deep_agent(
 
     return create_agent(
         model,
-        system_prompt=system_prompt + "\n\n" + BASE_AGENT_PROMPT if system_prompt else BASE_AGENT_PROMPT,
+        system_prompt=_merge_system_prompt(system_prompt),
         tools=tools,
         middleware=deepagent_middleware,
         response_format=response_format,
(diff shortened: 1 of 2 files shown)


libs/deepagents/deepagents/graph.py
```diff
@@ -8,32 +8,48 @@
 from langchain.agents.middleware.summarization import SummarizationMiddleware
 from langchain.agents.middleware.types import AgentMiddleware
 from langchain.agents.structured_output import ResponseFormat
-from langchain.chat_models import init_chat_model
 from langchain_anthropic import ChatAnthropic
 from langchain_anthropic.middleware import AnthropicPromptCachingMiddleware
```
eyurtsev, line 12: Before any change to create_deep_agent interface can be considered:

1. Logic needs to be applied across our middleware as most of the middleware is still accessing system_message.content rather than content_blocks -- you can see how it's handled in langchain middleware that deal with system message content blocks correctly
2. We'd want to test that things work across model the main providers

One of the main reasons to introduce a structured SystemMessage is for content blocks that support prompt caching of a **part** of the system message -- so we'd want to make sure that this feature is actually supported

sydney-runkle:
Closing as a dupe of https://github.com/langchain-ai/deepagents/pull/612, thank you for the contribution though!

ps06756:
@sydney-runkle This is a really bad open source etiquette, you essentially took my code change, added extra things without any kind of attribution.

At the very least, you should have provided attribution or built on top of original commit

Nobody is going to spend their time contributing if the maintainers take the commit, do a bunch of minor fixes and raise a new PR without any kind of attribution.

Same thing happened with another one of your repositories deepagentjs @eyurtsev

eyurtsev:
@ps06756 you're right! And thank you so for putting your time into the project already, we really appreciate it. I'll pass the feedback along to the rest of the team

ps06756:
@eyurtsev Can attribution be added for the commit ?  I had collaborated with @RussellLuo on this as well.

eyurtsev:
Not sure I'll investigate next week

sydney-runkle:
Hey @ps06756,

Thanks so much for your contribution! Sorry about the missing attribution, that was my bad. I forgot to add you as a co-author when I squash/merged my PR.

I've added you both here: https://github.com/langchain-ai/deepagents/pull/847

ps06756:
Thanks a lot @sydney-runkle ! Really appreciate it :)

sydney-runkle:
@ps06756,

Sure thing. Thanks for calling out that we missed adding you. That's definitely the kind of thing we don't want to miss!

PR #723, opened 2026-01-12
fix(sdk): fixing the `_get_backend` runtime argument type to match with request.runtime

keenborder786:
Description: Simple fix to make sure that `_get_backend` matches with the type of `request.runtime`.
Issue: https://github.com/langchain-ai/langchain/issues/34666


keenborder786:
@mdrxy

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -36,6 +38,13 @@
     truncate_if_too_long,
 )
 
+# Keep runtime-visible for pydantic/StructuredTool annotation resolution.
+_TYPE_RESOLUTION_DEPS = (ToolRuntime, ToolCallRequest, Awaitable, Callable, Sequence)
```
sydney-runkle: we don't need these

keenborder786:
@sydney-runkle I have moved the imports now to TYPE_CHECKING block.

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -996,7 +1003,7 @@ def __init__(
             tool_token_limit_before_evict,
         )
 
-    def _get_backend(self, runtime: ToolRuntime) -> BackendProtocol:
+    def _get_backend(self, runtime: Runtime) -> BackendProtocol:
```
aaron-seq, line 481: This change is correct for the FilesystemMiddleware._get_backend method, but there's a critical issue causing CI/CD failures:

The standalone _get_backend function (around line 381) still uses ToolRuntime:
```python
def _get_backend(backend: BACKEND_TYPES, runtime: ToolRuntime) -> BackendProtocol:
```

Since ToolRuntime is now in the TYPE_CHECKING block, it's not available at runtime, causing NameError in tests. This function is used by all the tool generators (_ls_tool_generator, _read_file_tool_generator, etc.) which are called at runtime.

There are two solutions:
1. Change the standalone _get_backend function to also use Runtime instead of ToolRuntime
2. Or keep ToolRuntime as a regular import (not in TYPE_CHECKING) since it's needed at runtime

Based on the issue description, I believe solution 1 is correct - both the standalone function and the class method should use Runtime to match request.runtime type.

keenborder786: The simplest solution was to move it out and use noqa to ignore the `TC002`. All tests are passing now.

sydney-runkle:
Unfortunately this is a quirk of our current backend protocol design, we need to redesign in order to fix. Going to close for now.

PR #735, opened 2026-01-13
feat: add agent name when creating subagents

sydney-runkle:
before
Screenshot 2026-01-13 at 5 09 38 PM

after
Screenshot 2026-01-13 at 5 08 41 PM

should work w/ the LS team to make `weather` visible as an agent...

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -251,7 +251,7 @@ def _get_subagents(
             system_prompt=DEFAULT_SUBAGENT_PROMPT,
             tools=default_tools,
             middleware=general_purpose_middleware,
-            name="subagent:general-purpose",
+            name="general-purpose",
         )
         agents["general-purpose"] = general_purpose_subagent
         subagent_descriptions.append(f"- general-purpose: {DEFAULT_GENERAL_PURPOSE_DESCRIPTION}")


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -352,7 +354,7 @@ def task(
             allowed_types = ", ".join([f"`{k}`" for k in subagent_graphs])
             return f"We cannot invoke subagent {subagent_type} because it does not exist, the only allowed types are {allowed_types}"
         subagent, subagent_state = _validate_and_prepare_state(subagent_type, description, runtime)
-        result = subagent.invoke(subagent_state, runtime.config)
```
sydney-runkle, line 386: passing parent config messes up metadata because this config takes prio over `subagent._config`

things like tags, metadata, etc are inherently passed along through context vars

if we want to pass this config, we need to explicitly merge w/ prio to the subagent config

eyurtsev:
Could we link to the PR that fixed this, so we can ask the author what would break for them?

sydney-runkle:
We originally merged https://github.com/langchain-ai/deepagents/pull/602

@Mr-Squirrel, what specifically breaks for you with the above change?

Things like tags, metadata, etc should be inherited from the parent config.

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -277,6 +278,7 @@ def _get_subagents(
             system_prompt=agent_["system_prompt"],
             tools=_tools,
             middleware=_middleware,
+            name=f"subagent:{agent_['name']}",
```
sydney-runkle:
```suggestion
            name=f"{agent_['name']}",
```

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -251,6 +251,7 @@ def _get_subagents(
             system_prompt=DEFAULT_SUBAGENT_PROMPT,
             tools=default_tools,
             middleware=general_purpose_middleware,
+            name="subagent:general-purpose",
```
sydney-runkle:
```suggestion
            name="general-purpose",
```

Mr-Squirrel:
@sydney-runkle before I answer that question I just want to make sure I understand what just happened here

In a PR named "feat: add agent name when creating subagents", from a branch called `sr/deepagents-orientation`, you arbitrarily decided to radically alter how subagents behave and delete already merged and published functionality because you don't know how to merge a dict? 

Is that the gist of what just happened?

When you add a comment like 
> passing parent config messes up metadata because this config takes prio over subagent._config
> things like tags, metadata, etc are inherently passed along through context vars
> if we want to pass this config, we need to explicitly merge w/ prio to the subagent config

I expect you to either a) open an issue to address this, or b) address this properly in this pr.

I do NOT expect you to randomly undo a serious fix that will completely break multiple enterprise pipelines unless we remain frozen forever.

---

Now to your question

The reason why I **need** the config to be passed along to subagents is that, as per the official [langchain docs on RunnableConfig](https://reference.langchain.com/v0.3/python/_modules/langchain_core/runnables/config.html#RunnableConfig), the config is intended to be used to configure the top level call, AND any sub calls as evidenced by the multiple doc strings in the source:
- "Tags for this call and any sub-calls (eg. a Chain calling an LLM)."
- "Metadata for this call and any sub-calls (eg. a Chain calling an LLM)."
- "Runtime values for attributes previously made configurable on this Runnable, or sub-Runnables"
- "Callbacks for this call and any sub-calls (eg. a Chain calling an LLM)."

Further evidence that the runtime config is explicitly intended to be passed along can be seen in their provision of
```python
COPIABLE_KEYS = [
    "tags",
    "metadata",
    "callbacks",
    "configurable",
]
```

As well as
```python
var_child_runnable_config: ContextVar[RunnableConfig | None] = ContextVar(
    "child_runnable_config", default=None
)
```

---

I firmly believe in the principle of "least surprise", and having an official langchain library incorrectly use its own core classes is definitely surprising. I'm honestly appalled you actually merged this PR without even waiting a single day for an answer.

This kind of rock-star behaviour is exactly why langchain has such a problem with reputation in the python ecosystem. "Important tool, crap code" I believe is the message I hear most.

@eyurtsev You were the one that reviewed and merged my PR, can you elaborate on how you are completely fine undoing it not even a month later? Does langchain have any kind of adherence to the common established practices and conventions around software development, semver, and API designs? Or is this completely up to whoever decides to merge for the day?

I would immediately revert this PR, and split it into two separate ones:

1. feat: add agent name when creating subagents -> a PR that adds agent names when creating subagents
2. fix: ensure subagent metadata overrides are merged with parent graph

Actually embarrassing, wow

libs/deepagents/deepagents/middleware/subagents.py
Mr-Squirrel: Where is `context` coming from? I don't see langchain Runnables use this at all, how am I supposed to know to use it? what is in it? how does it replace the documented `RunnableConfig`?


---
 ## `ainvoke()`

  ```python
  async def ainvoke(
      input: Input,
      config: RunnableConfig | None = None,
      **kwargs: Any
  ) -> Output

```

  ### Parameters

  #### `input`
  - **Type:** `Input`
  - **Required:** Yes
  - **Description:** The input to the Runnable.

  #### `config`
  - **Type:** `RunnableConfig | None`
  - **Default:** `None`
  - **Description:** A config to use when invoking the Runnable. The config supports standard keys like:
    - `'tags'` — for tracing purposes
    - `'metadata'` — for tracing purposes
    - `'max_concurrency'` — for controlling how much work to do in parallel
    - and other keys
  - **Reference:** [RunnableConfig Documentation](https://api.python.langchain.com/en/latest/runnables/langchain_core.runnables.config.RunnableConfig.html)

  ---

  ### Returns

  - **Type:** `Output`
  - **Description:** The output of the Runnable.
---

https://reference.langchain.com/python/langchain_core/runnables/#langchain_core.runnables.base.Runnable.ainvoke

**Comment by sydney-runkle** [comment]
@Mr-Squirrel, under the hood, we use `var_child_runnable_config` to pull the config from the parent. So it's still sent to subagent calls :) explicitly passing in the parent config actually is messing up the "merging" that happens under the hood.

can you please point to a specific example of code you're using that breaks because of this change? happy to address an MRE.

we tagged you explicitly to follow up knowing that you helped add this LOC. your original PR had no description of a specific problem you were solving, so we wanted to get more info.


would ask that you be a bit kinder w/ your next reply - we're here to help you!

**Comment by mdrxy** [comment]
@Mr-Squirrel 

> Does langchain have any kind of adherence to the common established practices and conventions around software development, semver, and API designs?

Yes, in fact. We have a comprehensive [versioning page](https://docs.langchain.com/oss/python/versioning).

According to the official Semantic Versioning `2.0.0` specification, the rules for `0.x.x` are distinctly different from versions `1.0.0` and above.

The specification explicitly states:

> "Anything MAY change at any time. The public API SHOULD NOT be considered stable."

`deepagents` has not yet been published as `1.x.x`.

**Comment by eyurtsev** [comment]
@Mr-Squirrel Could you confirm that you are running async + python 3.10 and the subagent is using a custom Runnable (rather than a Pregel object from langgraph)

And what's failing is inheritance of tags/metadata/ custom callbacks?

**Comment by Mr-Squirrel** [comment]
@eyurtsev I really cannot share my code since its under a lot of NDAs, but here is the gist

I develop and maintain several complex langchain agents what are used both standalone and as subagents.

To allow for this interop, all my agents extend `langchain_core.runnables.Runnable`, and then have a method to convert to a `CompiledSubAgent` as needed:
```python
 def to_compiled_sub_agent(self) -> CompiledSubAgent:
       return {
           "name": self.name,
           "description": self.description,
           "runnable": self,
       }
```

this then allows my agents to be passed to `deepagents.create_deep_agent` as so
```python
# as a subagent
create_deep_agent(
    ...
    subagents=[agent.to_compiled_sub_agent() for agent in self.sub_agents]),
    ...
)
```

`deepagents.create_deep_agent` clearly types subagents to be
```python
subagents: list[SubAgent | CompiledSubAgent] | None = None
    
# where CompiledSubAgent is clearly typed as
class CompiledSubAgent(TypedDict):
    """A pre-compiled agent spec."""

    name: str
    """The name of the agent."""

    description: str
    """The description of the agent."""

    runnable: Runnable
    """The Runnable to use for the agent."""

# where `Runnable` is directly imported from `"langchain-core>=1.2.6,<2.0.0"`
from langchain_core.runnables import Runnable
# a production version
```

This then means that `deepagents` by contract should invoke `Runnables` as per their specification, which among others clearly states the propagation of `RunnableConfig` to sub-calls which is exactly what a subagent is. This is further enforced by your own typing in the [specific method](https://github.com/langchain-ai/deepagents/blob/master/libs/deepagents/deepagents/middleware/subagents.py#L394) in question:

```python
result = subagent.invoke(subagent_state, context=runtime.context)

# where `subagent` is explicitly typed as a `langchain_core.runnables.Runnable`.
```

I use all the `COPIABLE_KEYS` if the config, and I access it through the officially typed parameter from `langchain_core.runnables.Runnable` [public production API](https://reference.langchain.com/python/langchain_core/runnables/#langchain_core.runnables.base.Runnable.ainvoke)

---

@sydney-runkle 

> explicitly passing in the parent config actually is messing up the "merging" that happens under the hood.

**How** is it messing it up, **Where** does this happen, **What** is the impact? 
Also, if this is so then it should be easy to verify with a simple unit test right? Prove that deepagents is correctly invoking `langchain_core.runnables.Runnable` 

> would ask that you be a bit kinder w/ your next reply - we're here to help you!

Im not being mean or kind, im being frank. If you object to my messaging then argue on its merit, not some weird emotional tangent. there's no obligation to sugar-coat things in open-source, there is an obligation to the community and the code.

Using commit [05a34a75466e3807b03fd68114c6a26ddeff3875](https://github.com/langchain-ai/deepagents/commit/05a34a75466e3807b03fd68114c6a26ddeff3875), prove that subagents `ainvoke()` receives the runtime config as a parameter from the parent with at least the copiable keys passed along

---

@mdrxy Either you don't fully understand what the issue here is, or you are confused. You're literally reaffirming my statement. **langchain agents public API** is 1.x, and then as per your own comments and reference

> 1.0.0: First stable release with production-ready APIs

I am arguing that since `deepagents` is an official langchain package, and y'all are langchain employees, it stands to reason that your usage of agents within the code should adhere to the public production API of langchain agents? 
Now you can argue that you consider the current langchain agents api to be inadequate or bad, and thus are developing a parallel or competing implementation, but that would in my mind not be in "principle of least surprise".

---

I reiterate my objection that 
> In a PR named "feat: add agent name when creating subagents", from a branch called sr/deepagents-orientation, you arbitrarily decided to radically alter how subagents behave and delete already merged and published functionality because you don't know how to merge a dict?

and that 
> I would immediately revert this PR, and split it into two separate ones:
> - feat: add agent name when creating subagents -> a PR that adds agent names when creating subagents
> - fix: ensure subagent metadata overrides are merged with parent graph

Either this is **your** repo that is primarily focused on satisfying your own opinions of how things should be, or this is an **open source** repo which then focuses on what is best for the ecosystem and community at large.

Honestly you decide, I think the actions of this PR speak for themselves.

eyurtsev:
@Mr-Squirrel Appreciate the feedback and quick response.

> Could you confirm that you are running async + python 3.10 and the subagent is using a custom Runnable (rather than a Pregel object from langgraph)
And what's failing is inheritance of tags/metadata/ custom callbacks?

Could you please confirm (yes/no)

1. async + python **3.10**?
2. custom Runnable is not a langgraph graph? (I think this is yes based on your explanation.)

Some context:

1. LangChain uses Python `contextvars` to propagate config automatically. For async code, this only works reliably starting in Python 3.11 due to asyncio limitations. Async LangChain users should upgrade to Python 3.11+.
2. The `runnable: Runnable` annotation was intended to support graph-based subagents (Pregel objects that subclass `Runnable`), so the type annotation is wrong. For new subagent code, using LangGraph graphs is recommended over subclassing `Runnable` (we'll update documentation to explain this)

sydney-runkle:
Re proof that copiable keys are propagated, here's a MRE:

```py
from dotenv import load_dotenv
from langchain_core.tools import tool

load_dotenv()


@tool
def get_weather(city: str) -> str:
    """Get the weather in a city."""
    return f"The weather in {city} is sunny."


agent = create_deep_agent(
    model="anthropic:claude-sonnet-4-20250514",
    system_prompt="Use the weather subagent to answer weather questions.",
    subagents=[
        {
            "name": "weather",
            "description": "Gets weather information for cities.",
            "system_prompt": "Use the get_weather tool to answer.",
            "tools": [get_weather],
        }
    ],
    name="supervisor",
)


# result = agent.invoke({"messages": [HumanMessage(content="What's the weather in Tokyo?")]})
# print(result["messages"][-1].content)


for ns, (chunk, metadata) in agent.stream(
    {"messages": [HumanMessage(content="What's the weather in Tokyo?")]},
    stream_mode="messages",
    subgraphs=True,
    config={"tags": ["proof"]},
):
    print(metadata.get("tags"))
```

Specifically, note that `['proof']` is included in `metadata.get(['tags'])` even for the stream chunks from the subagent.

eyurtsev:
This has taken far more maintainer time than it should have, so we’re going to update the documentation to help keep users on the intended happy path. Subclassing Runnable isn’t part of that happy path and isn’t something we plan to support.

The supported approach is documented here: https://docs.langchain.com/oss/python/deepagents/subagents#using-compiledsubagent

After reviewing the issue, it looks like the problem comes from the user's custom Runnable implementation rather than anything related to Python 3.10. Deepagents requires Python 3.11, so discussion about 3.10 isn’t relevant.

The core issue is that the user code subclasses Runnable, which isn’t a pattern we designed for or want to encourage. We’ll make this clearer in the docs.

@Mr-Squirrel, please make sure your Runnable implementation calls ensure_config (as other runnables do). The relevant code is here: https://github.com/langchain-ai/langchain/blob/master/libs/core/langchain_core/runnables/base.py#L2046

Mr-Squirrel:
**Python:** `3.11.8`
**Langchain:** `1.2.7`
**deepagents:** `0.3.1`

@eyurtsev 
> The runnable: Runnable annotation was intended to support graph-based subagents (Pregel objects that subclass Runnable), so the type annotation is wrong.

That sure does explain a whole lot. I would argue this should have been caught in my original PR, especially since I clearly stated I didn't know what more I needed to do to verify the fix.

> This has taken far more maintainer time than it should have,

I'm sorry that the community is such a drag, hopefully it becomes smaller in the future.

---

@sydney-runkle s proof is at least enough to understand what we "should" be doing, thanks for taking the time to provide the MRE. I expanded your example to include proof that the metadata is actually present in the runnable config argument as well, which is critical for my implementation.

```python
@tool
def get_weather(city: str, config: RunnableConfig) -> str:
    """Get the weather in a city."""
    print("\n=== TOOL CONFIG ===")
    print(f"Tags: {config.get('tags', [])}")
    print(f"Metadata: {config.get('metadata', {})}")
    print(f"Configurable: {config.get('configurable', {})}")
    print(f"Run name: {config.get('run_name', 'N/A')}")
    print("===================\n")
    return f"The weather in {city} is sunny."
```

---

@mdrxy comments also clarifies that the `deepagents` project as it stands should not be considered a tool that can be used to do work, but rather it is a fun toy to experiment with. A real shame since this could have been something that delivered real value, but hey it is what it is.

---


I still think the way this was handled is absolutely ludicrous, but at the end of the day y'all are obviously powering ahead in whatever direction you decide and there's not much we can do to influence that.

Would be great to see this handled through a proper issue and subsequent PR, but I guess `git bisect` will continue to be useful...

eyurtsev:
> This has taken far more maintainer time than it should have,
> I'm sorry that the community is such a drag, hopefully it becomes smaller in the future.

This was not directed at you and apologies if you read it this way. I was trying to say that this has taken too much developer time to debug and could've been prevented with better docs to guide users toward the happy path.

Apologies for interactions/issues and you probably have more important things to do than respond to PR comments all day long.

PR #740, opened 2026-01-13
chore: automatically sort imports

mdrxy:


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -22,8 +22,8 @@
 from deepagents.backends import StateBackend
 
 # Re-export type here for backwards compatibility
```
eyurtsev, line 24: nit: comment in the wrong place

libs/acp/pyproject.toml
```diff
@@ -54,5 +54,17 @@ Issues = "https://github.com/langchain-ai/deepagents/issues"
...
+line-length = 100
+
+[tool.ruff.lint]
+select = ["E", "F", "I"]
+ignore = []
+
+[tool.ruff.lint.isort]
+force-single-line = false
+combine-as-imports = true
+known-first-party = ["deepagents_acp", "deepagents", "deepagents_cli"]
```
eyurtsev, line 67: these are 3rd party actually

eyurtsev: err 2nd party -> they should show up in the 2nd group

PR #742, opened 2026-01-13
feat: summarization offloading

mdrxy:
Depends on https://github.com/langchain-ai/langchain/pull/34763

Extends langchain's base `SummarizationMiddleware` to persist conversation history to a backend before summarization, enabling retrieval of full context when needed by an agent.

If offloading fails, we intentionally do nothing so as to allow errors to be raised by APIs. This is probably better than silently dropping messages, as it would be very obvious very quickly, allowing us to investigate why the offloading failed.


libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -0,0 +1,515 @@
...
+
+        Args:
+            msg: Message to check.
+
+        Returns:
+            `True` if this is a summary message from a previous summarization.
+        """
+        if not isinstance(msg, HumanMessage):
+            return False
+        additional_kwargs = getattr(msg, "additional_kwargs", {})
```
ccurme: don't need `getattr` here

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -0,0 +1,515 @@
...
+    def _get_history_path(self, runtime: Runtime) -> str:
+        """Generate path for storing conversation history.
+
+        Args:
+            runtime: Runtime context.
+
+        Returns:
+            Path string like `'/conversation_history/{thread_id}/{timestamp}.json'`
+        """
+        thread_id = self._get_thread_id(runtime)
```
ccurme: do we need the thread ID? I don't see anything comparable in [MemoryMiddleware](https://github.com/langchain-ai/deepagents/blob/master/libs/deepagents/deepagents/middleware/memory.py)

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -0,0 +1,515 @@
...
+        payload = {
+            "timestamp": datetime.now(UTC).isoformat(),
+            "thread_id": thread_id,
+            "message_count": len(filtered_messages),
+            "messages": json.loads(self._serialize_messages(filtered_messages)),
+            "messages_text": get_buffer_string(filtered_messages),
+        }
+
+        try:
+            result = backend.write(path, json.dumps(payload, indent=2, default=str))
```
ccurme: I'm not sure we want to put this in json. Would it be more LLM friendly to use markdown? e.g.,
```
# Metadata
Timestamp: ...
Message count: ...

# Conversation history
System: ...
Human: ...
AI: ...
```

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -0,0 +1,515 @@
...
+
+        # Filter out previous summary messages to avoid redundant storage
+        filtered_messages = self._filter_summary_messages(messages)
+
+        payload = {
+            "timestamp": datetime.now(UTC).isoformat(),
+            "thread_id": thread_id,
+            "message_count": len(filtered_messages),
+            "messages": json.loads(self._serialize_messages(filtered_messages)),
+            "messages_text": get_buffer_string(filtered_messages),
```
ccurme: why not just use `get_buffer_string`? why do we have another version using pydantic serialization?

libs/deepagents/deepagents/middleware/summarization.py
mdrxy: this is debatable and something we should benchmark

libs/deepagents/deepagents/middleware/summarization.py
mdrxy: Thread ID organization is intentional. Conversation history is inherently per-thread, unlike `MemoryMiddleware` which loads the `AGENTS.md` global context. Without `thread_id`, all history files go to one directory, making it hard to find specific conversations.

libs/deepagents/deepagents/middleware/summarization.py
mdrxy: will keep as JSON for now since it's lossless

libs/deepagents/deepagents/middleware/summarization.py
mdrxy: https://github.com/langchain-ai/deepagents/pull/742#discussion_r2692964462

mdrxy:
Tests failing since it doesn’t have the dependent changes available in CI

libs/deepagents/tests/integration_tests/test_summarization.py
```diff
@@ -84,8 +97,70 @@ def test_summarize_continues_task(tmp_path: Path, model_name: str) -> None:
...
+
+    # Should have timestamp header(s) from summarization events
+    assert "## Summarized at" in content, "Missing timestamp header in markdown file"
+
+    # Should contain human-readable message content (from get_buffer_string)
+    assert "Human:" in content or "AI:" in content, "Missing message content in markdown file"
+
+    # Verify the summary message references the conversation_history path
+    summary_message = result["messages"][0]
+    assert "conversation_history" in summary_message.content
```
ccurme, line 163: this line appears to fail if we summarize more than once. example trace: https://smith.langchain.com/public/196add86-4e5c-4f9a-a322-6d52ff249607/r

can you make sure that case is functional and unit tested?

libs/deepagents/tests/integration_tests/test_summarization.py
```diff
@@ -84,8 +97,70 @@ def test_summarize_continues_task(tmp_path: Path, model_name: str) -> None:
...
+    )
+
+
+@pytest.mark.parametrize(
+    "model_name",
+    [
+        pytest.param("anthropic:claude-sonnet-4-5-20250929", id="claude-sonnet"),
+    ],
+)
+def test_summarization_offloads_to_filesystem(tmp_path: Path, model_name: str) -> None:
```
ccurme, line 127: this is more a unit test, there isn't really a demonstration of what practically this enables an agent to do. can you add a "needle in the haystack" follow-up question? e.g., ask for the first thing we import in the file, check that we read from the conversation history and get the correct answer.

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -0,0 +1,498 @@
...
+            keep=("messages", 20),
+        )
+        ```
+    """
+
+    def __init__(
+        self,
+        model: str | BaseChatModel,
+        *,
+        backend: BACKEND_TYPES | None = None,
```
eyurtsev: Can we make the backends required?

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -0,0 +1,498 @@
...
+    from langchain.agents.middleware.types import AgentState
+    from langchain.chat_models import BaseChatModel
+    from langgraph.runtime import Runtime
+
+    from deepagents.backends.protocol import BACKEND_TYPES, BackendProtocol
+
+logger = logging.getLogger(__name__)
+
+
+class SummarizationMiddleware(BaseSummarizationMiddleware):
```
sydney-runkle, line 67: should we actually inherit from langchain middleware here? does that slow down dev at all? what if we just build our own here and then port changes back to langchain when relevant?

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -0,0 +1,498 @@
...
+            runtime: Runtime context for factory functions.
+
+        Returns:
+            Resolved backend instance, or None if no backend configured.
+        """
+        if self._backend is None:
+            return None
+
+        if callable(self._backend):
+            tool_runtime = ToolRuntime(
```
sydney-runkle, line 145: a little crazy to me that we artificially create these

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -0,0 +1,498 @@
...
+        self,
+        model: str | BaseChatModel,
+        *,
+        backend: BACKEND_TYPES | None = None,
+        trigger: ContextSize | list[ContextSize] | None = None,
+        keep: ContextSize = ("messages", _DEFAULT_MESSAGES_TO_KEEP),
+        token_counter: TokenCounter = count_tokens_approximately,
+        summary_prompt: str = DEFAULT_SUMMARY_PROMPT,
+        trim_tokens_to_summarize: int | None = _DEFAULT_TRIM_TOKEN_LIMIT,
+        history_path_prefix: str = "/conversation_history",
```
sydney-runkle, line 80: do we need this if backends are required? should this be an attr on the backend / filesystem?

sydney-runkle (commented):
initial notes!

libs/deepagents/deepagents/middleware/summarization.py
mdrxy: yes

libs/deepagents/deepagents/middleware/summarization.py
mdrxy: can revisit down road as needed

libs/deepagents/deepagents/middleware/summarization.py
mdrxy: ya seems worth revisiting this API

mdrxy: doesn't fit clean in middleware context

libs/deepagents/deepagents/middleware/summarization.py
mdrxy: `history_path_prefix` should be on the backend but it isn't

Issue #731, opened 2026-01-13
TypeError: 'list' object is not a mapping in _file_data_reducer when using FilesystemMiddleware


madhan-palani-zoomrx:
### Description
When using deepagents with FilesystemMiddleware, the [_file_data_reducer](cci:1://file:///Users/madhan/research-bot/.venv/lib/python3.13/site-packages/deepagents/middleware/filesystem.py:58:0-91:17) function in [deepagents/middleware/filesystem.py](cci:7://file:///Users/madhan/research-bot/.venv/lib/python3.13/site-packages/deepagents/middleware/filesystem.py:0:0-0:0) throws a TypeError because it receives a list instead of the expected dictionary type.

### Error Details
```
TypeError: 'list' object is not a mapping
```

### Stack Trace
```python
Traceback (most recent call last):
  File "/Users/madhan/research-bot/app/services/chat_completions.py", line 291, in stream_chat_completion
    async for chunk in stream:
    ...<123 lines>...
  File "/Users/madhan/research-bot/.venv/lib/python3.13/site-packages/langgraph/pregel/main.py", line 2986, in astream
    loop.after_tick()
  File "/Users/madhan/research-bot/.venv/lib/python3.13/site-packages/langgraph/pregel/_loop.py", line 542, in after_tick
    self.updated_channels = apply_writes(
        self.checkpoint,
        ...<3 lines>...
        self.trigger_to_nodes,
    )
  File "/Users/madhan/research-bot/.venv/lib/python3.13/site-packages/langgraph/pregel/_algo.py", line 296, in apply_writes
    if channels[chan].update(vals) and next_version is not None:
  File "/Users/madhan/research-bot/.venv/lib/python3.13/site-packages/langgraph/channels/binop.py", line 122, in update
    self.value = self.operator(self.value, value)
  File "/Users/madhan/research-bot/.venv/lib/python3.13/site-packages/deepagents/middleware/filesystem.py", line 86, in _file_data_reducer
    result = {**left}
TypeError: 'list' object is not a mapping
```

### Root Cause
The [_file_data_reducer](cci:1://file:///Users/madhan/research-bot/.venv/lib/python3.13/site-packages/deepagents/middleware/filesystem.py:58:0-91:17) function expects its `left` parameter to be either `None` or a dictionary:

```python
def _file_data_reducer(left: dict[str, FileData] | None, right: dict[str, FileData | None]) -> dict[str, FileData]:
    # ...
    if left is None:
[...]

PR #754, opened 2026-01-14
chore(deepagents): add end to end tests to confirm file reducer working properly in state backend

sheing-google:
Adds more test coverage for: https://github.com/langchain-ai/deepagents/issues/731

However, it's still unclear as to what triggers the underlying issue

### DIFF
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -54,10 +54,7 @@ class FileData(TypedDict):
     """ISO 8601 timestamp of last modification."""
 
 
-def _file_data_reducer(
-    left: dict[str, FileData] | list[dict[str, FileData]] | None,
-    right: dict[str, FileData | None],
-) -> dict[str, FileData]:
+def _file_data_reducer(left: dict[str, FileData] | None, right: dict[str, FileData | None]) -> dict[str, FileData]:
     """Merge file updates with support for deletions.
 
     This reducer enables file deletion by treating `None` values in the right
@@ -65,8 +62,7 @@ def _file_data_reducer(
     state management where annotated reducers control how state updates merge.
 
     Args:
-        left: Existing files dictionary. May be `None` during initialization,
-            or a list of dicts when multiple tool calls update state.
+        left: Existing files dictionary. May be `None` during initialization.
         right: New files dictionary to merge. Files with `None` values are
             treated as deletion markers and removed from the result.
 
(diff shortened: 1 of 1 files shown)


eyurtsev:
Hi @sheing-google thanks for the PR. Depending on your comfort level with the library -- would you be willing to do an end to end test with parallel tool calling that reproduces the full issue end to end? there's an example of such tests for testing subagents. They set up a fake LLM for the agent and force the agent to take some actions

eyurtsev:
Here is reference code: 

https://github.com/langchain-ai/deepagents/blob/master/libs/deepagents/tests/unit_tests/test_subagents.py

sheing-google:
@eyurtsev Updated it. Passed with `make test` https://github.com/langchain-ai/deepagents/pull/754/commits/cff68318e7dda269f3af58d2797876a9fe1b6e97

eyurtsev:
@christian-bromann we'll likely need to port this to JS.

The write path has a race condition, but I think this is fine as its consistent with how a regular file system works. So the LLM should understand what it's doing.

I suspect that our edits just bug out though? And we might have to start sending diffs into the reducer?

* We can do a short term solution to prevent silent failures for edits: add an `after_model` hook that applies in this situation that will reject editing collisions with an artifical ToolMessage and jump back to the model.
* Longer terms solution is to accept Edit or diffs instructions in the reducer directly? (this would require some thinking)

eyurtsev:
@sheing-google thanks for the PR -- I am taking over it to add tests and logic for handling the edit scenario

libs/deepagents/tests/unit_tests/middleware/test_filesystem.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    # Fake model will issue two write_file tool calls in a single turn
+    fake_model = GenericFakeChatModel(
+        messages=iter(
+            [
+                AIMessage(
+                    content="",
+                    tool_calls=[
+                        {
+                            "name": "write_file",
```
eyurtsev: @sheing-google this test doesn't seem to fail if I remove the reducer logic.

Could you explain how you're encountering this problem?

madhan-palani-zoomrx:
@eyurtsev Thanks for working on the fix. I have gone through the discussion and code changes happened so far, but in my scenario I have encountered this even when there is no tool call. 

Also this issue has happened for the *right* parameter as well.

https://github.com/langchain-ai/deepagents/issues/731#issuecomment-3771328953

eyurtsev:
@sheing-google I undid the changes, added another unit test (similar to the one you had originally) and merging. I am still unclear as to what the root cause is for the bug and we need to get an MRE first before we can address it.

PR #521, opened 2025-12-01
feat(cli): inject local context into system prompt via `LocalContextMiddleware`

rlancemartin:
Adds `LocalContextMiddleware` to inject comprehensive local context into the agent's system prompt, including git state, current directory, file listings, and directory tree structure. 

## Example: 

trace: 
https://smith.langchain.com/public/04763d34-9fe1-4db5-a6d3-1efe3f88db52/r

result:
```
## Local Context

**Current Directory**: `/Users/rlm/Desktop/Code/deepagents/libs/deepagents-cli`

**Git**: Current branch `rlm/git-middleware`, main branch available: `master`

**Files** (14 shown):
- Makefile
- README.md
- agent.py
- cli-banner.jpg
- deepagents_cli/
- deepagents_cli.egg-info/
- examples/
- final_report.md
- hello_world_poem.txt
- langgraph_agent.py
- pyproject.toml
- research_request.md
- tests/
- uv.lock
... (3 more files)

**Tree** (3 levels):
deepagents-cli/
├── deepagents_cli/
│   ├── integrations/
│   │   ├── __init__.py
│   │   ├── daytona.py
│   │   ├── modal.py
│   │   ├── runloop.py
│   │   └── sandbox_factory.py
│   ├── skills/
│   │   ├── __init__.py
│   │   ├── commands.py
│   │   ├── load.py
│   │   └── middleware.py
│   ├── __init__.py
│   ├── __main__.py
│   ├── agent.py
│   ├── agent_memory.py
│   ├── commands.py
│   ├── config.py
│   ├── default_agent_prompt.md
│   ├── execution.py
│   ... (truncated)
... (truncated)
```

Test w/ this change -- 
https://smith.langchain.com/public/b4450671-ab71-4a59-81b0-5cbda5f3099b/r
https://smith.langchain.com/public/389c9ef1-a5be-49f7-a0bf-1721fa6c6a5d/r

make one `shell` tool call:
```
{
  "command": "cd /Users/rlm/Desktop/Code/deepagents/libs/deepagents-cli && make format"
}
```

On `master` it doesn't supply `cwd` -- 
https://smith.langchain.com/public/5cc9039b-ffeb-42f3-9fd8-361e2140eb85/r

make one `shell` tool call:
```
{
  "command": "make format"
}
```

PR #770, opened 2026-01-15
feat(cli): model identity

mdrxy:
Related: #521

* Modified system prompt creation to include a "Model Identity" section, providing the model name, provider, and context window size if available.

---

Theorizing that by grounding the model in its identity, it may hallucinate less or better understand its constraints. This *might* (keyword *might*) help the agent avoid context overflow if it is smart enough to use data like it's context window size.

Relevant: [arXiv:2512.24661](https://arxiv.org/abs/2512.24661)

---

Future items:

- [ ] Give model a hook to see current token usage
- [ ] Give model a hook to trigger summarization
- [ ] `ContextEditingMiddleware`?


libs/deepagents-cli/deepagents_cli/config.py
```diff
@@ -490,23 +491,40 @@ def create_model(model_name_override: str | None = None) -> BaseChatModel:
...
             max_tokens=None,
         )
+    else:
+        # Should not reach here due to earlier validation
+        console.print(f"[bold red]Error:[/bold red] Unknown provider: {provider}")
+        sys.exit(1)
+
+    # Extract context limit from model profile (if available)
+    if (
+        hasattr(model, "profile")
```
eyurtsev: this is a lot of defensive code is all of it necessary could we do just?

model.profile.get("max_input_tokens")

libs/cli/deepagents_cli/agent.py
```diff
@@ -105,6 +105,21 @@ def get_system_prompt(assistant_id: str, sandbox_type: str | None = None) -> str
     """
     agent_dir_path = f"~/.deepagents/{assistant_id}"
 
+    # Build model identity section
+    model_identity_section = ""
+    if settings.model_name and isinstance(settings.model_name, str):
+        model_identity_section = f"""### Model Identity
```
eyurtsev, line 126: Why do we want this in the system prompt?

mdrxy: see updated issue description

PR #778, opened 2026-01-15
Enhance BaseSandbox.ls_info() to return absolute paths

### LINKED ISSUE #775: _list_skills() cannot fetch skills because BaseSandbox.ls_info() is returning relative paths
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

```python
1. Create an agent, using a sandbox-based backend, and configure skills.
2. Let the agent perform tasks related to the skill.
```

```shell

```

### Description

Expected to see "skills" listed in the system prompt, but it actually says "(No skills available yet. You can create skills in ...)".

### Environment / System Info

_No response_

RussellLuo:
Closes #775.

### DIFF
--- a/libs/deepagents/deepagents/backends/sandbox.py
+++ b/libs/deepagents/deepagents/backends/sandbox.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import base64
 import json
+from pathlib import Path
 import shlex
 from abc import ABC, abstractmethod
 
@@ -190,7 +191,12 @@ except PermissionError:
                 continue
             try:
                 data = json.loads(line)
-                file_infos.append({"path": data["path"], "is_dir": data["is_dir"]})
+                file_infos.append(
+                    {
+                        "path": str(Path(path) / data["path"]),
+                        "is_dir": data["is_dir"],
+                    }
+                )
             except json.JSONDecodeError:
                 continue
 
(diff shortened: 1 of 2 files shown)


eyurtsev:
I'll close the PR but you're welcome to open a new one with the appropriate fix

RussellLuo:
Yeah, that fix was kinda messed up. Thanks for pointing it out!

I think we should tweak the command that runs in the sandbox:

```diff
async def als_info(self, path: str) -> list[FileInfo]:
        """Structured listing with file metadata using os.scandir."""
        cmd = f"""python3 -c "
import os
import json

path = '{path}'

try:
    with os.scandir(path) as it:
        for entry in it:
            result = {{
-                'path': entry.name,
+                'path': os.path.join(path, entry.name),
                'is_dir': entry.is_dir(follow_symlinks=False)
            }}
            print(json.dumps(result))
except FileNotFoundError:
    pass
except PermissionError:
    pass
" 2>/dev/null"""
```

If this is the right direction, I'd be more than happy to submit a new PR.

eyurtsev:
Feel free to make a PR if quick. I don't remember the correct semantics, and a bit short on time right now, so may be a few days before I review more carefully -- I basically don't remember if we decided that ls should behave like ls and give relative paths w/ information about the CWD just being surfaced through the system prompt. -- is this an option for your use case?

PR #784, opened 2026-01-15
fix(deepagents): truncate lines on read

ccurme:
The description for the read tool includes

> - Any lines longer than 2000 characters will be truncated

and there is a constant `MAX_LINE_LENGTH = 2000` in the module, but this constant is unused and the logic is not implemented.

Here we implement the truncation as described.

### DIFF
--- a/libs/deepagents/tests/unit_tests/middleware/test_filesystem.py
+++ b/libs/deepagents/tests/unit_tests/middleware/test_filesystem.py
@@ -1,0 +1,89 @@
+"""Unit tests for filesystem middleware helper functions."""
+
+import pytest
+
+from deepagents.middleware.filesystem import _truncate_lines
+
+
+def test_no_truncation_when_below_limit() -> None:
+    """Test that lines shorter than max_line_length are not modified."""
+    text = "short line\nanother short line"
+    result = _truncate_lines(text, max_line_length=100)
+    assert result == text
+
+
+def test_truncation_without_suffix() -> None:
+    """Test basic truncation without suffix."""
+    text = "a" * 100
+    result = _truncate_lines(text, max_line_length=10)
+    assert result == "a" * 10
+
+
+def test_truncation_with_suffix() -> None:
+    """Test truncation with suffix appended."""
+    text = "a" * 100
+    result = _truncate_lines(text, max_line_length=10, suffix="...")
+    assert result == "a" * 7 + "..."
+    assert len(result) == 10  # noqa: PLR2004
+
+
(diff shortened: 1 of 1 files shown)


libs/deepagents/tests/unit_tests/middleware/test_filesystem.py
```diff
@@ -0,0 +1,95 @@
+"""Unit tests for filesystem middleware functions."""
```
eyurtsev: Could we either:

1. Swap to not file system backend (so we don't default to testing everything with this backend)
2. Generalize it across backends

---

Also maybe this test we should do actually end to end with the full agent given that it's a read_tool?

Full end to end tests are in `test_subagents` (unit tests)

I kind of like having some of these end to end since we often uncover other issues. With the end to end tests we can also parameterize some cases across all backends

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -373,7 +421,8 @@ def sync_read_file(
         """Synchronous wrapper for read_file tool."""
         resolved_backend = _get_backend(backend, runtime)
         file_path = _validate_path(file_path)
-        return resolved_backend.read(file_path, offset=offset, limit=limit)
+        content = resolved_backend.read(file_path, offset=offset, limit=limit)
+        return _truncate_lines(content, suffix="...")
```
eyurtsev: nit: Do we need to be more precise about what `...` means ? (will the LLM know it's 
truncated? do we have anything in the prompts that denotes trunctation? e.g., [truncated])

libs/deepagents/tests/unit_tests/middleware/test_filesystem.py
ccurme: moved to `test_end_to_end` and parametrized backends

eyurtsev (approved):

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -349,6 +351,52 @@ async def async_ls(runtime: ToolRuntime[None, FilesystemState], path: str) -> st
     )
 
 
+def _truncate_lines(
```
eyurtsev, line 354: Nice! thanks for all the parameterized tests. We should actually test this function itself via unit tests (i.e., just the algo itself) without any runtimes etc

eyurtsev (approved):
Looks great! Apologies if I misread the previous tests and you already had tests at the level of _truncate_lines (I think they started from the file system backend).

PR #790, opened 2026-01-15
chore: update `Makefile` targets, add pre-commit hooks

mdrxy:
porting infra from `langchain-ai/langchain`

also switch to `ty` but don't activate in `Makefile` targets yet


libs/deepagents-cli/pyproject.toml
```diff
@@ -38,10 +38,19 @@ test = [
     "pytest-mock>=3.14.0",
     "pytest-socket>=0.7.0",
     "pytest-timeout>=2.3.1",
+    "pytest-watcher>=0.3.4,<1.0.0",
+    "pytest-xdist>=3.8.0,<4.0.0",
     "responses>=0.25.0",
+]
+test_integration = []
+
+lint = [
```
eyurtsev: Could we merge typing and lint into test? the distinction isn't helpful anymore (likely over-complicated in langchain mono-repo)

libs/deepagents-cli/pyproject.toml
```diff
@@ -38,10 +38,19 @@ test = [
...
     "ruff>=0.9.7",
 ]
 
+typing = [
+    "mypy>=1.10.0,<2.0.0",
```
eyurtsev: we don't expect to pass w/ my-py and if we're picking up a new static type checker, i'd like to get ty a try

mdrxy: will add `ty`

PR #803, opened 2026-01-16
feat: support `SystemMessage` for parity w/ `create_agent`

sydney-runkle:
* fixes https://github.com/langchain-ai/deepagents/issues/597
* enables adding cache control to messages explicitly

### DIFF
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -982,7 +982,7 @@ class FilesystemMiddleware(AgentMiddleware):
             system_prompt = "\n\n".join(prompt_parts)
 
         if system_prompt:
-            new_system_message = append_to_system_message(request.system_message, f"\n\n{system_prompt}")
+            new_system_message = append_to_system_message(request.system_message, system_prompt)
             request = request.override(system_message=new_system_message)
 
         return handler(request)
@@ -1030,7 +1030,7 @@ class FilesystemMiddleware(AgentMiddleware):
             system_prompt = "\n\n".join(prompt_parts)
 
         if system_prompt:
-            new_system_message = append_to_system_message(request.system_message, f"\n\n{system_prompt}")
+            new_system_message = append_to_system_message(request.system_message, system_prompt)
             request = request.override(system_message=new_system_message)
 
         return await handler(request)
--- a/libs/deepagents/deepagents/middleware/memory.py
+++ b/libs/deepagents/deepagents/middleware/memory.py
@@ -366,7 +366,7 @@ class MemoryMiddleware(AgentMiddleware):
         contents = request.state.get("memory_contents", {})
         agent_memory = self._format_agent_memory(contents)
 
-        new_system_message = append_to_system_message(request.system_message, f"\n\n{agent_memory}")
+        new_system_message = append_to_system_message(request.system_message, agent_memory)
 
         return request.override(system_message=new_system_message)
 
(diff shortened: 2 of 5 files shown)


libs/deepagents/tests/integration_tests/test_deepagents.py
```diff
@@ -194,3 +222,35 @@ async def test_with_memory_middleware(self):
         assert_all_deepagent_qualities(agent)
         result = await agent.ainvoke({"messages": [HumanMessage(content="What is your name?")]})
         assert "Jackson" in result["messages"][-1].content
+
+    def test_deep_agent_with_system_message(self):
```
eyurtsev: let's move this to unit tests

eyurtsev: there are several tests that do end to end testing there that you can adapt (end to end, subagents, read file operations etc)

libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -354,23 +354,29 @@ async def abefore_agent(self, state: MemoryState, runtime: Runtime, config: Runn
...
         contents = request.state.get("memory_contents", {})
         agent_memory = self._format_agent_memory(contents)
 
-        if request.system_prompt:
-            system_prompt = agent_memory + "\n\n" + request.system_prompt
+        # Memory is prepended (before other system content)
+        new_system_content: list[str | dict[str, str]]
+        if request.system_message is not None:
+            new_system_content = [
+                {"type": "text", "text": f"{agent_memory}\n\n"},
```
eyurtsev: doesn't this mess up prompt caching?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -981,7 +982,8 @@ def wrap_model_call(
             system_prompt = "\n\n".join(prompt_parts)
 
         if system_prompt:
-            request = request.override(system_prompt=request.system_prompt + "\n\n" + system_prompt if request.system_prompt else system_prompt)
+            new_system_message = append_to_system_message(request.system_message, f"\n\n{system_prompt}")
```
eyurtsev: Could we push `\n\n` into the function as well?

eyurtsev: `\n\n` should only be added if system-messages isn't empty

eyurtsev: So we want the utility function to take care of that edge case

libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -354,23 +355,20 @@ async def abefore_agent(self, state: MemoryState, runtime: Runtime, config: Runn
...
+            Modified request with memory injected into system message.
         """
         contents = request.state.get("memory_contents", {})
         agent_memory = self._format_agent_memory(contents)
 
-        if request.system_prompt:
-            system_prompt = agent_memory + "\n\n" + request.system_prompt
-        else:
-            system_prompt = agent_memory
+        new_system_message = append_to_system_message(request.system_message, f"\n\n{agent_memory}")
```
eyurtsev: There's a small buglet here -- `\n\n` should only be added if `request.system_message` isn't empty

sydney-runkle: Ah good call, thanks!

shell-nlp:
`SystemMessage.content = [{"type":"text","text":"xxx"}] `
This format is not compatible with OpenAI's API.

PR #804, opened 2026-01-16
chore(deepagnets): end to end tests for agent writing editing files

eyurtsev:
* Added unit tests that are end to end with agent using backend tools.
* These unit tests also prove that parallel writes are already handled without changes to the current reducer.


eyurtsev:
@sheing-google fyi this includes your unit-test from the other PR (thank you for those!). I'm actually not sure whether it'll give you credit for them or not b/c our PRs get squash merged by default. If it doesn't I'll try and work in another commit on top of your other PR for a unit test (just FYI)

PR #806, opened 2026-01-16
feat(deepagents): truncate old write / edit calls in message history

ccurme:
(description added by @mdrxy)

Truncates large `write_file` and `edit_file` tool call arguments in older messages to reduce context size without losing important structural information.

As conversations grow, old file write/edit operations consume significant context with content that's no longer relevant. This PR truncates these large arguments while preserving the tool call structure.

**Example:**

Before truncation:
```python
{
    "name": "write_file",
    "args": {
        "path": "/src/components/Button.tsx",
        "content": "import React from 'react';\n\nexport interface ButtonProps {\n  label: string;\n  onClick: () => void;\n  disabled?: boolean;\n}\n\nexport const Button: React.FC<ButtonProps> = ({ label, onClick, disabled }) => {\n  return (\n    <button onClick={onClick} disabled={disabled}>\n      {label}\n    </button>\n  );\n};\n"
    }
}
```

After truncation (default `max_length=100`):
```python
{
    "name": "write_file",
    "args": {
        "path": "/src/components/Button.tsx",
        "content": "import React from 'rea...(argument truncated)"
    }
}
```

The model can still understand *what* was written and *where*, without the full file contents consuming context.

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -150,8 +150,8 @@ def create_deep_agent(
         trigger = ("fraction", 0.85)
         keep = ("fraction", 0.10)
         truncate_args_settings = {
-            "trigger": ("fraction", 0.60),
-            "keep": ("fraction", 0.20),
+            "trigger": ("fraction", 0.85),
+            "keep": ("fraction", 0.10),
         }
     else:
         trigger = ("tokens", 170000)


libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -106,9 +106,108 @@ def __init__(
...
+                    ("fraction", 0.3)
+                    ```
+            token_counter: Function to count tokens in messages.
+            summary_prompt: Prompt template for generating summaries.
+            trim_tokens_to_summarize: Maximum tokens to keep when preparing messages for
+                the summarization call.
+
+                Pass `None` to skip trimming entirely.
+            clean_messages_trigger: Threshold to trigger message cleaning (truncating tool args).
+                If None, cleaning is disabled.
```
sydney-runkle: I didn't anticipate "cleaning" to mean "truncating tool args"

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -106,9 +106,108 @@ def __init__(
         summary_prompt: str = DEFAULT_SUMMARY_PROMPT,
         trim_tokens_to_summarize: int | None = _DEFAULT_TRIM_TOKEN_LIMIT,
         history_path_prefix: str = "/conversation_history",
+        clean_messages_trigger: ContextSize | None = None,
+        clean_messages_keep: ContextSize = ("messages", 20),
+        max_tool_arg_length: int = 100,
```
sydney-runkle: what if instead of `int` we used an alias `Characters = int` for readability and documentation

sydney-runkle (commented):
If I understand correctly, we're adding 4 args related to "cleaning" / tool arg truncation

this is a lot of args! to me this suggests we could go in a direction of (either or)
1. middleware w/ middleware - summarization middleware has a) general summarization b) tool arg truncation c) tool message truncation, and various other plugins
2. grouped args, so something like `truncate_args_settings={...}` which has all 4 of these new guys

libs/deepagents/deepagents/middleware/summarization.py
ccurme: updated naming everywhere

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -94,6 +111,20 @@ def __init__(
             trim_tokens_to_summarize: Max tokens to include when generating summary.
 
                 Defaults to 4000.
+            truncate_args_settings: Settings for truncating large tool arguments in old messages.
+
+                Provide a [`TruncateArgsSettings`][deepagents.middleware.summarization.TruncateArgsSettings]
```
mdrxy, line 116: nice x-ref

PR #807, opened 2026-01-16
chore: improve filesystem and subagents tool descriptions

sydney-runkle:
Fixes https://github.com/langchain-ai/deepagents/issues/416

Specifically, we want to expose clear argument descriptions to the model via json schemas.

Confirmed via langsmith that the new descriptions are rendering nicely.
Also confirmed via testing that JSON schemas have types and descriptions for args, and that these match for sync + async versions of these tools.

Would love to have benchmarks we could use to actually determine improvements here, this is a bit of a shot in the dark.

### DIFF
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -214,17 +214,29 @@ Examples:
 EXECUTE_TOOL_DESCRIPTION = """Executes a shell command in an isolated sandbox environment.
 
 Usage:
-- Always quote file paths containing spaces with double quotes
-- Before creating files/directories, verify parent directory exists with ls
-- Returns combined stdout/stderr output with exit code
-- Output may be truncated if very large
-- Use '&&' to chain dependent commands, ';' for independent sequential commands
-- Prefer absolute paths over cd to maintain working directory
-
-IMPORTANT: Use dedicated tools instead of shell equivalents:
-- Use glob tool instead of find
-- Use grep tool instead of grep/rg
-- Use read_file instead of cat/head/tail
+Executes a given command in the sandbox environment with proper handling and security measures.
+Before executing the command, please follow these steps:
+1. Directory Verification:
+   - If the command will create new directories or files, first use the ls tool to verify the parent directory exists and is the correct location
+   - For example, before running "mkdir foo/bar", first use ls to check that "foo" exists and is the intended parent directory
+2. Command Execution:
+   - Always quote file paths that contain spaces with double quotes (e.g., cd "path with spaces/file.txt")
+   - Examples of proper quoting:
+     - cd "/Users/name/My Documents" (correct)
+     - cd /Users/name/My Documents (incorrect - will fail)
+     - python "/path/with spaces/script.py" (correct)
+     - python /path/with spaces/script.py (incorrect - will fail)
+   - After ensuring proper quoting, execute the command
+   - Capture the output of the command
+Usage notes:
+  - Commands run in an isolated sandbox environment
+  - Returns combined stdout/stderr output with exit code
+  - If the output is very large, it may be truncated
+  - VERY IMPORTANT: You MUST avoid using search commands like find and grep. Instead use the grep, glob tools to search. You MUST avoid read tools like cat, head, tail, and use read_file to read files.
+  - When issuing multiple commands, use the ';' or '&&' operator to separate them. DO NOT use newlines (newlines are ok in quoted strings)
+    - Use '&&' when commands depend on each other (e.g., "mkdir dir && cd dir")
+    - Use ';' only when you need to run commands sequentially but don't care if earlier commands fail
+  - Try to maintain your current working directory throughout the session by using absolute paths and avoiding usage of cd
 
 Examples:
   Good examples:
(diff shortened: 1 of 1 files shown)


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -193,12 +193,14 @@ class FilesystemState(AgentState):
 
 WRITE_FILE_TOOL_DESCRIPTION = """Writes to a new file in the filesystem.
 
-Usage:
-- The file_path parameter must be an absolute path, not a relative path
-- The content parameter must be a string
-- The write_file tool will create the a new file.
-- Prefer to edit existing files over creating new ones when possible."""
+Parameters:
```
eyurtsev: This is different from all the other tools in structure. Could we take a look as to how it looks when materialized as tools chema?

eyurtsev:
I would suggest we take a look at 1-2 traces and examine the full representatio of the tools that the LLM receives (i.e., are the arguments properly documented etc)

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -17,6 +17,7 @@
 from langchain_core.messages import ToolMessage
 from langchain_core.tools import BaseTool, StructuredTool
 from langgraph.types import Command
+from pydantic import Field
```
eyurtsev: Why is Pydantic necessary here for the Annotation? Does Annotated[str, "description"] directly not work?

eyurtsev:
@sydney-runkle would you be able to generate a unit test somewhere that genertates the actual json schema of the main tools?

libs/deepagents/deepagents/middleware/filesystem.py
sydney-runkle: It does, nice

libs/deepagents/tests/unit_tests/middleware/test_tool_schemas.py
```diff
@@ -0,0 +1,82 @@
...
+                assert has_type, f"Arg '{arg_name}' in tool '{tool_name}' is missing type in JSON schema"
+
+                # Check description is present
+                assert "description" in arg_schema, (
+                    f"Arg '{arg_name}' in tool '{tool_name}' is missing description in JSON schema. "
+                    f"Add an Annotated type hint with a description string."
+                )
+
+    def test_sync_async_schema_parity(self) -> None:
+        """Verify sync and async functions produce identical JSON schemas.
```
sydney-runkle, line 57: Added to make sure that a dev doesn't actually change the description / params for one and not the other

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -182,81 +179,52 @@ class FilesystemState(AgentState):
...
-  - The command parameter is required
-  - Commands run in an isolated sandbox environment
-  - Returns combined stdout/stderr output with exit code
-  - If the output is very large, it may be truncated
-  - VERY IMPORTANT: You MUST avoid using search commands like find and grep. Instead use the grep, glob tools to search. You MUST avoid read tools like cat, head, tail, and use read_file to read files.
-  - When issuing multiple commands, use the ';' or '&&' operator to separate them. DO NOT use newlines (newlines are ok in quoted strings)
-    - Use '&&' when commands depend on each other (e.g., "mkdir dir && cd dir")
-    - Use ';' only when you need to run commands sequentially but don't care if earlier commands fail
-  - Try to maintain your current working directory throughout the session by using absolute paths and avoiding usage of cd
+EXECUTE_TOOL_DESCRIPTION = """Executes a shell command in an isolated sandbox environment.
```
sydney-runkle, line 214: do we want this to be as consolidated as it is?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -182,81 +179,52 @@ class FilesystemState(AgentState):
...
 
 Usage:
-- You must use your `Read` tool at least once in the conversation before editing. This tool will error if you attempt an edit without reading the file.
-- When editing text from Read tool output, ensure you preserve the exact indentation (tabs/spaces) as it appears AFTER the line number prefix. The line number prefix format is: spaces + line number + tab. Everything after that tab is the actual file content to match. Never include any part of the line number prefix in the old_string or new_string.
-- ALWAYS prefer editing existing files. NEVER write new files unless explicitly required.
-- Only use emojis if the user explicitly requests it. Avoid adding emojis to files unless asked.
-- The edit will FAIL if `old_string` is not unique in the file. Either provide a larger string with more surrounding context to make it unique or use `replace_all` to change every instance of `old_string`.
-- Use `replace_all` for replacing and renaming strings across the file. This parameter is useful if you want to rename a variable for instance."""
+- You must read the file before editing. This tool will error if you attempt an edit without reading the file first.
```
vtrivedy, line 182: it will?

libs/deepagents/deepagents/middleware/filesystem.py
eyurtsev: What do you mean?

eyurtsev (approved):

libs/deepagents/deepagents/middleware/filesystem.py
sydney-runkle: We consolidated the description (it's not as verbose). Is that ok?

libs/deepagents/deepagents/middleware/filesystem.py
eyurtsev: I don't think it does actually. I think it's probably a prompt to trick the LLM

sydney-runkle: yeah it doesn't haha

PR #809, opened 2026-01-16
fix(deepagents): respect continuation markers when reading files

ccurme:
Reverts https://github.com/langchain-ai/deepagents/pull/784

Currently, when we have a file in the filesystem with long lines:
- If you specify `limit=100`, we read 100 lines of the file
- We convert this into an intermediate representation by adding line numbers and breaking up lines over 10,000 characters. For example, if line 3 has 30,000 characters, it will become 3 lines numbered 3.1, 3.2, 3.3.
- The tool returns the entire result.

Here we update the `read_file` tools so that, if `limit` is specified, we respect these continuation markers.

Before: call with `limit=4` returns
```
line 1
line 2
line 3.1
line 3.2
line 3.3
line 4
```

After: call with `limit=4` returns
```
line 1
line 2
line 3.1
line 3.2
```
This lets the model navigate the file using the continuation markers, and reduces the risk of overflow.

### DIFF
--- a/libs/deepagents/tests/unit_tests/test_middleware.py
+++ b/libs/deepagents/tests/unit_tests/test_middleware.py
@@ -770,7 +770,7 @@ class TestFilesystemMiddleware:
         assert "     3\tshort line 3" in lines[2]
 
     def test_format_content_with_line_numbers_long_line_with_continuation(self):
-        """Test that long lines (>10000 chars) are split with continuation markers."""
+        """Test that long lines (>5000 chars) are split with continuation markers."""
         from deepagents.backends.utils import format_content_with_line_numbers
 
         long_line = "a" * 25000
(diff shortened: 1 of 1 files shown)


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -33,10 +33,8 @@
     sanitize_tool_call_id,
     truncate_if_too_long,
 )
-from deepagents.middleware._utils import append_to_system_message
 
 EMPTY_CONTENT_WARNING = "System reminder: File exists but has empty contents"
-MAX_LINE_LENGTH = 2000
 LINE_NUMBER_WIDTH = 6
 DEFAULT_READ_OFFSET = 0
 DEFAULT_READ_LIMIT = 500
```
nhuang-lc: Thoughts on dropping this to something lower like 50?

PR #822, opened 2026-01-17
feat(middleware): add ToolSearchMiddleware for on-demand tool discovery

hooman-bayer:
Implements a tool search middleware that enables on-demand tool discovery when agents have access to many tools (50+). This solves token bloat and accuracy degradation issues by deferring tool definitions until needed (Inspired from claude code recent toolsearch).

## Problem
- Tool definitions consume 10-20K+ tokens upfront, wasting context window
- Tool selection accuracy degrades beyond 30-50 tools
- Most tools unused per session, leading to inefficient token spend

## Solution
A hybrid middleware + meta-tool approach:
1. Index all tools at agent initialization using BM25
2. If tools exceed token threshold (default 10% of context), activate "deferred mode"
3. In deferred mode: only `search_tools` meta-tool + `always_include` tools available
4. Agent calls `search_tools(query)` to discover relevant tools
5. Discovered tools are "expanded" and become available for subsequent calls

## Features
- BM25 search algorithm (zero external dependencies)
- Regex pattern matching with 200-char limit
- Hybrid mode combining BM25 + regex
- Configurable token threshold (float or absolute)
- Always-include tools that bypass search
- Thread-safe initialization with double-checked locking
- Full async/sync support

## Spec Compliance
Based on Anthropic's MCP Tool Search specification:
- MAX_TOOL_CATALOG_SIZE = 10,000
- MAX_SEARCH_RESULTS = 5
- MAX_REGEX_PATTERN_LENGTH = 200
- BM25 IDF formula matches spec exactly

## Performance
- Token savings: 83-88% in typical scenarios
- Indexes 500 tools in <5 seconds
- Search completes in <100ms

## Tests
- 71 unit tests covering BM25, ToolIndex, middleware behavior
- 16 integration tests with realistic MCP server simulations
- Thread safety and concurrent access tests
- Edge case and error handling tests


eyurtsev (changes requested):
@hooman-bayer this implementation would in theory go to langchain rather than deepagnets. 

LangChain already has an implementation for finding tools on demand:
https://docs.langchain.com/oss/python/langchain/middleware/built-in#llm-tool-selector

The selection is based on an LLM (rather than bm25).

eyurtsev:
I'm going to close this PR as it's the wrong library for it. We're keeping deepagents middleware focus as much as possible on deepagents functionality. The ToolSearchMiddleware can be applied to any agent (not just a deepagent).

I don't know if we'd be interested in maintaining this implementation in langchain. I've considered adding bm25 a few months ago and decided against it as I couldn't convince myself that it made sense for the **average use case** when compared to selection via LLMs. The non-average use case, would likely require something more custom and appropriate evals to make sure that things work properly

hooman-bayer:
Thanks @eyurtsev - I thought given Claude code has recently added it and here you guys have CLI might be nice to have it - but I totally understand the concern ( just wanna mention this implement tried to mimic as close as possible Claude Code recent `ToolSearch` capability -  I imagine could be a nice addition to langchain)

Issue #819, opened 2026-01-17
Support for VertexAI and HuggingFace for Deepgents-CLI


AlsonYang:
### Checked other resources

- [x] This is a feature request, not a bug report.
- [x] I searched existing issues and didn't find this feature.
- [x] I checked the docs and README for existing functionality.
- [x] This request applies to this repo (deepagents) and not an external package.

### Area (Required)

- [ ] deepagents (SDK)
- [x] cli

### Feature description

I can specify my own model (via vertexai or huggingface Inference Hub) and create a deepagents within SDK. But when using the cli, it doesn't allow me to do so, and requires api_keys. Would love the cli to support vertexai authentication via project out of box.

Jordanh1996:
This should add support at least for Vertex AI
https://github.com/langchain-ai/deepagents/pull/828

eyurtsev:
I am not sure we can merge this as-is. I think we need a different design to support additional models. In particular, relying entirely on environment variables for configuration is not ideal for a CLI. We will likely want a standard, well-defined config file instead.

PR #828, opened 2026-01-17
feat(cli): Support VertexAI provider on the CLI

Jordanh1996:
## Description
Adds support on the CLI to use VertexAI models(both gemini and Anthropic Claude via Vertex).

## Changes
- Added `langchain-google-vertexai` as an optional dependency to avoid bloating the CLI with more dependencies
- Added detection to see if the user intended to use VertexAI as a provider(having `GOOGLE_CLOUD_PROJECT` env and no `GOOGLE_API_KEY` set)
- Supports both Gemini and Claude(hosted via Vertex)

Partially addresses #819 (VertexAI support)


juanwisz:
please also add aws bedrock and so on

eyurtsev:
I am not sure we can merge this as-is. I think we need a different design to support additional models. In particular, relying entirely on environment variables for configuration is not ideal for a CLI. We will likely want a standard, well-defined config file instead.

eyurtsev:
If the CLI is used in interactive mode:
* We can use a widget / screen for model configuration that will update a standard config file
* In non interactive mode, we can pass a path to a config file explicitly --config.

We'd need to spec this out to really define the desired behavior for both interactive and non interactive modes

Jordanh1996:
@eyurtsev 
That does sound like a great idea.
I do find the env var configuration awkward but as for now when we dont even have a spec why is this a blocker for this?
Since adding another provider and configuring it differently sounds like different deliverables/tasks

eyurtsev:
Fair enough might, we might add.

> we dont even have a spec why is this a blocker for this?

The spec sounds fairly simple, but yes I'm not going to get around to it for a while longer.

At some point, we'll make a breaking change in the CLI and stop reading env variables, but I think that's fine

libs/deepagents-cli/deepagents_cli/config.py
```diff
@@ -510,3 +550,34 @@ def create_model(model_name_override: str | None = None) -> BaseChatModel:
...
+                    "[bold red]Error:[/bold red] langchain-google-vertexai package is required for this model"
+                )
+                console.print("\nInstall it with:")
+                console.print("  pip install deepagents-cli[vertexai]", markup=False)
+                sys.exit(1)
+
+            return ChatAnthropicVertex(
+                # Remove version tag (e.g., "claude-haiku-4-5@20251015" -> "claude-haiku-4-5")
+                # ChatAnthropicVertex expects just the base model name without the @version suffix
+                model_name=model_name.split("@")[0],
```
eyurtsev: Is the version just ignored entirely?

Jordanh1996: If the version is included it breaks
image


I think we can either:
1. ignore it and let it fail as attached
2. validate it and warn/quit on startup etc
3. strip it just like I did

Maybe another idea? I'm not really prone to anything

Jordanh1996:
> Fair enough might, we might add.
> 
> > we dont even have a spec why is this a blocker for this?
> 
> The spec sounds fairly simple, but yes I'm not going to get around to it for a while longer.
> 
> At some point, we'll make a breaking change in the CLI and stop reading env variables, but I think that's fine

I definitely agree the current env var configuration and provider being inferred is relatively awkward.
I'm willing to help with speccing and implementation if you need

Jordanh1996:
> please also add aws bedrock and so on

I'm not against adding it. It's just that I don't have any AWS account + bedrock to validate that anything works. Therefore I can't do that as of now.

libs/deepagents-cli/deepagents_cli/config.py
eyurtsev: Could we let it fail loudly for now, and can relax later as needed? I would prefer a loud failure until we decide how to handle correctly

Jordanh1996: Sure thing. I removed that version stripping logic @eyurtsev

Jordanh1996:
@eyurtsev @mdrxy 
Is it safe to merge? any other notes?

eyurtsev (approved):

Jordanh1996:
@eyurtsev 
CI and conflicts resolved. can be merged

eyurtsev (approved):

eyurtsev:
Thank you very much!

PR #831, opened 2026-01-18
Localize text-to-sql example and use local deps

hhscc-lz:
Translate the text-to-sql example docs to Chinese. Document the local deepagents/langchain dependencies and manual langchain clone workflow. Ignore the local langchain clone in git.


hhscc-lz:
Opened against the wrong repo by mistake. Will reopen against hhscc-lz/deepagents.

PR #832, opened 2026-01-18
Localize text-to-sql example and use local deps

hhscc-lz:
Translate the text-to-sql example docs to Chinese. Document the local deepagents/langchain dependencies and manual langchain clone workflow. Ignore the local langchain clone in git.


eyurtsev:
The maintainers work in English and our code is written entirely in English. If you need to adapter behavior for chinese, you should be able to adjust the system prompt w/ instructions accordingly (llms arefluent in all languages and have no trouble understanding english and responding in chinese)

PR #836, opened 2026-01-18
fix(cli): unify /quit and /exit with alias support in autocomplete

### LINKED ISSUE #651: `quit`/`exit` should both share same dropdown option
Image

these should be on the same line since they are effectively aliases

shuaixr:
Update `SLASH_COMMANDS` to support aliases
Refactor `SlashCommandController` to search by aliases
Remove redundant `/exit` from the dropdown
Add unit tests for alias lookup

close #651


shuaixr:
emm. I just found some cmd configs in https://github.com/langchain-ai/deepagents/blob/6dca0edc3bd7ac863d07f1b5a44f52565812ea1d/libs/deepagents-cli/deepagents_cli/config.py#L59-L65
but they seem like dead code, right?

mdrxy:
Thank you for the contribution. For now, we'll just remove `/exit` as an option.

PR #860, opened 2026-01-21
feat: add optional `system_prompt` to `SkillsMiddleware` init

duanckham:
## Summary

Add an optional `system_prompt` argument to `SkillsMiddleware.__init__` so callers can override the built-in skills system prompt.

## Problem

`SkillsMiddleware` always used the hardcoded `SKILLS_SYSTEM_PROMPT`. There was no way to:
- Customize the skills section (tone, layout, or extra instructions)

## Solution

- Add `system_prompt: str | None = SKILLS_SYSTEM_PROMPT` to `__init__`.
- Store it as `self.system_prompt_template` and use it in `modify_request` when formatting the skills section.
- Omit the argument to keep the existing default; passing a custom string overrides it.

## Changes

**File:** `libs/deepagents/deepagents/middleware/skills.py`

- **`__init__` (L505–516):**
  - New parameter: `system_prompt: str | None = SKILLS_SYSTEM_PROMPT`
  - Docstring: `system_prompt: Optional custom system prompt override.`
  - Assignment: `self.system_prompt_template = system_prompt`
- **`modify_request`:** Unchanged; it already uses `self.system_prompt_template.format(skills_locations=..., skills_list=...)`.

## Backward compatibility

- `system_prompt` is optional and defaults to `SKILLS_SYSTEM_PROMPT`.
- All existing `SkillsMiddleware(backend=..., sources=...)` call sites continue to work without changes.

### DIFF
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -502,17 +502,18 @@ class SkillsMiddleware(AgentMiddleware):
 
     state_schema = SkillsState
 
-    def __init__(self, *, backend: BACKEND_TYPES, sources: list[str]) -> None:
+    def __init__(self, *, backend: BACKEND_TYPES, sources: list[str], system_prompt: str | None = SKILLS_SYSTEM_PROMPT) -> None:
         """Initialize the skills middleware.
 
         Args:
             backend: Backend instance or factory function that takes runtime and returns a backend.
                      Use a factory for StateBackend: `lambda rt: StateBackend(rt)`
             sources: List of skill source paths (e.g., ["/skills/user/", "/skills/project/"]).
+            system_prompt: Optional custom system prompt override.
         """
         self._backend = backend
         self.sources = sources
-        self.system_prompt_template = SKILLS_SYSTEM_PROMPT
+        self.system_prompt_template = system_prompt
 
     def _get_backend(self, state: SkillsState, runtime: Runtime, config: RunnableConfig) -> BackendProtocol:
         """Resolve backend from instance or factory.


eyurtsev:
The philosophy in deepagents in general is to aim for good defaults and remove the ability of developers to customize certain things.

* What is failing that you think requires a better prompt for middleware?
* Can we just improve the skills system prompt?

We*d **much** rather improve the defaults, then to make things configurable in general

duanckham:
> The philosophy in deepagents in general is to aim for good defaults and remove the ability of developers to customize certain things.
> 
> * What is failing that you think requires a better prompt for middleware?
> * Can we just improve the skills system prompt?
> 
> We*d **much** rather improve the defaults, then to make things configurable in general

First, we keep all prompts that may need tweaking in the cloud instead of local files. This lets us roll out hot updates quickly and run gray-scale A/B tests.

Second, we’re not entirely sure whether this matches industry best practices. But given how fast AI tech is evolving, it’s normal for different approaches to emerge. In our experience, explicitly including both good and bad examples in skills makes skill loading much more stable.

There are also a few smaller issues. For example, for very simple tasks (just 1–2 tool calls), we’d like the planning module to invoke tools directly. But for other simple tasks—even ones that could technically be done with a single tool call—we prefer to load a skill first to handle them.

That said, I understand your intention to keep the core capabilities clean and minimal. Still, I think it’s important to leave room for extensibility around key concepts like skills and sub-agents, so they can better support different business scenarios.

eyurtsev:
> Second, we’re not entirely sure whether this matches industry best practices. But given how fast AI tech is evolving, it’s normal for different approaches to emerge. In our experience, explicitly including both good and bad examples in skills makes skill loading much more stable.

> There are also a few smaller issues. For example, for very simple tasks (just 1–2 tool calls), we’d like the planning module to invoke tools directly. But for other simple tasks—even ones that could technically be done with a single tool call—we prefer to load a skill first to 
handle them.

In both cases, you should be able to update the system message of the agent itself with further instructions without having to access the middleware prompt message.

You can also create custom middleware that updates the system message if you need more control about the placement of your text w/ respect to other content blocks that are present in the system message.

---

system_prompt at the level of the SkillsMiddleware has hidden information (interpolated string paarameters) and I'd like to keep this non configurable, so when we can change the structure of the prompts and improve them without risking breaking user code

duanckham:
Ok, thanks a lot for your response. I will close this pr.

PR #863, opened 2026-01-21
fix(sdk): Improve usability by adjusting the default value of the general-purpose agent.

CLFutureX:
## Background
The `SubAgentMiddleware` currently creates a general-purpose agent by default. In scenarios where custom sub-agents are defined, we typically only care about these custom sub-agents and do not actively configure `default_tools` and `general_purpose_agent` . Additionally, even if `default_tools=[]` is set, the general-purpose agent is still created by default.
Meanwhile, the prompt of the general-purpose agent conflicts with the number of available tools.

For example, in the test case provided in the current project (a general-purpose agent with zero tools is still created at this time):
```python
def test_defined_subagent(self):
    agent = create_agent(
        model="claude-sonnet-4-20250514",
        system_prompt="Use the task tool to call a subagent.",
        middleware=[
            SubAgentMiddleware(
                default_model="claude-sonnet-4-20250514",
                default_tools=[],  # Empty list, but general-purpose agent is still created
                subagents=[
                    {
                        "name": "weather",
                        "description": "This subagent can get weather in cities.",
                        "system_prompt": "Use the get_weather tool to get the weather in a city.",
                        "tools": [get_weather],
                    }
                ],
            )
        ],
    )
```

## Optimization
1. To maintain backward compatibility (keep default values unchanged), add a check for `default_tools` during agent creation: skip creating the general-purpose agent if `default_tools` is empty.


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -286,7 +286,7 @@ def _get_subagents(
     subagent_descriptions = []
 
     # Create general-purpose agent if enabled
-    if general_purpose_agent:
+    if general_purpose_agent and default_tools:
```
eyurtsev: This logic does not look self-consistent...

1. middleware can define tools
2. the function signature uses `general_purpose_agent` as the parameter that controls whether a general purpose agent is included or not, conditioning on another parameter breaks these semantics.

CLFutureX: > This logic does not look self-consistent...
> 
> 1. middleware can define tools
> 2. the function signature uses `general_purpose_agent` as the parameter that controls whether a general purpose agent is included or not, conditioning on another parameter breaks these semantics.

You're  right,  
Actually, the original purpose of this PR was to enable on-demand instantiation of the general-purpose agent. For sub-agents, a general-purpose agent is typically unnecessary in most cases. However, the current setup creates one by default — users have to manually add extra configuration items if they want to disable it.
Given this issue, we might as well adjust the default value to False sooner rather than later (it will cause a backward compatibility issue).

CLFutureX: > You're right, Actually, the original purpose of this PR was to enable on-demand instantiation of the general-purpose agent. For sub-agents, a general-purpose agent is typically unnecessary in most cases. However, the current setup creates one by default — users have to manually add extra configuration items if they want to disable it. Given this issue, we might as well adjust the default value to False sooner rather than later (it will cause a backward compatibility issue).

Additionally, regarding the point you mentioned – "middleware can define tools" – I’ve identified a potential issue:
In the current implementation of create_deep_agent, the subagent_middleware does not inherit the user-defined middleware.
This could lead to a mismatch between the tools available to the Main Agent and the general-purpose agent.
As a result, it would contradict the prompt description of the general-purpose agent, which states: "This agent has access to all tools as the main agent."
Eventually, this tool capability gap might cause task execution failures.

sydney-runkle:
Closing for now, we don't want to change the default in the short term.

PR #880, opened 2026-01-22
feat(cli): add Koyeb as a sandbox provider 

jenperson:
[Koyeb Sandboxes](https://www.koyeb.com/docs/sandboxes) provide fully isolated, ephemeral environments on CPU or GPU. They are a great addition to the options available for running LangChain Deep Agents.

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/integrations/koyeb.py
+++ b/libs/deepagents-cli/deepagents_cli/integrations/koyeb.py
@@ -129,14 +129,12 @@ class KoyebBackend(BaseSandbox):
                         path, encoding="base64"
                     )
                 except SandboxFilesystemError as e:
-                    # If base64 decoding failed, try UTF-8
                     error_msg = str(e).lower()
                     if "base64" in error_msg or "invalid" in error_msg:
                         file_info = await self._sandbox.filesystem.read_file(
                             path, encoding="utf-8"
                         )
                     else:
-                        # Not a base64 error, re-raise to outer handler
                         raise
 
                 # Process the file content
@@ -159,7 +157,6 @@ class KoyebBackend(BaseSandbox):
                     )
                 )
             except SandboxFilesystemError:
-                # Generic filesystem error - use invalid_path as catch-all
                 responses.append(
                     FileDownloadResponse(
                         path=path,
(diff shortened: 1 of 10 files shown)


eyurtsev:
We are not currently accepting 3rd party sandbox integrations into the CLI. We'll post guidelines within ~1-2 weeks on how add a 3rd party sandbox integration. it'll be an optional dependency that you'll need to manage and maintain

libs/deepagents-cli/deepagents_cli/integrations/koyeb.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        self._sandbox = sandbox
+        self._timeout: int = 30 * 60  # 30 mins
+
+    @property
+    def id(self) -> str:
+        """Unique identifier for the sandbox backend."""
+        return self._sandbox.id
+
+    def execute(
```
eyurtsev, line 41: If your sandbox is capable of async code, then you should define the async path instead, and can leave the sync path unimplemented

eyurtsev (changes requested):
Not ready to accept integrations at the moment

libs/deepagents-cli/deepagents_cli/integrations/koyeb.py
jenperson: Correct me if I'm wrong, but the integration tests are all synchronous, so I can't leave the sync path unimplemented

mdrxy:
Hey @jenperson — thank you for your patience here, and apologies that we left this hanging. Eugene mentioned we'd post guidelines for third-party sandbox integrations within a couple of weeks, and we didn't follow through on that — we're sorry for the false start, especially given that you put in the work to address the review feedback.

We've since reassessed and have decided to close this PR. The CLI is likely entering maintenance mode, and rather than continuing to accept integrations directly, we want to first establish better scaffolding for how third-party providers get added and maintained. That's the work that needs to happen before something like this could land cleanly.

If you're open to it, we'd encourage you to open an issue to track Koyeb as a sandbox provider. That gives us a place to revisit this once the integration story is more mature, and helps us gauge community interest in the meantime.

Thanks again for the contribution — we appreciate the effort you put in.

PR #891, opened 2026-01-23
feat: add Moonshot (Kimi) model support

markshao:


eyurtsev:
We'll add support for this in a different way once we add model configuration.

PR #897, opened 2026-01-23
feat(deepagents): truncate tool messages from read_file

ccurme:


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -367,6 +369,18 @@ def sync_read_file(
             lines = lines[:limit]
             result = "".join(lines)
 
+        # Check if result exceeds token threshold and truncate if necessary
+        if token_limit_before_truncation and len(result) >= 4 * token_limit_before_truncation:
```
sydney-runkle: 4 feels like a magic number here, why 4?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -367,6 +369,18 @@ def sync_read_file(
...
+            truncate_threshold = 4 * token_limit_before_truncation
+            result = result[:truncate_threshold]
+            result += (
```
sydney-runkle, line 381: can centralize and use in both?

libs/deepagents/deepagents/middleware/filesystem.py
sydney-runkle: I see we use elsewhere, can hard code somewhere as a var w/

```
# Check if content exceeds eviction threshold
        # Using 4 chars per token as a conservative approximation (actual ratio varies by content)
        # This errs on the high side to avoid premature eviction of content that might fit
```

PR #900, opened 2026-01-23
feat(sdk,cli): sandbox provider interface

eyurtsev:
# Add SandboxProvider Abstraction

Adds a `SandboxProvider` protocol for third-party sandbox integrations. Migrated existing Modal, Runloop, and Daytona implementations to use it.

## Usage

```python
from deepagents.backends.sandbox import SandboxProvider

class MyProvider(SandboxProvider):
    def list(self, *, cursor=None, **kwargs):
        # Return available sandboxes
        return {"items": [...], "cursor": None}
    
    def get_or_create(self, *, sandbox_id=None, **kwargs):
        # Create new or connect to existing sandbox
        if sandbox_id:
            sandbox = self._client.get(sandbox_id)
        else:
            sandbox = self._client.create()
        return MySandboxBackend(sandbox)
    
    def delete(self, sandbox_id, **kwargs):
        # Clean up sandbox
        self._client.terminate(sandbox_id)

# Use it
provider = MyProvider()
sandbox = provider.get_or_create()
result = sandbox.execute("echo hello")
provider.delete(sandbox.id)
```

The deepagents CLI now uses this pattern for Modal, Runloop, and Daytona


libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -23,6 +30,340 @@
...
+        Returns:
+            An object implementing SandboxBackendProtocol.
+        """
+        return await asyncio.to_thread(self.get_or_create, sandbox_id=sandbox_id, **kwargs)
+
+    async def adelete(
+        self,
+        sandbox_id: str,
+        **kwargs: Any,
+    ) -> None:
```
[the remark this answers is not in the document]
sydney-runkle: good suggestion

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -23,6 +30,340 @@
...
+            self, *, cursor=None, **kwargs: Any
+        ) -> SandboxListResponse[ProviderMetadata]:
+            # Extract kwargs as needed
+            status = kwargs.get("status")
+            ...
+    ```
+"""
+
+
+class SandboxInfo(TypedDict, Generic[MetadataT]):
```
sydney-runkle, line 58: I like this method for extension

libs/cli/deepagents_cli/integrations/sandbox_factory.py
```diff
@@ -341,8 +132,33 @@ def get_default_working_dir(provider: str) -> str:
...
+        provider_name: Name of the provider ("modal", "runloop", "daytona")
+
+    Returns:
+        SandboxProvider instance
+
+    Raises:
+        ValueError: If provider_name is unknown
+        ImportError: If required SDK is not installed
+    """
+    if provider_name == "modal":
```
sydney-runkle, line 155: should we do a dict lookup?

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -23,6 +30,340 @@
     WriteResult,
 )
 
+# Type variable for provider-specific metadata
+MetadataT = TypeVar("MetadataT", covariant=True)
```
sydney-runkle, line 34: I think this should have a default value (perhaps `None` or `dict[str, Any]`). We don't want implementations to have to specify an untyped dict :/

eyurtsev: Implementations should specify a typed dict. Is there an easy way to do that? (i.e., bound to typed dict?)

libs/deepagents/deepagents/backends/sandbox.py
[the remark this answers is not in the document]
eyurtsev: Can change -- I actually think the `*` in this case is annoying unless we support sandbox_id: Optionals[str] = None (to add support for delete by query)

(i.e., the `*` was intentionally left out in this case)

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -23,6 +30,340 @@
...
+            if response["cursor"]:
+                next_response = provider.list(cursor=response["cursor"])
+
+            # With filters (if provider supports them)
+            running = provider.list(status="running")
+            ```
+        """
+
+    @abstractmethod
+    def get_or_create(
```
ccurme, line 220: why combine these into one method? I was confused initially about what to expect if I pass a `sandbox_id` that doesn't exist (does it create one and assign it that ID vs. error)

eyurtsev: The reason to combine into 1, is to avoid 2 network calls when a single network call can achieve the desired result.

We can document the convention that if a sandbox-id  doesn't exist, it should raise. Implementations can actually add a kwarg called "create_if_not_exists" to support a different behavior. But the default should be to raise

libs/cli/deepagents_cli/integrations/sandbox_factory.py
eyurtsev: not worth changes right now -- it's an entirely internal refactor

libs/deepagents/deepagents/backends/sandbox.py
sydney-runkle: Unfortunately not, that's why we have those odd things in langgraph (typed dict meta)

sydney-runkle: Right but evne your implementations aren't specifying a typed dict, so we should have a default?

libs/deepagents/deepagents/backends/sandbox.py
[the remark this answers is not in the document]
eyurtsev: Added `*`

PR #901, opened 2026-01-23
feat(deepagents): truncate middle of content instead of end on tool response eviction

ccurme:


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -834,11 +834,40 @@ def _get_filesystem_tools(
 You can do this by specifying an offset and limit in the read_file tool call.
 For example, to read the first 100 lines, you can use the read_file tool with offset=0 and limit=100.
 
-Here are the first 10 lines of the result:
+Here is a preview showing the head and tail of the result:
 {content_sample}
 """
 
 
+def _create_content_preview(content_str: str, head_lines: int = 5, tail_lines: int = 5) -> str:
```
eyurtsev: Use `*`

```suggestion
def _create_content_preview(content_str: str, *, head_lines: int = 5, tail_lines: int = 5) -> str:
```

PR #902, opened 2026-01-23
Add soccer deepagent + Runloop workflow

omarubilla:
lit


vtrivedy:
nice example! we won't be adding to main repo

PR #909, opened 2026-01-24
feat(cli): non-interactive mode w/ shell allow-listing

jkennedyvz:
Add a non-interactive execution mode (`-n`/`--non-interactive`) that runs a single task and exits without launching the UI, along with a restricted, shell command allow-list feature to control which local shell commands the agent can auto-approve.

Shell is disabled entirely in non-interactive mode unless explicitly opted in via `--shell-allow-list`. This also filters auto-approval in the interactive mode.

### Key Changes:

**Non-Interactive Mode**
- New `-n "message"` flag as mentioned above. This will stream tool calls and responses to stdout!

**Shell Security (Allow-List)**
- **Shell disabled by default** in non-interactive mode for security
- New `--shell-allow-list` flag to enable specific shell commands
  - `--shell-allow-list recommended` enables a curated safe list (read-only commands like `ls`, `cat`, `grep`, `head`, `tail`, `wc`, etc.). You should always audit this for your particular use case.
  - `--shell-allow-list ls,cat,grep` for custom allow-lists
- Environment variable support: `DEEPAGENTS_SHELL_ALLOW_LIST`

**Injection Prevention**
- Blocks dangerous shell patterns before parsing (command substitution, redirects, process substitution, here-docs, ANSI-C quoting, etc.)
- Validates all segments in piped/chained commands

**Other**
- Extract `get_langsmith_project_name()` and `fetch_langsmith_project_url()` from `welcome.py` into `config.py` so the non-interactive header can render clickable LangSmith thread links

image


jkennedyvz:
It may be worth disabling the shell tool by default in noni

mdrxy (changes requested):
```python
from deepagents_cli.config import _parse_shell_allow_list
```

Should this be made public?

`_make_hitl_decision()` - I'm not seeing this tested, same with auto-approve logic in `app.py`. Can we also directly test `_contains_dangerous_patterns()`?

Docstring in `non_interactive.py` claims function "auto-approves all actions" but actual behavior depends on shell_allow_list configuration.

nb: can we reduce use of `Any` types?

mdrxy (approved):

PR #916, opened 2026-01-25
feat(cli): custom subagents

vtrivedy:
### Summary                                                                                                                                                                                                             
                                                                                                                                                                                                                      
  Adds filesystem-based custom subagent support. Users define subagents as markdown files with YAML frontmatter in .deepagents/agents/{name}/{name}.md. These are automatically loaded at CLI startup and available   
  via the task tool.                                                                                                                                                                                                  
                                                                                                                                                                                                                      
  ### How It Works                                                                                                                                                                                                        
                                                                                                                                                                                                                      
  - User creates .deepagents/agents/researcher/researcher.md with frontmatter (name, description, optional model) and body (system prompt)                                                                            
  - CLI loads these at startup via list_subagents()                                                                                                                                                                   
  - Passed to SubAgentMiddleware which injects them into the system prompt and task tool description                                                                                                                  
  - When invoked, subagent receives its custom system prompt and inherits main agent's tools                                                                                                                          
                                                                                                                                                                                                                      
  ### Implementation                                                                                                                                                                                                      
                                                                                                                                                                                                                      
  - deepagents/middleware/subagents.py - Make tools optional in SubAgent, inject subagent list into sys prompt                                                                                                     
  - deepagents_cli/subagents.py - New loader for markdown+frontmatter files                                                                                                                                           
  - deepagents_cli/config.py - Add get_user_agents_dir(), get_project_agents_dir()                                                                                                                                    
  - deepagents_cli/agent.py - Load subagents, pass to create_deep_agent()                                                                                                                                             
                                                                                                                                                                                                                      
  ### Tests                                                                                                                                                                                                               
  - all unit tests                                                                                                                                                                                                                   
  - frontmatter parsing, directory scanning, and user/project loading


srtab:
@vtrivedy Wouldn't it be interesting to define this in the SDK, similar to skills?

vtrivedy:
> @vtrivedy Wouldn't it be interesting to define this in the SDK, similar to skills?

yes, once we validate the pattern in CLI we'll look to port to SDK

hwchase17:
i think it should be `.deepagents/agents/{name}/AGENTS.md`. this makes it easy to port a main agent to a subagent? would also kind of be nice if it didn't require adding the frontmatter maybe?

hwchase17:
not sure if we want it to, but right now the main agent doesnt know where in the filesystem the subagents are defined, so it cant update them

ccurme (approved):

PR #918, opened 2026-01-25
fix(cli,sdk): harden path hardening

jkennedyvz:
Normalize all virtual filesystem paths to use forward slashes regardless of OS, fixing breakage on Windows where `Path.relative_to()` and `str(Path)` produce backslash-separated strings. The old approach used manual string-prefix stripping to convert absolute paths to virtual paths — fragile and platform-dependent. Also consolidates the duplicated `_validate_path` function into a single canonical location and fixes a false-positive traversal rejection for filenames containing `..` as a substring (e.g., `foo..bar.txt`).

## Before / After

**Virtual paths on Windows:**
```python
# Before: str(Path) produces backslashes on Windows
"/src\utils\file.py"

# After: .as_posix() always produces forward slashes
"/src/utils/file.py"
```

**Filenames containing `..` as a substring:**
```python
# Before: substring check rejects legitimate filenames
_validate_path("foo..bar.txt")  # ValueError: Path traversal not allowed

# After: component check only rejects actual traversal
_validate_path("foo..bar.txt")   # "/foo..bar.txt"
_validate_path("foo/../etc")     # ValueError (real traversal)
```

**Autocomplete scoring with backslash paths:**
```python
# Before: backslash paths scored differently / broke filename extraction
_fuzzy_score("helper", "src\\utils\\helper.py")  # low/broken score

# After: normalized before scoring, matches identically
_fuzzy_score("helper", "src\\utils\\helper.py")  # same score as "src/utils/helper.py"
```


mdrxy (approved):
good for CLI, @eyurtsev or @sydney-runkle for SDK

PR #920, opened 2026-01-26
fix(sdk): summarization offloading

ccurme:


libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -92,7 +93,7 @@ def __init__(
         token_counter: TokenCounter = count_tokens_approximately,
         summary_prompt: str = DEFAULT_SUMMARY_PROMPT,
         trim_tokens_to_summarize: int | None = _DEFAULT_TRIM_TOKEN_LIMIT,
-        history_path_prefix: str = "/conversation_history",
+        history_path_prefix: str = "conversation_history",
```
ccurme: not sure if correct fix but this is the source of the error

libs/deepagents-cli/tests/unit_tests/test_end_to_end.py
```diff
@@ -138,6 +138,45 @@ def test_cli_agent_with_fake_llm_basic(self, tmp_path: Path) -> None:
             final_ai_message = ai_messages[-1]
             assert "Task completed successfully!" in final_ai_message.content
 
+    def test_cli_agent_summarizes(self, tmp_path: Path) -> None:
```
ccurme, line 141: this test fails on master

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -658,8 +659,10 @@ def before_model(
         backend = self._get_backend(state, runtime)
         file_path = self._offload_to_backend(backend, messages_to_summarize)
         if file_path is None:
-            # Offloading failed - don't proceed with summarization to preserve messages
-            return None
+            warnings.warn(
+                "Offloading conversation history to backend failed during summarization.",
+                stacklevel=2,
+            )
```
mdrxy, line 665: I think I disagree with this still

mdrxy (commented):
I think we should merge with the `history_path_prefix` fixed and then follow up with a PR changing the behavior when offloading fails once we decide

libs/deepagents/deepagents/middleware/summarization.py
ccurme: should we remove the exception handling here and change the return type to str? https://github.com/langchain-ai/deepagents/blob/629c2a316330e4304fdcfd4264cb87c12b224ce4/libs/deepagents/deepagents/middleware/summarization.py#L455C10-L455C21

currently we don't error, I think warning is a step up from that, but I'm OK to raise as well.

eyurtsev: Raise would be my default expectation

eyurtsev: Raise would be my default expectation

eyurtsev (approved):

PR #931, opened 2026-01-26
feat(sdk): add allowed_paths to filesystembackend, rename virtual_mode (wip)

eyurtsev:
Deprecates `virtual_mode` in FilesystemBackend in favor of two new parameters: `restrict_to_root` (bool) restricts access to root_dir, and `allowed_paths` (list[str|Path]) allows additional directories. The deprecation includes proper warnings and will be removed in deepagents 0.5, at which point `restrict_to_root` will default to True.


libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -60,90 +61,180 @@ class FilesystemBackend(BackendProtocol):
...
     """
 
     def __init__(
         self,
         root_dir: str | Path | None = None,
-        virtual_mode: bool = False,
+        virtual_mode: bool | None = None,
+        *,
+        restrict_to_root: bool | None = None,
+        allowed_paths: list[str | Path] | None = None,
```
sydney-runkle, line 100: Should we support patterns here?

libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -60,90 +61,180 @@ class FilesystemBackend(BackendProtocol):
...
+
+        if restrict_to_root is not None:
+            # User explicitly set restrict_to_root - no warnings needed
+            # (This is the desired end state)
+            pass
+        elif virtual_mode is not None:
+            # User is using deprecated virtual_mode parameter
+            warnings.warn(
+                "The 'virtual_mode' parameter is deprecated and will be removed in a future version. "
+                "Use 'restrict_to_root' instead for the same behavior.",
```
sydney-runkle, line 178: This could be more explicit (this line specifically)

libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -60,90 +61,180 @@ class FilesystemBackend(BackendProtocol):
...
+            restrict_to_root = virtual_mode
+        else:
+            # Neither parameter specified - warn about missing explicit choice
+            warnings.warn(
+                "The default value of 'restrict_to_root' will change to True in a future version for security. "
+                "Explicitly set restrict_to_root=False or restrict_to_root=True.",
+                FutureWarning,
+                stacklevel=2,
+            )
+            restrict_to_root = False
```
sydney-runkle, line 192: Why not default to True now? Bc breaking?

libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -60,90 +61,180 @@ class FilesystemBackend(BackendProtocol):
...
+            # Only /data and /tmp are accessible
+            backend = FilesystemBackend(
+                root_dir="/project",
+                restrict_to_root=False,  # Optional: makes intent explicit
+                allowed_paths=["/data", "/tmp"],
+            )
+            ```
         """
+        # Handle parameter migration and defaults
+        # Goal: Migrate from virtual_mode AND require explicit restrict_to_root
```
sydney-runkle: What is the timeline for this?

libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -60,90 +61,180 @@ class FilesystemBackend(BackendProtocol):
...
+        4. Use `restrict_to_root=True` with `root_dir` to enable path-based
+           access restrictions (blocks `..`, `~`, and absolute paths outside root).
+           In a future version, `restrict_to_root` will default to `True` for safety.
     """
 
     def __init__(
         self,
         root_dir: str | Path | None = None,
-        virtual_mode: bool = False,
+        virtual_mode: bool | None = None,
```
sydney-runkle, line 97: Should we add an overload so that when `virtual_mode` is specified, type checkers raise a warning?

sydney-runkle (commented):
Noice, API seems reasonable :)

libs/deepagents/deepagents/backends/filesystem.py
eyurtsev: Yes -- it's breaking

eyurtsev:
Re: pattern support for allowed_paths - that's an interesting idea! We could support glob patterns like `/home/*/projects` or `/data/**`. However, I think for the initial implementation, keeping it simple with exact directory paths is clearer and easier to reason about for security purposes.

If there's a strong use case for patterns, we could add it in a follow-up. What do you think? @sydney-runkle

---
^ Deepagents generated comment :)

libs/deepagents/deepagents/backends/filesystem.py
eyurtsev: No, I don't think we should at the moment. We'd prefer to keep this as simple as possible to reason about

eyurtsev:
Reviewing logic w/ respect to virtual_mode as it's also re-writing paths relative to the root directory (which is orthogonal to access control)

PR #933, opened 2026-01-27
fix(sdk): prevent skills state disorder when leader and subagent both have skills

### LINKED ISSUE #798: In version 0.3.6 of DeepAgents, when both the leader and sub-agent have skills, the skills state becomes disordered.
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

```python
I discovered that in the implementation of skillsMiddleware, when the modify_request method is executed, it retrieves the current skills_locations, but the actual skills loaded are those in the request.
```

```shell

```

### Description

Image

### Environment / System Info

In deepagents, both leader and sub_agent are created using deep_agent but reference different skills. When the leader triggers the sub_agent via the task tool, the leader’s skills_meta_data gets replaced/overwritten by the sub_agent’s skills metadata.  This causes the leader to lose its own skills context after the sub-agent execution.

shilicqupt:
… skills

When both leader and subagent have skills configured, the subagent's skills_metadata was overwriting the leader's skills context after subagent execution. This happened because skills_metadata was shared between agents and not properly isolated.

This fix adds 'skills_metadata' to _EXCLUDED_STATE_KEYS to ensure:
- Subagents start with fresh skills state and load their own skills independently
- Subagent skills metadata doesn't get returned to the parent
- Leader maintains its own skills context throughout execution

Fixes #798

🤖 Generated with [Claude Code]

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -110,7 +110,9 @@ DEFAULT_SUBAGENT_PROMPT = "In order to complete the objective that the user asks
 # 1. The messages key is handled explicitly to ensure only the final message is included
 # 2. The todos and structured_response keys are excluded as they do not have a defined reducer
 #    and no clear meaning for returning them from a subagent to the main agent.
-_EXCLUDED_STATE_KEYS = {"messages", "todos", "structured_response"}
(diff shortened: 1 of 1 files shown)


eyurtsev:
I believe another PR from the community was merged to address this issue

eyurtsev:
(Appreciate you taking the time to work on the fix @shilicqupt )

PR #938, opened 2026-01-28
fix(sdk): win use fileSystem backend, exists excape root dir

dhammubalakrishnan-collab:
when i use file system use as backend on win,it occur root dir escape and not run my code right.
current_script_path = Path(__file__).resolve()
work_dir = current_script_path.parent/'workdir'
fs_backend = FilesystemBackend(root_dir=str(work_dir))

agent = create_deep_agent(
    model=llm,

    tools=[python_tool],
    skills=['/skills/csv-data-summarizer'],
    system_prompt=research_instructions,
    backend=fs_backend,
    debug=True
)
it occur the error:
image
,i have already put the file in workspace

### DIFF
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -153,12 +153,30 @@ def _validate_path(path: str, *, allowed_prefixes: Sequence[str] | None = None) 
     normalized = os.path.normpath(path)
     normalized = normalized.replace("\\", "/")
 
-    if not normalized.startswith("/"):
-        normalized = f"/{normalized}"
-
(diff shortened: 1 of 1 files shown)


vinsonws:
I’m running into the same issue: skills are not loading correctly when using FilesystemBackend and passing a skills directory to create_deep_agent.
However, it seems that deepagents officially does not plan to support Windows for the time being.（https://github.com/langchain-ai/deepagents/issues/889）

For me, it only has a slight impact on development, as I'm currently using a Linux virtual machine for development.

eyurtsev:
We do not plan on supporting windows until we set up testing on a windows machine on CI otherwise we'll end up with regressions. If you're on windows use WSL for now

PR #943, opened 2026-01-28
feat(sdk): add `MiddlewareConfig` for disabling/replacing built-in middleware

joshb1050:
Adds a new `middleware_config` parameter to `create_deep_agent()` that allows users to disable or replace built-in middleware. Each middleware can be set to:
- `True` (default): use the built-in middleware
- `False`: disable the middleware entirely
- `AgentMiddleware` instance: replace with a custom implementation

This addresses the limitation where users could only add middleware but not remove or suppress the built-in ones.


christophercolumbusdog:
This would be incredibly useful for some projects I'm working on, thanks for pushing this!

eyurtsev:
Could you open an issue for this for discussion?

We've had a number of similar requests for configuration of middleware, but at the moment are not planning on supporting configuration of middleware.

(This PR is somewhat close to how it would look like if we were to add the functionality.)

eyurtsev (changes requested):
At the moment, the position of the package is to not support any configurability of the default middleware that ships with the agent. This is a philosophical position as the goal is to configure good defaults from the get-go so users don't need to change anything.

Could you open an issue w/ a discussion so we can centralize the conversation there and explain what your need is for disabling/replacing middleware? (i.e., now what the feature does, but why specifically do you need it for your use case?)

joshb1050:
@eyurtsev Ah, looks like this is already raised here: https://github.com/langchain-ai/deepagents/discussions/655#discussion-9312143

One thing I want to emphasize is that configurability here isn’t about forcing users to tune things by default. Deepagents can (and should) work sensibly out of the box for most people.

The concern is that without *any* ability to customize, e.g. disabling middleware, changing the inputs to summarization, or swapping the prompt, advanced users quickly hit a ceiling. At that point, the path of least resistance is often to abandon deepagents entirely and go back to a more explicit ReAct-style agent where these decisions are controllable.

The fact that middleware exists at all implies this is intended to be a tunable framework. Given that, I think it’s reasonable to expect some level of granular configurability rather than treating middleware as fixed policy.

From my perspective, exposing these hooks preserves deepagents as a flexible abstraction rather than a closed one, without harming the default experience. This is explicitly an optional feature so that most users do not need to configure it.

eyurtsev:
@joshb1050 thanks for the response

Did you need this feature for a specific use case? If so, what default did you need to change and why?

joshb1050:
@eyurtsev yes, for example tuning when the summarization middleware takes effect and what the prompt is for it.

sydney-runkle:
@joshb1050, what do you want to change specifically?

joshb1050:
@sydney-runkle we were looking to change the thresholds to which it kicks in, and actually just be able to disable some middleware so we can write our own. Being able to experiment here with different settings we would find very valuable.

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #955, opened 2026-01-30
fix(sdk): string interpolation in BaseSandbox write/edit tools 

trindadetiago:
The heredoc-based templates introduced in #872 contain f-string placeholders like {e} and {file_path} that are incorrectly processed by Python's str.format() method, causing KeyError exceptions.

This fix escapes these placeholders as {{e}} and {{file_path}} so they become literal {e} and {file_path} after format() processing, allowing the embedded Python code to work correctly.

Fixes regression introduced in:
https://github.com/langchain-ai/deepagents/pull/872

Changes:
- _WRITE_COMMAND_TEMPLATE: escape {e} and {file_path} in exception/error handlers
- _EDIT_COMMAND_TEMPLATE: escape {e} in exception handler
- Add test_sandbox_backend.py with tests for template formatting

### DIFF
--- a/libs/deepagents/tests/unit_tests/backends/test_sandbox_backend.py
+++ b/libs/deepagents/tests/unit_tests/backends/test_sandbox_backend.py
@@ -1,0 +1,141 @@
+"""Tests for BaseSandbox backend template formatting.
+
+These tests verify that the command templates in BaseSandbox can be properly
+formatted without raising KeyError due to unescaped curly braces.
+
+Related issue: https://github.com/langchain-ai/deepagents/pull/872
+The heredoc templates introduced in PR #872 contain {e} in exception handlers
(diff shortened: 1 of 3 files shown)


libs/deepagents/tests/unit_tests/backends/test_sandbox_backend.py
```diff
@@ -0,0 +1,141 @@
+"""Tests for BaseSandbox backend template formatting.
```
eyurtsev, line 1: Nice thank you -- I'll take over the tests, we have a full test suite in the deepagents-cli, I'm in the process of porting it over into the SDK so it runs on CI, i'll merge the tests together in a bit

eyurtsev:
@trindadetiago thank you! I also ported over our CLI tests so that we run the full test suite as part of CI

PR #959, opened 2026-01-30
fix(cli): use `LocalShellBackend` for shell access, allowing subagent access

vtrivedy:
* CLI will use now use `LocalShellBackend` for shell access instead of custom `ShellMiddleware`
* Subagents can now also use the shell


libs/deepagents/deepagents/graph.py
```diff
@@ -42,12 +42,13 @@ def get_default_model() -> ChatAnthropic:
...
 
 
-def create_deep_agent(
+def create_deep_agent(  # noqa: PLR0912
     model: str | BaseChatModel | None = None,
     tools: Sequence[BaseTool | Callable | dict[str, Any]] | None = None,
     *,
     system_prompt: str | SystemMessage | None = None,
     middleware: Sequence[AgentMiddleware] = (),
+    subagent_middleware: Sequence[AgentMiddleware] | None = None,
```
eyurtsev: I'd like to avoid creating an extra parameter.

We can add middleware for subagents using SubAgent type.

For the general purpose agent, we need to decide if we basically want it to inherit all middleware by default (except for subagents) or if it's restrict to some subset of middleware. But which subset and why?

mdrxy: > For the general purpose agent, we need to decide if we basically want it to inherit all middleware by default (except for subagents) or if it's restrict to some subset of middleware. But which subset and why?

A general-purpose subagent is essentially "the same agent with a fresh context window for a subtask" - it should have the same capabilities but doesn't need to re-initialize session state.

We should exclude `SubAgentMiddleware` to prevent infinite recursion. Also `MemoryMiddleware` since memory is session context; a subagent gets relevant context via the prompt.

PR #961, opened 2026-01-30
feat(cli): add configurable timeout to `ShellMiddleware`

mdrxy:
Introduces per-command timeout overrides to allow for long-running commands.

### DIFF
--- a/libs/cli/deepagents_cli/shell.py
+++ b/libs/cli/deepagents_cli/shell.py
@@ -11,7 +11,7 @@ from langchain.tools import ToolRuntime, tool
 from langchain_core.messages import ToolMessage
 from langchain_core.tools.base import ToolException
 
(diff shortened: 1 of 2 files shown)


libs/cli/deepagents_cli/shell.py
```diff
@@ -50,21 +56,27 @@ def __init__(
...
-            f"be truncated if they exceed the configured timeout or output limits."
+            f"be truncated if they exceed the configured timeout ({self._default_timeout}s) "
+            f"or output limits. Use the optional timeout parameter for long-running commands."
         )
 
         @tool(self._tool_name, description=description)
         def shell_tool(
             command: str,
             runtime: ToolRuntime[None, AgentState],
+            timeout: float | None = None,
```
jkennedyvz: nit: should this be int?

PR #965, opened 2026-01-31
fix(cli): terminal virtualize scrolling to stop perf issues

vtrivedy:
Long conversations cause the CLI to slow down because every chat message stays in the DOM as a Textual widget forever — layout recalculation scales with widget count. This adds a virtualization layer that caps DOM widgets at 50, archives older messages as lightweight dataclasses, and re-hydrates them on scroll-back.

## Solution
Based on Textual's `Log`/`RichLog` patterns:
- `MessageStore` holds all message data (lightweight dataclass)
- Only 50 most recent messages exist as widgets
- Older widgets pruned from DOM, data retained
- Scroll up near top → older messages re-hydrated from data
- Streaming messages protected from pruning

<details>
  <summary>More details</summary>

## Summary
- Adds message virtualization to prevent CLI slowdown when accumulating many messages
- Keeps max 50 widgets in DOM at any time, stores all message data for scroll-back
- Hydrates older messages on-demand when user scrolls up

## Problem
The CLI becomes slow and unresponsive after long conversations because:
- Every message is a widget kept in DOM forever
- Textual's layout recalculates for all widgets on every change

## Changes
- Add `MessageStore` and `MessageData` in `widgets/message_store.py` — a sliding-window manager that stores all messages as dataclasses and tracks which are currently mounted as widgets. `WINDOW_SIZE` (50) and `HYDRATE_BUFFER` (15) control the window bounds.
- `MessageData.from_widget` / `to_widget` round-trip all six message types (`UserMessage`, `AssistantMessage`, `ToolCallMessage`, `ErrorMessage`, `AppMessage`, `DiffMessage`). Tool messages use deferred state (`_deferred_status`, `_deferred_output`, `_deferred_expanded`) to restore status after mount without restarting animations.
- `_mount_message` now serializes each widget into the store on mount and calls `_prune_old_messages` to evict the oldest widgets when the window is exceeded. Pruned widget IDs are tracked so the visible window index stays correct.
- `on_scroll_up` triggers `_check_hydration_needed`, which schedules `_hydrate_messages_above` when the user scrolls within 2x viewport height of the top. Hydrated widgets are inserted before the first visible child, and scroll position is adjusted by a hardcoded per-message height estimate (60px) to prevent visual jumps.
- Streaming messages are protected from pruning via `set_active_message` — the adapter marks the current `AssistantMessage` (now assigned a `uuid`-based ID) as active during streaming and clears it in `_flush_assistant_text_ns`.
- `_clear_messages` now also resets the `MessageStore`, and `ErrorMessage`/`AppMessage` constructors store `_content` for serialization.
- Tests added in `libs/cli/tests/test_message_store.py` covering `MessageData` round-trip serialization and `MessageStore` window management.
</details>


mdrxy:
@vtrivedy I made some updates.. take a look when you get a chance. I need to sit with this PR more before merging probably -- unless it's creating issues more widely than I understand?

PR #975, opened 2026-01-31
fix(sdk): grep should perform literal search instead of regex

eyurtsev:
The grep tool was documented to perform literal text search, but the implementation in some backends treated patterns as regex. This caused errors when searching for common code patterns like `def __init__(self):` or `str | int`, which would fail with "unterminated subpattern" errors. The bug affected FilesystemBackend, StateBackend, and StoreBackend (SandboxBackend was already correct).


eyurtsev:
We need to review whether we want regexp support or just literals

PR #983, opened 2026-02-02
chore(sdk,cli): add `VERSION`

mdrxy:
& also add CI check to prevent drift


eyurtsev (approved):
Why not a unit test per package?

PR #991, opened 2026-02-02
feat(sdk): enable type checking in deepagents and resolve most linting issues

sydney-runkle:
on the back of https://github.com/langchain-ai/langchain/pull/34978

* removing a bunch of the "per file" ignores in `pyproject.toml` and moving to labeled inline ignores
* adding `ty` as part of CI for `libs/deepagents`
* resolving a bunch of type checking issues

notes:
* want to move towards abstractmethod long term
* eventually would like to make everything in the `BackendProtocol` keyword only
* the backend factory stuff is super type unsafe and is the cause of a lot of issues (we artificially create tool runtimes). hopefully will be resolved w/ a backend refactor.


libs/deepagents/deepagents/backends/composite.py
```diff
@@ -412,7 +420,7 @@ def edit(
         file_path: str,
         old_string: str,
         new_string: str,
-        replace_all: bool = False,
+        replace_all: bool | None = None,  # noqa: FBT001
```
sydney-runkle: moving away from boolean default, eventually this should be kw only to resolve FBT001

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -172,6 +172,7 @@ class BackendProtocol(abc.ABC):
     }
     """
 
+    @abc.abstractmethod
```
sydney-runkle: callout, we're actually adding these decorators now

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -549,7 +557,7 @@ def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadRespons
                     error=batch_responses[i].error if i < len(batch_responses) else None,
                 )
 
-        return results  # type: ignore[return-value]
+        return [r for r in results if r is not None]
```
sydney-runkle: this looks a little concerning, but there's not any `None`s bc of above logic

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -578,7 +578,7 @@ def _truncate_args(
         """
         counted_messages = [system_message, *messages] if system_message is not None else messages
         try:
-            total_tokens = self.token_counter(counted_messages, tools=tools)
+            total_tokens = self.token_counter(counted_messages, tools=tools)  # ty: ignore[unknown-argument]
```
sydney-runkle, line 581: tools should be known, maybe we need to bump version?

libs/deepagents/deepagents/graph.py
```diff
@@ -276,12 +268,12 @@ def create_deep_agent(
...
-            *system_prompt.content_blocks,
-            {"type": "text", "text": f"\n\n{BASE_AGENT_PROMPT}"},
-        ]
-        final_system_prompt = SystemMessage(content=new_content)
+        # SystemMessage: append BASE_AGENT_PROMPT to content
+        if isinstance(system_prompt.content, str):
+            final_system_prompt = SystemMessage(content=system_prompt.content + f"\n\n{BASE_AGENT_PROMPT}")
+        else:
+            new_content = [*system_prompt.content, {"type": "text", "text": f"\n\n{BASE_AGENT_PROMPT}"}]
+            final_system_prompt = SystemMessage(content=new_content)
```
sydney-runkle: should we be making one w/ content blocks?

libs/deepagents/tests/unit_tests/backends/test_protocol.py
```diff
@@ -6,15 +6,98 @@
...
+    WriteResult,
+)
 
 
 class BareBackend(BackendProtocol):
-    """Minimal subclass that implements nothing."""
+    """Minimal subclass that delegates to super() so NotImplementedError propagates."""
+
+    def ls_info(self, path: str) -> list[FileInfo]:
+        return super().ls_info(path)
```
sydney-runkle: notable that you can't do this anymore, we should consider

sydney-runkle (commented):
some initial thoughts

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -34,6 +34,23 @@
...
 
 
+def _remap_grep_path(m: GrepMatch, route_prefix: str) -> GrepMatch:
+    """Create a new GrepMatch with the route prefix prepended to the path."""
+    return GrepMatch(path=f"{route_prefix[:-1]}{m['path']}", line=m["line"], text=m["text"])
+
+
+def _remap_file_info_path(fi: FileInfo, route_prefix: str) -> FileInfo:
+    """Create a new FileInfo with the route prefix prepended to the path."""
+    result: FileInfo = {"path": f"{route_prefix[:-1]}{fi['path']}"}
```
eyurtsev: Is this not a shallow copy? Except for path?

libs/deepagents/deepagents/backends/composite.py
eyurtsev: Could we keep as was? And turn off the rule for this line? Having a none as an option is arguably worse

libs/deepagents/deepagents/backends/composite.py
eyurtsev: A cast works as well if so as it looks like the type checker fails to follow? The extra for loop is there to satisfy the type checker

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -327,12 +332,13 @@ async def awrite(
...
         return await asyncio.to_thread(self.write, file_path, content)
 
+    @abc.abstractmethod
     def edit(
         self,
         file_path: str,
         old_string: str,
         new_string: str,
-        replace_all: bool = False,
+        replace_all: bool | None = None,  # noqa: FBT001
```
eyurtsev: Let's avoid the None here

libs/deepagents/deepagents/middleware/summarization.py
sydney-runkle: this is just a quirk of our new token counting, leaving for now

.github/workflows/_lint.yml
```diff
@@ -47,3 +47,9 @@ jobs:
         working-directory: ${{ inputs.working-directory }}
         run: |
           make lint
+
+      - name: "🔎 Run Type Checker"
+        working-directory: ${{ inputs.working-directory }}
```
eyurtsev: enable it as part of linting?

libs/harbor/deepagents_harbor/backend.py
```diff
@@ -178,7 +178,7 @@ async def aedit(
         file_path: str,
         old_string: str,
         new_string: str,
-        replace_all: bool = False,
+        replace_all: bool | None = None,  # noqa: FBT001
```
eyurtsev: this should be bool = False

libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -69,7 +69,7 @@ class FilesystemBackend(BackendProtocol):
     def __init__(
         self,
         root_dir: str | Path | None = None,
-        virtual_mode: bool = False,
+        virtual_mode: bool | None = None,  # noqa: FBT001  # Boolean arg is part of BackendProtocol API
```
eyurtsev: None

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -158,7 +158,7 @@ class EditResult:
     occurrences: int | None = None
 
 
-class BackendProtocol(abc.ABC):
+class BackendProtocol(abc.ABC):  # noqa: B024 — methods raise NotImplementedError instead of using @abstractmethod to avoid breaking subclasses that only implement a subset
```
eyurtsev: fwiw I'd put the reason above the abc.ABC -- this is definitely above 120 characters limit (i think we're at 150 right now, but would be great to adjust down at some point)

libs/deepagents/deepagents/backends/store.py
```diff
@@ -529,9 +526,10 @@ async def aedit(
     def grep_raw(
         self,
         pattern: str,
-        path: str = "/",
+        path: str | None = None,
```
eyurtsev, line 529: Is this change correct?

sydney-runkle: yes we control default downstream

eyurtsev (approved):

Issue #951, opened 2026-01-29
Requesting more examples


saadazhar7G:
### Checked other resources

- [x] This is a feature request, not a bug report.
- [x] I searched existing issues and didn't find this feature.
- [x] I checked the docs and README for existing functionality.
- [x] This request applies to this repo (deepagents) and not an external package.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Feature description

There are too few projects of deep agents and examples to lean off of. I think this deep agents are best way to make agents work for you solving issues. but the documentation and examples needs much more work and examples

keenborder786:
Don't you think, following resources are enough:
- Examples: https://github.com/langchain-ai/deepagents/tree/master/examples
- Youtube Video: https://www.youtube.com/watch?v=5tn6O0uXYEg
- Main Docs: https://docs.langchain.com/oss/python/deepagents/overview

nicholaskuechler:
> Don't you following resources are enough:
> 
>     * Examples: https://github.com/langchain-ai/deepagents/tree/master/examples
> 

The first example I looked at: https://github.com/langchain-ai/deepagents/blob/master/examples/ralph_mode/ralph_mode.py gives me an exception when I try to run it. It seems like the deepagents repo went through a refactor, but some of the examples maybe didn't get refactored? Here's an example exception in ralph_mode.py:

```
from deepagents_cli.execution import execute_task
ModuleNotFoundError: No module named 'deepagents_cli.execution'
```

keenborder786:
@nicholaskuechler I have personally checked each example, and all of them seem to be working except for `ralph_mode`, which you also pointed out. That has now been taken care of.

saadazhar7G:
> Don't you think, following resources are enough:
> 
> * Examples: https://github.com/langchain-ai/deepagents/tree/master/examples
> * Youtube Video: https://www.youtube.com/watch?v=5tn6O0uXYEg
> * Main Docs: https://docs.langchain.com/oss/python/deepagents/overview

nope, there should be more usecases and versatility covered to make this library more adaptable

keenborder786:
@saadazhar7G sounds good, let's keep this issue open, I will also see if I can contribute some examples.
cc: @mdrxy

saadazhar7G:
I can give one idea here
[...]

PR #1008, opened 2026-02-02
fix(cli): Ralph Example fixed for newer version of DeepAgent CLI

keenborder786:
- Simple Change to make sure that the ralph_mode example reflects the newer deep agent cli, as the community was complaining that the current version is not working properly, as pointed out in issue #951


examples/ralph_mode/image_1.png
mdrxy, line 1: screenshot has an artifact on the top right from your system, also the text is rather small

examples/ralph_mode/ralph_mode.py
```diff
@@ -1,14 +1,11 @@
 #!/usr/bin/env python3
 """
-Ralph Mode - Autonomous looping for Deep Agents
-
-Ralph is an autonomous looping pattern created by Geoff Huntley.
-Each loop starts with fresh context. The filesystem and git serve as memory.
+Ralph Mode - Single run for DeepAgents
```
mdrxy:
```suggestion
Ralph Mode - Single run for Deep Agents
```

examples/ralph_mode/ralph_mode.py
```diff
@@ -37,86 +33,32 @@ async def ralph(task: str, max_iterations: int = 0, model_name: str = None):
...
-    for f in sorted(Path(work_dir).rglob("*")):
-        if f.is_file() and ".git" not in str(f):
-            console.print(f"  {f.relative_to(work_dir)}", style="dim")
+    await app.run_async()
 
 
 def main():
     parser = argparse.ArgumentParser(
-        description="Ralph Mode - Autonomous looping for Deep Agents",
+        description="Ralph Mode - Single run for DeepAgents",
```
mdrxy:
```suggestion
        description="Ralph Mode - Single run for Deep Agents",
```

keenborder786:
@mdrxy I have resolved the above issues.

keenborder786:
@mdrxy  any isssues with this one?

mdrxy:
photo is still hard to read, thinking something like this (but don't use this one since it's in local mode)
Screenshot

keenborder786:
@mdrxy okay let me see

keenborder786:
@mdrxy please check now, and for this example we are supposed to use local sandbox.

mdrxy:
@keenborder786 can you describe how this is addressing the problem? it looks like the actual ralph logic has been lost? would like to know more about the motivations driving the proposed changes

keenborder786:
@Mason Daugherty (mdrxy), please see https://github.com/langchain-ai/deepagents/issues/951#issuecomment-3829783996.
The current example is completely outdated and does not run.
The updated DeepAgent app inherently works on an `autonomous looping pattern`, so this example has been updated to reflect that and make it work as well.

mdrxy:
@keenborder786 kindly, can you describe what you mean by:

> The updated DeepAgent app inherently works on an autonomous looping pattern

keenborder786:
@mdrxy, the core autonomous looping pattern (just like for ralph mode) has been moved to `DeepAgentsApp`  itself, specifically in `execute_task_textual`.
Therefore, just running the `DeepAgentsApp` itself now reflects Ralph mode.

keenborder786:
@mdrxy all clear now?

mdrxy:
@keenborder786 no, this fundamentally changes the behavior of the example and loses the Ralph loop which necessitates it in the first place! Appreciate the attempt but closing; we will handle updating this example

keenborder786:
@mdrxy Okay, I still don’t understand how this fundamentally changes the behavior, since DeepAgent also enters the same loop pattern. But if you say so, this definitely needs to be fixed — the current example is completely broken.

mdrxy:
@keenborder786 one thing first: for changes like this, it's worth opening an issue or starting a discussion before putting up a PR. it saves everyone time — a quick back-and-forth upfront would have caught the conceptual mismatch here before 15 commits and a few weeks of rebasing. please keep that in mind for future contributions, both here and across other repos.

the core issue with this PR is that it removes the loop entirely, which defeats the purpose of the example.

the Ralph pattern is specifically about running the agent in an outer loop with fresh context each iteration, while the filesystem (and git) serve as persistent memory across runs. see [Geoff Huntley's original write-up](https://ghuntley.com/ralph/) for more.

each iteration picks up where the last left off by reading what's already on disk, not by holding state in a single long-running session.

what this PR does is replace that with a single-shot call — which is just a normal interactive agent session. The `--iterations` flag is gone, the per-iteration prompt that tells the agent to check existing files and keep building is gone, and the `ArgumentParser` description was even updated to say "Single run for Deep Agents". **this is no longer a Ralph example.**

the API changes look like they're heading in the right direction, but the looping logic needs to be preserved and adapted around the new API. specifically:

- the outer while loop)
- spawning a fresh agent/app instance each iteration (resetting state between runs)
- the prompt that contextualizes each iteration
- the --iterations CLI arg
- the README's "How It Works" section explaining the loop semantics

keenborder786:
@mdrxy Thank you so much for the comprehensive feedback and for sharing [Geoff Huntley’s original write-up](https://ghuntley.com/ralph/)
.

I’d be happy to update this example in a new PR, and I’ll also create a separate issue for it. Since updating this example seems really important for the community.

PR #1009, opened 2026-02-02
feat(sdk): add namespace templating to StoreBackend

eyurtsev:
Adds flexible namespace templating to StoreBackend inspired by langmem's approach. StoreBackend now accepts an optional namespace_template parameter that supports variable substitution from config['configurable'], enabling user-scoped or multi-level namespacing patterns like ('filesystem', '{user_id}') or ('workspace', '{workspace_id}', 'user', '{user_id}'). The implementation preserves backwards compatibility by falling back to legacy assistant_id detection when no template is provided.


sydney-runkle:
Addressed via https://github.com/langchain-ai/deepagents/pull/1169

PR #1017, opened 2026-02-02
fix(sdk): handle exact file paths in grep and glob operations

eyurtsev:
Fixes a bug where grep and glob operations fail when called with exact file paths, commonly occurring when agents search evicted large tool results.

## The Bug

When agents grep an exact file path like `/large_tool_results/toolu_01ABC`, the operation fails with "No matches found" despite the file existing. Root cause: path normalization was adding trailing slashes to all paths, breaking exact file matches since file keys in state have no trailing slash.

## The Fix

Replaced `_validate_path` with `_normalize_path` that strips trailing slashes (except for root `/`). Added `_filter_files_by_path` helper that intelligently handles both exact file matches and directory prefixes by checking for exact matches first before falling back to directory prefix matching.


eyurtsev:
> Why do we also have _validate_path in FileSystemMiddleware?

We might need to do another clean up for this

Issue #802, opened 2026-01-16
`create_deep_agent` subagents cannot customize `SummarizationMiddleware` parameters


nick-youngblut:
### Checked other resources

- [x] This is a feature request, not a bug report.
- [x] I searched existing issues and didn't find this feature.
- [x] I checked the docs and README for existing functionality.
- [x] This request applies to this repo (deepagents) and not an external package.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Feature description

When using `create_deep_agent` with custom subagents, the `SummarizationMiddleware` configuration is inherited from the main agent via `default_middleware`. This creates problems when subagents use different models with smaller context windows, as the summarization trigger thresholds are calculated based on the main agent's model.

### Use case

```python
# Main agent uses a large model
agent = create_deep_agent(
    model="anthropic:claude-sonnet-4-5-20250929",  # 200k context
    subagents=[
        {
            "name": "fast_processor",
            "description": "Quick processing tasks",
            "system_prompt": "You process data quickly.",
            "tools": [],
            "model": "openai:gpt-4o-mini",     # 128k context - SMALLER
            # Cannot customize SummarizationMiddleware trigger for this smaller context!
        }
    ],
)
```

In this case, the subagent's `SummarizationMiddleware` will use trigger thresholds appropriate for 200k context, not 128k.

## Option A: Allow middleware override in SubAgent spec

Add logic to detect and replace middleware of the same type rather than merging:

```python
def _get_subagents(...):
    for agent_ in subagents:
        if "middleware" in agent_:
            # Build middleware list, allowing subagent to override specific types
            _middleware = _merge_middleware_with_override(
                default_subagent_middleware, 
                agent_["middleware"]
            )
        else:
            _middleware = [*default_subagent_middleware]
```

### Option B: Recalculate summarization trigger per subagent model

When a subagent specifies a different model, recalculate the SummarizationMiddleware trigger based on that model's context window:

```python
def _get_subagents(...):
    for agent_ in subagents:
        subagent_model = agent_.get("model", default_model)
        
        # Recalculate trigger if model differs
        if subagent_model != default_model:
            _middleware = _build_middleware_for_model(subagent_model, ...)
        else:
            _middleware = [*default_subagent_middleware]
```

### Option C: Add `middleware_overrides` field to SubAgent

Allow targeted overrides without requiring full middleware specification:

```python
class SubAgent(TypedDict):
    # ... existing fields ...
[...]

PR #1018, opened 2026-02-02
poc(sdk): cleaner subagent middleware

sydney-runkle:
Our middlewares are rapidly changing, I propose we have private interfaces for them in the short term so that we can easily make breaking changes.

This simplifies things by:
* Not special casing logic for the "general purpose" subagent - we just build it within `create_deep_agent`
* Passing the `backend` directly to the subagent middleware so that we can build dependent middlewares there (which we'll need for subagent skills)
* Uses subagent models for summarization, rather than the main agent model (fixes bug reported in https://github.com/langchain-ai/deepagents/issues/802)
* Moving summarization param logic into a helper function to declutter `create_deep_agent`

The new subagents file removes ~150 LOC compared to old one

old signature

```py
def __init__(
        self,
        *,
        # these are all just for gen purpose dude
        default_model: str | BaseChatModel,  
        default_tools: Sequence[BaseTool | Callable | dict[str, Any]] | None = None,
        general_purpose_agent: bool = True,
        general_purpose_middleware: list[AgentMiddleware] | None = None,
        default_interrupt_on: dict[str, bool | InterruptOnConfig] | None = None,
        # these are valuable
        default_middleware: list[AgentMiddleware] | None = None,
    ) -> None:
```

new signature

```py
def __init__(
        self,
        *,
        subagents: list[SubAgent | CompiledSubAgent],
        task_description: str | None = TASK_TOOL_DESCRIPTION,
    ) -> None:
```


libs/deepagents/deepagents/middleware/_subagents.py
```diff
@@ -0,0 +1,417 @@
...
+                "trigger": ("fraction", 0.85),
+                "keep": ("fraction", 0.10),
+            }
+        else:
+            trigger = ("tokens", 170000)
+            keep = ("messages", 6)
+            truncate_args_settings = {
+                "trigger": ("messages", 20),
+                "keep": ("messages", 20),
+            }
```
sydney-runkle: should be moved to summarization middleware

libs/deepagents/deepagents/middleware/_subagents.py
```diff
@@ -0,0 +1,399 @@
...
+
+from deepagents.backends.protocol import BackendFactory, BackendProtocol
+from deepagents.middleware._utils import append_to_system_message
+from deepagents.middleware.filesystem import FilesystemMiddleware
+from deepagents.middleware.patch_tool_calls import PatchToolCallsMiddleware
+from deepagents.middleware.subagents import CompiledSubAgent, SubAgent
+from deepagents.middleware.summarization import SummarizationMiddleware, compute_summarization_defaults
+
+# State keys excluded from subagent state transfer
+_EXCLUDED_STATE_KEYS = {"messages", "todos", "structured_response", "skills_metadata"}
```
eyurtsev, line 26: We likely need memory here?

libs/deepagents/deepagents/middleware/_subagents.py
```diff
@@ -0,0 +1,399 @@
...
+        if system_prompt and subagent_descriptions:
+            agents_section = "\n\nAvailable subagent types:\n" + "\n".join(subagent_descriptions)
+            self.system_prompt = system_prompt + agents_section
+        else:
+            self.system_prompt = system_prompt
+
+        # Create the task tool
+        self.tools = [self._create_task_tool(task_description)]
+
+    def _build_base_middleware(self, model: BaseChatModel) -> list[AgentMiddleware]:
```
eyurtsev, line 236: Why not in the init? (The knowledge of what's "default" feels like it should be localized in one place if the default for the main agent and subagents is very similar?)

libs/deepagents/deepagents/middleware/_subagents.py
sydney-runkle: I don't think so, subagents don't have memory at the moment

But yes, if we allow them to

PR #1020, opened 2026-02-03
feat(sdk): Support queuing messages when thread is busy

aran-yogesh:
Implementation

1. Thread active check (`webapp.py`)
- Added `is_thread_active()` function that checks if thread status is `busy`
- Webhook handler now checks thread status before creating a new run

2. Message queuing to store (`webapp.py`)
- Added `queue_message_for_thread()` function
- Messages stored with namespace `("queue", thread_id)` and key `"pending_messages"`
- **Supports multiple queued messages** - stores as a list (FIFO order) instead of overwriting

3. Before model middleware (`server.py`)
- Added `check_message_queue_before_model()` decorated with `@before_model`
- Looks up queued messages in the store
- Extracts all messages, combines them with clear formatting, injects into state
- Clears the queue after processing

### DIFF
--- a/libs/deepagents-cli/deepagents_cli/server.py
+++ b/libs/deepagents-cli/deepagents_cli/server.py
@@ -13,6 +13,7 @@ logger = logging.getLogger(__name__)
 from langchain.agents.middleware import AgentState, after_agent, after_model, before_model
 from langgraph.config import get_config, get_store
 from langgraph.graph.state import RunnableConfig
+from langgraph.pregel import Pregel
 from langgraph.runtime import Runtime
 
 warnings.filterwarnings("ignore", module="langchain_core._api.deprecation")
@@ -23,9 +24,9 @@ import asyncio
 warnings.filterwarnings("ignore", message=".*Pydantic V1.*", category=UserWarning)
 
 # Now safe to import agent (which imports LangChain modules)
-from deepagents_cli.agent import create_server_agent
+from deepagents.backends.sandbox import SandboxBackendProtocol
 
-# CRITICAL: Import config FIRST to set LANGSMITH_PROJECT before LangChain loads
+from deepagents_cli.agent import create_server_agent
 from deepagents_cli.config import settings
 from deepagents_cli.encryption import decrypt_token
 from deepagents_cli.integrations.sandbox_factory import create_sandbox_async
@@ -42,6 +43,13 @@ client = get_client()
 SANDBOX_CREATING = "__creating__"
 SANDBOX_CREATION_TIMEOUT = 180
 SANDBOX_POLL_INTERVAL = 1.0
+
+# HTTP status codes
+HTTP_CREATED = 201
+HTTP_UNPROCESSABLE_ENTITY = 422
+
+# Message count thresholds
+MIN_MESSAGES_FOR_PREV_CHECK = 2
 
 _SANDBOX_BACKENDS: dict[str, Any] = {}
 
@@ -102,13 +110,13 @@ async def create_github_pr(
 
             pr_data = pr_response.json()
 
-            if pr_response.status_code == 201:
+            if pr_response.status_code == HTTP_CREATED:
                 pr_url = pr_data.get("html_url")
                 pr_number = pr_data.get("number")
                 logger.info("PR created successfully: %s", pr_url)
                 return pr_url, pr_number
 
-            if pr_response.status_code == 422:
+            if pr_response.status_code == HTTP_UNPROCESSABLE_ENTITY:
                 logger.error("GitHub API validation error (422): %s", pr_data.get("message"))
             else:
                 logger.error(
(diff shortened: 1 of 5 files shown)


libs/deepagents-cli/deepagents_cli/server.py
```diff
@@ -107,6 +107,121 @@ class LinearNotifyState(AgentState):
...
+        if queued_item is None:
+            return None
+
+        # Extract all queued messages
+        queued_value = queued_item.value
+        queued_messages = queued_value.get("messages", [])
+
+        if not queued_messages:
+            # Empty message list, clear and skip
+            await store.adelete(namespace, "pending_messages")
```
bracesproul: we delete here & a few lines below. it would be a little cleaner to call the `adelete` method once

libs/deepagents-cli/deepagents_cli/server.py
```diff
@@ -107,6 +107,121 @@ class LinearNotifyState(AgentState):
...
+        if not combined_content_parts:
+            return None
+
+        combined_content = "\n\n---\n\n".join(combined_content_parts)
+
+        # Create a new human message with the queued content
+        # Frame it so the agent knows to COMPLETE current task first, then handle follow-ups
+        if len(queued_messages) > 1:
+            intro = (
+                f"[IMPORTANT: {len(queued_messages)} follow-up requests received]\n\n"
```
bracesproul: i wouldn't constrain it to this. they could be telling the model to completely change direction, or something else

i would just inject the messages into the state as is, since they're human messages the LLM should know how to handle them

libs/deepagents-cli/deepagents_cli/server.py
```diff
@@ -107,6 +107,121 @@ class LinearNotifyState(AgentState):
...
+        )
+
+        # Clear the queue to prevent re-processing
+        await store.adelete(namespace, "pending_messages")
+
+        # Build combined message content from all queued messages (FIFO order)
+        combined_content_parts = []
+        for i, msg in enumerate(queued_messages, 1):
+            message_content = msg.get("content", "")
+            queued_at = msg.get("queued_at", "unknown")
```
bracesproul: im not sure i'd do this. similar to my comment before, the model doesn't need to know the messages were queued, that's an implementation detail. just injecting them as human messages should be good enough

libs/deepagents-cli/deepagents_cli/server.py
```diff
@@ -352,32 +467,79 @@ async def open_pr_if_needed(state: AgentState, runtime: Runtime) -> dict[str, An
...
-                -d '{{"title":"{pr_title}","head":"{target_branch}","base":"{base_branch}","body":"{pr_body}"}}'
-            """
+            logger.info("Creating PR: head=%s, base=%s, repo=%s/%s", target_branch, base_branch, repo_owner, repo_name)
 
-            pr_result = await asyncio.to_thread(sandbox_backend.execute, create_pr_cmd)
+            # Debug: Check token type (first few chars only for security)
+            token_prefix = github_token[:10] if github_token else "None"
+            logger.info("GitHub token prefix: %s... (length: %d)", token_prefix, len(github_token) if github_token else 0)
+            
+            # Check what scopes the token has by calling GitHub API
```
bracesproul: we can drop this. good for debugging but we shouldn't make this req every time

libs/deepagents-cli/deepagents_cli/server.py
```diff
@@ -352,32 +467,79 @@ async def open_pr_if_needed(state: AgentState, runtime: Runtime) -> dict[str, An
...
+                "title": pr_title,
+                "head": target_branch,
+                "base": base_branch,
+                "body": pr_body
+            }
+            
+            logger.info("PR payload: %s", pr_payload)
+            
+            try:
+                async with httpx.AsyncClient() as http_client:
```
bracesproul: lets refactor this into a reusable func instead of defining inline

libs/deepagents-cli/deepagents_cli/webapp.py
```diff
@@ -334,6 +335,102 @@ def generate_thread_id_from_issue(issue_id: str) -> str:
...
+        True if successfully queued, False otherwise
+    """
+    langgraph_client = get_client(url=LANGGRAPH_URL)
+    try:
+        namespace = ("queue", thread_id)
+        key = "pending_messages"
+        
+        new_message = {
+            "content": message_content,
+            "issue_data": issue_data,
```
bracesproul: doesn't look like this was used in the middleware, we can omit for now (unless it is used and im missing something?)

libs/deepagents-cli/deepagents_cli/webapp.py
```diff
@@ -334,6 +335,102 @@ def generate_thread_id_from_issue(issue_id: str) -> str:
...
+        value = {"messages": existing_messages}
+        
+        logger.info(
+            "Attempting to queue message for thread %s with namespace=%s, key=%s (total queued: %d)",
+            thread_id,
+            namespace,
+            key,
+            len(existing_messages),
+        )
+        # Store the queued messages list with namespace ("queue", thread_id)
```
bracesproul: can drop a lot of these comments, unless the code is doing something where the dev wouldn't actually know why without a comment, you can avoid it.

e.g. some crazy bug where if they delete the line below, everything breaks. in that case, add a comment. but for something like this, the code is pretty self explanatory. you should go through the rest of this PR and remove a lot of the comments

bracesproul (changes requested):
most nits, looks pretty good overall. very excited for this!

libs/deepagents-cli/deepagents_cli/server.py
```diff
@@ -107,6 +186,87 @@ class LinearNotifyState(AgentState):
...
+        content_parts = [msg.get("content", "") for msg in queued_messages if msg.get("content")]
+
+        if not content_parts:
+            return None
+
+        combined_content = "\n\n".join(content_parts)
+
+        new_message = {
+            "role": "user",
+            "content": combined_content,
```
bracesproul: nit, content can be an array of content blocks. we should do that:
```
[
  {
    "type: "text",
    "text": {content}
  }
]
```

instead of joining each msg from the queue

PR #1038, opened 2026-02-03
fix(cli): resolve text highlight auto-copy timing issue

### LINKED ISSUE #703: CLI Textual Highlight to Copy is Broken
Highlighting should auto copy but doesn't add to clipboard in the CLI.  Paste is Ctrl+V on Mac as well

ARUNKUMAR069:
Fixes #703

Summary

This PR fixes the issue where highlighted text in the CLI was not being copied to the clipboard. The problem was caused by a timing mismatch between Textual mouse events and the widget’s selection state updates.

Changes Made

Deferred clipboard copy after mouse selection

Updated the on_mouse_up handler to call the clipboard copy logic using call_after_refresh()

Ensures the text selection state is fully committed before attempting to copy

Added comprehensive test coverage

Tests for single and multiple text selections

Edge case handling (empty selection, errors)

Clipboard fallback behavior (pyperclip)

All tests passing 

Root Cause

on_mouse_up is triggered immediately when the mouse button is released, but Textual updates the text_selection state asynchronously.
As a result, the clipboard copy logic was running before the selection was finalized, leading to nothing being copied.

Deferring the operation using call_after_refresh() ensures the selection state is stable and correct.

Notes

Paste keybindings (e.g. Cmd+V on macOS) are handled internally by Textual’s TextArea widget. Any paste issues are outside the scope of the CLI code and not addressed in this PR.


mdrxy:
Hi @ARUNKUMAR069, is this different from #707?

ARUNKUMAR069:
Hi @mdrxy — good question.

Yes, this is different from #707.

PR #707 addressed the symptom by triggering the copy logic directly on mouse events, but it still relied on the selection state being immediately available.

This PR fixes the underlying root cause: Textual updates `text_selection` asynchronously, so `on_mouse_up` can fire before the selection is finalized.

By deferring the clipboard copy using `call_after_refresh()`, this ensures the selection state is fully committed before copying. I also added test coverage for multi-selection and empty-selection edge cases, which were not covered previously.

Happy to adjust or rebase if you’d prefer this logic incorporated differently.

ARUNKUMAR069:
@mdrxy is there any issue in the code ?

mdrxy:
Thanks for the contribution @ARUNKUMAR069.

Closing this for now — I haven't been able to reproduce the original issue (#703), and without a reliable repro it's hard to validate the fix.

The `call_after_refresh` change is theoretically reasonable, but the test suite here is entirely mocked and doesn't exercise the actual timing race condition that motivates the change. The tests validate `copy_selection_to_clipboard` behavior in isolation, which already works — the claimed bug is about *when* that function runs relative to Textual's selection state update, and no test here covers that.

Happy to revisit if someone can provide reliable reproduction steps (terminal emulator, OS, specific interaction pattern that fails).

PR #1039, opened 2026-02-03
feat(infra): have `release-please` auto regenerate `uv.lock` files

mdrxy:


.github/workflows/release-please.yml
```diff
@@ -24,13 +24,53 @@ jobs:
...
+
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v7
+
+      - name: Update lockfiles
+        run: |
+          for dir in $(find . -name "uv.lock" -type f -exec dirname {} \;); do
+            echo "Updating $dir"
+            if [ "$dir" = "./libs/acp" ]; then
+              uv lock --directory "$dir" --python 3.14
```
eyurtsev, line 57: I don't believe we support python 3.14?

.github/workflows/release-please.yml
eyurtsev: Could you isolate this by step

.github/workflows/release-please.yml
mdrxy: why not

PR #1041, opened 2026-02-03
feat(sdk): subagent skills, middleware refactor

sydney-runkle:
1. Add support for custom skills to subagents

```py
                                                                                                       
researcher = {                                                                                           
    "name": "researcher",                                                                                
    "description": "Research assistant",                                                                 
    "tools": [web_search],
    "skills": [".agents/skills/researcher/"],                                                            
}                          

agent = create_deep_agent(
    model="anthropic:claude-sonnet-4-20250514",
    subagents=[researcher],
    skills=[".agents/skills/main/"],
)
```

3. Test that skills aren't inherited for custom subagents, but are for the general purpose subagent
4. Refactor subagent middleware to take `backend` arg and move away from treating `general-purpose` subagent as first class (we'll want to add other prebuilts)
5. Abstract away summarization trigger logic into a helper function
6. Fix https://github.com/langchain-ai/deepagents/issues/802 by using subagent models for summarization rather than the main model
7. Ensure we're not passing skills or memory metadata to subagents

```py
def __init__(
        self,
        *,
        # these are all just for gen purpose dude
        default_model: str | BaseChatModel,  
        default_tools: Sequence[BaseTool | Callable | dict[str, Any]] | None = None,
        general_purpose_agent: bool = True,
        general_purpose_middleware: list[AgentMiddleware] | None = None,
        default_interrupt_on: dict[str, bool | InterruptOnConfig] | None = None,
        # these are valuable
        default_middleware: list[AgentMiddleware] | None = None,
    ) -> None:
```

new signature

```py
def __init__(
        self,
        *,
        subagents: list[SubAgent | CompiledSubAgent],
        task_description: str | None = TASK_TOOL_DESCRIPTION,
        **deprecated_kwargs: Any
    ) -> None:
```

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -151,8 +151,21 @@ def create_deep_agent(
 
     backend = backend if backend is not None else (lambda rt: StateBackend(rt))
 
-    # Build general-purpose subagent
-    gp_middleware: list[AgentMiddleware] = []
+    # Build general-purpose subagent with default middleware stack
+    gp_middleware: list[AgentMiddleware] = [
+        TodoListMiddleware(),
+        FilesystemMiddleware(backend=backend),
+        SummarizationMiddleware(
+            model=model,
+            backend=backend,
+            trigger=summarization_defaults["trigger"],
+            keep=summarization_defaults["keep"],
+            trim_tokens_to_summarize=None,
+            truncate_args_settings=summarization_defaults["truncate_args_settings"],
+        ),
+        AnthropicPromptCachingMiddleware(unsupported_model_behavior="ignore"),
+        PatchToolCallsMiddleware(),
+    ]
     if skills is not None:
         gp_middleware.append(SkillsMiddleware(backend=backend, sources=skills))
     if interrupt_on is not None:
@@ -165,18 +178,44 @@ def create_deep_agent(
         "middleware": gp_middleware,
     }
 
-    # Process user-provided subagents to fill in defaults for model and tools
+    # Process user-provided subagents to fill in defaults for model, tools, and middleware
     processed_subagents: list[SubAgent | CompiledSubAgent] = []
     for spec in subagents or []:
         if "runnable" in spec:
             # CompiledSubAgent - use as-is
             processed_subagents.append(spec)
         else:
-            # SubAgent - fill in defaults
+            # SubAgent - fill in defaults and prepend base middleware
+            subagent_model = spec.get("model", model)
+            if isinstance(subagent_model, str):
+                subagent_model = init_chat_model(subagent_model)
+
+            # Build middleware: base stack + skills (if specified) + user's middleware
+            subagent_summarization_defaults = _compute_summarization_defaults(subagent_model)
+            subagent_middleware: list[AgentMiddleware] = [
+                TodoListMiddleware(),
+                FilesystemMiddleware(backend=backend),
+                SummarizationMiddleware(
+                    model=subagent_model,
+                    backend=backend,
+                    trigger=subagent_summarization_defaults["trigger"],
+                    keep=subagent_summarization_defaults["keep"],
+                    trim_tokens_to_summarize=None,
+                    truncate_args_settings=subagent_summarization_defaults["truncate_args_settings"],
+                ),
+                AnthropicPromptCachingMiddleware(unsupported_model_behavior="ignore"),
+                PatchToolCallsMiddleware(),
+            ]
+            subagent_skills = spec.get("skills")
+            if subagent_skills:
+                subagent_middleware.append(SkillsMiddleware(backend=backend, sources=subagent_skills))
+            subagent_middleware.extend(spec.get("middleware", []))
+
             processed_spec: SubAgent = {
                 **spec,
-                "model": spec.get("model", model),
+                "model": subagent_model,
                 "tools": spec.get("tools", tools or []),
+                "middleware": subagent_middleware,
             }
             processed_subagents.append(processed_spec)
 
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -5,11 +5,10 @@ from collections.abc import Awaitable, Callable, Sequence
 from typing import Annotated, Any, NotRequired, TypedDict, Unpack, cast
 
 from langchain.agents import create_agent
-from langchain.agents.middleware import HumanInTheLoopMiddleware, InterruptOnConfig, TodoListMiddleware
+from langchain.agents.middleware import HumanInTheLoopMiddleware, InterruptOnConfig
 from langchain.agents.middleware.types import AgentMiddleware, ModelRequest, ModelResponse
 from langchain.chat_models import init_chat_model
 from langchain.tools import BaseTool, ToolRuntime
-from langchain_anthropic.middleware import AnthropicPromptCachingMiddleware
 from langchain_core.language_models import BaseChatModel
 from langchain_core.messages import HumanMessage, ToolMessage
 from langchain_core.runnables import Runnable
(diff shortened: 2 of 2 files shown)


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -456,98 +513,255 @@ class SubAgentMiddleware(AgentMiddleware):
...
-        default_model: str | BaseChatModel,
-        default_tools: Sequence[BaseTool | Callable | dict[str, Any]] | None = None,
-        default_middleware: list[AgentMiddleware] | None = None,
-        default_interrupt_on: dict[str, bool | InterruptOnConfig] | None = None,
+        backend: BackendProtocol | BackendFactory | None = None,
         subagents: list[SubAgent | CompiledSubAgent] | None = None,
         system_prompt: str | None = TASK_SYSTEM_PROMPT,
-        general_purpose_agent: bool = True,
         task_description: str | None = None,
+        **deprecated_kwargs: Unpack[_DeprecatedKwargs],
```
eyurtsev, line 552: What happens if I type "fooofoobar=2" -- does this correctly raise an exception? (i.e., kwargs should be fully absorbed)

libs/deepagents/deepagents/graph.py
```diff
@@ -141,50 +146,42 @@ def create_deep_agent(
...
-            "trigger": ("messages", 20),
-            "keep": ("messages", 20),
-        }
-
-    # Build middleware stack for subagents (includes skills if provided)
-    subagent_middleware: list[AgentMiddleware] = [
-        TodoListMiddleware(),
-    ]
+    # Compute summarization defaults based on model profile
+    summarization_defaults = compute_summarization_defaults(model)
```
eyurtsev: `compute_summarization_defaults` is internal I believe? (i.e., we might want to move logic into the middleware itself for fallback behavior?

libs/deepagents/deepagents/graph.py
```diff
@@ -141,50 +146,42 @@ def create_deep_agent(
...
+    for spec in subagents or []:
+        if "runnable" in spec:
+            # CompiledSubAgent - use as-is
+            processed_subagents.append(spec)
+        else:
+            # SubAgent - fill in defaults
+            processed_spec: SubAgent = {
+                **spec,
+                "model": spec.get("model", model),
+                "tools": spec.get("tools", tools or []),
```
eyurtsev, line 217: ah does the current code on master inconsistently handle missing middleware vs. missing tools? (i.e., if tools are unspecified we use main agent tools, but if middleware is missing we don't use any middleware?)

libs/deepagents/tests/integration_tests/test_subagent_middleware.py
```diff
@@ -223,61 +202,316 @@ def test_defined_subagent_custom_runnable(self):
...
                             "tools": [get_weather],
                             "interrupt_on": {"get_weather": True},
                         },
                     ],
                 )
             ],
         )
-        # This would error if the default middleware was accumulated
-        assert True
+        # This would error if the middleware was accumulated incorrectly
```
eyurtsev, line 232: Should we delete the integration test? did you run it? (It's not hooked up on CI right now -- i've been relying on the equivalent unit tests that are basically end to end)

libs/deepagents/pyproject.toml
```diff
@@ -105,20 +105,21 @@ ignore-var-parameters = true
...
 "deepagents/backends/protocol.py" = ["B024", "B027", "FBT001", "FBT002"]
-"deepagents/backends/sandbox.py" = ["FBT001", "FBT002", "PLC0105", "PLR2004"]
+"deepagents/backends/sandbox.py" = ["FBT001", "FBT002", "PLR2004"]
 "deepagents/backends/state.py" = ["ANN204", "D102", "D205", "EM101", "FBT001", "FBT002", "PERF401"]
 "deepagents/backends/store.py" = ["A002", "ANN204", "BLE001", "D102", "D205", "F821", "FBT001", "FBT002", "PERF401"]
 "deepagents/backends/utils.py" = ["D301", "E501", "EM101", "FBT001", "RET504", "RUF005", "TRY003"]
 "deepagents/middleware/filesystem.py" = ["EM102", "TRY003"]
 "deepagents/middleware/memory.py" = ["E501", "EM102", "G004", "PERF401", "SIM108", "T201", "TC002", "TC003", "TRY003"]
 "deepagents/middleware/skills.py" = ["EM101", "SIM108", "TC002", "TC003", "TRY003"]
+"deepagents/graph.py" = ["PLR0912"]
```
eyurtsev, line 122: intentional?

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -80,6 +80,51 @@ class TruncateArgsSettings(TypedDict, total=False):
...
 
+class SummarizationDefaults(TypedDict):
+    """Default settings computed from model profile."""
+
+    trigger: ContextSize
+    keep: ContextSize
+    truncate_args_settings: TruncateArgsSettings
+
+
+def compute_summarization_defaults(model: BaseChatModel) -> SummarizationDefaults:
```
eyurtsev: Should we mark as internal for now?

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -456,98 +513,255 @@ class SubAgentMiddleware(AgentMiddleware):
...
+
+        # Warn about any deprecated kwargs that were provided
+        provided_deprecated = [key for key in deprecated_kwargs if key != "general_purpose_agent"]
+        if "general_purpose_agent" in deprecated_kwargs and not general_purpose_agent:
+            provided_deprecated.append("general_purpose_agent")
+
+        if provided_deprecated:
+            warnings.warn(
+                f"The following SubAgentMiddleware arguments are deprecated and will be removed "
+                f"in a future version: {', '.join(provided_deprecated)}. "
```
eyurtsev: If you'd like we can mark for deprecation in deepagents 0.5.0

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -106,11 +120,15 @@ class CompiledSubAgent(TypedDict):
...
 # When returning updates:
 # 1. The messages key is handled explicitly to ensure only the final message is included
-# 2. The todos, structured_response, and skills_metadata keys are excluded as they do not have
-#    a defined reducer and no clear meaning for returning them from a subagent to the main agent.
-#    Each agent loads its own skills independently based on its middleware configuration.
+# 2. The todos and structured_response keys are excluded as they do not have a defined reducer
+#    and no clear meaning for returning them from a subagent to the main agent.
+# 3. The skills_metadata key is automatically excluded from subagent output via PrivateStateAttr
+#    annotations on the state schema. However, it must ALSO be explicitly filtered from
+#    runtime.state when invoking a subagent to prevent parent skills from leaking to child agents
```
eyurtsev: and we don't have the same issue w/ memory?

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -325,45 +350,38 @@ def _get_subagents(
...
-def _create_task_tool(
-    *,
-    default_model: str | BaseChatModel,
-    default_tools: Sequence[BaseTool | Callable | dict[str, Any]],
-    default_middleware: list[AgentMiddleware] | None,
-    default_interrupt_on: dict[str, bool | InterruptOnConfig] | None,
-    subagents: list[SubAgent | CompiledSubAgent],
-    general_purpose_agent: bool,
+def _build_task_tool(
+    subagent_graphs: dict[str, Runnable],
```
eyurtsev: This is internal so doesn't really matter. but the function signature is a bit odd given that subagetn_graphs is decoupled from subagent_descriptions... should be a list of dicts here to group related information together, the map can be generated within the body of the function

mdrxy:
Need to take a closer look but does #1059 affect anything?

libs/deepagents/pyproject.toml
sydney-runkle: yeah complexity

will work to clean up a bunch of these in a diff PR

libs/deepagents/deepagents/middleware/subagents.py
sydney-runkle: Going to just add the key. We don't have memory for subagents right now but it can't hurt to pass a clean state. Also users could theoretically add custom memory middleware.

Thanks for revisiting

libs/deepagents/deepagents/middleware/subagents.py
sydney-runkle: Agreed, will refactor

libs/deepagents/deepagents/graph.py
sydney-runkle: No the current code handles it properly, it's just in the subagent middleware, whereas here we enforce default model + tools in create_deep_agent

PR #1050, opened 2026-02-03
poc(sdk): move away from backend factory

sydney-runkle:
1. Remove recommended pattern of `lambda rt: StateBackend(rt)`
2. Remove all artificial construction of `ToolRuntime`
3. Add a `bind` method to `BackendProtocol` that accepts a `BackendContext` object with state and runtime

I actually still feel conflicted about the fact that we need to pass around state and runtime to backends, that makes them feel quite coupled w/ langgraph?

currently... breaking :/
fixes a bunch of linting issues as well :)

### DIFF
--- a/libs/deepagents/deepagents/backends/state.py
+++ b/libs/deepagents/deepagents/backends/state.py
@@ -1,8 +1,11 @@
 """StateBackend: Store files in LangGraph agent state (ephemeral)."""
 
-from typing import TYPE_CHECKING
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Any
 
 from deepagents.backends.protocol import (
+    BackendContext,
     BackendProtocol,
     EditResult,
     FileDownloadResponse,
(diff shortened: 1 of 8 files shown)


libs/deepagents/deepagents/backends/state.py
```diff
@@ -35,11 +38,39 @@ class StateBackend(BackendProtocol):
...
+        """Initialize StateBackend."""
+        self._context: BackendContext | None = None
+
+    def bind(self, ctx: BackendContext) -> None:
+        """Bind the backend to an execution context.
+
+        Args:
+            ctx: The context containing state and runtime for the current execution.
+        """
+        self._context = ctx
```
eyurtsev, line 62: mutation in place post init, very likely to cause race conditions

libs/deepagents/deepagents/graph.py
```diff
@@ -55,7 +55,7 @@ def create_deep_agent(
     context_schema: type[Any] | None = None,
     checkpointer: Checkpointer | None = None,
     store: BaseStore | None = None,
-    backend: BackendProtocol | BackendFactory | None = None,
+    backend: BackendProtocol | None = None,
```
eyurtsev, line 58: Is this a breaking change?

sydney-runkle: yes

we probably want to make this backwards compatible in the short term, just wanted to demonstrate value w/ POC

PR #1059, opened 2026-02-04
feat(cli): support  .`agents/skills` dir alias

mdrxy:
Skills can now be placed in `.agents/skills/` (project) or `~/.agents/skills/` (user) as a generic alternative to `.deepagents/skills/`. The `.agents/` paths take precedence over `.deepagents/` when skills have the same name.

When a skill with the same name exists in multiple directories, the skill from the **highest-precedence directory wins completely**. There is no merging of skill properties - the entire skill definition is replaced.

### Recommended Usage

- Use `.agents/skills/` for skills you want to share across different AI CLI tools
- Use `.deepagents/skills/` for Deep Agents-specific skills
- Project-level skills override user-level skills of the same name

### DIFF
--- a/libs/cli/deepagents_cli/config.py
+++ b/libs/cli/deepagents_cli/config.py
@@ -423,6 +423,37 @@ class Settings:
             return None
         return self.project_root / ".deepagents" / "agents"
 
+    @property
+    def user_agents_dir(self) -> Path:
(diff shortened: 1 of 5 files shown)


eyurtsev:
This is good -- there might be some bugs w/ knowing which skills folder to update if there are many, so would just recommend seeing if they're listed in order

eyurtsev (approved):

PR #1074, opened 2026-02-04
fix(sdk): resolve silent exception handling in backends and middleware

kaushikkumarkr:
This PR fixes multiple instances where exceptions were being silently swallowed across the SDK, making debugging difficult. I've replaced these silent failures with proper DEBUG-level logging to improve observability without changing the runtime behavior.

Changes:
store.py: Added logging when retrieving the namespace fails or when encountering invalid items during store searches.
local_shell.py: Added logging with full stack traces for failed shell command executions.
summarization.py: Added logging when the thread ID cannot be retrieved from the config.

### DIFF
--- a/libs/deepagents/deepagents/backends/local_shell.py
+++ b/libs/deepagents/deepagents/backends/local_shell.py
@@ -7,6 +7,7 @@ on the host machine with full system access.
 
 from __future__ import annotations
 
+import logging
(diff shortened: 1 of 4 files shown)


mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #1079, opened 2026-02-04
fix(deepagents): `_validate_skill_name` method consistently returns false on Windows 

EugeneTechLab:
root cause: In the if name != directory_name check, directory_name was holding the directory path instead of the directory name, which caused the validation logic to fail on Windows.

fix: updated the logic to correctly derive and compare the directory name, ensuring _validate_skill_name behaves as expected across platforms.


eyurtsev:
Windows is only supported through WSL at the moment.

PR #1085, opened 2026-02-04
feat(cli): add working_dir and LangSmith template customization

aran-yogesh:
Changes:
- Add working_dir parameter to get_system_prompt() and create_cli_agent() to override default sandbox working directory
- Add template and template_image parameters to LangSmithProvider.get_or_create() for custom sandbox templates
- Add template and template_image pass-through in create_sandbox() factory function
- Support env vars LANGSMITH_SANDBOX_TEMPLATE_NAME and LANGSMITH_SANDBOX_TEMPLATE_IMAGE for defaults

### DIFF
--- a/libs/cli/deepagents_cli/integrations/langsmith.py
+++ b/libs/cli/deepagents_cli/integrations/langsmith.py
@@ -20,51 +20,36 @@ from deepagents.backends.sandbox import (
 )
 
 if TYPE_CHECKING:
-    from langsmith.sandbox import Sandbox, SandboxClient
-
+    from langsmith.sandbox import Sandbox, SandboxClient, SandboxTemplate
 
 # Default template configuration
-DEFAULT_TEMPLATE_NAME = "deepagents-cli"
-DEFAULT_TEMPLATE_IMAGE = "python:3"
+DEFAULT_TEMPLATE_NAME = os.getenv("LANGSMITH_SANDBOX_TEMPLATE_NAME", "deepagents-cli")
+DEFAULT_TEMPLATE_IMAGE = os.getenv("LANGSMITH_SANDBOX_TEMPLATE_IMAGE", "python:3")
 
 
 class LangSmithBackend(BaseSandbox):
-    """LangSmith backend implementation conforming to SandboxBackendProtocol.
-
-    This implementation inherits all file operation methods from BaseSandbox
-    and only implements the execute() method using LangSmith's API.
-    """
+    """LangSmith backend implementation."""
 
     def __init__(self, sandbox: Sandbox) -> None:
-        """Initialize the LangSmithBackend with a sandbox instance.
-
-        Args:
-            sandbox: LangSmith Sandbox instance
-        """
+        """Initialize with sandbox instance."""
         self._sandbox = sandbox
-        self._timeout: int = 30 * 60  # 30 mins default
+        self._timeout: int = 30 * 60
 
     @property
     def id(self) -> str:
-        """Unique identifier for the sandbox backend."""
+        """Return sandbox name as ID."""
         return self._sandbox.name
 
     def execute(self, command: str) -> ExecuteResponse:
-        """Execute a command in the sandbox and return ExecuteResponse.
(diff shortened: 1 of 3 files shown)


libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -20,12 +20,11 @@
...
 
 if TYPE_CHECKING:
-    from langsmith.sandbox import Sandbox, SandboxClient
+    from langsmith.sandbox import Sandbox, SandboxClient, SandboxTemplate
 
-
-# Default template configuration
-DEFAULT_TEMPLATE_NAME = "deepagents-cli"
-DEFAULT_TEMPLATE_IMAGE = "python:3"
+# Default template configuration - can be overridden via environment variables
```
bracesproul: dont use env vars, see eugene's req in slack

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -83,10 +82,11 @@ def download_files(self, paths: list[str]) -> list[FileDownloadResponse]:
 
         Returns:
             List of FileDownloadResponse objects, one per input path.
-            Response order matches input order.
 
-        TODO: Map LangSmith API error strings to standardized FileOperationError codes.
-        Currently only implements happy path.
+        Note:
```
bracesproul, line 85: why'd you change this docstring? dont think it should be changed

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -102,7 +102,6 @@ def download_files(self, paths: list[str]) -> list[FileDownloadResponse]:
     def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadResponse]:
         """Upload multiple files to the LangSmith sandbox.
 
-        Leverages LangSmith's native file write API for efficiency.
```
bracesproul, line 105: same with all the docstring changes. dont make changes unless absolutely necessary for the PR so we can keep the diff size as small as possible

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -168,55 +174,70 @@ def get_or_create(
         *,
         sandbox_id: str | None = None,
         timeout: int = 180,
+        template: SandboxTemplate | str | None = None,
+        template_image: str | None = None,
```
bracesproul: eugege asked for this to be inside `kwargs`

instead of defining args, you should just check to see if `kwargs` is a dict, if so check if it has those fields, and if so, use them

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -168,55 +174,70 @@ def get_or_create(
...
                 sandbox = self._client.get_sandbox(name=sandbox_id)
             except Exception as e:
                 msg = f"Failed to connect to existing sandbox '{sandbox_id}': {e}"
                 raise RuntimeError(msg) from e
             return LangSmithBackend(sandbox)
 
-        # Create new sandbox - ensure template exists first
-        self._ensure_template()
+        # Resolve template name from various input types
+        template_name = self._resolve_template_name(template)
```
bracesproul: this should prob be resolve template name & image

if template_image isn't defined, we should still check to see if template is an instance of `SandboxTemplate`, and if so, we can use the `template.image` field there to get the image

libs/cli/deepagents_cli/integrations/sandbox_factory.py
```diff
@@ -68,6 +68,8 @@ def create_sandbox(
     *,
     sandbox_id: str | None = None,
     setup_script_path: str | None = None,
+    template: str | None = None,
+    template_image: str | None = None,
```
bracesproul: same comment as above, but i think this should just be `kwargs` then you pass `kwargs` to `get_or_create`

bracesproul (changes requested):

PR #1087, opened 2026-02-04
poc(sdk): deprecate factory pattern for backends

sydney-runkle:
1. deprecate factory pattern in favor of making langgraph context accessible in individual methods
2. introduce `namespace` arg to `StoreBackend` to support namespace gen as a function of langgraph context
3. rename unintuitively named tools `ls_info` -> `ls`, for example

need to convince myself that this is much better and mostly backwards compatible

### New `BackendContext` for method-level context
```python
@dataclass
class BackendContext:
    state: dict[str, Any]
    runtime: Runtime

# Usage
ctx = BackendContext(state=state, runtime=runtime)
result = backend.read("/path/to/file", ctx=ctx)
```

### Cleaner method names with deprecation warnings
```python
# Old (deprecated, warns)
backend.ls_info("/path")
backend.grep_raw("pattern")
backend.glob_info("*.py")

# New
backend.ls("/path", ctx=ctx)
backend.grep("pattern", ctx=ctx)
backend.glob("*.py", ctx=ctx)
```

### Flexible namespace factory for `StoreBackend`
```python
# Old: relied on assistant_id detection from metadata
backend = StoreBackend(runtime)

# New: explicit namespace factory
backend = StoreBackend(
    namespace=lambda ctx: ("filesystem", ctx.runtime.context.user_id)
)
```

### Before: Tools required `_get_backend()` factory resolution
```python
# Middleware had to store factory and resolve at call time
class FilesystemMiddleware:
    def __init__(self, backend: BACKEND_TYPES | None = None):
        # Had to support both instances AND factories
        self.backend = backend if backend is not None else (lambda rt: StateBackend(rt))
    
    def _get_backend(self, runtime: ToolRuntime) -> BackendProtocol:
        # Factory resolution on every tool call
        if callable(self.backend):
            return self.backend(runtime)
        return self.backend
    
    # Tool implementation
    def _ls_sync(self, runtime: ToolRuntime, path: str) -> str:
        resolved_backend = self._get_backend(runtime)  # resolve factory
        infos = resolved_backend.ls_info(validated_path)
        ...
```

### After: Pass context directly to backend methods
```python
# Middleware stores backend instance directly
class FilesystemMiddleware:
    def __init__(self, backend: BackendProtocol | None = None):
        self.backend = backend if backend is not None else StateBackend()
    
    # Tool implementation - context passed at call site
    def _ls_sync(self, runtime: ToolRuntime, path: str) -> str:
        infos = self.backend.ls(validated_path, ctx=BackendContext(runtime.state, runtime))
        ...
```

### Before: Hooks needed runtime at construction
```python
# Backend needed runtime upfront to access state
class StateBackend:
    def __init__(self, runtime: ToolRuntime):
        self.runtime = runtime
    
    def ls_info(self, path: str) -> list[FileInfo]:
        files = self.runtime.state.get("files", {})  # accessed via stored runtime
        ...
```

### After: Hooks receive context per-call
```python
# Backend is stateless, receives context when needed
class StateBackend:
    def __init__(self):
        pass  # no runtime needed
    
    def ls(self, path: str, *, ctx: BackendContext | None = None) -> list[FileInfo]:
        files = ctx.state.get("files", {})  # accessed via passed context
        ...
```

### DIFF
--- a/libs/deepagents/deepagents/backends/composite.py
+++ b/libs/deepagents/deepagents/backends/composite.py
@@ -101,7 +101,7 @@ class CompositeBackend(BackendProtocol):
         self,
         path: str,
         *,
-        _ctx: BackendContext | None = None,
+        ctx: BackendContext | None = None,  # noqa: ARG002
     ) -> list[FileInfo]:
         """List directory contents (non-recursive).
 
@@ -161,7 +161,7 @@ class CompositeBackend(BackendProtocol):
         self,
         path: str,
         *,
-        _ctx: BackendContext | None = None,
+        ctx: BackendContext | None = None,  # noqa: ARG002
     ) -> list[FileInfo]:
         """Async version of ls."""
         # Check if path matches a specific route
@@ -205,7 +205,7 @@ class CompositeBackend(BackendProtocol):
         offset: int = 0,
         limit: int = 2000,
         *,
-        _ctx: BackendContext | None = None,
+        ctx: BackendContext | None = None,  # noqa: ARG002
     ) -> str:
         """Read file content, routing to appropriate backend.
 
@@ -227,7 +227,7 @@ class CompositeBackend(BackendProtocol):
         offset: int = 0,
         limit: int = 2000,
         *,
-        _ctx: BackendContext | None = None,
+        ctx: BackendContext | None = None,  # noqa: ARG002
     ) -> str:
         """Async version of read."""
         backend, stripped_key = self._get_backend_and_key(file_path)
@@ -239,7 +239,7 @@ class CompositeBackend(BackendProtocol):
         path: str | None = None,
         glob: str | None = None,
         *,
-        _ctx: BackendContext | None = None,
+        ctx: BackendContext | None = None,  # noqa: ARG002
     ) -> list[GrepMatch] | str:
         """Search files for literal text pattern.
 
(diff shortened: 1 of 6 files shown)


mdrxy:
Supersedes #1050?

sydney-runkle:
yeah gonna move forward w/ this one bc it's backwards compat ish

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -96,23 +97,29 @@ def _get_backend_and_key(self, key: str) -> tuple[BackendProtocol, str]:
 
         return self.default, key
 
-    def ls_info(self, path: str) -> list[FileInfo]:
+    def ls(
+        self,
+        path: str,
+        *,
+        ctx: BackendContext | None = None,  # noqa: ARG002
```
eyurtsev, line 104: Should we keep ARG002 or get rid of it? Seems like it's largely false positives

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -96,23 +97,29 @@ def _get_backend_and_key(self, key: str) -> tuple[BackendProtocol, str]:
...
+    ) -> list[FileInfo]:
```
eyurtsev, line 105: Are there additional changes we need to incorporate (e.g., return types) before we commit to new names?

sydney-runkle: yeah we should think about return types before doing v0.4

libs/deepagents/deepagents/backends/composite.py
sydney-runkle: it's that this is "unused" but I want it to match the signature of the protocol... can probably ignore at the file level? wdyt

PR #1097, opened 2026-02-04
fix(cli): show full shell command in error output

### LINKED ISSUE #1080: cli truncates shell command input
### Privileged issue

### Issue Content

I'd like to be able to see the full shell command. Currently shell command is failing and I don't know what the CLI is trying to run (so don't know how to fix what's happening)

### Area (Required)

- [ ] deepagents (SDK)
- [x] cli

mdrxy:
- When a shell command fails, the full command is now prepended to the error output with a `$` prefix
- Applies to shell, bash, and execute tool types
- The command line is styled in dim grey for visual distinction from the error message

Fixes #1080

### DIFF
--- a/libs/cli/deepagents_cli/file_ops.py
+++ b/libs/cli/deepagents_cli/file_ops.py
@@ -15,7 +15,7 @@ from deepagents_cli.config import settings
 logger = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
-    from deepagents.backends.protocol import BACKEND_TYPES
+    from deepagents.backends.protocol import BackendProtocol
 
 FileOpStatus = Literal["pending", "success", "error"]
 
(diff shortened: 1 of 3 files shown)


libs/cli/deepagents_cli/file_ops.py
```diff
@@ -274,7 +274,7 @@ class FileOpTracker:
     """Collect file operation metrics during a CLI interaction."""
 
     def __init__(
-        self, *, assistant_id: str | None, backend: BACKEND_TYPES | None = None
```
mdrxy, line 277: technically breaking but if you weren't using `BackendProtocol` originally it would've been wrong

eyurtsev: Cli doesn't have a public interface ATM wouldn't worry about breaking changes at all... Public interface is more with external conventions like skills/memory etc

eyurtsev:
Do we render it on "input" I think the main issue is not being able to fully see it **before** approving it (though seeing after is good as well)

mdrxy:
@eyurtsev yes

PR #1099, opened 2026-02-05
feat(cli): support custom working directories and LangSmith sandbox templates

aran-yogesh:
When creating agents programmatically, you can now customize the sandbox working directory and LangSmith template:

```diff
# Custom working directory
agent, backend = create_cli_agent(
    ...,
+    working_dir="/home/user/my-project",
)

# Custom LangSmith template and API key
with create_sandbox(
    "langsmith",
+    template="my-template",
+    template_image="my-image:latest",
    api_key="lsak_...",
) as sandbox:
    ...
```

- Added `working_dir` parameter to `get_system_prompt` and `create_cli_agent`
  - Override the directory injected into the agent's system prompt -- callers can now point the agent at a specific path inside the sandbox (e.g. a cloned repo)
- Support `template`/`template_image` via `kwargs` for LangSmith sandbox -- choose which sandbox template and image to use
  - Use hardcoded defaults with auto-creation if template missing


libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -192,17 +196,18 @@ def get_or_create(
                 raise RuntimeError(msg) from e
             return LangSmithBackend(sandbox)
 
+        # Resolve template name and image from kwargs
```
bracesproul: unnecessary comment, would drop

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -183,6 +183,10 @@ def get_or_create(
         Raises:
             RuntimeError: Sandbox connection or startup failed
         """
+        # Extract template params from kwargs
+        template = kwargs.get("template")
+        template_image = kwargs.get("template_image")
```
bracesproul: question for Daniel, should we add an arg for `template_image`, or is `template` enough?

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -192,17 +196,18 @@ def get_or_create(
                 raise RuntimeError(msg) from e
             return LangSmithBackend(sandbox)
 
+        # Resolve template name and image from kwargs
+        template_name, resolved_image = self._resolve_template(template, template_image)
```
bracesproul: also would unify the naming here:
`resolved_template_name`, `resolved_image_name`

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -233,30 +238,53 @@ def delete(self, *, sandbox_id: str, **kwargs: Any) -> None:  # noqa: ARG002
...
+        template: SandboxTemplate | str | None,
+        template_image: str | None = None,
+    ) -> tuple[str, str | None]:
+        """Resolve template name and image from kwargs.
+
+        Returns:
+            Tuple of (template_name, template_image).
+        """
+        if template is None:
+            return DEFAULT_TEMPLATE_NAME, template_image
```
bracesproul: prob best to make this template_image if defined, else default template image

this way we can ensure `_resolve_template` always returns a value for both template name and image, and it'll either be the name/image from args, or the default values

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -233,30 +238,53 @@ def delete(self, *, sandbox_id: str, **kwargs: Any) -> None:  # noqa: ARG002
...
+        # SandboxTemplate object - extract image if not provided
+        image = template_image
+        if image is None and hasattr(template, "image"):
+            image = template.image
+        return template.name, image
+
+    def _ensure_template(
+        self,
+        template_name: str | None = None,
+        template_image: str | None = None,
```
bracesproul: if we make sure `_resolve_template` always returns a value (either from `kwargs` or default) we can make this always accept a str and we wont need to fallback to none

bracesproul (commented):
nice! couple small comments, after resolving we can req review from Eugene & Daniel

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -168,7 +168,7 @@ def get_or_create(
         *,
         sandbox_id: str | None = None,
         timeout: int = 180,
-        **kwargs: Any,  # noqa: ARG002
+        **kwargs: Any,
```
eyurtsev, line 173: Document the params explicitly here -- same as timeout

do something like 

```python
template: FullTypeAnnotation = None,
template_image: FUll Type annotaiton = ...
```

And then **kwargs

and in the body of the function do:

```python
if kwargs:
  raise TypeError('Received unsupported arguments
```

We want the client to type all the options explicitly, and unfortunately this is the only way to express this in a pythonic way

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -233,7 +242,32 @@ def delete(self, *, sandbox_id: str, **kwargs: Any) -> None:  # noqa: ARG002
...
+            Tuple of (template_name, template_image). Always returns values,
+            using defaults if not provided.
+        """
+        resolved_image = template_image or DEFAULT_TEMPLATE_IMAGE
+        if template is None:
+            return DEFAULT_TEMPLATE_NAME, resolved_image
+        if isinstance(template, str):
+            return template, resolved_image
+        # SandboxTemplate object - extract image if not provided
+        if template_image is None and hasattr(template, "image"):
```
eyurtsev: why is `hasattr` being used?

eyurtsev: Unless there's something weird going on python code should avoid using 
`hasattr`, `getattr` and `setattr` -- those are reserved for special occasions.

libs/cli/deepagents_cli/agent.py
```diff
@@ -411,6 +416,7 @@ def create_cli_agent(
     tools: Sequence[BaseTool | Callable | dict[str, Any]] | None = None,
     sandbox: SandboxBackendProtocol | None = None,
     sandbox_type: str | None = None,
+    working_dir: str | None = None,
```
eyurtsev: This information doesn't belong in the create_cli_agent entrypoint -- it's a property of the sandbox

libs/cli/deepagents_cli/integrations/langsmith.py
DanielKneipp: I have nothing against allowing a custom image tbh. The only thing to keep in mind I guess is that this will add divergence from the other sandbox integrations (but not too big of a deal imo)

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -136,16 +136,21 @@ def __init__(self, api_key: str | None = None) -> None:
...
-            ValueError: If LANGSMITH_API_KEY environment variable not set
+            ValueError: If no API key is provided and neither env var is set.
         """
         from langsmith import sandbox
 
-        self._api_key = api_key or os.environ.get("LANGSMITH_API_KEY")
+        self._api_key = (
+            api_key
+            or os.environ.get("LANGSMITH_API_KEY")
+            or os.environ.get("LANGSMITH_API_KEY_PROD")
```
bracesproul: they wont want this in the deepagents code, this should live in our open-swe code, and then we pass this via the class/func args
```suggestion
```

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -168,21 +173,29 @@ def get_or_create(
...
+        template: str | None = None,
+        template_image: str | None = None,
+        **kwargs: Any,
     ) -> SandboxBackendProtocol:
         """Get existing or create new LangSmith sandbox.
 
         Args:
             sandbox_id: Optional existing sandbox name to reuse
             timeout: Timeout in seconds for sandbox startup (default: 180)
-            **kwargs: Additional LangSmith-specific parameters
```
bracesproul, line 178: remember to add back this docstring b/c the arg is still there

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -168,21 +173,29 @@ def get_or_create(
...
         Returns:
             LangSmithBackend instance
 
         Raises:
             RuntimeError: Sandbox connection or startup failed
+            TypeError: If unsupported keyword arguments are provided
         """
+        if kwargs:
+            msg = f"Received unsupported arguments: {list(kwargs.keys())}"
+            raise TypeError(msg)
```
bracesproul, line 193: why are we throwing? did they ask for you to do this?

libs/cli/deepagents_cli/integrations/sandbox_factory.py
```diff
@@ -77,19 +82,33 @@ def create_sandbox(
...
 
     Yields:
         SandboxBackendProtocol instance
+
+    Raises:
+        TypeError: If unsupported keyword arguments are provided
     """
+    if kwargs:
+        msg = f"Received unsupported arguments: {list(kwargs.keys())}"
+        raise TypeError(msg)
```
bracesproul: same here, why throw?

libs/cli/deepagents_cli/integrations/sandbox_factory.py
```diff
@@ -77,19 +82,33 @@ def create_sandbox(
         provider: Sandbox provider ("daytona", "langsmith", "modal", "runloop")
         sandbox_id: Optional existing sandbox ID to reuse
         setup_script_path: Optional path to setup script to run after sandbox starts
+        api_key: LangSmith API key (langsmith provider only)
+        template: Template name for sandbox (langsmith provider only)
+        template_image: Docker image for template (langsmith provider only)
```
bracesproul: needs kwargs docstring

mdrxy:
I refined the PR body. Please edit if anything is inaccurate.

libs/cli/deepagents_cli/agent.py
eyurtsev: @copilot update working_dir to be `_working_dir` and mark it for internal use only. (Say that this is for updating the prompt for sandboxes and implementation will change in the future, so end users shouldn't be using this.)

libs/cli/deepagents_cli/integrations/sandbox_factory.py
eyurtsev: We want to throw on any unabsorbed kwargs -- this behavior is correct for python. Otherwise incorrect variable names passed will end up being silently ignored

libs/cli/deepagents_cli/integrations/sandbox_factory.py
eyurtsev: if we're not allowing other kwargs, OK not to document them -- it's more of catch all for any typos in this case

libs/cli/deepagents_cli/integrations/langsmith.py
eyurtsev: Throwing is correct in this case unless we're actually propagating them into some API. Without throwing it will silently fail

eyurtsev (approved):

eyurtsev (dismissed):
there should be no changes to create_sandbox

libs/cli/deepagents_cli/integrations/sandbox_factory.py
```diff
@@ -68,6 +69,10 @@ def create_sandbox(
     *,
     sandbox_id: str | None = None,
     setup_script_path: str | None = None,
+    api_key: str | None = None,
```
eyurtsev: @copilot undo the changes in sandbox factory to what they were before. we don't need any changes in create_sandbox

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -136,16 +136,16 @@ def __init__(self, api_key: str | None = None) -> None:
...
+            ValueError: If no API key is provided and env var is not set.
         """
         from langsmith import sandbox
 
         self._api_key = api_key or os.environ.get("LANGSMITH_API_KEY")
         if not self._api_key:
-            msg = "LANGSMITH_API_KEY environment variable not set"
+            msg = "LangSmith API key not set"
             raise ValueError(msg)
         self._client: SandboxClient = sandbox.SandboxClient()
```
bracesproul: need to pass here

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -136,16 +136,16 @@ def __init__(self, api_key: str | None = None) -> None:
...
 
         Raises:
-            ValueError: If LANGSMITH_API_KEY environment variable not set
+            ValueError: If no API key is provided and env var is not set.
         """
         from langsmith import sandbox
 
         self._api_key = api_key or os.environ.get("LANGSMITH_API_KEY")
         if not self._api_key:
-            msg = "LANGSMITH_API_KEY environment variable not set"
```
bracesproul, line 148: revert err message

libs/cli/deepagents_cli/agent.py
bracesproul, line 1: revert all changes in this file

bracesproul (approved):
lgtm thanks!

PR #1108, opened 2026-02-05
feat(sdk): add sandbox provider exceptions

eyurtsev:
Adds a small exception hierarchy for sandbox providers (SandboxError and SandboxNotFoundError). SandboxError supports standard exception chaining (raise ... from e) and exposes the chained exception via original_exc for callers that need provider-specific inspection.


eyurtsev:
@copilot remove the typing.cast that is imported in that file and isn't needed

PR #1127, opened 2026-02-05
feat(cli): model switcher & arbitrary chat model support

mdrxy:
Closes #624
Closes #658
Closes #819
Closes #958
Closes #1071

> [!WARNING]
> Certain proprietary model features may be dropped as a result of switching models mid-thread (e.g. Gemini thought signatures, server side tools)

## Breaking changes
- Signature of `create_model` -- no longer use `model_name_override`
- Renames provider identifiers from `'google'`/`'vertexai'` to `'google_genai'`/`'google_vertexai'` to match `init_chat_model`'s expected provider strings

---

## Overview

Adds a `/model` command that lets users switch LLM providers and models mid-conversation without restarting the CLI. The switch is backed by `init_chat_model` from `langchain`, replacing the previous provider-specific model instantiation with a unified factory. Model preferences persist across sessions via a TOML config at `~/.deepagents/config.toml`.

* `/model` slash command has two modes:
  * `/model` opens an interactive modal selector with keyboard navigation, search filtering, and provider grouping
  * `/model provider:model` switches directly
* Can hot-swap the active agent mid-thread; preserves the existing checkpointer and conversation history
* Can switch to any `BaseChatModel` subclass

# Full documentation

The Deep Agents CLI supports any [chat model provider compatible with LangChain](https://docs.langchain.com/oss/python/integrations/providers/overview), unlocking use for virtually any LLM that supports tool calling. Any service that exposes an OpenAI-compatible or Anthropic-compatible API also works out of the box — see [Compatible APIs](#compatible-apis).

## Quick start

The CLI integrates automatically with the [following model providers](#provider-reference) — no extra configuration needed beyond installing the relevant provider package.

1. **Install provider packages**

    Each model provider requires installing its corresponding LangChain integration package. These are available as optional extras when installing the CLI:

    ```bash
    # Install with one provider
    uv tool install 'deepagents-cli[anthropic]'

    # Install with multiple providers at once
    uv tool install 'deepagents-cli[anthropic,openai,groq]'

    # Add additional packages at a later date
    uv tool upgrade deepagents-cli --with langchain-ollama

    # All providers
    uv tool install 'deepagents-cli[anthropic,bedrock,cohere,deepseek,fireworks,google-genai,groq,huggingface,ibm,mistralai,nvidia,ollama,openai,perplexity,vertexai,xai]'
    ```

3. **Set your API key**

    Most providers require an API key. Set the appropriate environment variable listed in the table below. Refer to each integration package's docs for details.

### Provider reference

[...]

PR #1131, opened 2026-02-05
Chinese Model provider

Malowking:
feat(cli): add Qwen and DeepSeek model support via ChatOpenAI

Adds Qwen and DeepSeek model support using langchain_community.chat_models.ChatOpenAI base implementation. Extends model providers while maintaining existing functionality.

Fixes #ISSUE_NUMBER

Changes:
- Added Qwen and DeepSeek to model providers
- Implemented ChatOpenAI-based model creation
- Added API key validation
- Preserved backward compatibility

Verification:
- `make format` and `make lint` passed
- Tested model instantiation
- Confirmed error handling works

No breaking changes.

### DIFF
--- a/libs/cli/deepagents_cli/config.py
+++ b/libs/cli/deepagents_cli/config.py
@@ -854,7 +854,7 @@ def create_model(model_name_override: str | None = None) -> BaseChatModel:
         model_name = os.environ.get("DEEPSEEK_MODEL", "deepseek-chat")
     elif settings.has_qwen:
         provider = "qwen"
-        model_name = os.environ.get("QWEN_MODEL", "qwen-plus")
+        model_name = os.environ.get("QWEN_MODEL", "qwen3-max")
     else:
         console.print("[bold red]Error:[/bold red] No credentials configured.")
         console.print("\nPlease set one of the following environment variables:")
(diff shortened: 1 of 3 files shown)


eyurtsev (changes requested):
We want to support this by allowing folks to configure something like

--completions-api --base-url --api-key etc

rather than using a lot of specific env variables per provider -- since a lot of providers rely on the exact same conventions

Using the completions API will be basically unlock all models at once that rely on completions API compatibility

Malowking:
> 我们希望通过允许用户配置类似的功能来支持这一点
> 
> --完成 --基础-URL --api-key等
> 
> 而不是每个供应商使用大量特定的环境变量——因为很多供应商依赖完全相同的约定
> 
> 使用完备API基本上可以一次性解锁所有依赖完备API兼容性的模型



> We want to support this by allowing folks to configure something like
> 
> --completions-api --base-url --api-key etc
> 
> rather than using a lot of specific env variables per provider -- since a lot of providers rely on the exact same conventions
> 
> Using the completions API will be basically unlock all models at once that rely on completions API compatibility

Even if we add this kind of configuration when starting the service, Google and Anthropic still have their own API invocation patterns. This means we would still need to add a lot of conditional logic to distinguish between the Completions API and other API styles.

If we introduce a .env file and switch to configuring the model through environment variables instead, the create_model method would only need to read parameters from the file, without requiring changes in many other places.

I’m not sure whether this approach is feasible, so I would greatly appreciate your advice. Thank you.

mdrxy:
Closing in favor of #1127

PR #1138, opened 2026-02-05
fix(deepagents): refactor summarization middleware

ccurme:
Migrate `SummarizationMiddleware` from the inherited `before_model`/`abefore_model` pattern to the `wrap_model_call` protocol. Instead of mutating LangGraph state with `RemoveMessage(REMOVE_ALL_MESSAGES)` on each summarization, the middleware now tracks a `SummarizationEvent` in private state and reconstructs the effective message list on each call, avoiding full state rewrites.

Sample trace: https://smith.langchain.com/public/19d8beae-abb7-4d47-b8b4-18b284804b84/r


mdrxy:
~description plz 🥺 ~

**edit: I added one**

ccurme:
> ~description plz 🥺 ~
> 
> **edit: I added one**

Thank you!

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -187,7 +214,8 @@ def __init__(
             )
             ```
         """
-        super().__init__(
+        # Initialize langchain helper for core summarization logic
+        self._lc_helper = LCSummarizationMiddleware(
```
sydney-runkle, line 220: should we just copy the logic over? Feels less fickle

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -211,36 +241,56 @@ def __init__(
...
-
-            tool_runtime = ToolRuntime(
-                state=state,
-                context=runtime.context,
-                stream_writer=runtime.stream_writer,
-                store=runtime.store,
-                config=config,
-                tool_call_id=None,
-            )
-            return self._backend(tool_runtime)
```
sydney-runkle, line 243: why did this change?

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -644,66 +730,58 @@ async def _aoffload_to_backend(
...
-                    RemoveMessage(id=REMOVE_ALL_MESSAGES),
-                    *truncated_messages,
-                ]
-            }
+            return handler(request.override(messages=truncated_messages))
 
         # If no truncation and no summarization
         if not should_summarize:
-            return None
+            return handler(request.override(messages=truncated_messages))
```
sydney-runkle, line 792: this is misleading (if not should summarize, override w/ truncated messages)

ccurme: truncated in this case refers to the truncated tool calls.

libs/deepagents/deepagents/middleware/summarization.py
ccurme: what's fickle about it? I didn't want to maintain two sets of tests for helper functions, initialization, etc.

PR #1143, opened 2026-02-05
chore(acp): rename to langchain-acp prior to release

eyurtsev:
Rename to langchain-acp for initial release


eyurtsev:
keeping as deepagents-acp for now

PR #1151, opened 2026-02-05
fix(cli): harden dictionary iteration and HITL fallback handling

### LINKED ISSUE #956: Bug: 'dictionary changed size during iteration' during parallel tool calls in subagent
## Description

When a subagent executes multiple tool calls in parallel (e.g., 8 concurrent `run_sql` calls), the application crashes with:

```
Error: dictionary changed size during iteration
```

## Environment

- **deepagents-cli version**: 0.0.9 (also checked latest 0.0.14 source - issue appears to persist)
- **deepagents version**: 0.2.7+
- **Python**: 3.11+
- **OS**: Windows 11 / WSL2

## Steps to Reproduce

1. Create an agent with a subagent that has access to a tool like `run_sql`
2. Give the subagent a task that requires validating multiple tables
3. The subagent attempts to execute 8 parallel `run_sql` calls (one per table)
4. Error occurs during tool result processing

Example scenario:
```python
# Subagent receives task to validate 8 tables
# It fires 8 parallel run_sql calls:
run_sql("SELECT COUNT(*) FROM table1")
run_sql("SELECT COUNT(*) FROM table2")
# ... 6 more parallel calls
# Error: dictionary changed size during iteration
```

## Suspected Root Cause

In `textual_adapter.py`, there are two dictionary iterations without `list()` protection:

**Line ~413:**
```python
for interrupt_id, hitl_request in pending_interrupts.items():  # Unprotected
```

**Line ~468:**
```python
for tool_msg in adapter._current_tool_messages.values():  # Unprotected
```

Compare with protected patterns elsewhere in the same file:
```python
for ns_key, pending_text in list(pending_text_by_namespace.items()):  # Safe
for tool_msg in list(adapter._current_tool_messages.values()):  # Safe (line 495)
```

## Suggested Fix

Wrap the unprotected iterations with `list()`:

```python
# Line 413
for interrupt_id, hitl_request in list(pending_interrupts.items()):

# Line 468  
for tool_msg in list(adapter._current_tool_messages.values()):
```

## Workaround

Instruct the subagent to execute tool calls **sequentially** rather than in parallel, or batch multiple queries into single calls using `UNION ALL`.

## Additional Context

This issue affects any workflow where a subagent needs to validate or query multiple resources concurrently. The parallel execution is a natural optimization that LLMs tend to use, making this a common failure mode.

---
*Reported via DeepMig project (https://github.com/ai4data/deepmig)*

saakshigupta2002:
Fixes #956

Defensive hardening of dictionary iterations in `textual_adapter.py`, plus fixes for incomplete cleanup in HITL fallback branches.

**Dictionary iteration:** Wraps all `_current_tool_messages` and `pending_interrupts` iterations with `list()` to snapshot before iterating.

**HITL fallback branches:** The `else` branches for unexpected decision types previously fell through to reject without updating tool widget state or clearing `_current_tool_messages`. These now properly call `set_rejected()` and `.clear()`, matching the explicit reject branch.

**Logging:** Replaces `with suppress(Exception)` with `try/except` + `logger.debug` so cancellation-cleanup failures are visible. Adds `logger.warning` for unexpected HITL decision types that previously failed silently.


mdrxy:
@saakshigupta2002 could you add a test that fails without these changes?

saakshigupta2002:
> @saakshigupta2002 could you add a test that fails without these changes?

sure! will do that soon

saakshigupta2002:
> @saakshigupta2002 could you add a test that fails without these changes?

Hey @mdrxy! 

Just pushed a commit with regression tests. I added a `TestDictIterationSafety` class with 6 tests that cover both `.items()` and `.values()` iteration patterns.

The key idea: I created `_MutatingItemsDict` and `_MutatingValuesDict` subclasses that delete a key mid-iteration, which deterministically triggers the same `RuntimeError: dictionary changed size during iteration` that happens in production when async tool callbacks fire during the HITL loop.

Two of the tests (`test_items_iteration_fails_without_list` and `test_values_iteration_fails_without_list`) prove that bare iteration crashes. The remaining tests prove that `list()` wrapping fixes it and that `_build_interrupted_ai_message` correctly uses the pattern.

 Let me know if you'd like any changes! :)

PR #1154, opened 2026-02-06
feat(sdk): add per-command `timeout` override to `execute()`

mdrxy:
Restore per-command timeout control that was lost when #1107 replaced the CLI's `ShellMiddleware` with `LocalShellBackend`.

The SDK's `execute()` previously only accepted `command: str` with the timeout fixed at init time, so the LLM couldn't extend it for long-running commands. This adds `timeout` as a keyword-only parameter across the entire `SandboxBackendProtocol` hierarchy and removes the CLI's `CLIShellBackend` wrapper that is no longer needed.

## Changes

- Add keyword-only `timeout: int | None = None` to `SandboxBackendProtocol.execute()` / `aexecute()` and propagate through `BaseSandbox`, `LocalShellBackend`, `CompositeBackend`, and all partner implementations
- Expose `DEFAULT_EXECUTE_TIMEOUT = 120` as a named constant
- Add `max_timeout` parameter to `FilesystemMiddleware.__init__` (default 3600s) that caps per-command overrides from the LLM

## (negligible) Breaking changes

`LocalShellBackend.__init__` timeout type narrowed from `float` to `int`.


libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -448,13 +448,21 @@ def id(self) -> str:
     def execute(
         self,
         command: str,
+        *,
+        timeout: int | None = None,
```
eyurtsev, line 453: This makes it entirely LLM controlled... which a DOS attack? If we're introducing this, is there a way to introduce a max timeout?

mdrxy: I can add a `max_timeout` init parameter (defaulting to e.g. `3600`) that caps the per-command value. If the LLM exceeds it, the middleware returns an explicit error (same pattern as the `<= 0` validation already in this PR):

```python
if timeout is not None and timeout > self._max_timeout:
    return f"Error: timeout {timeout}s exceeds maximum allowed
({self._max_timeout}s)."
```

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -462,6 +464,7 @@ def __init__(
         system_prompt: str | None = None,
         custom_tool_descriptions: dict[str, str] | None = None,
         tool_token_limit_before_evict: int | None = 20000,
+        max_timeout: int = 3600,
```
eyurtsev: Perhaps a more specific name given that timeout only applies to exec (not to read or write etc.)?

eyurtsev:
We're good to merge this one

Would need to merge against main, run make update-snapshots and potentially improve any prompts that need improving

Can help get this over the finish line on Friday

mdrxy:
> run make update-snapshots

had no effect @eyurtsev

PR #1169, opened 2026-02-06
feat(sdk): refactor namespace gen for StoreBackend

sydney-runkle:
## Summary                                                                                  
                                                                                             
* Add a namespace parameter to `StoreBackend` that accepts a factory function (`NamespaceFactory`) for flexible, user-controlled namespace resolution           
* Introduce BackendContext dataclass that provides the factory with access to both runtime state and context (e.g., user ID, workspace ID)                                   
* Deprecate the legacy implicit namespace behavior (assistant_id detection from metadata) with a DeprecationWarning

## Namespace Factory Usage

Admittedly the `ctx.runtime.context.user_id` is a bit ugly but this data structure is extensible and we are likely to change the `StoreBackend` signature anyways (moving away from accepting runtime.

```python
from deepagents.backends import StoreBackend, BackendContext

# 1) Simplest: static namespace
be = StoreBackend(rt, namespace=lambda _ctx: ("filesystem",))

# 2) User-scoped isolation via a named function
def user_namespace(ctx: BackendContext) -> tuple[str, ...]:
    return ("filesystem", ctx.runtime.context.user_id)

be = StoreBackend(rt, namespace=user_namespace)

# 3) Multi-level: workspace + user isolation
be = StoreBackend(
    rt,
    namespace=lambda ctx: (
        "workspace",
        ctx.runtime.context.workspace_id,
        "user",
        ctx.runtime.context.user_id,
    ),
)
```

### DIFF
--- a/libs/deepagents/deepagents/backends/store.py
+++ b/libs/deepagents/deepagents/backends/store.py
@@ -107,10 +107,14 @@ class StoreBackend(BackendProtocol):
             runtime: The ToolRuntime instance providing store access and configuration.
             namespace: Optional callable that takes a BackendContext and returns
                 a namespace tuple. This provides full flexibility for namespace resolution.
+                We forbid * which is a wild card for now.
                 If None, uses legacy assistant_id detection from metadata (deprecated).
 
                 .. note::
                     This parameter will be **required** in version 0.5.0.
+
+                .. warning::
+                    This API is subject to change in a minor version.
 
         Example:
                     namespace=lambda ctx: ("filesystem", ctx.runtime.context.user_id)


libs/deepagents/deepagents/backends/store.py
```diff
@@ -24,6 +29,67 @@
...
+
+    state: StateT
+    runtime: "Runtime[ContextT]"
+
+
+# Type alias for namespace factory functions
+NamespaceFactory = Callable[[BackendContext[Any, Any]], tuple[str, ...]]
+
+# Allowed characters in namespace components: alphanumeric, plus characters
+# common in user IDs (hyphen, underscore, dot, @, +, colon, tilde).
```
eyurtsev, line 49: Perhaps we inline somewhere that we forbid `*` which is a wild card for now.

libs/deepagents/deepagents/backends/store.py
```diff
@@ -34,13 +100,23 @@ class StoreBackend(BackendProtocol):
...
+                If None, uses legacy assistant_id detection from metadata (deprecated).
+
+                .. note::
+                    This parameter will be **required** in version 0.5.0.
+
+        Example:
+                    namespace=lambda ctx: ("filesystem", ctx.runtime.context.user_id)
         """
         self.runtime = runtime
+        self._namespace = namespace
```
eyurtsev, line 123: Should we do a warning here if no namespace is provided?

PR #1176, opened 2026-02-06
feat(deepagents): simplify sandbox interface

eyurtsev:
* Deprecate SandboxProvider (we'll delete it in 0.4.0)
* Introduce SandboxClient (slightly simpler interface)
* Update implementations in the CLI and daytona to use the simpler interface


libs/partners/daytona/tests/integration_tests/sandboxes.py
```diff
@@ -0,0 +1,339 @@
+"""Integration tests for the deepagents `SandboxClient` abstraction.
```
eyurtsev, line 1: This is here temporarily to help develop against new abstraction... I'll move it back to langchain_tests in a bit

hntrl (approved):
stamp

PR #1189, opened 2026-02-07
fix(sdk): harden skills metadata parsing, improve spec compliance

mdrxy:
- `_parse_skill_metadata` now rejects whitespace-only skill descriptions (previously accepted)
- Broaden `_validate_skill_name` to accept Unicode lowercase characters (e.g., `café`, `über-tool`) instead of ASCII-only, aligning with the Agent Skills spec definition of "lowercase alphanumeric"
- Add validation and truncation for the `compatibility` field (`MAX_SKILL_COMPATIBILITY_LENGTH = 500`), coerce `metadata` to `dict[str, str]`, and handle `allowed-tools` as both YAML list and space-delimited string
- Display license and compatibility annotations in the system prompt skill listing
- Fix edge cases (whitespace-only descriptions, boolean license values)

https://agentskills.io/specification

### DIFF
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -286,7 +286,7 @@ def _parse_skill_metadata(
         logger.warning("Skipping %s: frontmatter is not a mapping", skill_path)
         return None
 
-    name = frontmatter_data.get("name")
+    name = str(frontmatter_data.get("name", "")).strip()
     description = str(frontmatter_data.get("description", "")).strip()
     if not name or not description:
         logger.warning("Skipping %s: missing required 'name' or 'description'", skill_path)


libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -168,26 +204,38 @@ class SkillsStateUpdate(TypedDict):
...
+
+            Error message is empty if valid.
     """
     if not name:
         return False, "name is required"
     if len(name) > MAX_SKILL_NAME_LENGTH:
         return False, "name exceeds 64 characters"
-    # Pattern: lowercase alphanumeric, single hyphens between segments, no start/end hyphen
-    if not re.match(r"^[a-z0-9]+(-[a-z0-9]+)*$", name):
+    if name.startswith("-") or name.endswith("-") or "--" in name:
```
vtrivedy, line 232: fyi this diverges from cli's `deepagents skills create`, which is ascii only.  should square away formatting across them

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -237,10 +286,8 @@ def _parse_skill_metadata(
         logger.warning("Skipping %s: frontmatter is not a mapping", skill_path)
         return None
 
-    # Validate required fields
     name = frontmatter_data.get("name")
```
vtrivedy: also wrap in string like description?

libs/deepagents/deepagents/middleware/skills.py
mdrxy: will be ported

PR #1207, opened 2026-02-09
feat: add partner packages for modal and runloop

eyurtsev:
* Add partner packages for modal and run loop
* We'll move away form the SandboxProvider API as it's unnecessary easier to just let the user work with the underlying client directly to generate the sandbox.


eyurtsev:
There's some issue with the lock files workflow. I'm going to force commit and see if it's due to new file addition somehow.

PR #1234, opened 2026-02-10
release(deepagents): 0.4

ccurme:
Notable changes:
- Refactor the SDK's summarization middleware. Summarization now happens in the model node via `wrap_model_call` events. Due to this we retain the full message history in the graph state.
- System messages and tool definitions are accurately counted toward summarization triggers.
- Add exception handling for context overflows: ChatOpenAI and ChatAnthropic now raise a standard LangChain exception when input exceeds a model's token limit. We catch this error and trigger summarization.
- When model strings in `create_deep_agent` are prefixed with `openai:`, we default to the Responses API.


sydney-runkle (approved):
LGTM pending PR description

PR #1264, opened 2026-02-11
chore: migrate to uv workspace

sydney-runkle:
Migrate to a uv workspace with a single lockfile, and simplify CI checks to match.

## Summary

- Adds a root `pyproject.toml` that declares a **uv workspace** with lib packages as members (`libs/deepagents`, `libs/cli`, `libs/acp`, `libs/partners/*`)
- Replaces **6 per-package `uv.lock` files** with a **single workspace lockfile** at the repo root
- Converts inter-package path dependencies (`{ path = "../deepagents" }`) to **workspace references** (`{ workspace = true }`)
- Consolidates scattered `override-dependencies` and `constraint-dependencies` into the root `pyproject.toml`
- Simplifies `check_lockfiles.yml` CI — one `uv lock --check` for the workspace; removes harbor and example lockfile checks
- Removes the PR-level SDK pin check (`check_sdk_pin` job) — the same check already gates CLI releases in `release.yml`
- Updates `uv_setup` action cache globs to reference the root lockfile
- **Harbor is excluded** from the workspace because its upstream `harbor` PyPI dep requires Python `>=3.12`, and uv workspaces enforce all members' `requires-python` at sync time — excluding it lets the rest of the workspace support Python 3.11. Harbor keeps its own `uv.lock`.
- Examples (`examples/*`) remain independent with their own lockfiles

## Test plan

- [x] `uv lock --check` passes (workspace)
- [x] `uv lock --check --directory libs/harbor` passes (standalone)
- [x] `uv sync --package deepagents` / `deepagents-cli` / `deepagents-harbor` all succeed
- [x] `deepagents` unit tests pass (586 passed)
- [x] `deepagents-cli` unit tests pass (938 passed)
- [x] `make test` works from member directories
- [x] Example lockfiles remain independent
- [x] CI passes end-to-end


ccurme (approved):
Approving with caveat that if we start running into issues resolving all dependencies for all packages in a single env, we should revisit (e.g., if we add more standboxes or other integrations).

This will speed us up for now while we don't have many packages here.

PR #1295, opened 2026-02-11
feat(cli): expand local context & implement via bash for sandbox support

vtrivedy:
Rewrite `LocalContextMiddleware` to run environment detection as a bash script via `backend.execute()` instead of Python's `subprocess`/`pathlib` calls. The old implementation only worked locally — running detection through the backend means the same context-gathering logic works in remote sandboxes too.

## Changes
- Replace all previous Python-based detection methods with a single bash script built from composable section functions
- `LocalContextMiddleware` now requires a `backend` argument implementing the `_ExecutableBackend` protocol
- Move middleware registration in `create_cli_agent` outside the local/remote branch so it activates for any backend that supports shell execution, gated by `isinstance(backend, _ExecutableBackend)`
- Add new detection capabilities not in the old implementation: runtime version reporting (Python/Node versions), uncommitted change count, and `npm test` command detection via `package.json` scripts

The loop detection middleware is ported from the Harbor benchmark harness where it ran across 445 TB2 trials, adapted for CLI's human-in-the-loop context.


libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,231 @@
...
+        self._file_edit_counts: dict[str, int] = defaultdict(int)
+        self._warnings_shown: set[str] = set()
+        self._pending_reflection: tuple[str, int] | None = None
+        self._last_msg_count: int | None = None
+
+    # -- summarization reset ------------------------------------------------
+
+    def _reset_counts(self) -> None:
+        """Reset all tracking state (called after summarization)."""
+        self._file_edit_counts.clear()
```
eyurtsev: This is doing mutation in place rather than updating langgraph state. Because of the mutation in place the likelihood of bugs is **very** high. It'll likely break sub-agents which will be relying on the same instance or if ported to the sdk it'll break any concurrency on the main agent when running within a server

aaron-seq (changes requested):
**Disclosure:** This review was conducted with the assistance of AI tools, and I have reviewed and verified all findings.

## Summary

This PR adds valuable improvements to the CLI local context system with runtime version detection and git dirty file count tracking. However, there are critical CI/CD failures that must be addressed before merging.

## CI/CD Failures

### 1. Lint Errors (libs/cli/deepagents_cli/local_context.py)

**Line 148:** Line too long (109 > 88 characters)
```python
[line for line in result.stdout.strip().split("\n") if line.strip()]
```

**Recommendation:** Break this list comprehension into multiple lines for better readability:
```python
[
    line
    for line in result.stdout.strip().split("\n")
    if line.strip()
]
```

**Line 435 & 453:** Missing docstring parameter descriptions
- Line 435: Missing description for `cmd` and `label` parameters in the loop
- Line 453: Missing proper docstring format

### 2. Test Failures (test_local_context.py::TestLocalContextMiddleware)

**Critical Issue:** `test_get_git_info_no_main_branches` is failing with `StopIteration` error.

**Root Cause:** The `_get_git_info()` method now makes THREE `subprocess.run()` calls:
1. Get current branch
2. Get available branches  
3. **NEW:** Get git status for dirty count (lines 138-145)

However, the test mock only provides TWO side effects:
```python
mock_run.side_effect = [mock_branch_result, mock_branches_result]
```

When the code calls `subprocess.run()` a third time for git status, the mock exhausts its side effects and raises `StopIteration`.

**Fix Required:** Update ALL test cases in `test_local_context.py` to include a third mock for the git status call:
```python
mock_status_result = Mock()
mock_status_result.returncode = 0
mock_status_result.stdout = ""  # or mock dirty files

mock_run.side_effect = [
    mock_branch_result,
    mock_branches_result, 
    mock_status_result  # NEW
]
```

## Code Quality Observations

### Positive Aspects
1. Good security practices with `shutil.which()` validation and noqa comments
2. Proper timeout handling (2 seconds) prevents hanging
3. Clean separation of concerns with `_detect_runtime_versions()` method
4. Comprehensive docstrings explaining the feature

### Suggestions
1. The loop detection middleware file is well-documented and follows good practices
2. Consider adding unit tests for the new runtime version detection feature
3. The mutation concern raised by @eyurtsev in the conversation tab should be addressed

## Verdict

Requesting changes due to:
1. Lint violations that must be fixed
2. Test failures caused by missing mock side effects
3. These issues prevent the PR from passing CI/CD checks

Once these are resolved, the PR will be ready for approval.

libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,231 @@
+"""Middleware for detecting when the agent is stuck in an editing loop.
+
+NOT YET WIRED IN — this file is a proposal for review. To enable, add
+``LoopDetectionMiddleware()`` to the middleware list in ``agent.py:create_cli_agent()``.
```
ccurme: Why in CLI and not the SDK?

ccurme (commented):
(needs tests)

ccurme:
No worries if not easily accessible but attribution on score improvement from either middleware would be interesting.

vtrivedy:
@ccurme @eyurtsev scoped this down just for extending local context logic and implementing it via bash

libs/cli/deepagents_cli/local_context.py
```diff
@@ -619,4 +407,8 @@ async def awrap_model_call(
         return await handler(modified_request or request)
 
 
-__all__ = ["LocalContextMiddleware"]
+__all__ = [
+    "DETECT_CONTEXT_SCRIPT",
```
eyurtsev: Should probably be only LocalContextMiddleware?

libs/cli/deepagents_cli/local_context.py
```diff
@@ -17,571 +18,366 @@
...
+    echo "**Tree** (3 levels):"
+    echo '```text'
+    echo "$T"
+    echo '```'
+    echo ""
+  fi
+fi"""
+
+
+def _section_makefile() -> str:
```
eyurtsev, line 267: Do we have the path of the makefile included? (mostly wondering about mono-repo structure). Do we need to find the makefile?

eyurtsev (commented):
This looks pretty reasonable to me.

I think we mostly need to test with the CLI to make sure nothing breaks as the bash code has higher chance of having some subtle bugs.

cc @mdrxy what do you think? what would make you comfortable w/ functionality here?

mdrxy (approved):

PR #1297, opened 2026-02-11
feat(cli): `--acp` mode to run CLI agent as ACP server

eyurtsev:
Add a new `--acp` flag to `deepagents` (CLI) so the existing coding agent can run as an Agent Client Protocol (ACP) server over stdio, without launching the Textual UI.

## Why

- ACP clients (editors and orchestrators) can now drive the same agent used by `deepagents` interactive mode.
- We avoid a separate ACP-only agent entrypoint and reuse existing CLI model/agent configuration.

## Changes

- Added `--acp` to CLI argument parsing and help output.
- Added ACP runtime path in `cli_main` that creates the configured CLI agent, wraps it with `AgentServerACP`, and runs `acp.run_agent(...)` over stdio.
- Skipped Textual/UI dependency checks when `--acp` is used.
- Added runtime dependency: `deepagents-acp>=0.0.4`.

## Quick Tutorial

1. Start the ACP server:
   ```bash
   deepagents --acp
   ```

2. Connect from an ACP-capable client (for example: Zed, VS Code ACP extension, ACP UI).

3. In the client, create a session with your project directory as `cwd`, then send a prompt.
   Example prompts:
   - "Summarize this repository."
   - "Find failing tests and propose a fix."

4. Optional model configuration still works:
   ```bash
   deepagents --acp --agent agent -M gpt-5.2 --model-params '{"temperature":0.2}'
   ```

### DIFF
--- a/libs/cli/deepagents_cli/main.py
+++ b/libs/cli/deepagents_cli/main.py
@@ -897,7 +897,7 @@ def cli_main() -> None:
     # Check dependencies first
     check_cli_dependencies()
 
-    from deepagents_cli.config import console
+    from deepagents_cli.config import console, settings
 
     try:
         args = parse_args()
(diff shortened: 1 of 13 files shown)


libs/cli/deepagents_cli/main.py
```diff
@@ -644,6 +641,40 @@ def cli_main() -> None:
                 )
                 sys.exit(1)
 
+        if getattr(args, "acp", False):
```
eyurtsev, line 1049: Going to take a bit more time to hook up to the full parameterization since there are other things printing to stdout/stdin.

We need to plumb through all the shell config without polluting stdout/stin with console prints etc.

PR #1266, opened 2026-02-11
feat(cli): add fuzzy search to model switcher

peterkolcza:
## Summary

Replace substring-based filtering in the `/model` selector with Textual's built-in `Matcher` (`textual.fuzzy.Matcher`) for fzf-style fuzzy matching. Users can now find models with partial or approximate names (e.g., "cs45" matches "claude-sonnet-4-5", "gpt4" matches "gpt-4o"). Results are ranked by match quality.

Fixes #1228

## Changes

- **`libs/cli/deepagents_cli/widgets/model_selector.py`**: Replace substring `in` check with `Matcher` scoring in `_update_filtered_list()`. No new dependencies — `Matcher` is part of the existing `textual` dependency. No public API changes.
- **`libs/cli/tests/unit_tests/test_model_selector.py`**: Add 11 new tests covering fuzzy subsequence matching, cross-hyphen matching, case insensitivity, ranking, empty/whitespace filters, no-match edge case, selection clamping, and navigation after filtering.

## How was this verified?

- `make format` — passes
- `make lint` — passes
- `make test` — 949 tests pass (11 existing selector tests + 11 new fuzzy tests)
- No new dependencies added

## Areas requiring careful review

- The `Matcher` import path (`textual.fuzzy.Matcher`) — confirmed via Textual source and docs
- Flat sort by score relies on `_update_display()` to regroup by provider for rendering

## AI disclaimer

This contribution was developed with the assistance of AI agents (Claude) for planning, research, code generation, and review.

PR #1301, opened 2026-02-12
feat(cli): add substring matching to command history navigation

### LINKED ISSUE #1092: Substring matching in command history
### Privileged issue

### Issue Content

History contains:

```bash
git checkout main
docker compose up
npm run build
```

You type:

```bash
up
```

Press up arrow and we jump to:

```bash
docker compose up
```

Similar to `zsh-history-substring-search`

### Area (Required)

- [ ] deepagents (SDK)
- [x] cli

peterkolcza:
Fixes #1092

---

**Before:** typing `docker` and pressing up-arrow used `startswith(prefix)` — only matched entries *beginning with* the typed text. Typing `compose` wouldn't find `docker compose up`.

**After:** typing `compose` and pressing up-arrow uses case-insensitive `in` matching — finds any entry *containing* the query anywhere. The query is frozen on the first up-arrow press so the filter doesn't shift as the displayed text changes. Same behavior as `zsh-history-substring-search`.

## Changes

- Replace `startswith(prefix)` with case-insensitive `in` matching in `HistoryManager.get_previous` and `get_next`, with the query normalized to lowercase via `.strip().lower()` at capture time
- Capture the search query once on the first `get_previous` call of a navigation session (`_current_index == -1`) and freeze it — subsequent up-arrow presses reuse the same filter even as the displayed text changes
- Convert `prefix` positional parameter to keyword-only `query` on `get_previous`; remove the now-unused `prefix` parameter from `get_next` (internal class, not exported)
- Pass `query=event.current_text` from `ChatInput.on_chat_text_area_history_previous` to wire the feature into the UI
- Clear `_query` in `reset_navigation` so stale filters don't leak across sessions

> [!NOTE]
> This contribution was developed with the assistance of AI coding agents (Claude Code).


peterkolcza:
Hey @mdrxy 👋 This one is ready for review as well — builds on top of the same CLI UX improvements as #1266. Let me know if anything needs adjustment. Thanks!

mdrxy (approved):

PR #1305, opened 2026-02-12
feat: use `skill` tool

sydney-runkle:


sydney-runkle:
delaying context: fork implementation, can be added on later

sanathk12:
Curious, why have we delayed this feature ?

PR #1306, opened 2026-02-12
feat(sdk): reimplement skill loading via `skill` tool calling

sydney-runkle:
* implement skill loading via a `skill` tool like claude code and open code do :)
* support reading full skill.md content (instead of default cap at 100 lines that the read file tool uses)
* improves tracing experience for skill loading - you can easily see on the waterfall view when a skill is loaded + the skill content

before:
Screenshot 2026-02-12 at 2 21 04 PM

after:
Screenshot 2026-02-12 at 2 44 50 PM

future notes:
* paves the way for invoking skills in subagents long term, if we want
* this would provide a nice interface for "skills backends" or "skills sources" in the future. `SkillsMiddleware` could accept a spec for where to load skills from. In `before_agent` we could call `source.list_skills()` and in the tool we could call `source.load_skill(name=...)`. Right now skills sources are coupled with backends, but we could expand in the long term.

example trace showing `skill` tool, loaded content (> 100 lines): https://smith.langchain.com/public/fc18ae05-27f8-4ca5-b144-4ce949849954/r

### DIFF
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -726,8 +726,8 @@ class SkillsMiddleware(AgentMiddleware):
         """
         middleware = self
 
-        def _find_skill(skill_name: str, runtime: ToolRuntime) -> SkillMetadata | None:
-            """Find a skill by name from the current state."""
+        def _lookup_skill_metadata(skill_name: str, runtime: ToolRuntime) -> SkillMetadata | None:
+            """Look up a skill by name from the already-loaded metadata in state."""
             skills_metadata: list[SkillMetadata] = runtime.state.get("skills_metadata", [])
             for skill in skills_metadata:
                 if skill["name"] == skill_name:
@@ -738,7 +738,7 @@ class SkillsMiddleware(AgentMiddleware):
             name: Annotated[str, "The name of the skill to load."],
             runtime: ToolRuntime,
         ) -> str:
-            skill_meta = _find_skill(name, runtime)
+            skill_meta = _lookup_skill_metadata(name, runtime)
             if skill_meta is None:
                 available = [s["name"] for s in runtime.state.get("skills_metadata", [])]
                 return f"Skill '{name}' not found. Available skills: {', '.join(available) or 'none'}"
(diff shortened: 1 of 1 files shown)


libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -671,6 +705,84 @@ def _get_backend(self, state: SkillsState, runtime: Runtime, config: RunnableCon
...
+    def _get_backend_from_tool_runtime(self, runtime: ToolRuntime) -> BackendProtocol:
+        """Resolve backend from tool runtime context.
+
+        Args:
+            runtime: The tool runtime context.
+
+        Returns:
+            Resolved backend instance.
+        """
+        if callable(self._backend):
```
eyurtsev, line 717: no async right?

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -671,6 +705,84 @@ def _get_backend(self, state: SkillsState, runtime: Runtime, config: RunnableCon
...
+
+    def _create_skill_tool(self) -> StructuredTool:
+        """Create the `skill` tool for loading skills by name.
+
+        Returns:
+            A StructuredTool that loads skills by name.
+        """
+        middleware = self
+
+        def _find_skill(skill_name: str, runtime: ToolRuntime) -> SkillMetadata | None:
```
eyurtsev: Could we improve this name -- it sounds like it's working against the current state specifically for a cache look up. name right now sounds like it reachees to the backend

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -124,7 +124,7 @@ class CompiledSubAgent(TypedDict):
 #    be explicitly filtered from runtime.state when invoking a subagent to prevent parent state
 #    from leaking to child agents (e.g., the general-purpose subagent loads its own skills via
 #    SkillsMiddleware).
-_EXCLUDED_STATE_KEYS = {"messages", "todos", "structured_response", "skills_metadata", "memory_contents"}
+_EXCLUDED_STATE_KEYS = {"messages", "todos", "structured_response", "skills_metadata", "memory_contents", "loaded_skills"}
```
eyurtsev, line 127: Is loaded_skills here correct?

wirs1214-alt:
@sydney-runkle 
Does the skill loading tool take into account the resource loading of the reference?

PR #1314, opened 2026-02-12
feat(sdk): add base agent prompt

vtrivedy:
## Summary

  - Replace the one-line base prompt ("In order to complete the objective...you have access to a number of standard tools") with a `base_prompt.md` file covering core behavior, task workflow, tool usage, and
   file reading best practices
  - Load prompt from file via `Path(__file__).parent / "base_prompt.md"` in `graph.py`
  - Update test assertions to match new prompt content

  ## Motivation

  The SDK ships with essentially no behavioral guidance — the old prompt was a single sentence. Every other agent harness (Claude Code, Cursor, Codex CLI, etc.) includes baseline behavioral rules. This adds
  general-purpose guidance that applies to any deep agent, not just coding agents. The CLI layer adds coding-specific rules on top.

  ## What's in the prompt

  - **Core behavior** — concise, no preamble, ask before acting on ambiguity
  - **Professional objectivity** — accuracy over validation
  - **Following conventions** — read before editing, mimic existing style
  - **Task workflow** — understand → act → verify
  - **Tool usage** — prefer specialized tools, parallelize independent calls
  - **File reading** — pagination best practices
  - **Progress updates** — brief status for longer tasks

  ## Test plan

  - [x] Existing tests pass with updated assertions
  - [x] `base_prompt.md` included in package data (already covered by `*.md` in pyproject.toml)

### DIFF
--- a/libs/deepagents/deepagents/base_prompt.md
+++ b/libs/deepagents/deepagents/base_prompt.md
@@ -1,6 +1,4 @@
-You are a Deep Agent, an AI assistant that helps users accomplish tasks
-using tools. You respond with text and tool calls. The user can see your
-responses and tool outputs in real time.
+You are a Deep Agent, an AI assistant that helps users accomplish tasks using tools. You respond with text and tool calls. The user can see your responses and tool outputs in real time.
 
 ## Core Behavior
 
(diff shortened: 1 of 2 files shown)


libs/deepagents/deepagents/graph.py
```diff
@@ -33,7 +34,7 @@
 )
 from deepagents.middleware.summarization import SummarizationMiddleware, _compute_summarization_defaults
 
-BASE_AGENT_PROMPT = "In order to complete the objective that the user asks of you, you have access to a number of standard tools."
+BASE_AGENT_PROMPT = (Path(__file__).parent / "base_prompt.md").read_text()
```
eyurtsev, line 37: I usually try to inline content into SDKs rather than do a file at run time. I am not entirely sure there's any difference though, so fine as is...

There is an importlib.resources API: https://docs.python.org/3/library/importlib.resources.html


Might be good to explicitly specify encoding -- as default encoding may depend on the OS in older python versions (I'd check) this 

```python
.read_text(encoding='utf-8')
```

libs/deepagents/deepagents/base_prompt.md
```diff
@@ -0,0 +1,50 @@
...
+
+Keep working until the task is fully complete. Don't stop partway and explain what you would do — just do it. Only yield back to the user when the task is done or you're genuinely blocked.
+
+**When things go wrong:**
+- If something fails repeatedly, stop and analyze *why* — don't keep retrying the same approach.
+- If you're blocked, tell the user what's wrong and ask for guidance.
+
+## Tool Usage
+
+- Use specialized tools over shell equivalents when available (e.g., `read_file` over `cat`, `edit_file` over `sed`)
```
eyurtsev, line 38: This part of the prompt isn't correct since `cat` and `sed` are NOT available for all the regular backends. These are only available when `execute` is available. 

I suspect this will cause more issues for tool hallucination

PR #1327, opened 2026-02-13
POC: stateless loop detection middleware

vtrivedy:
- Rewrites the loop detection middleware proposed in #1295 to be fully stateless, addressing review feedback from @eyurtsev and @ccurme
- Edit counts and warning dedup are derived by scanning `state["messages"]` on each hook invocation — no mutable instance state
- Thread-safe, sub-agent safe (shared instance, different state dicts), and summarization-aware (old messages compress → counts naturally reset)

## How it works

- `wrap_tool_call`: after each `edit_file`/`write_file`, counts edits from messages. At soft threshold (8), appends a nudge to the ToolMessage content. Agent can acknowledge or ignore.
- `after_model`: counts edits from messages. At hard threshold (12) with no pending tool calls, injects a HumanMessage telling the agent to stop and ask the user, then forces a new model turn via `jump_to: "model"`.
- Warning dedup: checks if marker text already exists in message history. No sets or dicts on the instance.
- Path resolution: uses `Path.resolve()` so relative and absolute paths to the same file are counted together.
- Only instance attrs are `soft_threshold` and `hard_threshold` (immutable config).

Not yet wired into `agent.py` — kept as a standalone proposal with tests.

### DIFF
--- a/libs/cli/deepagents_cli/loop_detection.py
+++ b/libs/cli/deepagents_cli/loop_detection.py
@@ -1,0 +1,241 @@
+"""Stateless middleware that detects edit loops on the same file.
+
+NOT YET WIRED IN — add ``LoopDetectionMiddleware()`` to the middleware
+list in ``agent.py:create_cli_agent()`` to enable.
+
+Soft warning at 8 edits (appended to tool output), hard warning at 12
+(injected HumanMessage + ``jump_to: "model"``).  All counts derived
+from ``state["messages"]`` — no mutable instance state.
+"""
+
+from __future__ import annotations
+
+import logging
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+from langchain.agents.middleware.types import (
+    AgentMiddleware,
+    AgentState,
+    hook_config,
+)
+from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
+
+if TYPE_CHECKING:
+    from collections.abc import Awaitable, Callable
+
+    from langchain.tools.tool_node import ToolCallRequest
+    from langgraph.runtime import Runtime
+    from langgraph.types import Command
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
+# Warning templates
+# ---------------------------------------------------------------------------
+
+LOOP_WARNING_SOFT = """
+**NOTE**: You've edited `{file_path}` {count} times. If you're stuck, consider:
+- Is there a fundamentally different approach?
+- Can you use a different library or tool?
+- Should you step back and re-read the original requirements?
+"""
+
+LOOP_WARNING_HARD = """
+You've edited `{file_path}` {count} times without resolving the issue.
+
+Stop and ask the user what to do — describe what you've tried and what's
+not working so they can help you course-correct.
+"""
+
+# Tool names we track
+_EDIT_TOOL_NAMES = frozenset({"edit_file", "write_file"})
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _count_file_edits(messages: list[Any]) -> dict[str, int]:
+    """Count per-file edit_file/write_file calls from AIMessage tool_calls."""
+    counts: dict[str, int] = {}
+    for msg in messages:
+        if not isinstance(msg, AIMessage):
+            continue
+        for tc in getattr(msg, "tool_calls", []) or []:
+            if tc.get("name") not in _EDIT_TOOL_NAMES:
+                continue
+            args = tc.get("args") or {}
+            raw_path = args.get("file_path") or args.get("path") or "unknown"
+            path = str(Path(raw_path).resolve())
+            counts[path] = counts.get(path, 0) + 1
+    return counts
+
+
+def _soft_warning_already_shown(messages: list[Any], file_path: str) -> bool:
+    """Check if a soft warning for *file_path* was already appended."""
+    marker = f"**NOTE**: You've edited `{file_path}`"
+    for msg in messages:
+        if isinstance(msg, ToolMessage) and marker in str(msg.content):
+            return True
+    return False
+
+
+def _hard_warning_already_shown(messages: list[Any], file_path: str) -> bool:
+    """Check if a hard warning for *file_path* was already injected."""
+    marker = f"You've edited `{file_path}`"
+    for msg in messages:
+        if isinstance(msg, HumanMessage) and marker in str(msg.content):
+            return True
+    return False
+
+
+# ---------------------------------------------------------------------------
+# Middleware
+# ---------------------------------------------------------------------------
+
+
+class LoopDetectionMiddleware(AgentMiddleware):
+    """Detect edit loops on the same file via message scanning."""
+
+    DEFAULT_SOFT_THRESHOLD = 8
+    DEFAULT_HARD_THRESHOLD = 12
+
+    def __init__(
+        self,
+        *,
+        soft_threshold: int = DEFAULT_SOFT_THRESHOLD,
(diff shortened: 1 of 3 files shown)


libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,241 @@
...
+    counts: dict[str, int] = {}
+    for msg in messages:
+        if not isinstance(msg, AIMessage):
+            continue
+        for tc in getattr(msg, "tool_calls", []) or []:
+            if tc.get("name") not in _EDIT_TOOL_NAMES:
+                continue
+            args = tc.get("args") or {}
+            raw_path = args.get("file_path") or args.get("path") or "unknown"
+            path = str(Path(raw_path).resolve())
```
eyurtsev, line 71: A very minor security issue here:
* the path may be against a state backend, and it resolves against a local path.
* let's not use `Path(..).resolve()` to resolve anything. Given that our APIs already require absolute paths, could we just pass the path as is?

libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,241 @@
...
+                continue
+
+            logger.warning(
+                "Loop detection: %s edited %d times (hard warning)",
+                file_path,
+                count,
+            )
+            return {
+                "messages": [
+                    HumanMessage(
```
eyurtsev, line 229: Can we find a way to do this without mutating the message history with a HumanMessage?

This could be confusing (considered a bug) if a UI app attempts to render teh chat history and suddenly there's a message from a user but the user never typed it

eyurtsev: The way I'd propose to do this is instead using wrap_model_call if we really need to!

We re-prompt the model, but only surface the updated AI message in history.

libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,241 @@
...
+        return self._check_hard_threshold(state)
+
+    def _check_hard_threshold(self, state: AgentState[Any]) -> dict[str, Any] | None:
+        """Check if any file has hit the hard threshold and inject warning."""
+        messages = state.get("messages", [])
+
+        # Don't inject between AI tool_call and tool_result
+        if messages:
+            last_msg = messages[-1]
+            if hasattr(last_msg, "tool_calls") and last_msg.tool_calls:
```
eyurtsev, line 211: let's avoid using `hasattr`. We want as little as possible `hasattr` or `getattr` code in the SDK

Instead we want to (1) verify that the message is an AIMessage and if it's an AIMessage then it definitely has the `last_msg.tool_calls` attribute.

libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def __init__(
+        self,
+        *,
+        soft_threshold: int = DEFAULT_SOFT_THRESHOLD,
+        hard_threshold: int = DEFAULT_HARD_THRESHOLD,
+    ) -> None:
+        """Initialize with edit-count thresholds.
+
+        Raises:
```
eyurtsev, line 113: Nit: i'd remove documentation for raises if the exception is a common one

libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,241 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Awaitable, Callable
+
+    from langchain.tools.tool_node import ToolCallRequest
+    from langgraph.runtime import Runtime
+    from langgraph.types import Command
+
+logger = logging.getLogger(__name__)
+
+# ---------------------------------------------------------------------------
```
eyurtsev, line 33: nit: i'm personally not a big fan of the `---` sections for short files. (usually indicates that this code was all llm written)

libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def _count_file_edits(messages: list[Any]) -> dict[str, int]:
+    """Count per-file edit_file/write_file calls from AIMessage tool_calls."""
+    counts: dict[str, int] = {}
+    for msg in messages:
+        if not isinstance(msg, AIMessage):
+            continue
+        for tc in getattr(msg, "tool_calls", []) or []:
```
eyurtsev, line 66: Could we avoid `getattr`

if the msg is an AIMessage tool_calls are always defined on it

eyurtsev (commented):
Major: 
* HumanMessage addition in after model
* Path(path).resolve() usage (security)

libs/cli/tests/unit_tests/test_loop_detection.py
```diff
@@ -0,0 +1,519 @@
...
+    _hard_warning_already_shown,
+    _soft_warning_already_shown,
+)
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _ai_with_tool_calls(
```
eyurtsev, line 25: Could we instead inline `AIMessage(content=, ...)` directly? 

The helper feels like it's introducing round-about code

libs/cli/deepagents_cli/loop_detection.py
```diff
@@ -0,0 +1,241 @@
...
+        result: ToolMessage | Command[Any],
+    ) -> ToolMessage | Command[Any]:
+        """Append soft warning to tool result if threshold is reached."""
+        tool_name = request.tool_call.get("name", "")
+        if tool_name not in _EDIT_TOOL_NAMES:
+            return result
+
+        args = request.tool_call.get("args") or {}
+        raw_path = args.get("file_path") or args.get("path") or "unknown"
+        file_path = str(Path(raw_path).resolve())
```
eyurtsev, line 158: Do we allow relative paths?

`str(Path(raw_path).resolve())` is relative to the local file system, but the backend may be in a sandbox

eyurtsev:
What would happen for CLI coding tasks when I'm doing a refactor and updating the same file... I think it can be common to have many edits in just one file?

mdrxy:
Fresh use case argues for reviving (and reshaping).

**Today's incident:** eval CI hung on `test_chain_search_then_email` against `baseten:moonshotai/Kimi-K2.6`. Trace shows **712 `web_search` calls** with identical args before the 6h job timeout would've kicked in. Pattern: the model read a stub's echo response as "empty results," reformulates in text output, but re-emits the same `tool_calls.args`.

The backstop this PR provides is exactly the right shape — but two gaps stop it from helping here:

1. **Allowlist scope.** `_EDIT_TOOL_NAMES = {"edit_file", "write_file"}` skips everything else. Today's repeat was `web_search`; tomorrow's could be any tool whose output the model misreads.
2. **CLI-only.** Lives in `libs/cli/`; evals and other `create_deep_agent` callers don't pick it up.

**Proposal — revive, but generalize:**

- Lift to `libs/deepagents/deepagents/middleware/` so it ships with the SDK, not just the CLI shell.
- Switch from "per-file edit count" to **"identical `(tool_name, args)` in a row, ≥ N times"**. This:
  - Catches today's bug (90 captured tool calls all had `args={"query": "LangGraph 0.3 release notes"}`).
  - Defuses the legitimate-refactor concern @eyurtsev raised earlier — a real refactor varies args every call and never trips, even at 50+ edits to one file.
  - Lets the threshold drop much lower (5–8 in a row vs. 12 cumulative) without false positives.

Existing review feedback still applies and carries over:
- Drop `Path.resolve()` — moot if we remove the file-tracking dimension.
- No synthetic `HumanMessage`; re-prompt via `wrap_model_call` instead.
- Type-narrow on `AIMessage` rather than `hasattr`/`getattr`.
- Style nits.

Will be superseding with a PR.

PR #1334, opened 2026-02-14
feat(deepagents): Support gitignore for glob tool in filesystem backend

jacoblee93:
`glob` matches things like `node_modules` which most people wouldn't want


jacoblee93:
I actually think this should be the default behavior CC @vtrivedy

jacoblee93:
> Can't node modules be used to reference the implementation of the underlying library? I use this functionality all the time with .venv when in the cli

I was envisioning something where if it had to look there, it could use the shell

But yeah may need to rethink it more

PR #1339, opened 2026-02-14
fix(cli): preserve path context in file tool display

### LINKED ISSUE #1142: Full path is dropped when rendering tools
### Privileged issue

### Issue Content

edit file doesn't show full path... this is very confusing in a mono repo , especially when working with worktrees

Image

## Root Cause

`abbreviate_path` in `libs/cli/deepagents_cli/ui.py` falls back to `path.name` (basename only) for long absolute paths. In monorepos/worktrees, multiple packages contain identically named files — showing only the basename is ambiguous.

## Fix

PR #1339 replaces the basename-only fallback with middle truncation (`truncate_path_middle`), preserving both the leading directory context and the filename joined by an ellipsis (`…`).

### Area (Required)

- [ ] deepagents (SDK)
- [x] cli

weiguangli-io:
Fix the path display in CLI tool call output so users can distinguish files in monorepos and worktrees. Previously, `abbreviate_path` fell back to showing only the basename (e.g., `file.py`) for long absolute paths — useless when multiple packages contain identically named files.

Fixes #1142

## Changes
- Add `truncate_path_middle` helper inside `format_tool_display` that preserves both the leading directory context and the filename, joining them with an ellipsis (`…`) when the path exceeds 60 characters — replaces the old basename-only fallback in `abbreviate_path`
- Both the error/fallback branch and the normal long-path branch in `abbreviate_path` now route through `truncate_path_middle` instead of `truncate_value` / `path.name`

## Testing
- Add `test_read_file_long_absolute_path_keeps_context` asserting that a deeply nested absolute path retains directory context and the ellipsis glyph in the formatted output

---

### DIFF
--- a/libs/cli/deepagents_cli/ui.py
+++ b/libs/cli/deepagents_cli/ui.py
@@ -92,6 +92,21 @@ def format_tool_display(tool_name: str, tool_args: dict) -> str:
     """
     prefix = get_glyphs().tool_prefix
 
+    def truncate_path_middle(path_str: str, max_length: int = 60) -> str:
+        """Truncate a path while preserving both prefix and filename.
+
+        Returns:
+            A potentially shortened path string with middle ellipsis.
+        """
+        if len(path_str) <= max_length:
+            return path_str
+        ellipsis = get_glyphs().ellipsis
+        if max_length <= len(ellipsis) + 2:
+            return truncate_value(path_str, max_length)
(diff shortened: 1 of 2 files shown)


weiguangli-io:
Closing — this has merge conflicts and hasn't received review. Happy to resubmit if the issue is still relevant.

Issue #1348, opened 2026-02-15
`_python_search` glob matching silently drops all files when pattern contains `**`


cordialgerm:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

---
# Bug: `_python_search` glob matching silently drops all files when pattern contains `**`

## Summary

`FilesystemBackend.grep_raw()` returns zero results whenever a glob pattern
containing `**` is passed (e.g. `**/*.py`), but only when ripgrep (`rg`) is
not available as a system binary. The fallback Python search implementation
has two bugs in its glob matching that cause every file to be filtered out.

## Environment

- `deepagents` backends (`FilesystemBackend`, `CompositeBackend`)
- Any system where `rg` is not on `PATH` as a real binary (shell aliases
  don't count — `subprocess.run(["rg", ...])` can't resolve them)

## Root Cause

In `deepagents/backends/filesystem.py`, the `_python_search` method (the
fallback used when ripgrep is unavailable) filters files on line 397:

```python
if include_glob and not wcglob.globmatch(fp.name, include_glob, flags=wcglob.BRACE):
    continue
```

Two problems:

fp.name is just the filename (e.g. "main.py"), not the relative
path. Glob patterns like **/*.py expect path components to match
against. A bare filename will never match a pattern that contains /.

The GLOBSTAR flag is missing. Without it, wcmatch treats ** as
a literal pattern segment rather than a recursive wildcard. Even if the
full relative path were used, **/*.py still wouldn't match because
** isn't interpreted as "any number of directories."

The ripgrep path (_ripgrep_search) handles --glob **/*.py correctly, so
this bug is completely hidden on systems with rg installed as a real binary.

Reproduction

```python
from pathlib import Path
from deepagents.backends import FilesystemBackend

# Create a backend with files
tmp = Path("/tmp/test_glob_bug")
tmp.mkdir(exist_ok=True)
(tmp / "top.py").write_text("def hello(): pass\n")
sub = tmp / "pkg"
sub.mkdir(exist_ok=True)
(sub / "module.py").write_text("def world(): pass\n")

backend = FilesystemBackend(root_dir=str(tmp), virtual_mode=True)

# Works — no glob
[...]

PR #1350, opened 2026-02-15
fix(sdk): fix glob matching with ** in _python_search fallback

dotuananh0712:
## Summary

Fixed issue #1348 where glob patterns containing `**` silently return zero results when ripgrep is unavailable.

## Problem

The `_python_search` fallback method had two bugs:
1. Used `fp.name` (filename only) instead of relative path for glob matching
2. Missing `GLOBSTAR` flag, so `**` wasn't treated as recursive wildcard

## Fix

```python
# Before (broken):
if include_glob and not wcglob.globmatch(fp.name, include_glob, flags=wcglob.BRACE):
    continue

# After (fixed):
if include_glob:
    rel_path = str(fp.relative_to(root))
    if not wcglob.globmatch(
        rel_path, include_glob, flags=wcglob.BRACE | wcglob.GLOBSTAR
    ):
        continue
```

Now matches against relative path and adds GLOBSTAR flag for proper `**` matching.

## Testing

- Reproduced the bug with test case from issue
- Verified fix works with glob patterns like `**/*.py`

AI assistance used for code review and implementation

### DIFF
--- a/libs/deepagents/tests/unit_tests/backends/test_filesystem_backend.py
+++ b/libs/deepagents/tests/unit_tests/backends/test_filesystem_backend.py
@@ -54,7 +54,7 @@ def test_filesystem_backend_virtual_mode(tmp_path: Path, monkeypatch: pytest.Mon
     write_file(f1, "hello virtual")
     write_file(f2, "content")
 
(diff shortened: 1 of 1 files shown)


jacoblee93:
Thank you!

PR #1365, opened 2026-02-17
chore: enable type checking for all other packages

sydney-runkle:
* enable typechecking
* move away from blanket ignores, per line ignores have a reasoning comment
* enforced in CI :)

### DIFF
--- a/libs/acp/deepagents_acp/server.py
+++ b/libs/acp/deepagents_acp/server.py
@@ -1,7 +1,10 @@
+"""ACP server implementation for Deep Agents."""
+
+from __future__ import annotations
+
 import json
-from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Any, Optional
+from typing import TYPE_CHECKING, Any
 from uuid import uuid4
 
 from acp import (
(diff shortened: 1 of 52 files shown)


libs/acp/deepagents_acp/server.py
```diff
@@ -365,45 +381,44 @@ def _create_tool_call_start(
...
             return start_tool_call(
                 tool_call_id=tool_id,
                 title=title,
                 kind=tool_kind,
                 status="pending",
                 raw_input=tool_args,
             )
-        elif tool_name == "execute" and isinstance(tool_args, dict):
-            command = tool_args.get("command", "")
+        if tool_name == "write_file" and isinstance(tool_args, dict):
```
eyurtsev, line 392: Could you double check logic here is correct -- a bit hard to tell from the review... probably OK

eyurtsev (approved):
looks good, could we get @mdrxy to review the CLI?

libs/acp/deepagents_acp/server.py
sydney-runkle: yep, easier if you strip whitespace

sydney-runkle: ya it's right bc there's returns

mdrxy (approved):

PR #1369, opened 2026-02-17
feat(sdk): allow specifying agent name param for general purpose sub-agent

hwchase17:
## Summary

Allow users to override the default general-purpose subagent configuration in `create_deep_agent` by passing a subagent with the name `"general-purpose"` in the `subagents` list.

Previously, the built-in general-purpose subagent was always prepended, making it impossible to customize its model, system prompt, or other settings. Now, if a user-provided subagent shares the `"general-purpose"` name, it replaces the default instead of duplicating it.

## Changes

- **`libs/deepagents/deepagents/graph.py`**: Check whether `processed_subagents` already contains a `"general-purpose"` entry before prepending the default spec. Also inlines the base agent prompt and adds a minor noqa annotation.
- **`libs/deepagents/tests/unit_tests/test_subagents.py`**: Adds `test_general_purpose_subagent_override` to verify that a user-supplied general-purpose subagent fully replaces the default.

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -79,7 +79,7 @@ def get_default_model() -> ChatAnthropic:
     )
 
 
-def create_deep_agent(  # noqa: C901, PLR0912  # Complex graph assembly logic with many conditional branches
+def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly logic with many conditional branches
     model: str | BaseChatModel | None = None,
     tools: Sequence[BaseTool | Callable | dict[str, Any]] | None = None,
     *,


libs/deepagents/deepagents/graph.py
```diff
@@ -48,13 +48,14 @@ def get_default_model() -> ChatAnthropic:
...
 
-def create_deep_agent(  # noqa: C901, PLR0912  # Complex graph assembly logic with many conditional branches
+def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly logic with many conditional branches
     model: str | BaseChatModel | None = None,
     tools: Sequence[BaseTool | Callable | dict[str, Any]] | None = None,
     *,
     system_prompt: str | SystemMessage | None = None,
     middleware: Sequence[AgentMiddleware] = (),
     subagents: list[SubAgent | CompiledSubAgent] | None = None,
+    subagent_model: str | BaseChatModel | None = None,
```
eyurtsev: Do we have a way to specify t his through subagents instead?

General agent is forcing this arg to appear at top level of API, but it spreads subagent configuration into multiple locations

sydney-runkle (approved):

PR #829, opened 2026-01-18
feat(cli): add interactive step-into mode for subagents

juanwisz:
## Problem

Subagent delegation is currently fire-and-forget. You specify a task, the agent runs in isolation, and you get a summary back. If the results are wrong, you start over.

This works for simple tasks but fails for anything requiring nuance. The user can't:
- Course-correct mid-exploration
- Say "not that direction, focus here"
- Guide what ends up in the summary

**Context isolation was conflated with loss of control.** They're orthogonal concerns.

## Solution

Add a "Step into" option to the HITL prompt when the agent invokes the `task` tool:

```
[Approve]  [Reject]  [Step into]  [Auto-accept all]
```

- **Approve** = fire-and-forget (existing behavior)
- **Step into** = enter the subagent conversation interactively

When stepping in:
- User enters an interactive session with the subagent
- Prompt shows depth: `[general-purpose:1] >`
- Summary file created at `~/.deepagents/{agent}/branches/{id}/summary.md`
- `/return` exits and sends summary to parent context
- `/summary` edits the summary file
- `/context` shows position in context stack

Context stays isolated. User stays in control. Both concerns addressed.

## Changes

- `config.py`: Add `ConversationContext`, refactor `SessionState` to use context stack
- `execution.py`: Add "Step into" to HITL, create branch context
- `main.py`: Context routing, step-in/return handling  
- `commands.py`: Add `/return`, `/summary`, `/context` commands
- `input.py`: Dynamic prompt with depth indicator
- `STEP_INTO.md`: Documentation

## Example

```
You: Research the auth system

Agent: I'll delegate this to a subagent.

[Select "Step into"]

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Stepped into: general-purpose subagent
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

[general-purpose:1] Agent: Found 3 auth approaches. Which to focus on?

[general-purpose:1] You: JWT only.

[general-purpose:1] Agent: [explores JWT, reports findings]

[general-purpose:1] You: Add that to the summary.

[general-purpose:1] You: /return

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
✓ Returned to root
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

Agent: Based on the findings... [uses summary you helped craft]
```

[...]

PR #1372, opened 2026-02-17
feat(code): add step-into mode for subagents

juanwisz:
Adds interactive **step-into mode** for subagent invocations in the CLI. Instead of fire-and-forget delegation, users can enter an interactive session with a subagent while maintaining context isolation.

- When the HITL approval prompt appears for a `task` tool call, a new **"Step into"** option lets users enter the subagent conversation interactively
- New slash commands: `/return` (exit subagent, send summary to parent), `/summary` (view/edit summary file), `/context` (show conversation stack)
- Context stack tracks nested subagent sessions with isolated thread IDs
- Summary file created per branch for curating what returns to the parent

This is a re-port of #829 against the current Textual-based CLI architecture (the original PR targeted the old Rich+prompt_toolkit code in `libs/deepagents-cli/` which has since been replaced by `libs/cli/`).


mdrxy:
cool, thank you @juanwisz for re-opening on top of the recent changes.

would you mind attaching a screen recording of the functionality to the PR description? you should be able to drag and drop a video file and it will upload inline

juanwisz:
https://github.com/user-attachments/assets/c2cf7ec1-bb33-4fb6-97ab-f0c7116f2c70

@mdrxy done

mdrxy:
Interesting, thank you @juanwisz for the recording

2 things that stand out immediately
1. We should probably more clearly indicate that stepping into a conversation = implies approving the action (this is ambiguous right now)
2. Whilst "stepped in", I'd like to see some visual indicator to inform the user that they are inside of a subagent. Open to ideating on this

juanwisz:
@mdrxy I will work on the changes for this. Regarding point 1, you’re right.

As for point 2, if you watch the video closely (at 0:08), you’ll see that we indicate when we step into a specific subagent. However, we could make this much clearer by always displaying the current node and the depth level.

I think this feature is very strong. Today, it really comes down to two things: good context management and avoiding unnecessary interface changes, even something as small as opening another CLI. This feature lets you micro-manage context very efficiently by only injecting what is actually necessary, and it allows you to do everything from the same terminal.

mdrxy:
scratchpad
- markdown bold formatting on `/context` not working
- `running...` spinner for root invocation not stopping timer after task finishes

mdrxy:
I'd like this to possibly be a sub-chat window with clear visual indication and persistent tooltips for the available commands (currently you need to scroll all the way up to see)

juanwisz:
@mdrxy All feedback addressed in the latest commits:

1. **Step-into implies approval** — option text now reads `Step into (approve + interactive) (s)` to make it unambiguous.

2. **Visual indicator while stepped-in** — added a persistent `SubagentBanner` widget docked above the chat input. It shows the subagent type, nesting depth, and available commands (`/return`, `/summary`, `/context`) at all times while inside a subagent. No more scrolling up to find them. The prompt indicator also changes to `[subagent-type:depth] >`.

3. **Markdown bold on `/context`** — fixed. `AppMessage` wraps content in a Rich `Text` object which doesn't parse markdown syntax. Switched to using Rich `Text` with `style="bold"` directly.

4. **Spinner not stopping** — fixed. The spinner wasn't being cleared when the `execute_task_textual` function returned early on step-into. Added the `_set_spinner(None)` guard before returning.

Also merged latest `origin/main` (was 200+ commits behind) and added **25 unit tests across 7 test classes** covering: adapter flow with fake agent streaming + HITL interrupts, banner rendering via Textual `run_test()`, approval menu option/keybinding logic, Rich Text context output validation, and edge cases (args preservation, thread_id compat, stack isolation).

Happy to record a new demo video if that would help!

mdrxy:
awesome, thanks @juanwisz 

one bug I've found is that if I reject an interrupt while inside a stepped-into agent, it bubbled up to the top-level task request -- this doesn't make sense, since the subagent isn't necessarily done with execution. does that make sense? pressing `esc` should scope the rejection to the current HITL interrupt

juanwisz:
@mdrxy Nice catch, fixed in the latest commit.

The issue was that `execute_task_textual` treated all rejections the same way - it set `suppress_resumed_output = True` and returned early. That works fine at root level where you want the user to type a new message, but inside a subagent it kills the whole agent loop.

The fix checks `session_state.depth` before returning early. If we are inside a subagent (depth > 0), the rejection goes through `Command(resume=hitl_response)` with the `RejectDecision` instead of returning. The agent gets the rejection, adapts, and keeps going. User sees "Rejected - agent will continue." Root level behavior stays the same.

Added two tests: one verifying root rejection still returns early, one verifying subagent rejection resumes the agent (called twice instead of once).

juanwisz:
Did a deeper audit of the step-into lifecycle and found 4 more bugs that I've fixed:

1. **parent_tool_call_id never wired up** - the field existed on ConversationContext but was always None. Now it captures the interrupt ID when stepping in, so /return can properly resume the parent thread.

2. **auto_approve leaked across contexts** - toggling auto-approve inside a subagent changed it globally. Now push_context saves the parent's value and pop_context restores it.

3. **context not popped on subagent error** - if the subagent crashed, the user got stuck in a dead context with the banner still showing. Now the except block pops the context automatically.

4. **branch directories never cleaned up** - summary files accumulated in ~/.deepagents/branches/ forever. Now /return cleans up after reading the summary.

All 32 tests passing. Added 7 new tests specifically for these bugs.

juanwisz:
@mdrxy hey, since the rejection scoping fix we added 34 unit tests covering the full step-into lifecycle and re-merged latest origin/main. CI is green, ready for review when you get a chance.

mdrxy:
nice thanks @juanwisz though it appears there are some conflicts still - would you mind resolving?

mdrxy:
hey @juanwisz — checking in on this. are you still able to resolve the merge conflicts?

on our end, we want to spend a bit more time thinking through the architecture and UX/layout before pushing this through — specifically around the sub-chat window approach and how the visual hierarchy feels in practice. want to make sure we land on something solid since this will set the pattern for how users interact with subagents going forward.

appreciate the work you've put into this so far — the rejection scoping fix and test coverage are great. will circle back with more concrete feedback once we've had time to test further.

juanwisz:
@mdrxy conflicts resolved and pushed. Also added two things on top:

- **Recursive step-into**: if a subagent spawns another subagent and the user steps into that one too, the loop now handles it properly instead of dropping the nested context.
- **Breadcrumbs in the banner**: at depth 3+, the banner shows the full path (e.g. `root > researcher > coder`) instead of just the current node. At depth 1-2 it stays as `[type:depth]` to avoid clutter.

Re: the sub-chat window approach and visual hierarchy, totally understand wanting to get that right before merging. A few thoughts on my end:

The current banner + prompt indicator combo is intentionally lightweight. It avoids splitting the terminal into panes or opening a separate view, which keeps the interaction model simple (one input, one output stream, context switches are explicit). That said, if you want a more visually distinct separation between parent and subagent conversations, one option would be a collapsible panel that holds the subagent exchange, with the parent conversation dimmed or scrolled above. That would make the boundary clearer without requiring a full sub-window.

Happy to prototype whatever direction you land on. Let me know what feels right after you've had time to test.

mdrxy:
Closing this for now to keep the PR queue tidy — not a rejection. The step-into mode is a feature we'll eventually want, and this is a solid reference for the implementation. We'll re-open / pick this back up once we're ready to build it against the current architecture.

Thanks @juanwisz for the re-port work — it won't go to waste. 🙏

PR #1379, opened 2026-02-18
feat(sdk,cli): add package version tracking to tracing metadata

mdrxy:
Depends on https://github.com/langchain-ai/langchain/pull/35295
Should consider bumping min version

---

LangSmith traces from `deepagents`/`deepagents-cli` don't include installed package versions, making it hard to correlate failures with specific versions. Every LangSmith run now automatically carries `deepagents_version` and (when using the CLI) `deepagents_cli_version` in run metadata.

### DIFF
--- a/libs/cli/deepagents_cli/config.py
+++ b/libs/cli/deepagents_cli/config.py
@@ -12,7 +12,11 @@ import sys
 import uuid
 from dataclasses import dataclass
 from enum import StrEnum
-from importlib.metadata import PackageNotFoundError, distribution
+from importlib.metadata import (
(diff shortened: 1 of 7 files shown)


libs/cli/deepagents_cli/config.py
sydney-runkle, line 1: this file should be private right (utils)

mdrxy:
Closing in favor of a fresh approach. Context: version metadata will be reworked after langchain-core PR #35295 lands with the JS-style `_add_version()` pattern (constructor-based, no child clobber risk). See Slack threads for full discussion.

PR #1398, opened 2026-02-18
feat(sdk): support `utf-8` and `base64` encodings in `read` / `write` paths

sydney-runkle:
## Summary
Changes file storage from `list[str]` (newline-split lines) to plain `str` with an `encoding` field to distinguish text from binary content. Fully backwards compatible.

## Backwards Compatibility
- **Reading old data**: Legacy `list[str]` content auto-converts to `str` with `DeprecationWarning`
- **Writing old format**: Set `file_format="v1"` to continue using `list[str]` (no `encoding` field)
- **Import compat**: `FileData` still available from `deepagents.middleware.filesystem`

## Motivation
The old `list[str]` format had several issues:
- **Semantic mismatch**: Files are strings/bytes, not line arrays. Splitting on `\n` is a presentation concern (for `read` tool), not a storage concern.
- **No binary support**: Couldn't handle images, PDFs, or any non-UTF-8 content in `read`
- **Unnecessary overhead**: Every write split strings, every read joined them back. Not high cost, but unnecessary.

### New `FileData` format
```python
# Old
{"content": ["hello", "world"], "created_at": ..., "modified_at": ...}

# New  
{"content": "hello\nworld", "encoding": "utf-8", "created_at": ..., "modified_at": ...}
```

| Field | Old | New |
|---|---|---|
| `content` | `list[str]` | `str` (UTF-8 text or base64 binary) |
| `encoding` | *(none)* | `"utf-8"` or `"base64"` |

### Backend API
Both `StateBackend` and `StoreBackend` accept a `file_format` parameter:
```python
# Default — new format
StateBackend(runtime)  # equivalent to file_format="v2"

# Legacy format for backwards compat
StateBackend(runtime, file_format="v1")
```

Using `Literal["v1", "v2"]` instead of a boolean allows future versions (e.g. `"v3"`) without API breakage.


libs/deepagents/deepagents/backends/store.py
```diff
@@ -198,44 +200,62 @@ def _get_namespace_legacy(self) -> tuple[str, ...]:
             return (assistant_id, namespace)
         return (namespace,)
 
-    def _convert_store_item_to_file_data(self, store_item: Item) -> dict[str, Any]:
+    def _convert_store_item_to_file_data(self, store_item: Item) -> FileData:
```
sydney-runkle, line 218: improvement!

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -106,6 +116,22 @@ class GrepMatch(TypedDict):
     text: str
 
 
+class FileData(TypedDict):
+    """Data structure for storing file contents with metadata."""
+
+    content: str
+    """File content as a plain string (utf-8 text or base64-encoded binary)."""
+
+    encoding: str
```
eyurtsev, line 130: Do we need encoding? Is the agreement that encoding is always based on file extension then encoding is probably not important?

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -160,17 +186,24 @@ class EditResult:
...
+    All file data is represented as dicts with the following structure::
+
+        {
+            "content": str,  # Text content (utf-8) or base64-encoded binary
+            "encoding": str,  # "utf-8" for text, "base64" for binary data
+            "created_at": str,  # ISO format timestamp
+            "modified_at": str,  # ISO format timestamp
+        }
+
+    .. note::
```
eyurtsev: we use mkdocs format

libs/deepagents/deepagents/backends/state.py
```diff
@@ -37,9 +41,33 @@ class StateBackend(BackendProtocol):
     This is indicated by the uses_state=True flag.
     """
 
-    def __init__(self, runtime: "ToolRuntime") -> None:
-        """Initialize StateBackend with runtime."""
+    def __init__(
+        self,
+        runtime: "ToolRuntime",
+        *,
+        file_format: FileFormat = "v2",
```
eyurtsev, line 50: If we proceed w/ migration then:
  
* Default should be None.  
* None -> selects v1 and emits a deprecation warning
* We inline the literal in the type annotation?

eyurtsev, line 50: Please note that `v2` still has to accept v1 format due to suspended threads

libs/deepagents/deepagents/backends/store.py
```diff
@@ -585,12 +635,16 @@ def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadRespons
         responses: list[FileUploadResponse] = []
 
         for path, content in files:
-            content_str = content.decode("utf-8")
-            # Create file data
-            file_data = create_file_data(content_str)
+            try:
```
eyurtsev, line 665: I don't believe we need a try / except here, we can probably just decode base on file extension

this was a bug in the code so the encoding to utf-8 would've failed for a lot of binary data

libs/deepagents/deepagents/backends/store.py
```diff
@@ -618,9 +671,10 @@ def download_files(self, paths: list[str]) -> list[FileDownloadResponse]:
                 continue
 
             file_data = self._convert_store_item_to_file_data(item)
-            # Convert file data to bytes
             content_str = file_data_to_string(file_data)
-            content_bytes = content_str.encode("utf-8")
```
sydney-runkle, line 623: bug -- lack of support for non-text files

libs/deepagents/deepagents/backends/store.py
```diff
@@ -585,12 +634,16 @@ def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadRespons
         responses: list[FileUploadResponse] = []
 
         for path, content in files:
-            content_str = content.decode("utf-8")
```
sydney-runkle, line 588: bug -- lack of support for non text files

PR #1399, opened 2026-02-18
feat: expose tool_token_limit_before_evict in create_deep_agent()

AI-Guru:
## Summary

- Add `tool_token_limit_before_evict` parameter to `create_deep_agent()` and forward it to all three internal `FilesystemMiddleware` instances (main agent, general-purpose subagent, and user-defined subagents).
- Default remains `20000` for backward compatibility.

## Motivation

When agents use MCP tools that return large results (e.g., web search, web scraping), a single tool call can inject 40,000–130,000+ tokens into context. The `FilesystemMiddleware` already has excellent eviction logic that offloads large results to `/large_tool_results/` and replaces them with a preview + file reference. However, the threshold is currently hardcoded inside `create_deep_agent()` with no way to override it.

**Example:** An autonomous agent spawns researcher subagents that make three parallel `web_search` MCP calls. Each returns ~45k tokens, inflating subagent context from 5k to 138k tokens in a single step. Lowering the eviction threshold to 5,000 tokens offloads each result to disk immediately and keeps context lean, saving both latency and API cost.

## Changes

One file changed: `libs/deepagents/deepagents/graph.py`

1. New keyword argument `tool_token_limit_before_evict: int | None = 20000`
2. Forwarded to all three `FilesystemMiddleware(backend=backend)` call sites
3. Docstring updated

## Test plan

- [ ] Verify default behavior unchanged (no argument = 20,000 token threshold)
- [ ] Verify custom limit applies to main agent, GP subagent, and user subagents
- [ ] Verify `None` disables eviction

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -63,6 +63,7 @@ def create_deep_agent(  # noqa: C901, PLR0912  # Complex graph assembly logic wi
     store: BaseStore | None = None,
     backend: BackendProtocol | BackendFactory | None = None,
     interrupt_on: dict[str, bool | InterruptOnConfig] | None = None,
+    tool_token_limit_before_evict: int | None = 20000,
     debug: bool = False,
     name: str | None = None,
     cache: BaseCache | None = None,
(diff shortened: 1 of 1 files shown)


eyurtsev (changes requested):
Thanks for the contribution -- we won't expose this configuration top level. We keep on updating the strategy as we're benchmarking code, so this parameter may not even be available in the future.

Did you find that this parameter is too small or too big for your use case?

AI-Guru:
@eyurtsev thanks for looking into this! I am running local LLMs - context length max 300K - and I found the value too big. I monkey-patched it in my project.

eyurtsev:
@AI-Guru which local llms? we rely on model profiles right now to associate context length with LLMs. Would love to knwo if we can fix the model profiles for you

eyurtsev:
cc @ccurme

mdrxy:
Closing; please feel free to convert this discussion to a new issue @AI-Guru

PR #1402, opened 2026-02-18
fix(sdk): add default chunked upload/download to `BaseSandbox`

AI-Guru:
## Summary

- Add concrete `upload_files()` and `download_files()` implementations to `BaseSandbox`, replacing the current `@abstractmethod` stubs
- Small files (<64KB) use a single `execute()` call (same as before)
- Large files are split into 64KB base64 chunks to avoid hitting the Linux kernel `ARG_MAX` limit (~128KB per argument)
- Downloads also chunk to avoid `execute()` output truncation
- Methods are non-abstract so backends with native file transfer (SSH/SFTP, Daytona REST) can still override them

## Problem

When using sandbox backends that route file transfers through `execute()` (Docker with tmpfs, microsandbox), uploading binary files larger than ~100KB fails with:

```
exec /bin/bash: argument list too long
```

This happens because the base64-encoded file content is embedded in the command string passed to `exec_run()`, which exceeds the kernel's `ARG_MAX` limit. Docker's `put_archive` API cannot be used as a workaround because tmpfs mounts are invisible to it.

The error is particularly problematic for autonomous workflows that generate binary artifacts (PDFs, images) inside sandboxed containers.

## Solution

`BaseSandbox` already provides default implementations of `read()`, `write()`, `edit()`, `ls_info()`, `glob_info()`, and `grep_raw()` via `execute()`. This PR extends that pattern to `upload_files()` and `download_files()`:

- **Upload**: Files over 64KB are base64-encoded, split into chunks, each chunk appended to a temp file inside the sandbox via separate `execute()` calls, then the assembled base64 is decoded to the final path
- **Download**: Files over 64KB are read in binary chunks inside the sandbox, each chunk base64-encoded and returned via separate `execute()` calls, then reassembled on the host

All operations go through `execute()`, respecting the full middleware chain.

## Test plan

- [x] Unit tests: 6 new tests covering small file upload, large file chunking, upload error handling, small file download, large file chunked download, and missing file error
- [x] All 15 sandbox backend tests pass
- [x] Verified end-to-end with Docker backend: 301KB PDF upload/download roundtrip succeeds
- [x] Verified 1MB random binary roundtrip with byte-level fidelity

### DIFF
--- a/libs/deepagents/deepagents/backends/sandbox.py
+++ b/libs/deepagents/deepagents/backends/sandbox.py
@@ -10,6 +10,7 @@ It also defines the BaseSandbox implementation used by the CLI sandboxes.
 from __future__ import annotations
 
 import base64
+import binascii
 import json
 import logging
 import shlex
@@ -437,10 +438,7 @@ except PermissionError:
         for path, content in files:
             b64 = base64.b64encode(content).decode("ascii")
 
-            if len(b64) <= self._UPLOAD_CHUNK_BYTES:
-                result = self._upload_single(path, b64)
-            else:
-                result = self._upload_chunked(path, b64)
+            result = self._upload_single(path, b64) if len(b64) <= self._UPLOAD_CHUNK_BYTES else self._upload_chunked(path, b64)
 
             if result.exit_code == 0:
                 responses.append(FileUploadResponse(path=path))
@@ -463,7 +461,7 @@ except PermissionError:
             ExecuteResponse from the sandbox.
         """
         cmd = (
-            "python3 -c \"\n"
+            'python3 -c "\n'
             "import base64, pathlib, sys\n"
             "b64 = sys.stdin.read().strip()\n"
             "p = pathlib.Path('" + file_path + "')\n"
@@ -493,10 +491,10 @@ except PermissionError:
 
         # Ensure parent directory exists.
         result = self.execute(
-            "python3 -c \""
+            'python3 -c "'
             "import pathlib; "
             "pathlib.Path('" + file_path + "').parent.mkdir(parents=True, exist_ok=True)"
-            "\""
+            '"'
         )
         if result.exit_code != 0:
             return result
@@ -506,7 +504,7 @@ except PermissionError:
         for i in range(0, len(b64), chunk_size):
             chunk = b64[i : i + chunk_size]
             append_cmd = (
-                "python3 -c \"\n"
+                'python3 -c "\n'
                 "import sys\n"
                 "chunk = sys.stdin.read().strip()\n"
                 "with open('" + tmp_b64 + "', 'a') as f:\n"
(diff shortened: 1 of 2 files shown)


libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -397,18 +401,227 @@ def glob_info(self, pattern: str, path: str = "/") -> list[FileInfo]:
...
+
+            if result.exit_code == 0:
+                responses.append(FileUploadResponse(path=path))
+            else:
+                log.warning("Failed to upload %s: %s", path, result.output)
+                responses.append(FileUploadResponse(path=path, error="permission_denied"))
+
+        return responses
+
+    def _upload_single(self, file_path: str, b64: str) -> ExecuteResponse:
```
eyurtsev, line 637: If you're interested in working on this feature, could you try to match the style of the other provided implementations (e.,g., write_file etc)?

offload as much as you can into the template which attempts to make it easier to see the boundary between python and shell code (makes it easier to verify that the implementation does not security issues)

Make sure that everything that needs to be escaped is escaped properly so there's no additional attack surface

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -397,18 +401,227 @@ def glob_info(self, pattern: str, path: str = "/") -> list[FileInfo]:
...
+        """Upload a small file in one execute() call.
+
+        Args:
+            file_path: Absolute path inside the sandbox.
+            b64: Base64-encoded file content (must fit within ARG_MAX).
```
eyurtsev: is ARG_MAX relevant if we're feeding through stdin?

libs/deepagents/deepagents/backends/sandbox.py
AI-Guru: Done! Refactored all upload/download methods to use module-level `_COMMAND_TEMPLATE` constants matching the existing `_WRITE_COMMAND_TEMPLATE` / `_EDIT_COMMAND_TEMPLATE` pattern.

Seven new templates added: `_UPLOAD_COMMAND_TEMPLATE`, `_UPLOAD_CHUNK_COMMAND_TEMPLATE`, `_UPLOAD_DECODE_COMMAND_TEMPLATE`, `_REMOVE_COMMAND_TEMPLATE`, `_DOWNLOAD_SIZE_COMMAND_TEMPLATE`, `_DOWNLOAD_COMMAND_TEMPLATE`, `_DOWNLOAD_CHUNK_COMMAND_TEMPLATE`.

All file paths are now base64-encoded before interpolation (safe charset `[A-Za-z0-9+/=]`), which also eliminates the shell injection risk from the previous string concatenation approach. Template format tests added for each.

libs/deepagents/deepagents/backends/sandbox.py
AI-Guru: Good question! ARG_MAX itself (typically ~2MB) is the total size limit for all arguments + environment passed to `execve()`. But the more relevant constraint here is `MAX_ARG_STRLEN` (typically `PAGE_SIZE * 32` = 128KB on Linux), which limits any *single* argument string.

When we use `bash -c 'script<<heredoc'`, the entire string — including the heredoc content — is passed as a single argument to `execve()`. So heredocs via `bash -c` don't bypass the limit the way a standalone heredoc in an interactive shell would (where bash itself reads from stdin).

With 64KB raw chunks → ~87KB base64, we stay safely under the 128KB `MAX_ARG_STRLEN` per-argument limit.

I've updated the comments in the code to reference `MAX_ARG_STRLEN` instead of `ARG_MAX` to be more precise about which kernel limit actually applies.

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #1413, opened 2026-02-19
fix(cli): escape `Rich` markup in shell command display

mdrxy:
Fixes a rendering crash when shell commands contain literal `Rich` tag-like text (for example `[/dim]`), which was being interpreted as markup instead of plain text.

**Before**  
`ApprovalMenu._get_command_display()` built a Rich-marked string by interpolating raw command text:

```python
f"[bold #f59e0b]{command}[/bold #f59e0b]"
```

If `command` contained `[` or `]` tag sequences, Rich parsed them.  

Example input:

```text
echo [/dim] [literal]
```

could raise:

```text
... closing tag '[/dim]' does not match any open tag
```

**After**  
`command` (and truncated command) is escaped before insertion into markup:

```python
f"[bold #f59e0b]{escape_markup(command)}[/bold #f59e0b]"
```

Result: Rich renders the command text literally, including bracket sequences, with no parse error.

### DIFF
--- a/libs/cli/deepagents_cli/app.py
+++ b/libs/cli/deepagents_cli/app.py
@@ -1311,6 +1311,10 @@ class DeepAgentsApp(App):
                 image_tracker=self._image_tracker,
             )
         except Exception as e:  # noqa: BLE001  # Resilient tool rendering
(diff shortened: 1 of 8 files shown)


eyurtsev (approved):
thank you!

PR #1420, opened 2026-02-19
feat(cli,sdk): compaction hook

mdrxy:
Add two paths for conversation compaction — a `compact_conversation` tool that the agent can invoke proactively, and a `/compact` slash command for manual use.

Both operate using `SummarizationMiddleware` internals, writing a summarization event into checkpointed state so older messages are virtually hidden without deletion.

Full history is written to backend storage before compaction (when possible).


libs/cli/deepagents_cli/app.py
```diff
@@ -1254,6 +1258,265 @@ def _scroll_after_command() -> None:
...
+
+        messages = state.values.get("messages", [])
+
+        try:
+            await self._set_spinner("Compacting")
+
+            from deepagents.middleware.summarization import (
+                _compute_summarization_defaults,  # noqa: PLC2701
+            )
+            from langchain.agents.middleware.summarization import (
```
ccurme: Why are we using this one instead of deepagents

libs/cli/deepagents_cli/compact_tool.py
```diff
@@ -0,0 +1,573 @@
...
+        logger.warning(
+            "Exception offloading compact history to %s", path, exc_info=True
+        )
+        return None
+
+    logger.debug("Offloaded %d messages to %s", len(filtered), path)
+    return path
+
+
+def _generate_summary(
```
ccurme: OK to duplicate code for now but would be good to refactor into the existing middleware at some point.

Middleware can just count tokens and invoke the tool.

libs/cli/deepagents_cli/compact_tool.py
```diff
@@ -0,0 +1,573 @@
...
+
+    if cutoff == 0:
+        return Command(
+            update={
+                "messages": [
+                    ToolMessage(
+                        content=(
+                            "Nothing to compact yet"
+                            " \u2014 conversation is within"
+                            " the token budget."
```
ccurme: What is the point of this check?

libs/cli/deepagents_cli/compact_tool.py
```diff
@@ -0,0 +1,573 @@
...
+    Args:
+        response: The model response object.
+
+    Returns:
+        Extracted summary text.
+
+    Raises:
+        RuntimeError: If the response contains no usable text.
+    """
+    content = response.content
```
ccurme: `response.text`?

ccurme (commented):
Do we provide any feedback to the model as to the percent of token budget used?

Might be worth an experiment, even without a working tool implementation, to see how / when it elects to compact.

mdrxy:
@ccurme 

> Do we provide any feedback to the model as to the percent of token budget used?

Not currently, but can do in a follow-up?

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -419,39 +425,286 @@ def _build_new_messages_with_path(self, summary: str, file_path: str | None) ->
...
+                            f"{exc}). The conversation has not been compacted "
+                            "— no messages were summarized or removed."
+                        ),
+                        tool_call_id=tool_call_id,
+                    )
+                ],
+            }
+        )
+
+    def _run_compact(self, runtime: ToolRuntime) -> Command:
```
ccurme: Why don't we truncate args as we do in wrap_model_call?

ccurme (commented):
We need end-to-end tests and at least one eval:
1. tests/unit_tests/test_end_to_end.py
2. evals/test_summarization.py

Would be good to see traces as well.

ccurme (approved):
Excluded SummarizationToolMiddleware from default middleware stack until 0.5 release (needs testing in AB frontend).

PR #1449, opened 2026-02-20
Sr/additional read file moves

sydney-runkle:
* move to `ReadResult`
* move line formatting into read file tool
* move away from "download files" in `read` and use b64 string support in `read`


ccurme:
Superseded by https://github.com/langchain-ai/deepagents/pull/1836

PR #1461, opened 2026-02-20
fix(sdk): guard timeout forwarding for older backend packages

mdrxy:
Backwards-compatible guard for the per-command `timeout` parameter added in #1154. Partner backend packages (`langchain-daytona`, `langchain-modal`, `langchain-runloop`) previously had no version constraint on the SDK, so upgrading the SDK without upgrading a partner package would `TypeError` on every `execute()` call, as the SDK unconditionally passed `timeout=` to backends whose signatures didn't accept it.

### DIFF
--- a/libs/deepagents/deepagents/backends/filesystem.py
+++ b/libs/deepagents/deepagents/backends/filesystem.py
@@ -5,6 +5,7 @@ import logging
 import os
 import re
 import subprocess
+import warnings
 from datetime import datetime
 from pathlib import Path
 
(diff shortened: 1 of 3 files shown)


libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -479,7 +484,58 @@ async def aexecute(
...
+            "Could not inspect signature of %s.execute; assuming timeout is not supported. This may indicate a backend packaging issue.",
+            cls.__qualname__,
+            exc_info=True,
+        )
+        return False
+    else:
+        return "timeout" in sig.parameters
+
+
+def execute_accepts_timeout(backend: SandboxBackendProtocol) -> bool:
```
eyurtsev: this looks weird

PR #1464, opened 2026-02-20
fix(sdk): correct imports and mock inheritance in filesystem middleware tests

mdrxy:


libs/deepagents/tests/integration_tests/test_filesystem_middleware.py
```diff
@@ -1,15 +1,16 @@
 import uuid
 
 import pytest
-from langchain.agents import AgentMiddleware, create_agent
+from langchain.agents import create_agent
+from langchain.agents.middleware.types import AgentMiddleware
```
mdrxy:
```suggestion
from langchain.agents.middleware import AgentMiddleware
```

PR #1473, opened 2026-02-20
feat(sdk): add LangSmith backend implementation

rahul-langchain:
We have this in the CLI but adding this to the actual package. I tested to make sure the import works by running `uv run python -c "from deepagents.backends import LangSmithBackend, LangSmithProvider; print('OK')"`


mdrxy:
#1983

PR #1484, opened 2026-02-21
fix(sdk): sanitize paths in BaseSandbox ls_info and read templates

### LINKED ISSUE #1483: BaseSandbox ls_info() and read() interpolate paths without sanitization
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

_No response_

```python
from deepagents.backends.sandbox import _READ_COMMAND_TEMPLATE

# A malicious path breaks out of the Python string literal
malicious_path = "'; import os; os.system('echo INJECTED'); #"

cmd = _READ_COMMAND_TEMPLATE.format(
    file_path=malicious_path, offset=0, limit=100
)
print(cmd)
# Output includes:
#   file_path = ''; import os; os.system('echo INJECTED'); #'
# The injected code would execute inside the sandbox.

# Same issue exists in BaseSandbox.ls_info() at line 214:
#   path = '{path}'  <-- raw interpolation
```

```shell

```

### Description

`BaseSandbox.ls_info()` (line 214) and `_READ_COMMAND_TEMPLATE` (line 155) interpolate `path`/`file_path` directly into Python code strings using `'{path}'` / `'{file_path}'` without any sanitization.

A crafted path containing single quotes can break out of the Python string literal and execute arbitrary code inside the sandbox.

Other methods in the same file already handle this correctly:
- `glob_info()` uses base64 encoding (line 369-370)
- `write()` uses base64 + JSON + heredoc stdin (line 271-273)
- `edit()` uses base64 + JSON + heredoc stdin (line 297-298)
- `grep_raw()` uses `shlex.quote()` (line 330, 341)

The Harbor backend (`libs/harbor/deepagents_harbor/backend.py`) also uses `shlex.quote()` for all equivalent path parameters.

The fix is to base64-encode paths before interpolation (matching the existing pattern in `_GLOB_COMMAND_TEMPLATE`) for both `_READ_COMMAND_TEMPLATE` and
`ls_info()`.

While the blast radius is limited to sandboxed environments (not the user's local machine), it could allow privilege escalation within a sandbox via prompt injection.

I'd like to submit a PR for this if a maintainer approves.

### Environment / System Info

OS: Windows 10 (10.0.26100)
Python: 3.12
deepagents: 0.4.2
deepagents-cli: 0.0.21

AdemBoukhris457:
`BaseSandbox.ls_info()` and `_READ_COMMAND_TEMPLATE` interpolate `path`/`file_path` directly into Python code strings using raw `'{path}'` formatting. A crafted path containing single quotes can break out of the string literal and execute arbitrary code inside the sandbox.

This applies the same base64-encoding pattern already used by `_GLOB_COMMAND_TEMPLATE`, `_WRITE_COMMAND_TEMPLATE`, and `_EDIT_COMMAND_TEMPLATE` in the same file.

Fixes #1483

- The `_READ_COMMAND_TEMPLATE` now takes `file_path_b64` instead of `file_path`, and `BaseSandbox.read()` encodes it before calling `.format()`. This is the same pattern as `glob_info()`.
- The `ls_info()` inline command now encodes `path` the same way.
- Two new tests verify that malicious paths do not appear in the generated command strings.

## How was this verified?

- Ran `make format`, `make lint`, and `make test` from `libs/deepagents/`
- Added `test_sandbox_ls_info_path_is_sanitized` and `test_sandbox_read_path_is_sanitized` to confirm malicious paths are not interpolated raw
- Updated `test_read_command_template_format` to use the new base64 interface
- Verified `write()`, `edit()`, `glob_info()`, and `grep_raw()` already use safe patterns (base64 or `shlex.quote()`) and are unchanged

### DIFF
--- a/libs/deepagents/tests/unit_tests/backends/test_sandbox_backend.py
+++ b/libs/deepagents/tests/unit_tests/backends/test_sandbox_backend.py
@@ -175,35 +175,3 @@ def test_sandbox_grep_literal_search() -> None:
     # Verify the command uses grep -rHnF for literal search (combined flags)
     assert sandbox.last_command is not None
     assert "grep -rHnF" in sandbox.last_command
-
-
-class EmptyOutputSandbox(MockSandbox):
-    """MockSandbox that returns empty output, suitable for ls_info/read tests."""
-
-    def execute(self, command: str) -> ExecuteResponse:
-        self.last_command = command
-        return ExecuteResponse(output="", exit_code=0, truncated=False)
-
-
-def test_sandbox_ls_info_path_is_sanitized() -> None:
-    """Test that ls_info base64-encodes paths to prevent injection."""
-    sandbox = EmptyOutputSandbox()
-
-    malicious_path = "'; import os; os.system('echo INJECTED'); #"
-    sandbox.ls_info(malicious_path)
-
-    assert sandbox.last_command is not None
-    assert malicious_path not in sandbox.last_command
-    assert "base64" in sandbox.last_command
-
-
-def test_sandbox_read_path_is_sanitized() -> None:
-    """Test that read base64-encodes paths to prevent injection."""
-    sandbox = EmptyOutputSandbox()
-
-    malicious_path = "'; import os; os.system('echo INJECTED'); #"
-    sandbox.read(malicious_path)
-
-    assert sandbox.last_command is not None
-    assert malicious_path not in sandbox.last_command
-    assert "base64" in sandbox.last_command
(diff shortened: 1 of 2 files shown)


eyurtsev (approved):

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -151,8 +151,9 @@
 _READ_COMMAND_TEMPLATE = """python3 -c "
 import os
 import sys
+import base64
 
-file_path = '{file_path}'
+file_path = base64.b64decode('{file_path_b64}').decode('utf-8')
```
eyurtsev, line 156: We should probably do another pass here to unpack file_path, offset and limit from a single payload

PR #1486, opened 2026-02-21
fix(sdk): deduplicate middleware when user passes a built-in type

### LINKED ISSUE #1485: create_deep_agent duplicates middleware when user passes a built-in type via middleware=
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

_No response_

```python
from deepagents.middleware.summarization import SummarizationMiddleware
from deepagents.backends import StateBackend

custom_summarization = SummarizationMiddleware(
    model="gpt-4o-mini",
    backend=StateBackend,
    trigger=("fraction", 0.85),
    keep=("fraction", 0.10),
)

agent = create_deep_agent(middleware=[custom_summarization])
# The agent now has TWO SummarizationMiddleware instances
```

```shell

```

### Description

`create_deep_agent` hardcodes a `SummarizationMiddleware` in its internal middleware stack (graph.py L258-265), then blindly appends user-provided middleware via `deepagent_middleware.extend(middleware)` at L271.

If a user passes a custom `SummarizationMiddleware` through the `middleware=` parameter (as shown in the `SummarizationMiddleware` module docstring example at summarization.py L22), the agent ends up with two instances of the same middleware type, causing double-summarization.

The same issue applies to any built-in middleware type (FilesystemMiddleware, PatchToolCallsMiddleware, etc.) and to subagent middleware (L230).

Expected: user-provided middleware of a built-in type should replace the default instance, not duplicate it.

The SummarizationMiddleware module docstring (L8-23) explicitly shows `create_deep_agent(middleware=[middleware])` as the intended customization pattern, confirming this is meant to work.

### Environment / System Info

OS: Windows 10 (10.0.26100)
Python: 3.12
deepagents: 0.4.2
deepagents-cli: 0.0.21

AdemBoukhris457:
## Summary

- `create_deep_agent` hardcodes a default middleware stack (including `SummarizationMiddleware`, `FilesystemMiddleware`, `PatchToolCallsMiddleware`, etc.) and then blindly appends any user-provided `middleware=` entries, causing duplicate instances when the user passes a built-in type.
- The `SummarizationMiddleware` module docstring ([summarization.py L8-23](../blob/main/libs/deepagents/deepagents/middleware/summarization.py#L8-L23)) explicitly shows `create_deep_agent(middleware=[middleware])` as the intended customization pattern, confirming this should work without duplication.
- This fix filters out built-in middleware instances whose type matches a user-provided one before appending, for both the main agent stack and subagent stacks.

Fixes #1485

## Why this approach

The dedup is type-based: if the user passes a `SummarizationMiddleware`, the built-in `SummarizationMiddleware` is removed and the user's instance takes its place. This is the minimal change that:

1. Preserves existing behavior for users who don't pass `middleware=`
2. Lets users override any built-in middleware by passing their own instance of the same type
3. Still allows users to append entirely new middleware types as before

## Changes

- **`libs/deepagents/deepagents/graph.py`**: Added type-based dedup logic before extending the middleware list, in both the main agent stack (L276) and the subagent stack (L230).
- **`libs/deepagents/tests/unit_tests/test_end_to_end.py`**: Added two unit tests — one structural test that passes a custom `SummarizationMiddleware` to `create_deep_agent`, and one that directly validates the dedup logic produces no duplicate types.

- The dedup uses `type(m)` equality, not `isinstance`. This is intentional — it avoids accidentally removing unrelated middleware that happens to share a base class. Reviewers should confirm this matches the project's middleware inheritance patterns.
- The user's middleware is appended at the end of the stack (after dedup). If positional ordering matters for certain middleware combinations, this could be a concern.


eyurtsev:
Hi @AdemBoukhris457. Summarization is non configurable by design -- we're improving it based on benchmark and want it to work out of the box by default. If you're running into issues w/ summarization please let us know and we can fix it.

PR #1490, opened 2026-02-22
perf(sdk): batch memory source loading into a single download_files call

### LINKED ISSUE #1489: perf(sdk): batch memory source loading into a single download_files call
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Feature description

`MemoryMiddleware.before_agent` and `abefore_agent` load each memory source file one at a time in a loop (memory.py L324-327, L353-356). Each iteration calls `backend.download_files([single_path])` -- one network round-trip per source.

The `download_files` API already accepts a list of multiple paths and returns a list of responses. The middleware just doesn't use this batching capability.

With N memory sources and a remote backend (Docker, Daytona, Modal, Runloop), this means N sequential round-trips instead of 1. At ~100ms per round-trip, 5 sources wastes ~400ms on agent startup.

Replace the per-source loop:

    for path in self.sources:
        content = self._load_memory_from_backend_sync(backend, path)
        if content:
            contents[path] = content

With a single batched call:

    results = backend.download_files(list(self.sources))
    for path, response in zip(self.sources, results):
        if response.error is not None:
            if response.error != "file_not_found":
                raise ValueError(f"Failed to download {path}: {response.error}")
            continue
        if response.content is not None:
            contents[path] = response.content.decode("utf-8")

Same change for the async path using `await backend.adownload_files(...)`.

No API changes. No behavioral changes. Same error handling. Just fewer round-trips.

_No response_

AdemBoukhris457:
Closes #1489

## Summary

- `MemoryMiddleware.before_agent` and `abefore_agent` called `download_files([single_path])` in a loop -- one network round-trip per memory source.
- Replaced with a single `download_files(all_paths)` call that fetches all sources in one batch.
- Same error handling: `file_not_found` is silently skipped, other errors raise `ValueError`.

## Why this approach

The `download_files` API already supports batching -- it takes `list[str]` and returns `list[FileDownloadResponse]`. The middleware simply wasn't using this capability. This is a zero-risk change: same inputs, same outputs, same error behavior, fewer round-trips.

With N memory sources and a remote backend at ~100ms per round-trip:
- Before: N round-trips (e.g., 5 sources = ~500ms)
- After: 1 round-trip (e.g., 5 sources = ~100ms)

## Changes

- **`libs/deepagents/deepagents/middleware/memory.py`**: Replaced per-source loop with a single batched `download_files` / `adownload_files` call in both `before_agent` and `abefore_agent`.
- **`libs/deepagents/tests/unit_tests/middleware/test_memory_middleware.py`**: Added tests verifying batch behavior and partial failure handling.

- The error handling is inlined instead of delegating to `_load_memory_from_backend_sync` / `_load_memory_from_backend`. Behavior is identical but reviewers should verify the error paths match.
- The private helper methods are now unused. Keeping them for now to minimize the diff; happy to remove if preferred.


eyurtsev:
Thank you!

PR #1506, opened 2026-02-24
fix(sdk): validate tool_call_id before subagent invocation

### LINKED ISSUE #1505: SubAgentMiddleware validates tool_call_id after subagent invocation instead of before
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

_No response_

```python
# In libs/deepagents/deepagents/middleware/subagents.py, the task() and atask()
# functions check `runtime.tool_call_id` AFTER invoking the subagent:

def task(description, subagent_type, runtime):
    # ... validation ...
    result = subagent.invoke(subagent_state)  # <-- expensive LLM call happens here
    if not runtime.tool_call_id:              # <-- check happens AFTER the call
        raise ValueError("Tool call ID is required for subagent invocation")
    return _return_command_with_state_update(result, runtime.tool_call_id)

# If tool_call_id is missing, the entire subagent execution (which may involve
# multiple LLM calls, tool use, ...) is wasted before the error is raised.
```

```shell
ValueError: Tool call ID is required for subagent invocation

The error is raised correctly, but only after the subagent has already completed
its full execution — wasting compute, tokens, and time.
```

### Description

In `libs/deepagents/deepagents/middleware/subagents.py`, both `task()` (line 442) and `atask()` (line 460) validate that `runtime.tool_call_id` is present only after `subagent.invoke()` / `subagent.ainvoke()` has already completed.

Since subagent invocation can be expensive (multiple LLM calls, tool use, etc.), the validation should happen before the invoke call to fail fast and avoid wasting resources.

The fix is to move the `if not runtime.tool_call_id` check before the `subagent.invoke()` / `subagent.ainvoke()` call in both functions.

### Environment / System Info

Python: 3.11+
deepagents: 0.4.2
deepagents-cli: 0.4.1

AdemBoukhris457:
## Summary

- Moves `tool_call_id` validation in `task()` and `atask()` to run **before** `subagent.invoke()` / `subagent.ainvoke()`
- Previously, the check happened after the full subagent execution completed, wasting compute and tokens when `tool_call_id` was missing
- Both sync and async paths are fixed identically

Fixes #1505

- The `if not runtime.tool_call_id` check is now the first thing after the `subagent_type` validation, before `_validate_and_prepare_state()` and `subagent.invoke()`. This is a strict ordering improvement with no behavioral change for valid inputs.


eyurtsev:
Merging, though I'm unclear on whether there's a code path that can actually result in this issue right now?

PR #1513, opened 2026-02-26
feat(cli): show thread working directory in thread list views

hntrl:
## Summary

Adds a "Location" column to both thread list views (CLI table and Textual modal) showing the working directory where each thread was last used. This makes it easy to identify which project a thread belongs to when resuming work.

**Before**
Screenshot 2026-02-25 at 6 46 53 PM

**After**
Screenshot 2026-02-25 at 6 47 12 PM

### Metadata recording (`textual_adapter.py`, `non_interactive.py`)

Both interactive and non-interactive session configs now include `"cwd"` in the LangGraph checkpoint metadata, recording the current working directory at session start. This is backward compatible — old threads simply won't have the field and will show a blank location.

### Data layer (`sessions.py`)

- `ThreadInfo` TypedDict gains an optional `cwd` field
- New `format_path()` helper renders paths relative to `~` when under the home directory, otherwise shows the fully qualified path
- `list_threads()` SQL queries now extract `json_extract(metadata, '$.cwd')` — returns `NULL` gracefully for old checkpoints without the field

### Thread selector widget (`widgets/thread_selector.py`)

- Added a `Location` column to the header and option labels
- Locations longer than 24 characters are truncated with a `…` prefix
- Widened the modal container from 80→100 columns to accommodate the new column

### CLI thread list (`sessions.py` — `list_threads_command`)

- Added a "Location" column to the Rich table output

### Tests

- `TestFormatPath` — covers `~` substitution, paths outside home, `None`/empty input, and edge cases
- Updated `temp_db` fixtures to include `cwd` in test metadata
- Added assertion that `list_threads` returns `cwd` values correctly
- Updated `_build_stream_config` tests to expect `cwd` in metadata
- Added `cwd` to `MOCK_THREADS` in thread selector tests

> **Note:** This PR was authored with the assistance of an AI agent.

### DIFF
--- a/libs/cli/deepagents_cli/non_interactive.py
+++ b/libs/cli/deepagents_cli/non_interactive.py
@@ -25,6 +25,7 @@ import logging
 import sys
 from dataclasses import dataclass, field
 from datetime import UTC, datetime
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
 
 from langchain.agents.middleware.human_in_the_loop import ActionRequest, HITLRequest
@@ -588,6 +589,7 @@ async def run_non_interactive(
             "assistant_id": assistant_id,
             "agent_name": assistant_id,
             "updated_at": datetime.now(UTC).isoformat(),
+            "cwd": str(Path.cwd()),
         },
     }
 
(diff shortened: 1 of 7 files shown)


mdrxy:
Haven't been able to test yet but I suspect this will break on narrow terminal windows

PR #1517, opened 2026-02-26
feat(backends): add RedisBackend for persistent file storage

shein341:
## Summary

- Adds `RedisBackend`, a new `BackendProtocol` implementation that stores agent virtual-filesystem files in a Redis Hash (`{namespace}:files`)
- Exports `RedisBackend` from the `deepagents.backends` public API
- Adds `redis>=5.0` as an optional dependency (`pip install deepagents[redis]`)

## Motivation

The existing `StateBackend` stores files in LangGraph in-memory state — all file content is lost when the process restarts. `RedisBackend` provides:

- **Persistence** — files survive process restarts and container re-deploys
- **Multi-session sharing** — multiple agent threads can share the same namespace via a shared `namespace` key
- **TTL support** — automatic session cleanup via Redis `EXPIRE`

## Implementation

`RedisBackend` implements all `BackendProtocol` methods (sync + async variants):

| Method | Description |
|--------|-------------|
| `ls_info` | List direct children of a directory path |
| `read` | Read file with line-number formatting and offset/limit |
| `write` | Create new file (fails if file already exists) |
| `edit` | Replace string occurrences (keyword-only `replace_all` flag) |
| `grep_raw` | Regex search across all files in namespace |
| `glob_info` | Glob pattern matching against stored file paths |

Async variants (`als_info`, `aread`, `awrite`, `aedit`, `agrep_raw`, `aglob_info`) delegate to their sync counterparts. They can be swapped for `redis.asyncio` in a future PR.

## Usage

```python
from deepagents.backends import RedisBackend

backend = RedisBackend(
    host="localhost",
    port=6379,
    namespace="my-agent-session",
    ttl_seconds=3600,
)
agent = create_deep_agent(backend=backend)
```

## Tests

- **16 unit tests** (`tests/unit_tests/backends/test_redis_backend.py`) — fully mocked with `unittest.mock`, no Redis required, 92% line coverage
- **4 integration tests** (`tests/integration_tests/backends/test_redis_backend_integration.py`) — auto-skip if Redis is unavailable (`pytest.importorskip` + connection check)

## Test plan

- [ ] `pytest tests/unit_tests/backends/test_redis_backend.py -v` → 16 passed
- [ ] `ruff check libs/deepagents/deepagents/backends/redis_backend.py` → All checks passed
- [ ] Integration tests pass against a local Redis instance (plain `redis:7` is sufficient — Redis Stack not required)

### DIFF
--- a/examples/deep_research/uv.lock
+++ b/examples/deep_research/uv.lock
@@ -1734,15 +1734,15 @@ wheels = [
 
 [[package]]
 name = "langgraph-checkpoint"
-version = "3.0.1"
+version = "4.0.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "langchain-core" },
     { name = "ormsgpack" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/0f/07/2b1c042fa87d40cf2db5ca27dc4e8dd86f9a0436a10aa4361a8982718ae7/langgraph_checkpoint-3.0.1.tar.gz", hash = "sha256:59222f875f85186a22c494aedc65c4e985a3df27e696e5016ba0b98a5ed2cee0", size = 137785, upload-time = "2025-11-04T21:55:47.774Z" }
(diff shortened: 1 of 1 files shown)


eyurtsev:
Hi @shein341 thank you for the contribution. We're not yet ready to accept integrations backends integrations into the SDK. 

There's a community maintained package that you could try to contribute the integration to https://github.com/DiTo97/deepagents-backends

PR #1022, opened 2026-02-03
fix(sdk): improve subagent message extraction

sandeepyadav1478:
## Summary

Replaces fragile single-line message extraction in subagents with a robust multi-tier approach that handles edge cases including Anthropic's multi-part content format, ToolMessage fallback for ToolStrategy, and proper whitespace handling. This fix resolves issues where subagents would return empty responses when using certain LLM providers or structured output strategies.

Fixes #979
Fixes #720

## Changes

**Core Logic (`subagents.py`):**
- Replaced 1-line `.text.rstrip()` extraction with 40-line robust multi-tier approach
- **Tier 1 (Priority):** Extract from AIMessage with backward iteration
  - Support both `.content` and `.text` attributes (prefers `.content`)
  - Handle string content with proper `strip()` (both leading/trailing whitespace)
  - Handle Anthropic's list-based content format (filters text blocks, ignores images/tool_use)
- **Tier 2 (Fallback):** Extract from ToolMessage when no AIMessage has content
  - Essential for ToolStrategy structured output scenarios
- **Tier 3 (Default):** Return empty string to prevent confusing error messages

**Tests (`test_subagents.py`):**
- Added `TestMessageExtractionFromSubagents` class with 2 critical tests:
  1. `test_multipart_content_with_empty_blocks`: Validates Anthropic multi-part content handling
  2. `test_toolmessage_extraction_for_toolstrategy`: Validates ToolMessage fallback for structured output

## Testing

✅ All 11 tests pass (9 existing + 2 new)
- Ran `pytest tests/unit_tests/test_subagents.py -v`

## Breaking Changes

None. This is a backward-compatible enhancement to the existing message extraction logic.

## Twitter handle (optional)

Twitter: @sandeepyadav1478

PR #1518, opened 2026-02-26
fix(sdk): robust message extraction from subagents

### LINKED ISSUE #979: `SubAgentMiddleware` crashes with `AttributeError` when extracting messages from subagents
Related: #948

### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

_No response_

```python
from deepagents.middleware.subagents import SubAgentMiddleware

# Create an agent with subagents
agent = create_deep_agent(
    model="anthropic:claude-3-5-sonnet-20241022",
    middleware=[
        SubAgentMiddleware(
            default_model="anthropic:claude-3-5-sonnet-20241022",
            subagents=[],
            general_purpose_agent=True,
        )
    ]
)

# Invoke the agent with a task that uses subagents
# The subagent may end with a ToolMessage or AIMessage with only tool_calls
result = agent.invoke({
    "messages": [
        {"role": "user", "content": "Use the task tool to research something"}
    ]
})

# This will crash with AttributeError when the subagent's last message
# is not an AIMessage with simple text content
```

```shell
Traceback (most recent call last):
  File "...", line X, in <module>
    result = agent.invoke(...)
  File ".../deepagents/middleware/subagents.py", line 377, in _return_command_with_state_update
    message_text = result["messages"][-1].text.rstrip() if result["messages"][-1].text else ""
AttributeError: 'ToolMessage' object has no attribute 'text'
```

### Problem

When using subagents via the `SubAgentMiddleware`, the message extraction logic in the `_return_command_with_state_update` function (line 377 of `subagents.py`) fails in several scenarios:

1. **Last message is a ToolMessage**: Crashes with `AttributeError: 'ToolMessage' object has no attribute 'text'`
2. **AIMessage uses `.content` instead of `.text`**: Returns empty string
3. **Multi-part message content**: Cannot extract text from list-based content blocks (e.g., Anthropic's message format)

### Current Code

The current implementation assumes the last message is always an AIMessage with a `.text` attribute:

```python
message_text = result["messages"][-1].text.rstrip() if result["messages"][-1].text else ""
```

### Expected Behavior

The subagent should return the actual text content from the last AIMessage with content, regardless of:
- Whether it's the last message in the list
- Whether it uses `.text` or `.content` attributes
- Whether content is a string or list of content blocks

### Actual Behavior

- Crashes with AttributeError when last message is a ToolMessage
- Returns empty string when message uses `.content` or list-based format
- Fails to extract text from multi-part messages

### Proposed Fix

The fix should:
1. Iterate backward through messages to find the last AIMessage with actual content
2. Check both `.content` and `.text` attributes
3. Handle both string content and list-based content blocks (for Anthropic-style messages)
4. Provide informative feedback when no text response is found

**Location:** `libs/deepagents/deepagents/middleware/subagents.py`, line 377 in `_return_command_with_state_update`

I have a working implementation that handles all these cases if you'd like me to submit a PR.

### Environment / System Info

OS: macOS / Linux / Windows
Python: 3.11+
deepagents: latest (as of the issue date)
langchain: latest

saeedseyfi:
Replace fragile `messages[-1].text` extraction in `_return_command_with_state_update` with backward-iteration that handles ToolMessage, empty AIMessage, and multi-part content.

Fixes #979
Based on @sandeepyadav1478's work in #1022 (which has merge conflicts against current main).

Ran `make lint` and `make test` from `libs/deepagents/` — all 21 tests pass.

Verified by adding 3 integration tests:
1. Subagent ending with ToolMessage after todo tool call (the original bug)
2. ToolStrategy pattern where conversation ends with ToolMessage
3. Empty content graceful error handling


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -443,7 +460,10 @@ def task(
             raise ValueError(value_error_msg)
         subagent, subagent_state = _validate_and_prepare_state(subagent_type, description, runtime)
         result = subagent.invoke(subagent_state)
-        return _return_command_with_state_update(result, runtime.tool_call_id)
+        try:
```
eyurtsev, line 463: This doesn't look right -- subagent should only be responding with AIMessage and AIMessage always has `text` defined on it. 

How can this situation arise currently?

libs/deepagents/tests/unit_tests/test_subagents.py
```diff
@@ -1591,3 +1592,165 @@ def test_valid_deprecated_kwargs_accepted(self) -> None:
...
+            return {"messages": state["messages"] + [ai_msg, tool_call_msg, tool_result]}
+
+        parent_agent = create_deep_agent(
+            model=parent_chat_model,
+            checkpointer=InMemorySaver(),
+            subagents=[
+                CompiledSubAgent(
+                    name="worker",
+                    description="Worker subagent.",
+                    runnable=RunnableLambda(subagent_ending_with_tool_message),
```
eyurtsev, line 1645: Are you using RunnableLambdas in your own system?

khteh: This looks strange to me. I only use `StateGraph.compile()` and/or `langchain.agents.create_agent()` in my own system.

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

Issue #444, opened 2025-11-21
Some models have a max_token that does not reach 170,000. Is it possible to enable the setting for max_tokens_before_summary?


lzcangel:
Some models have a max_token that does not reach 170,000. Is it possible to enable the setting for max_tokens_before_summary?

<img width="706" height="574" alt="Image" src="https://github.com/user-attachments/assets/1542c389-aed2-4df1-883d-5a7558d5b3a9" />

julienlesbegueriesperso:
I created a 
class CustomSummarizationMiddleware(SummarizationMiddleware):
    """Custom summarization middleware with 4000 token threshold."""
    pass

then used in create_deep_agent
CustomSummarizationMiddleware(
                    model=GPT_OSS_OLLAMA,
                    max_tokens_before_summary=4000,
                ),
            ],


hope it helps

eyurtsev:
We'll start configuring this off the context window of the model soon (which will be available with langchain 1.1 via model profiles)

sydney-runkle:
Closing this for now, we base this off proportion of available context window when the model profile is available!

Said profile should be available for all major models.

PR #1520, opened 2026-02-26
feat(cli): add model profile max input tokens

arthur-drozdov:
## Description

This PR allows users to set a `max_input_tokens` property inside their `~/.deepagents/config.toml` that acts as an explicit context limit for models (like OpenAI-compatible ones) that don't auto-report their profile sizes. 

The system dynamically instantiates string-declared models via `langchain.chat_models.init_chat_model`, allowing the context limit to be explicitly bound to the model profile upon initialization. This properly enables fraction-based triggering inside the internal CLI `SummarizationMiddleware` everywhere the deep agent operates, instead of defaulting to 170k.

Related issue: https://github.com/langchain-ai/deepagents/issues/444

## Changes
- Updated `create_cli_agent` to use `init_chat_model` for string models, passing the explicit `profile` kwarg if `model_context_limit` is set.
- Fixed documentation comments to clarify summarization applicability.
- Unit tests updated to verify correct `init_chat_model` patching and profile injection logic.


mdrxy:
Covered in #1605, #1603 - thanks for the initiative!

PR #1539, opened 2026-02-27
feat(sdk): swarm tool

eyurtsev:
1. **Moves `_build_swarm_tool` from a standalone function to a method on `SubAgentMiddleware`** — it now has access to `self._backend` directly instead of receiving it as a parameter.

2. **Uses the backend for file I/O instead of raw `os`/`open`** — the old code used `open()` to read the config file and write results. The new code uses `resolved_backend.adownload_files()` and `resolved_backend.aupload_files()`, making it work with non-local backends (sandboxes, etc.).

3. **Adds structured parsing with validation** — extracts `SwarmTaskSpec` (TypedDict) and `ParsedSwarmConfig`, and validates the JSON config against them using Pydantic's `TypeAdapter`. Invalid configs now return a `ToolMessage` with `status="error"` instead of crashing.

4. **Returns `ToolMessage` instead of plain `str`** — `aswarm` now returns a `ToolMessage` directly, giving control over `status` and `tool_call_id`.

5. **Handles `GraphInterrupt` from subagents** — new logic collects `Interrupt` objects from subagent exceptions and re-raises them, enabling human-in-the-loop to work through swarm tasks.

6. **Removes the sync `swarm()` wrapper** — the old code had both sync (via `ThreadPoolExecutor`) and async versions. Now it's async-only via `coroutine=aswarm`.

7. **Defaults `enable_swarm=False`** in `create_deep_agent` (was `True`).

8. **Adds docstring for `enable_swarm`** in `create_deep_agent` and the CLI's `create_cli_agent`.

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -2,12 +2,11 @@
 
 import asyncio
 import json
+import os
 import warnings
-from langgraph.types import Interrupt
-
 from collections.abc import Awaitable, Callable, Sequence
+from pathlib import Path
 from typing import Annotated, Any, NotRequired, TypedDict, Unpack, cast
-from langgraph.errors import GraphInterrupt
 
 from langchain.agents import create_agent
 from langchain.agents.middleware import HumanInTheLoopMiddleware, InterruptOnConfig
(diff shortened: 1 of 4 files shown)


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -808,10 +753,139 @@ def __init__(
...
+            output_dir_clean = output_dir.rstrip("/")
+
+            outputs: list[tuple[str, bytes]] = []
+            summaries: list[str] = []
+            for item in results:
+                if isinstance(item, Exception):
+                    summaries.append(f"Error: {item}")
+                    continue
+                task_id, result_text = item
+                output_path = f"{output_dir_clean}/{task_id}.txt"
```
eyurtsev, line 892: be careful here w/ potential path escapes etc (i.e., let's sanitisze task_id)

ccurme: believe this was addressed.

PR #1542, opened 2026-02-28
chore(sdk): remove duplicate defaultdict import in composite.py

AdemBoukhris457:
## Summary
- `from collections import defaultdict` is imported at module level (line 21) and again inside a method (line 528). The inner import is redundant.
- Removed the duplicate.


eyurtsev:
looks like an empty commit

PR #1544, opened 2026-02-28
chore(sdk): precompile frontmatter regex pattern in skills.py

AdemBoukhris457:
## Summary
- `_parse_skill_metadata()` created a raw regex string and called `re.match()` on every invocation. Moved the pattern to a module-level precompiled constant `_FRONTMATTER_PATTERN` so it's compiled once.


libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -130,6 +130,8 @@
 MAX_SKILL_DESCRIPTION_LENGTH = 1024
 MAX_SKILL_COMPATIBILITY_LENGTH = 500
 
+_FRONTMATTER_PATTERN = re.compile(r"^---\s*\n(.*?)\n---\s*\n", re.DOTALL)
```
eyurtsev, line 133: doesn't python  maintain an LRU cache for regexps by default?

mdrxy:
practically negligible unless this path is called at very large scale

Issue #1547, opened 2026-02-28
Extract shared sync/async logic in middleware to reduce duplication


AdemBoukhris457:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Reproduction Steps / Example Code (Python)

```python
# Example of the duplication pattern (from memory.py):

# Sync version
def before_agent(self, state, runtime, config):
    backend = self._get_backend(runtime)
    contents = []
    for source in self.memory_sources:
        content = self._load_memory_from_backend_sync(backend, source.path)
        if content:
            contents.append(content)
    return state

# Async version — identical logic, just `await` added
async def abefore_agent(self, state, runtime, config):
    backend = self._get_backend(runtime)
    contents = []
    for source in self.memory_sources:
        content = await self._load_memory_from_backend(backend, source.path)
        if content:
            contents.append(content)
    return state
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

Every middleware file has near-identical sync/async method pairs. This is the single biggest code quality issue in the SDK middleware layer.

Affected files:
- `memory.py`: `_load_memory_from_backend` / `_load_memory_from_backend_sync`, `before_agent` / `abefore_agent`, `wrap_model_call` / `awrap_model_call`
- `subagents.py`: `task` / `atask`, `wrap_model_call` / `awrap_model_call`
- `skills.py`: `_list_skills` / `_alist_skills`, `before_agent` / `abefore_agent`, `wrap_model_call` / `awrap_model_call`
- `summarization.py`: `wrap_model_call` / `awrap_model_call`, `_offload_to_backend` / `_aoffload_to_backend`
- `filesystem.py`: 7 tool pairs + `wrap_model_call` + `_process_large_message` + `_intercept_large_tool_result`

Total: ~800-900 lines of near-identical duplicated logic.

**Proposed fix:** Extract shared logic into sync helper methods that both sync and async versions call. The async-specific parts (backend calls) are handled after the shared preparation logic returns.

This will be done in 3 phases/PRs:
1. Simple pairs in `memory.py`, `subagents.py`, `skills.py`
2. Backend operation pairs in `skills.py`, `memory.py`, `summarization.py`
[...]

PR #1548, opened 2026-02-28
refactor(sdk): extract shared sync/async helpers in memory and skills

AdemBoukhris457:
## Summary

- Extract `MemoryMiddleware._process_download_response()` to deduplicate response validation shared by `_load_memory_from_backend` and `_load_memory_from_backend_sync`
- Extract `_build_skill_md_paths()` and `_parse_skill_responses()` to deduplicate directory scanning and response parsing shared by `_list_skills` and `_alist_skills`
- Simplify `_list_skills` / `_alist_skills` from ~40 lines each to ~8 lines each
- Add 16 unit tests covering all new helpers

Intentionally skipped extracting `before_agent`/`abefore_agent` and `wrap_model_call`/`awrap_model_call` pairs — their loop bodies are too small (3-5 lines) to justify callback indirection.

`subagents.py` was also left unchanged since its duplication is already minimized by existing helpers.

Part 1 of #1547

### DIFF
--- a/libs/deepagents/deepagents/middleware/memory.py
+++ b/libs/deepagents/deepagents/middleware/memory.py
@@ -302,7 +302,7 @@ class MemoryMiddleware(AgentMiddleware[MemoryState, ContextT, ResponseT]):
         results = backend.download_files([path])
         return self._process_download_response(results, path)
 
-    def before_agent(self, state: MemoryState, runtime: Runtime, config: RunnableConfig) -> MemoryStateUpdate | None:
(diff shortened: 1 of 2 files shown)


eyurtsev:
@AdemBoukhris457 thanks for the contribution! I would suggest avoiding making no-op refactors into the repo. We mostly want to focus on resolving bugs / issues.

This refactor actually adds more code than it removes -- even without reviewing doesn't look like something we'd want to change.

Generally, reviewing code takes time and we'd rather focus right now on bugs or new features rather than internal refactors

PR #1552, opened 2026-03-01
fix(sdk,cli): accept YAML list format for allowed-tools in skill frontmatter

jkennedyvz:
## Summary
- Skills authored with YAML list syntax for `allowed-tools` (e.g., `- Bash\n- Read`) were silently ignored with a warning
- Now supports both string format (`allowed-tools: Bash Read Write`) and YAML list format (`allowed-tools:\n  - Bash\n  - Read`)
- Non-string list items are coerced via `str()`, `None`/whitespace-only entries are dropped

## Test plan
- [x] `test_parse_skill_metadata_allowed_tools_yaml_list_accepted` — verifies list format produces correct output
- [x] `test_parse_skill_metadata_allowed_tools_yaml_list_coerces_non_strings` — verifies mixed types are coerced and empty entries dropped
- [x] Existing string-format test (`test_parse_skill_metadata_allowed_tools_multiple_spaces`) still passes

### DIFF
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -322,6 +322,9 @@ def _parse_skill_metadata(  # noqa: C901
             for t in raw_tools.split()
             if t.strip(",")
         ]
+    elif isinstance(raw_tools, list):
+        # Support YAML list format (e.g., `- Bash\n- Read`)
+        allowed_tools = [str(t).strip() for t in raw_tools if t is not None and str(t).strip()]
     else:
         if raw_tools is not None:
             logger.warning(
--- a/libs/deepagents/tests/unit_tests/middleware/test_skills_middleware.py
+++ b/libs/deepagents/tests/unit_tests/middleware/test_skills_middleware.py
@@ -437,7 +437,7 @@ def test_validate_metadata_valid_dict_passthrough() -> None:
     assert result == {"author": "acme"}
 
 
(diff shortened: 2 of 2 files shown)


mdrxy:
@eyurtsev has the most up to date info on this, deferring to him

eyurtsev:
@jkennedyvz It's unclear from the agents spec if a yaml formatted list is meant to be supported. Could you confirm that that's actually the intention of the spec? I had a suspicion that while it's presented as yaml, the goal is to make is a subset of the spec so it's easy to implement the parsers

jkennedyvz:
@eyurtsev I see this YAML list pattern ITW e.g.

https://raw.githubusercontent.com/trailofbits/skills/refs/heads/main/plugins/static-analysis/skills/semgrep/SKILL.md

Did not notice the error until -n spat out some warnings.

mdrxy:
Per the [Agent Skills spec](https://agentskills.io/specification#allowed-tools-field), `allowed-tools` is defined as a space-delimited string:

> A space-delimited list of tools that are pre-approved to run

```yaml
allowed-tools: Bash(git:*) Bash(jq:*) Read
```

We should fix the crash (not erroring on a list value), but I don't think we should silently accept YAML list syntax — that diverges from the spec. The skill in the original issue is authored incorrectly; accepting it here would paper over that.

We'll warn when `allowed-tools` is a list (as we do today), and not crash. The skill author should fix their frontmatter to match the spec.

PR #1553, opened 2026-03-01
feat(cli): add hook dispatch for external tool integration

jkennedyvz:
(description edited by @mdrxy)

Add a lightweight hook system that lets external tools subscribe to CLI lifecycle events.

Users configure hooks in `~/.deepagents/hooks.json` with optional event filters — matching commands receive a JSON payload on stdin and run in a background thread so the event loop is never blocked.

### Core (`hooks.py`)
- `_load_hooks()` — lazily loads and caches hook definitions from `~/.deepagents/hooks.json`; validates config is a JSON object with a `hooks` list; rejects non-dict roots, non-list `hooks` values, and non-list `command` entries with clear warning-level logs
- `_run_single_hook()` — runs a single hook via `subprocess.run` (not `Popen.communicate`) so child processes are automatically killed on timeout, preventing zombie/orphan process leaks
- `_dispatch_hook_sync()` — dispatches matching hooks concurrently via `ThreadPoolExecutor` when multiple hooks match; per-hook errors are isolated so one failure never blocks others
- `dispatch_hook()` — async public API; offloads subprocess work to a thread via `asyncio.to_thread`; top-level `try/except` ensures errors never propagate to callers
- `dispatch_hook_fire_and_forget()` — GC-safe wrapper that stores task references in a module-level set with `done_callback` cleanup; replaces bare `create_task` calls

### Event wiring
- **Interactive TUI** (`app.py`): `user.prompt` on chat submission, `context.compact` before summarization, `session.end` on exit (dispatched synchronously since the event loop is about to be torn down)
- **Non-interactive mode** (`non_interactive.py`): `session.start` before streaming, `input.required` on interrupt (via `dispatch_hook_fire_and_forget`), `task.complete` and `session.end` after the agent loop finishes
- **Textual streaming adapter** (`textual_adapter.py`): `session.start`, `input.required`, `task.complete`, `tool.error` with `tool_names` list, and `permission.request` with `tool_names` list before the approval dialog

### Error handling
- `FileNotFoundError` / `PermissionError` (persistent config errors) logged at `warning` level so users see feedback
- `TimeoutExpired` logged at `warning` level; child process auto-killed by `subprocess.run`
- All other exceptions logged at `debug` level
- `dispatch_hook` itself wraps everything in `try/except` to honor the "never propagate" contract

## Security — Evaluated and Dismissed

| Concern | Why Not a Vulnerability |
|---|---|
| Arbitrary command execution from `hooks.json` | By design. Same trust model as Git hooks, VS Code tasks, shell aliases. Attacker with write access to `~/.deepagents/` already owns the user account. |
| LLM-influenced `tool_name` in payload | Flows to stdin JSON only, never to command args. `json.dumps()` handles escaping. A receiving script would need to deliberately eval raw JSON values. |
| No schema validation on individual hook entries | Malformed entries produce caught exceptions, not security impact. `subprocess.run` with `shell=False` and a non-list command is now rejected at the validation layer. |


eyurtsev:
@jkennedyvz do we have a design doc to list the hooks we want?

eyurtsev:
* Majority of hooks can be implemented using a hook middleware
* Some hooks require custom logic similar to the one in this PR
* Would be good to make sure we have a full story here for the feature in terms of what we want to get

PR #1577, opened 2026-03-02
feat(cli): add nvidia api key support and default model

soumilinandi:
## Description

Adds NVIDIA API key support to Deep Agents CLI. Users can now launch deepagents with only `NVIDIA_API_KEY` set, and it will use the default NVIDIA model (`nvidia/nemotron-3-nano-30b-a3b`). Previously, users would get an error saying no credentials were configured, and missing package errors would show incorrect package names.

## Changes

- Added `nvidia_api_key` attribute and `has_nvidia` property to Settings class
- Added `NVIDIA_API_KEY` detection in `from_environment()`
- Set default NVIDIA model when only `NVIDIA_API_KEY` is configured
- Added `nvidia` to package_map for correct error messages
- Updated credential error message to include `NVIDIA_API_KEY`

## Verification

Tested by:
1. Setting only `NVIDIA_API_KEY` and verifying deepagents launches successfully
2. Confirming error messages show correct package name (`langchain-nvidia-ai-endpoints)` when package is missing
3. Verifying default model is used when no config.toml exists
4. Testing that `config.toml` model overrides work correctly


mdrxy:
@soumilinandi is this still a draft?

PR #1580, opened 2026-03-02
ci(sdk): skip some evals

eyurtsev:
Skipping some evals


eyurtsev:
cc @mdrxy fyi disabling these evals since they rely on resources I don't have access to. could we maybe introduce them as fixtures in the code?

eyurtsev:
or we could make the traces public if possible?

PR #1590, opened 2026-03-03
chore(sdk,cli): switch `claude-sonnet-4-5-20250929` -> `claude-sonnet-4-6`

mdrxy:
Technically breaking if someone wasn't setting `model` on `create_deep_agent()` and is expecting sonnet 4.5 behavior

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -71,11 +71,10 @@ def get_default_model() -> ChatAnthropic:
     """Get the default model for deep agents.
 
     Returns:
-        `ChatAnthropic` instance configured with Claude Sonnet 4.5.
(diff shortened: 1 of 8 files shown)


libs/deepagents/deepagents/graph.py
```diff
@@ -71,11 +71,10 @@ def get_default_model() -> ChatAnthropic:
     """Get the default model for deep agents.
 
     Returns:
-        `ChatAnthropic` instance configured with Claude Sonnet 4.5.
+        `ChatAnthropic` instance configured with Claude Sonnet 4.6.
     """
     return ChatAnthropic(
-        model_name="claude-sonnet-4-5-20250929",
-        max_tokens=20000,
```
mdrxy, line 78: i think setting `max_tokens` was unnecessary as it's inferred from `.profile`

PR #1594, opened 2026-03-03
chore(sdk): add json report to evals

eyurtsev:
This adds a json report for evals. Removes parallelism for now. we can add parallelization to evals later


libs/deepagents/Makefile
```diff
@@ -36,7 +36,7 @@ integration_test integration_tests:
 	uv run --group test pytest -n auto -vvv --timeout 30 $(TEST_FILE)
 
 evals: ## Run evals
-	LANGSMITH_TEST_SUITE=deepagents-evals uv run --group test pytest -n auto tests/evals
+	LANGSMITH_TEST_SUITE=deepagents-evals uv run --group test pytest tests/evals
```
eyurtsev, line 39: -n auto was removed intentionally since the report generation doesn't work (we can add support later now a big deal)

PR #1604, opened 2026-03-03
fix(infra): run evals across full model matrix

eyurtsev:
Removes the workflow_dispatch model dropdown and restores the evals workflow to always run across the full model matrix.

Created with Deep Agents CLI.


eyurtsev:
it seems gpt-5.2 can't figure this out. i'll add convenient parameterization later, for now we can run all models

PR #1612, opened 2026-03-03
fix(cli): apply profile overrides in `/compact`

mdrxy:
`/compact` was ignoring `--profile-override` and `config.toml` profile overrides when deciding whether to summarize. The command creates a fresh model via `create_model()` without forwarding the overrides, so the summarization middleware used the model's native `max_input_tokens` (e.g., 200K) instead of the user's override (e.g., 4096).

### DIFF
--- a/libs/cli/deepagents_cli/app.py
+++ b/libs/cli/deepagents_cli/app.py
@@ -1427,12 +1427,14 @@ class DeepAgentsApp(App):
                 )
                 return
 
-            # Apply profile override so summarization uses the same context
-            # limit shown by /tokens (from --profile-override or config.toml).
-            # create_model() above builds a fresh model without these overrides.
+            # create_model() applies config.toml overrides but not CLI
+            # --profile-override (the raw CLI dict isn't retained after
+            # startup). Patch settings.model_context_limit — which reflects
+            # both sources — into the fresh model
             ctx = settings.model_context_limit
             if ctx is not None:
-                # Intentionally over-defensively checking for profile attribute
+                # Guard against models that lack a profile dict
+                # (custom/non-standard providers)
                 profile = getattr(model, "profile", None)
                 native = (
                     profile.get("max_input_tokens")
--- a/libs/cli/deepagents_cli/app.py
+++ b/libs/cli/deepagents_cli/app.py
@@ -1427,6 +1427,27 @@ class DeepAgentsApp(App):
                 )
                 return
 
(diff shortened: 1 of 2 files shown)


libs/cli/deepagents_cli/app.py
```diff
@@ -1427,6 +1427,27 @@ async def _handle_compact(self) -> None:
                 )
                 return
 
+            # Apply profile override so summarization uses the same context
+            # limit shown by /tokens (from --profile-override or config.toml).
+            # create_model() above builds a fresh model without these overrides.
+            ctx = settings.model_context_limit
```
eyurtsev, line 1434: the devx for working with profiles isn't super convenient if users have to do all of the defensive typing

we can't do something like `model.profile.max_input_tokens`?

libs/cli/tests/unit_tests/test_compact.py
```diff
@@ -942,3 +942,169 @@ def test_format_fraction_limit_with_context(self) -> None:
...
+
+class TestCompactProfileOverride:
+    """Verify /compact respects profile overrides (--profile-override / config.toml).
+
+    When the user overrides `max_input_tokens` via a profile override, the
+    `/compact` command must use the overridden value — not the model's native
+    profile — when computing the retention budget and cutoff index.
+    """
+
+    @pytest.mark.asyncio
```
eyurtsev, line 955: generally don't need this line due to pytest auto mode

libs/cli/tests/unit_tests/test_compact.py
```diff
@@ -942,3 +942,169 @@ def test_format_fraction_limit_with_context(self) -> None:
...
+            def capture_defaults(model: MagicMock) -> dict[str, Any]:
+                captured_models.append(model)
+                return {"keep": ("fraction", 0.10)}
+
+            mock_mw = MagicMock()
+            mock_mw._determine_cutoff_index.return_value = 0
+            mock_mw._apply_event_to_messages.side_effect = lambda msgs, _ev: list(msgs)
+
+            with (
+                patch(_CREATE_MODEL_PATH, return_value=mock_result),
```
eyurtsev, line 979: there's a lot of MagicMock() and patch() -- if at all possible could be useful to pull into a single helper in the global namespace that is a custom context manager

patch code is a bit brittle since it breaks down whenever the patched object is refactored to a different path

i realize it's a bit less of a concern these days with llms... i've just spent a lot of my life debugging broken patch code :)

PR #1620, opened 2026-03-03
docs(sdk): middleware module docstring

mdrxy:


libs/deepagents/deepagents/middleware/__init__.py
```diff
@@ -1,4 +1,66 @@
...
+* Track state across turns
+* Be available to all SDK consumers (not just the CLI)
+
+Use a **plain tool** when:
+
+* The function is stateless and self-contained
+* No system-prompt or request modification is needed
+* The tool is specific to a single consumer (e.g. CLI-only)
+
+## Middleware in this package
```
eyurtsev: I don't know how much we want to adverise these -- they are technically private. I'm just worried that this will be surfaced in the API refs

PR #1627, opened 2026-03-03
test(sdk): add fixture for summarization tests

mdrxy:
Following #1580, un-skips tests


eyurtsev:
@mdrxy for the fixture, could we just inline a list of Messages without using any serde?

PR #1657, opened 2026-03-05
fix(sdk): improve path matching for `ls_info` method in root directory

lksr1201:
## Problem Description
In the deepagents/backends/store.py file, the ls_info method fails to list files directly in the root directory ("/") such as "test.md". This is because the current filtering logic requires file paths to start with the normalized path ("/"), but direct files like "test.md" don't meet this condition.

## Reproduction Steps
1. Create a file directly in the root directory, e.g., "test.md"
2. Call the ls_info("/") method
3. Observe the returned result, noting that "test.md" is not in the result list ## Expected Behavior
ls_info("/") should return all files and directories in the root directory, including files directly in the root directory (like "test.md").

## Actual Behavior
ls_info("/") only returns subdirectories in the root directory, not files directly in the root directory.

## Root Cause
In the ls_info method, the filtering logic at lines 314-316 is:

(Replace this entire block of text)

Read the full contributing guidelines: https://docs.langchain.com/oss/python/contributing/overview
If you paste a large clearly AI generated description here your PR may be IGNORED or CLOSED!

Thank you for contributing to Deep Agents! Follow these steps to have your pull request considered as ready for review.

1. PR title: Should follow the format: TYPE(SCOPE): DESCRIPTION

  - Examples:
    - fix(sdk): resolve flag parsing error
    - feat(cli): add multi-tenant support
    - test(acp): update API usage tests
  - Allowed TYPE and SCOPE values: https://github.com/langchain-ai/deepagents/blob/main/.github/workflows/pr_lint.yml#L15-L26

2. PR description:

  - Write 1-2 sentences summarizing the change.
  - If this PR addresses a specific issue, please include "Fixes #ISSUE_NUMBER" in the description to automatically close the issue when the PR is merged.
  - If there are any breaking changes, please clearly describe them.
  - If this PR depends on another PR being merged first, please include "Depends on #PR_NUMBER" in the description.

3. Run `make format`, `make lint` and `make test` from the root of the package(s) you've modified.

  - We will not consider a PR unless these three are passing in CI.

4. How did you verify your code works?

Additional guidelines:

  - We ask that if you use generative AI for your contribution, you include a disclaimer.
  - PRs should not touch more than one package unless absolutely necessary.
  - Do not update the `uv.lock` files or add dependencies to `pyproject.toml` files (even optional ones) unless you have explicit permission to do so by a maintainer.

## Social handles (optional)

Twitter: @
LinkedIn: https://linkedin.com/in/

### DIFF
--- a/libs/deepagents/deepagents/backends/store.py
+++ b/libs/deepagents/deepagents/backends/store.py
@@ -311,11 +311,17 @@ class StoreBackend(BackendProtocol):
 
         for item in items:
             # Check if file is in the specified directory or a subdirectory
-            if not str(item.key).startswith(normalized_path):
-                continue
-
-            # Get the relative path after the directory
-            relative = str(item.key)[len(normalized_path) :]
+            # Special case: if path is root ("/"), match files without leading slash
+            item_key_str = str(item.key)
+            if normalized_path == "/":
+                if "/" in item_key_str:
+                    continue
+                relative = item_key_str
+            else:
+                if not item_key_str.startswith(normalized_path):
+                    continue
+                # Get the relative path after the directory
+                relative = item_key_str[len(normalized_path) :]
 
             # If relative path contains '/', it's in a subdirectory
             if "/" in relative:


eyurtsev:
If easy would you be willing to add a unit test that shows the failure with store?

lksr1201:
[ls_unittest.py](https://github.com/user-attachments/files/25782735/ls_unittest.py)
# Unit Test for `ls_info` Root Directory Bug

I've created a comprehensive unit test that demonstrates the failure of the `ls_info` method when listing files directly in the root directory. Here's a brief introduction:

## Test Overview

This unit test verifies the bug in the `ls_info` method where files directly in the root directory (like "test.md") are not listed when calling `ls_info("/")`. The test:

1. **Creates a mock store** that returns two items:
   - A file "test.md" directly in the root directory
   - A file "subdir/file.txt" in a subdirectory

2. **Tests two scenarios**:
   - `test_ls_info_root_directory`: Verifies that `ls_info("/")` lists both the root file and subdirectory
   - `test_ls_info_subdirectory`: Verifies that `ls_info("/subdir")` lists files in subdirectories (as a control test)

3. **Includes debugging output** to show the flow of data and help diagnose the issue

## Expected Behavior

When running the test, you'll see that:
- The mock store correctly returns both items
- However, `ls_info("/")` returns an empty list instead of including "test.md"
- This confirms the bug where files directly in the root directory are filtered out

## Test Structure

The test uses a mock store and runtime to isolate the `ls_info` method behavior, making it easy for maintainers to reproduce and verify the issue without needing a full store setup.

The test code is self-contained and can be run directly to demonstrate the failure.

lksr1201:
@eyurtsev  👋 I've followed up on your request and added a comprehensive unit test to reproduce the issue.

The test file `ls_unittest.py` clearly demonstrates the failure of the `ls_info` method when listing files directly in the root directory. You can review the test to confirm the bug is properly reproduced.

Let me know if you need any adjustments to the test or have further feedback!

mdrxy:
Thanks for the contribution! I think this fix is targeting a non-issue and would actually break the existing root-listing behavior.

The convention in `StoreBackend` is that file paths are absolute (start with `/`). All callers of `write()` pass absolute paths, so item keys look like `/config.json`, `/test.md`, etc. — see the existing tests in `test_store_backend.py` (`test_store_backend_ls_nested_directories`, `test_store_backend_ls_trailing_slash`).

Tracing the proposed code for `ls("/")` against an existing root file `/config.json`:

```python
normalized_path = "/"
item_key_str = "/config.json"
if normalized_path == "/":
    if "/" in item_key_str:   # True — "/" is in "/config.json"
        continue              # file gets skipped
```

Every file in root contains `/` (the leading slash), so they'd all be filtered out under the new branch. The existing tests assert `/config.json` and `/file.txt` show up under `ls("/")` — those would now fail.

The current logic at the top of the loop already handles `/test.md` correctly because `"/test.md".startswith("/")` is True. Could you share a concrete reproducer (the exact `write()` call and the resulting `ls()` output)? My read is the only way to land in the broken state is to bypass `write()` and put a key without a leading slash directly into the store, which isn't a supported path.

PR #1659, opened 2026-03-05
fix(sdk): enforce path boundaries in `CompositeBackend` routing

### LINKED ISSUE #1654: [Bug] CompositeBackend route prefix matching fails when prefix lacks trailing slash
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

_No response_

```python
from deepagents.backends.composite import CompositeBackend

runtime = make_runtime()

# Route configured WITHOUT trailing slash
composite = CompositeBackend(
    default=StateBackend(runtime),
    routes={"/abcd": StoreBackend(runtime)}  # Note: no trailing "/"
)

# This path should NOT match the route "/abcd"
backend, stripped_key = composite._get_backend_and_key("/abcde/file.txt")

print(backend)       # Expected: default (StateBackend)
                     # Actual: StoreBackend (incorrect!)

print(stripped_key)  # Expected: "/abcde/file.txt"
                     # Actual: "/e/file.txt" (incorrect!)
```

```shell

```

## Bug Description

`CompositeBackend._get_backend_and_key()` uses `str.startswith()` for route prefix matching, which can cause incorrect routing when route prefixes are configured without a trailing slash.

## Steps to Reproduce

```python
from deepagents.backends.composite import CompositeBackend

runtime = make_runtime()

# Route configured WITHOUT trailing slash
composite = CompositeBackend(
    default=StateBackend(runtime),
    routes={"/abcd": StoreBackend(runtime)}  # Note: no trailing "/"
)

# This path should NOT match the route "/abcd"
backend, stripped_key = composite._get_backend_and_key("/abcde/file.txt")

print(backend)       # Expected: default (StateBackend)
                     # Actual: StoreBackend (incorrect!)

print(stripped_key)  # Expected: "/abcde/file.txt"
                     # Actual: "/e/file.txt" (incorrect!)
```

## Expected Behavior

- `/abcde/file.txt` should **NOT** match the route prefix `/abcd`
- The path should fall through to the default backend with the original path preserved

## Actual Behavior

- `/abcde/file.txt` incorrectly matches `/abcd` because `"/abcde/file.txt".startswith("/abcd")` returns `True`
- The stripped path becomes `/e/file.txt`, which is completely wrong
- Files are routed to the wrong backend

## Root Cause Analysis

The issue not only arises in the `_get_backend_and_key()` function, but also in other functions that employ similar processing logic to obtain backend and stripped_key::

```python
def _get_backend_and_key(self, key: str) -> tuple[BackendProtocol, str]:
    for prefix, backend in self.sorted_routes:
        if key.startswith(prefix):  # ← Doesn't enforce path boundary
            suffix = key[len(prefix):]
            stripped_key = f"/{suffix}" if suffix else "/"
            return backend, stripped_key
    return self.default, key
```

The `startswith()` check doesn't ensure that the match occurs at a path boundary (i.e., followed by `/` or end of string).

## Affected Scenarios

| Route Prefix | Input Path | `startswith` Result | Bug? |
|-------------|-----------|---------------------|------|
| `/abcd/` | `/abcde/file.txt` | `False` | ✅ Correct |
| `/abcd` | `/abcde/file.txt` | `True` | ❌ Bug |
| `/memories/` | `/memories-backup/file.txt` | `False` | ✅ Correct |
| `/memories` | `/memories-backup/file.txt` | `True` | ❌ Bug |

## Environment

- **Package version**: 0.4.5
- **File**: `deepagents/backends/composite.py`
- **Method**: `CompositeBackend._get_backend_and_key()`

## Additional Context

The docstring mentions that prefixes "should end with `/`", but this is only a recommendation and not enforced by the code:

> routes: Map of path prefixes to backends. Prefixes must start with "/" and **should** end with "/" (e.g., "/memories/").

This creates a potential footgun for users who forget the trailing slash.

### Environment / System Info

_No response_

dotuananh0712:
Fixes #1654. Previously, route prefixes like '/abcd' would incorrectly match '/abcde/file.txt' because it used 'startswith()'. This PR ensures that matches only occur at path boundaries (followed by '/' or exact match).


eyurtsev:
@dotuananh0712 thank you!

PR #1671, opened 2026-03-05
fix(cli): anchor `ChatInput` below scrollable area

jcuypers:
Move `ChatInput` out of the `VerticalScroll` container so it stays fixed above the status bar instead of scrolling away with chat messages. Previously, the input would drift offscreen as streaming content grew, requiring users to scroll down to find it.

## Changes

- Move `ChatInput` and its `bottom-app-container` wrapper outside the `VerticalScroll(id="chat")` block in `AgentApp.compose`, placing it between the scroll area and `StatusBar`
- Remove the `chat-spacer` `Static` widget that was only needed to fill space below the input when it lived inside the scroll area


mdrxy:
thanks @jcuypers, this is something i've been thinking about for some time.

would you mind adding a quick screen recording demonstrating?

jcuypers:
Thanks for the feedback.   Here's an example:
https://github.com/user-attachments/assets/cc7186b0-0644-4723-a37d-b7ff7899d48d

In addition, i would like to move some of the instructions like @ / etc to the top of the chatInput so its clear at any point in time.
for the header: The idea was to add thread summary titles to /threads so at least you know what it is as well as header titles.  something like opencode.
its also easy to add a side panel with some other info.    
If you are ok with the direction, i could work a bit on that

mdrxy:
cool, thanks!

I don't think we should had a header -- screen real estate is finite, and I think it would be better to instead overwrite the terminal's native title field (does that make sense? do you know what title bar I'm talking about? usually it says the name of the running process)

one issue we'll need to handle is the lack of anchoring. in your screen recording, when the LLM output expanded below the viewport, you had to scroll down to see it. I think we should stay anchored to the most recent generation

> i would like to move some of the instructions like @ / etc to the top of the chatInput

I think I'm on board with this.

jcuypers:
The header is a bit mixed I guess:  we should account for whatever system it is run in.   some have these automatic header updates, others don't.   The idea was that in the place of "Deep agents"  would be the autosummary of the initial prompt (just like opencode).   In the header we have full control how we would do that, anything else depends on whatever system / config.  It's just lets say you have tmux with 4 agents and then switching back and forth with threads inside of them :)   currently this is a nightmare (hence my atttempt).   but ok its a democracy, the best / better idea wins.

scrolling to the latest update:  ok I will have a look at it.   

Other improvement I will pass them by in a similar format so its easy to agree and move forward.

mdrxy:
gotcha, understood re: header. i think it should come in a separate PR / not block this PR; could you factor it out while we scope?

I think this PR could be good to go once scrolling is figured out

jcuypers:
Ok, it actually works as-is (the existing logic works):

1) new session:
  a) user enters request in the chatinput
  b) the streaming starts, when there is overflow it automatically scrolls to the newest content.

2) inside an existing session:
   a) the user is near the bottom of the message content -> auto-scroll start on updates
   b) the user has scrolled up to look at something (there is no auto-scroll)
   c) the user is at any position and adds instruction into the chatinput and confirms.  the content goes directly to the end and start streaming new content.

I think this is what we were talking about?

I will remove the header tomorrow and push the final code.

PR #1675, opened 2026-03-06
fix(sdk): deduplicate subagents by name in `SubAgentMiddleware`

IHEII:
## Issue description

When multiple subagent specs share the same name are registered, `_build_task_tool` previously built the `subagent_graphs` dict and description string directly from the raw list, resulting in duplicate entries in the tool description presented to the LLM.

## How this commit works

This fix deduplicates subagents by name (last-one-wins) before building the graphs dict and description string, ensuring each subagent appears only once.

## How was this verified?

Manually tested with a configuration that registers duplicate subagent names and confirmed the generated tool description no longer contains duplicates.

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -389,14 +389,15 @@ def _build_task_tool(  # noqa: C901
     """
     # Detect duplicate subagent names and raise an error.
     seen_names: set[str] = set()
+    duplicate_names: set[str] = set()
     for spec in subagents:
         if spec["name"] in seen_names:
-            msg = (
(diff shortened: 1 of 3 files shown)


eyurtsev:
Could you include a unit test and an MRE to explain how you're launching things?

It's more likely that we'll want to raise a run time error rather than fail silently with last to win

IHEII:
Hi @eyurtsev, I've updated the MR with a RuntimeError for duplicates and added unit tests. Could you take another look? Thanks!
(BTW, the original silent replacement was intended to allow overriding built-in subagents :)

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #1686, opened 2026-03-06
fix(cli): remove dead chat-spacer widget and resize handler

jcuypers:
Remove the `#chat-spacer` widget and all supporting logic from the interactive UI. The spacer was originally used to fill viewport space between the chat input and status bar before any messages existed, but is no longer needed.


mdrxy (approved):
TY!

PR #1704, opened 2026-03-07
feat(cli): experimental background task runtime

### LINKED ISSUE #1283: Support background tasks in the CLI
### Checked other resources

### Area (Required)

- [ ] deepagents (SDK)
- [x] cli

### Feature description

Background tasks/signals/auto-prompting are useful. They let an agent start other work and revisit later. They let an agent easily sleep until a signal fires. 

If we added to deepagents, it would be nice if it were a good abstraction so we could implement it using different queue backends (inmemory asyncio queue, LSD queue, kafka /idk)

_No response_

_No response_

chenwanqq:
Closes #1283

---

Introduces a new background-task capability using a TaskIQ-backed runtime.

Opt-in and disabled by default for now. When enabled, agents can submit and manage long-running shell tasks asynchronously without blocking the main conversation flow.

## Why

Long-running shell/tool operations can block agent progress and degrade interactive UX. 

This change adds a controlled background execution model so the agent can continue reasoning while tasks run, then explicitly wait for results when needed.

### Core runtime
- Added `BackgroundRuntime` in [`libs/cli/deepagents_cli/background_runtime.py`](/Users/chenwanqq/deepagents/libs/cli/deepagents_cli/background_runtime.py)
- Uses TaskIQ `InMemoryBroker` mode
- Provides:
  - submit shell background task
  - list/get task state
  - best-effort kill
  - event-driven wait (no sleep polling in tool layer)
- Tracks task lifecycle states:
  - `queued`, `running`, `succeeded`, `failed`, `killed`
- Includes thread-safe task/event bookkeeping

### Agent middleware
- Added `BackgroundMiddleware` in [`libs/cli/deepagents_cli/background_middleware.py`](/Users/chenwanqq/deepagents/libs/cli/deepagents_cli/background_middleware.py)
- Injects background tools:
  - `submit_background_task`
  - `list_background_tasks`
  - `kill_background_task`
  - `wait_background_task`
- `before_model` hook injects background status summary and guidance to call wait when results are required
- `wait_background_task` is orchestration-only and delegates waiting to runtime

### CLI / app integration
- Added CLI flags in [`libs/cli/deepagents_cli/main.py`](/Users/chenwanqq/deepagents/libs/cli/deepagents_cli/main.py):
  - `--background-tasks` / `--no-background-tasks`
  - `--background-runtime-mode` (currently `inmemory`)
- Runtime lifecycle is managed by CLI app lifecycle:
  - start before app run
  - shutdown in cleanup path
- App integration in [`libs/cli/deepagents_cli/app.py`](/Users/chenwanqq/deepagents/libs/cli/deepagents_cli/app.py):
  - background HITL bridge worker (serial approval handling)
  - TUI notification messages when background tasks complete/fail/kill
- Agent wiring in [`libs/cli/deepagents_cli/agent.py`](/Users/chenwanqq/deepagents/libs/cli/deepagents_cli/agent.py) to conditionally append middleware

### Help + dependency updates
- Help text updated in [`libs/cli/deepagents_cli/ui.py`](/Users/chenwanqq/deepagents/libs/cli/deepagents_cli/ui.py)
- Added dependency in [`libs/cli/pyproject.toml`](/Users/chenwanqq/deepagents/libs/cli/pyproject.toml):
  - `taskiq>=0.12.0,<1.0.0`

## Behavioral Notes
- Scope: interactive TUI mode only
- Default: background tasks disabled
- Kill semantics: cooperative + best-effort; state is converged to `killed` with explanatory text
- HITL for background tasks reuses existing main-session approval UI and is processed serially
- Wait timeout behavior is handled gracefully (task remains background-running; caller gets a clear “not completed yet” response path)

## Review Focus Areas
1. Runtime state transitions and cancellation semantics in `BackgroundRuntime`
2. Middleware hook behavior and tool return contracts
3. App-level HITL bridge serial processing and notification delivery timing
4. CLI lifecycle cleanup ordering (background runtime + sandbox + MCP manager)

### DIFF
--- a/libs/cli/deepagents_cli/agent.py
+++ b/libs/cli/deepagents_cli/agent.py
@@ -30,6 +30,7 @@ if TYPE_CHECKING:
     from langgraph.pregel import Pregel
     from langgraph.runtime import Runtime
 
+    from deepagents_cli.background_runtime import BackgroundRuntime
     from deepagents_cli.mcp_tools import MCPServerInfo
     from deepagents_cli.output import OutputFormat
 
@@ -553,6 +554,8 @@ def create_cli_agent(
     enable_skills: bool = True,
     enable_shell: bool = True,
     enable_ask_user: bool = False,
+    enable_background_tasks: bool = False,
+    background_runtime: BackgroundRuntime | None = None,
     checkpointer: BaseCheckpointSaver | None = None,
     mcp_server_info: list[MCPServerInfo] | None = None,
 ) -> tuple[Pregel, CompositeBackend]:
@@ -590,6 +593,9 @@ def create_cli_agent(
         enable_shell: Enable shell execution via `LocalShellBackend`
             (only in local mode). When enabled, the `execute` tool is available.
         enable_ask_user: Enable the `ask_user` tool for interactive questioning.
+        enable_background_tasks: Enable background-task middleware tools.
+        background_runtime: Shared background runtime used by
+            `BackgroundMiddleware`.
         checkpointer: Optional checkpointer for session persistence.
 
             If `None`, uses `InMemorySaver` (no persistence across
@@ -717,6 +723,11 @@ def create_cli_agent(
         agent_middleware.append(
             LocalContextMiddleware(backend=backend, mcp_server_info=mcp_server_info)
         )
+
+    if enable_background_tasks and background_runtime is not None:
+        from deepagents_cli.background_middleware import BackgroundMiddleware
+
+        agent_middleware.append(BackgroundMiddleware(background_runtime))
 
     # Get or use custom system prompt
     if system_prompt is None:
--- a/libs/cli/deepagents_cli/agent.py
+++ b/libs/cli/deepagents_cli/agent.py
@@ -727,6 +727,10 @@ def create_cli_agent(
     if enable_background_tasks and background_runtime is not None:
         from deepagents_cli.background_middleware import BackgroundMiddleware
 
+        # Wire the active backend so background tasks execute in the same
+        # environment as regular commands (local shell or remote sandbox).
+        if isinstance(backend, _ExecutableBackend):
+            background_runtime.set_backend(backend)  # type: ignore[arg-type]  # _ExecutableBackend implies SandboxBackendProtocol here
         agent_middleware.append(BackgroundMiddleware(background_runtime))
 
     # Get or use custom system prompt
--- a/libs/cli/deepagents_cli/app.py
+++ b/libs/cli/deepagents_cli/app.py
@@ -94,6 +94,8 @@ if TYPE_CHECKING:
     from textual.worker import Worker
 
     from deepagents_cli.ask_user import AskUserWidgetResult, Question
+    from deepagents_cli.background_runtime import BackgroundRuntime
+    from deepagents_cli.ask_user import AskUserWidgetResult, Question
     from deepagents_cli.mcp_tools import MCPServerInfo
 
 # iTerm2 Cursor Guide Workaround
@@ -564,6 +566,7 @@ class DeepAgentsApp(App):
         sandbox_type: str | None = None,
         mcp_server_info: list[MCPServerInfo] | None = None,
         profile_override: dict[str, Any] | None = None,
+        background_runtime: BackgroundRuntime | None = None,
         **kwargs: Any,
     ) -> None:
         """Initialize the Deep Agents application.
@@ -585,6 +588,7 @@ class DeepAgentsApp(App):
             mcp_server_info: MCP server metadata for the `/mcp` viewer.
             profile_override: Extra profile fields from `--profile-override`,
                 retained for model hot-swap and footer display.
+            background_runtime: Optional background task runtime.
             **kwargs: Additional arguments passed to parent
         """
         super().__init__(**kwargs)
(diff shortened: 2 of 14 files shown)


mdrxy:
nice, thanks! would you mind posting a screen recording showing demo usage? excited to review

chenwanqq:
> nice, thanks! would you mind posting a screen recording showing demo usage? 

Thanks! I’ll upload a screen recording demonstrating the usage later today.

chenwanqq:
> #1283

https://github.com/user-attachments/assets/c8e91f03-8dd0-4708-b353-045e44be49e8




hello, here is a demo!

mdrxy:
this is awesome! thank you for taking the time to implement and screen record a demo.

one thing to note:

Screenshot

I think "rejected by approval" is slightly confusing/ambiguous

If the task failed due to rejection, I think we should not print that it failed, since it never started to begin with, right?

open to your thoughts!

chenwanqq:
> this is awesome! thank you for taking the time to implement and screen record a demo.
> 
> one thing to note:
> 
> Screenshot
> I think "rejected by approval" is slightly confusing/ambiguous
> 
> If the task failed due to rejection, I think we should not print that it failed, since it never started to begin with, right?
> 
> open to your thoughts!

Good catch — thanks for pointing that out!

1. I agree that **"rejected by approval"** does sound a bit odd and could be confusing from a UX perspective.

2. I also agree that if a task is **rejected before it actually starts**, it probably doesn’t make sense to print it as a failed task. Since nothing really ran, surfacing it the same way as an actual failure could be misleading.

3. That said, I do think the **rejection signal itself should still be passed back to the agent internally**. Even if we don’t print it in the CLI output, the fact that a command was rejected by the user is still useful information for the agent’s subsequent reasoning and decision-making.

Happy to tweak the wording or output behavior here 👍

chenwanqq:
> this is awesome! thank you for taking the time to implement and screen record a demo.
> 
> one thing to note:
> 
> Screenshot
> I think "rejected by approval" is slightly confusing/ambiguous
> 
> If the task failed due to rejection, I think we should not print that it failed, since it never started to begin with, right?
> 
> open to your thoughts!

Thanks for the feedback! I’ve made some updates based on your suggestions. I also recorded a new demo video reflecting the updated behavior. Would love to hear your thoughts!

https://github.com/user-attachments/assets/eb4e164b-c94b-4f54-b10b-5b4c610d4b9a

mdrxy:
I think the feature direction is useful, but I'd like to reshape it so it builds on the existing backend/sandbox model instead of introducing a second execution path in the CLI.

The main issue with the current approach is that background tasks are implemented as a separate local subprocess runtime. That creates a mismatch with Deep Agents' existing execution model: if the user is in sandbox mode, background commands should also run in that sandbox, not on the host. I don't think this needs to move into the SDK yet, though. We can keep the entire task manager CLI-only and still align it with the current abstractions.

Plan forward:

- Keep the runtime/task registry in the CLI.
- Change background execution to call the active backend's `aexecute()` instead of spawning local subprocesses directly.
- Treat this as a session-local backgrounding feature, not a durable job system.
- Support `submit`, `list`, and `wait` first.
- Make `kill` best-effort in local mode, and either best-effort or disabled in sandbox mode until we have a real per-job cancellation abstraction.
- Make the feature explicitly inherit the current execution environment:
  - local backend => runs locally
  - sandbox backend => runs in the sandbox
- Update the prompt/help text so we are not telling the model "all execution happens in the sandbox" while background tasks actually run elsewhere.

I think that gets us most of the UX value with much less conceptual drift:
- no new execution model
- no SDK refactor required yet
- no conflict with the existing sandbox concept
- much simpler mental model for users and maintainers

The tradeoff is that this would still be "background within the current CLI session," not true remote jobs. So we would not get durable state, reconnection to in-flight commands after restart, live log streaming, or reliable remote cancellation. But that feels like a reasonable v1 boundary, and we can add a richer job abstraction later if this proves valuable.

mdrxy:
@chenwanqq see proposed rfc: https://github.com/chenwanqq/deepagents/pull/3

chenwanqq:
@mdrxy Sorry for the late reply — weekdays are usually a bit busy for me.

I merged your changes and made a few minor fixes for Python 3.11 compatibility. I also tested it in the Daytona sandbox. You can try other sandbox providers as well.

mdrxy:
Hey @chenwanqq — closing this out now that #1759 has landed.

I want to say genuinely, thank you for the time and effort you put into this. The implementation was solid, the demo recordings were great, and you were super responsive to every round of feedback (the approval-rejection UX fix was a nice touch). The architectural discussion we had here — around aligning background execution with the sandbox model and keeping it session-local — directly shaped the direction we ended up taking in #1759.

The core ideas you brought to the table (task lifecycle management, background execution without blocking the conversation, middleware-based tool injection) carried forward into the new design.

Would love to keep collaborating if you're interested. Thanks again for the contribution.

chenwanqq:
> Hey @chenwanqq — closing this out now that #1759 has landed.
> 
> I want to say genuinely, thank you for the time and effort you put into this. The implementation was solid, the demo recordings were great, and you were super responsive to every round of feedback (the approval-rejection UX fix was a nice touch). The architectural discussion we had here — around aligning background execution with the sandbox model and keeping it session-local — directly shaped the direction we ended up taking in #1759.
> 
> The core ideas you brought to the table (task lifecycle management, background execution without blocking the conversation, middleware-based tool injection) carried forward into the new design.
> 
> Would love to keep collaborating if you're interested. Thanks again for the contribution.

Great！ I also believe the client-server architecture will open up more possibilities for numerous new features—such as scheduled tasks, the expansion of more entry points, and agent teams. I will keep following this project closely!

mdrxy:
@chenwanqq yes I am quite excited - you may be interested in #1758

PR #1705, opened 2026-03-07
fix(cli): right-align token count next to model name in status bar

mdrxy:
The token count in the status bar was left-aligned next to the cwd, separated from the model name by a large `1fr` gap. Move the flexible space to a dedicated spacer between cwd and tokens so the token count and model name are grouped together on the right.


eyurtsev (approved):
gogogo

PR #1720, opened 2026-03-08
fix(harbor): `grep` literal search, implement missing `download_files`

zamal-db:
## Problem

I was setting up a Harbor eval pipeline where the agent processes mixed-content repositories (code + screenshots in READMEs, architecture diagrams, etc.). Two things broke:

**grep returning wrong results** - I noticed the agent's `grep` calls were matching lines they shouldn't. After digging in, the issue is that `HarborSandbox.agrep_raw` uses `grep -rHn` while `BaseSandbox` and the protocol docstring both specify `-rHnF` (fixed-string / literal mode). Without the `-F` flag, patterns like `def __init__(self):` treat parentheses as regex group operators, and `str | int` treats `|` as alternation, matching every line containing either word. The tool description even gives `def __init__(self):` as an example pattern, so this is clearly meant to be literal.

**image file reads crashing** - any time the agent tried to read a `.png` or `.jpg`, the run died with `NotImplementedError`. The cause: `FilesystemMiddleware` (which `create_deep_agent` always includes in the middleware stack) routes image reads through `download_files`/`adownload_files` on the backend. `HarborSandbox` never implemented these methods, so it falls through to `BackendProtocol`'s default `raise NotImplementedError`. Every other sandbox backend (Daytona, Modal, Runloop) has implementations. Harbor was the only one missing them.

## Fix

- `agrep_raw`: one-character fix, add the `-F` flag to `grep_opts` so it matches `BaseSandbox` and the protocol contract.
- `adownload_files`: implemented using base64-encode-in-shell then decode locally, which is the same pattern `awrite` already uses in this file. Handles missing files (exit 1), directories (exit 2), and strips the `stderr` marker that `aexecute` appends before decoding.

## Tests

Harbor had no real unit tests, just a `test_placeholder.py` import check. Added an async test suite with a `MockEnvironment` that stubs `BaseEnvironment.exec`, covering all existing operations (execute, read, write, edit, ls, grep, glob, download) plus sync-not-supported guards. The grep tests explicitly assert `-rHnF` is in the command string.


mdrxy:
We've migrated to ~/libs/evals, would you mind rebasing to match? Thanks!

zamal-db:
rebased onto main and updated tests to match the libs/evals migration. method names now use the new api (als, agrep, aglob) and assertions unwrap the new result types (ReadResult, LsResult, GrepResult, GlobResult). ready for review.

PR #1741, opened 2026-03-09
fix(cli): use max-height for tool-info-scroll to shrink-wrap content

### LINKED ISSUE #1120: Tool HITL window massive when not necessary
### Privileged issue

### Issue Content

Image

**Root cause:** `app.tcss:149` sets `.tool-info-scroll` to a fixed `height: 10`, so the scroll container is always 10 lines tall even when content is 1 line. Tools like `web_search` go through `GenericApprovalWidget` (key-value pairs), often producing just one line of output.

**Proposed fix** — single-line CSS change in `libs/cli/deepagents_cli/app.tcss`:

```diff
 /* Scrollable tool info area in approval menu */
 .approval-menu .tool-info-scroll {
-    height: 10;
+    max-height: 10;
     margin-top: 0;
 }
```

`max-height` lets the container shrink to fit content naturally while still capping at 10 lines for tools with large content (diffs, file writes). Scroll kicks in only when content exceeds 10 lines.

- `web_search`, `read_file`, other generic tools with small args → window shrinks to fit
- `edit_file` (diffs up to 50 lines) → still capped at 10 visible lines with scroll
- `write_file` (content up to 30 lines) → still capped at 10 visible lines with scroll
- `bash`/`shell`/`execute` → unaffected (use `_MINIMAL_TOOLS` path, skip scroll container)

### Area (Required)

- [ ] deepagents (SDK)
- [x] cli

Br1an67:
Fixes #1120

---

The Tool HITL (Human-in-the-Loop) approval window was always 10 lines tall even when content was just 1 line. This made the window unnecessarily large for tools like `web_search` that produce minimal output.

## Changes

- Changed `height: 10` to `max-height: 10` for `.approval-menu .tool-info-scroll` in `app.tcss`

This allows the container to shrink to fit its content naturally while still capping at 10 lines for tools with larger content (diffs, file writes). Scrolling kicks in only when content exceeds 10 lines.


mdrxy:
thanks, can you show a before/after screenshot? let's add it to the original PR body and reply when added please

mdrxy:
@Br1an67 I tried testing this and the window is now full-height

mdrxy:
Done in #1835

PR #1758, opened 2026-03-10
feat(sdk): add async subagent middleware for remote LangGraph servers

hwchase17:
Adds `AsyncSubAgentMiddleware` to the SDK, enabling agents to launch background jobs on remote LangGraph deployments via the LangGraph SDK. Unlike synchronous subagents that block until completion, async subagents return a job ID immediately — the main agent can continue working, check progress on demand, send follow-up instructions, or cancel jobs. 

The CLI reads async subagent definitions from `[async_subagents]` in `~/.deepagents/config.toml` and wires them in automatically.

## Changes
- Add `AsyncSubAgentMiddleware` in `deepagents/middleware/async_subagents.py` with five tools: `launch_async_subagent`, `check_async_subagent`, `update_async_subagent`, `cancel_async_subagent`, and `list_async_subagent_jobs` — each with sync + async implementations that return `Command` objects to persist job state
- Track jobs via `AsyncSubAgentState` with a `_jobs_reducer` that merges updates into a `dict[str, AsyncSubAgentJob]`, surviving context compaction
- Cache LangGraph SDK clients per `(url, headers)` tuple in `_ClientCache` to avoid reconnecting on every tool call
- Add `async_subagents` parameter to `create_deep_agent` in `graph.py` — when provided, appends `AsyncSubAgentMiddleware` to the middleware stack
- Add `load_async_subagents()` in the CLI to parse `config.toml` and pass specs through to `create_cli_agent` in both interactive and server modes
- Wire HITL interrupt config for `launch_async_subagent`, `update_async_subagent`, and `cancel_async_subagent` in the CLI's approval system
- Export `AsyncSubAgent`, `AsyncSubAgentJob`, and `AsyncSubAgentMiddleware` from `deepagents` and `deepagents.middleware`

### DIFF
--- a/libs/cli/deepagents_cli/agent.py
+++ b/libs/cli/deepagents_cli/agent.py
@@ -599,6 +599,11 @@ def _add_interrupt_on() -> dict[str, InterruptOnConfig]:
         "description": _format_task_description,  # type: ignore[typeddict-item]  # Callable description narrower than TypedDict expects
     }
 
+    async_subagent_interrupt_config: InterruptOnConfig = {
+        "allowed_decisions": ["approve", "reject"],
+        "description": "Launch or send a follow-up to a remote async subagent.",
+    }
+
     interrupt_map: dict[str, InterruptOnConfig] = {
         "execute": execute_interrupt_config,
         "write_file": write_file_interrupt_config,
@@ -606,6 +611,8 @@ def _add_interrupt_on() -> dict[str, InterruptOnConfig]:
         "web_search": web_search_interrupt_config,
         "fetch_url": fetch_url_interrupt_config,
         "task": task_interrupt_config,
+        "launch_async_subagent": async_subagent_interrupt_config,
+        "update_async_subagent": async_subagent_interrupt_config,
     }
 
     if REQUIRE_COMPACT_TOOL_APPROVAL:
--- a/libs/cli/deepagents_cli/agent.py
+++ b/libs/cli/deepagents_cli/agent.py
@@ -97,6 +97,10 @@ def load_async_subagents(config_path: Path | None = None) -> list[AsyncSubAgent]
             data = tomllib.load(f)
     except (tomllib.TOMLDecodeError, PermissionError, OSError) as e:
         logger.warning("Could not read async subagents from %s: %s", config_path, e)
+        console.print(
+            f"[bold yellow]Warning:[/bold yellow] Could not read async subagents "
+            f"from {config_path}: {e}",
+        )
         return []
 
     section = data.get("async_subagents")
@@ -605,7 +609,7 @@ def _add_interrupt_on() -> dict[str, InterruptOnConfig]:
 
     async_subagent_interrupt_config: InterruptOnConfig = {
         "allowed_decisions": ["approve", "reject"],
-        "description": "Launch or send a follow-up to a remote async subagent.",
+        "description": "Launch, update, or cancel a remote async subagent.",
     }
 
     interrupt_map: dict[str, InterruptOnConfig] = {
--- a/libs/cli/deepagents_cli/agent.py
+++ b/libs/cli/deepagents_cli/agent.py
@@ -7,6 +7,7 @@ import os
 import re
 import shutil
 import tempfile
+import tomllib
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
@@ -81,8 +82,6 @@ def load_async_subagents(config_path: Path | None = None) -> list[AsyncSubAgent]
     Returns:
         List of `AsyncSubAgent` specs (empty if section is absent or invalid).
     """
-    import tomllib
-
     if config_path is None:
         config_path = Path.home() / ".deepagents" / "config.toml"
 
@@ -100,7 +99,7 @@ def load_async_subagents(config_path: Path | None = None) -> list[AsyncSubAgent]
     if not isinstance(section, dict):
         return []
 
-    required = {"description", "url", "graph_id"}
+    required = {"description", "graph_id"}
     agents: list[AsyncSubAgent] = []
     for name, spec in section.items():
         if not isinstance(spec, dict):
@@ -115,9 +114,10 @@ def load_async_subagents(config_path: Path | None = None) -> list[AsyncSubAgent]
         agent: AsyncSubAgent = {
             "name": name,
             "description": spec["description"],
-            "url": spec["url"],
             "graph_id": spec["graph_id"],
         }
+        if "url" in spec and isinstance(spec["url"], str):
+            agent["url"] = spec["url"]
         if "headers" in spec and isinstance(spec["headers"], dict):
             agent["headers"] = spec["headers"]
         agents.append(agent)
--- a/libs/cli/deepagents_cli/agent.py
+++ b/libs/cli/deepagents_cli/agent.py
@@ -19,6 +19,7 @@ if TYPE_CHECKING:
     from collections.abc import Callable, Sequence
 
     from deepagents.backends.sandbox import SandboxBackendProtocol
+    from deepagents.middleware.async_subagents import AsyncSubAgent
     from deepagents.middleware.subagents import CompiledSubAgent, SubAgent
     from langchain.agents.middleware import InterruptOnConfig
     from langchain.agents.middleware.types import AgentState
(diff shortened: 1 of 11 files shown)


libs/cli/deepagents_cli/agent.py
```diff
@@ -61,6 +62,67 @@
...
+        url = "https://my-deployment.langsmith.dev"
+        graph_id = "agent"
+
+    Args:
+        config_path: Path to config file. Defaults to `~/.deepagents/config.toml`.
+
+    Returns:
+        List of `AsyncSubAgent` specs (empty if section is absent or invalid).
+    """
+    import tomllib
```
hinthornw: nit

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,431 @@
...
+                headers=_resolve_headers(spec),
+            )
+        return self._sync[name]
+
+    def async_(self, name: str) -> LangGraphClient:
+        """Get or create an async client for the named agent."""
+        if name not in self._async:
+            spec = self._agents[name]
+            self._async[name] = get_client(
+                url=spec["url"],
```
hinthornw: Should this default to `None`? In most cases we want to use the asgi transport, right?

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,431 @@
...
+        """Get or create a sync client for the named agent."""
+        if name not in self._sync:
+            spec = self._agents[name]
+            self._sync[name] = get_sync_client(
+                url=spec["url"],
+                headers=_resolve_headers(spec),
+            )
+        return self._sync[name]
+
+    def async_(self, name: str) -> LangGraphClient:
```
hinthornw: The interface naming is kinda funky

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,431 @@
...
+        """Get or create an async client for the named agent."""
+        if name not in self._async:
+            spec = self._agents[name]
+            self._async[name] = get_client(
```
hinthornw: Creating clients is expensive. I'd prefer not to create a new one for every sub-agent

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,431 @@
...
+    def launch_async_subagent(
+        description: Annotated[str, "A detailed description of the task for the async subagent to perform."],
+        subagent_type: Annotated[str, "The type of async subagent to use. Must be one of the available types."],
+    ) -> str:
+        error = _validate_agent_type(agent_map, subagent_type)
+        if error:
+            return error
+        spec = agent_map[subagent_type]
+        client = clients.sync(subagent_type)
+        thread = client.threads.create()
```
hinthornw: ooc: why do we need sync methods? Can we use the single async client and avoid the ssl cache hit and avoid the fact that sync client doesn't support asgi transport?

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,431 @@
...
+    async def alaunch_async_subagent(
+        description: Annotated[str, "A detailed description of the task for the async subagent to perform."],
+        subagent_type: Annotated[str, "The type of async subagent to use. Must be one of the available types."],
+    ) -> str:
+        error = _validate_agent_type(agent_map, subagent_type)
+        if error:
+            return error
+        spec = agent_map[subagent_type]
+        client = clients.async_(subagent_type)
+        thread = await client.threads.create()
```
hinthornw: I'd prefer to just create the thread inline (if_not_exist="Create") to avoid the doulbe requests

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,431 @@
...
+            thread_id=thread["thread_id"],
+            assistant_id=spec["graph_id"],
+            input={"messages": [{"role": "user", "content": description}]},
+        )
+        job_id = _format_job_id(subagent_type, thread["thread_id"], run["run_id"])
+        return f"Launched async subagent. job_id: {job_id}"
+
+    async def alaunch_async_subagent(
+        description: Annotated[str, "A detailed description of the task for the async subagent to perform."],
+        subagent_type: Annotated[str, "The type of async subagent to use. Must be one of the available types."],
```
hinthornw: What are the available types?

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,431 @@
...
+
+    async def acheck_async_subagent(
+        job_id: Annotated[str, "The exact job_id string returned by launch_async_subagent. Pass it verbatim."],
+    ) -> str:
+        agent_name, thread_id, run_id = _parse_job_id(job_id)
+        client = clients.async_(_resolve_client_name(agent_name, agent_map))
+        run = await client.runs.get(thread_id=thread_id, run_id=run_id)
+        result: dict[str, Any] = {"status": run["status"], "run_id": run["run_id"], "thread_id": thread_id}
+        if run["status"] == "success":
+            state = await client.threads.get_state(thread_id=thread_id)
```
hinthornw: Cheaper to just threads.get() and get the values if all you're trying to do is get the last message. Don't use get_state unless you need the additional data.

libs/cli/deepagents_cli/main.py
```diff
@@ -768,12 +773,17 @@ async def _run_acp_cli_async(
         sys.stderr.flush()
         return 1
 
+    from deepagents_cli.agent import load_async_subagents
```
hinthornw: why defer?

libs/cli/deepagents_cli/main.py
```diff
@@ -634,6 +634,10 @@ async def run_textual_cli_async(
                 console.print(Text(str(e), style="dim"))
                 sys.exit(1)
 
+        from deepagents_cli.agent import load_async_subagents
```
hinthornw: why defer?

libs/cli/deepagents_cli/main.py
```diff
@@ -634,6 +634,10 @@ async def run_textual_cli_async(
...
+
+        async_subagents = load_async_subagents() or None
```
hinthornw: ooc can we load from the server?

hinthornw (approved):

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: Agreed but the docstring calls out get or create which makes it sound like a singleton. We should double check this.

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,431 @@
...
+
+    return StructuredTool.from_function(
+        name="launch_async_subagent",
+        func=launch_async_subagent,
+        coroutine=alaunch_async_subagent,
+        description=description,
+    )
+
+
+def _resolve_client_name(agent_name: str, agent_map: dict[str, AsyncSubAgent]) -> str:
```
colifran: Looks like we have a legacy 2-part job id format and a 3-part job id format. I think in the case of the legacy 2-part job id when`agent_name` is `""` it makes sense to use `next(iter(agent_map))` but I'm wondering if we should be throwing an error for the case where we have an `agent_name` but it isn't in the `agent_map` instead of using `next(iter(agent_map))`

libs/cli/deepagents_cli/main.py
sydney-runkle: we shouldn't, @mdrxy lint should flag here?

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: Updated to get_async and get_sync

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: Updated

libs/cli/deepagents_cli/main.py
colifran: Updated

libs/cli/deepagents_cli/main.py
colifran: Updated.

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: Fixed.

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: `url` is now `NotRequired` in `AsyncSubAgent` and both `get_sync` and `get_async` use `spec.get("url")`, so omitting it falls through to ASGI transport. Does that make sense? Can you confirm?

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: If we key on (url, headers) instead of agent name then two agents with the same URL and headers would share a single client. Would that work here for a better client cache? Something like:
```py
def get_async(self, name: str) -> LangGraphClient:
    spec = self._agents[name]
    key = (spec.get("url"), frozenset(_resolve_headers(spec).items()))
    if key not in self._async:
        self._async[key] = get_client(url=spec.get("url"), headers=_resolve_headers(spec))
    return self._async[key]
```

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: Do you think we can just remove them?

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: What's the preferred approach for creating a thread id? UUID? Leave it as None?

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: Updated.

libs/deepagents/deepagents/middleware/async_subagents.py
colifran: Updated.

mdrxy (approved):

libs/deepagents/deepagents/middleware/async_subagents.py
```diff
@@ -0,0 +1,843 @@
...
+            )
+        return self._sync[key]
+
+    def get_async(self, name: str) -> LangGraphClient:
+        """Get or create an async client for the named agent."""
+        spec = self._agents[name]
+        key = self._cache_key(spec)
+        if key not in self._async:
+            self._async[key] = get_client(
+                url=spec.get("url"),
```
ccurme, line 198: in what circumstances does this work when `url` is None? if we deploy both the deepagent and a dedicated async subagent through langgraph.json, would this work?

ccurme (approved):

PR #1759, opened 2026-03-10
feat(cli): client-server architecture via `langgraph dev`

hwchase17:
(taken over by @mdrxy)

---

# 💫 North star: The CLI becomes a thin client over a LangGraph API server.

Today the CLI runs the agent in-process. This PR splits it into client (TUI, input handling, streaming display) and server (`langgraph dev` subprocess running the agent graph). The CLI auto-spawns the server transparently — same UX, but the architecture now supports:

* **Remote/daemon mode** — connect to an already-running server (e.g. `--server-url`) instead of spawning one
* **Persistent remote sessions** — server owns the checkpointer; no more local SQLite fallback
* **Multi-client** — multiple CLI instances (or other clients) can share one server, enabling collaborative/shared agent sessions down the line

(note these are still work in progress but goals we aim to explore)

---

Decouple agent execution from the CLI process by introducing a client-server architecture. The CLI now auto-starts a `langgraph dev` subprocess, passes configuration via `DA_SERVER_*` environment variables, and connects through `langgraph-sdk`'s `RemoteGraph` over HTTP+SSE. This lets the agent graph run in an isolated server process while the Textual TUI stays a pure UI client — a prerequisite for multi-session support, remote deployment, and cleaner process isolation.

## Changes

- Add `ServerProcess` in `server.py` to manage the `langgraph dev` subprocess lifecycle (start, health-poll via `/ok`, graceful `SIGTERM` shutdown, port collision recovery)
- Add `RemoteAgent` in `remote_client.py` wrapping `RemoteGraph` — handles thread-ID-to-UUID normalization, message-dict-to-LangChain-object conversion for the three message types (`ai`/`human`/`tool`), and interrupt object deserialization
- Add `ServerConfig` dataclass in `_server_config.py` with `to_env()`/`from_env()` symmetry so the CLI and server subprocess share a single schema for the `DA_SERVER_*` env-var communication channel
- Add `server_graph.py` as the `langgraph dev` entry point — reads `ServerConfig.from_env()`, resolves model/tools/MCP, and calls `create_cli_agent(checkpointer=False)` since the platform manages persistence
- Add `server_manager.py` with `server_session()` async context manager that orchestrates workspace scaffolding (generates `langgraph.json`, checkpointer module pointing at the CLI's SQLite DB, and a `pyproject.toml` referencing the local CLI package), server startup, and guaranteed cleanup
- Add `ConfigurableModelMiddleware` that reads `config["configurable"]["model"]` to swap the model per-invocation without recompiling the graph — replaces the old `create_cli_agent()` rebuild approach for `/model` switching. Needed or else we need to restart/rebuild CLI for mid-thread swaps. This simultaneously solves our prior problem of the app having a short hang for graph re-builds when swapping
  - Add `_deepagents_prepend` flag on middleware and `_split_middleware()` in the SDK's `graph.py` so `ConfigurableModelMiddleware` runs before the built-in stack
- Add `ProjectContext` dataclass in `project_utils.py` to carry the user's original cwd and project root through the server boundary, replacing scattered `Path.cwd()` / `find_project_root()` calls in `agent.py`, `mcp_tools.py`, and `config.py`
- Simplify `_switch_model` in `app.py` from ~100 lines of agent-rebuild-with-rollback to ~20 lines that set `self._model_override` for the middleware
- Rework `run_textual_cli_async` and `run_non_interactive` in `main.py`/`non_interactive.py` to use `server_session()` instead of inline agent/sandbox/MCP/checkpointer setup
- Add `_is_remote()` guard in `DeepAgentsApp` and thread-history fallback that reads the SQLite checkpointer directly when the dev server returns empty state for previously persisted threads

---

Credit: This work was informed by @chenwanqq's background task runtime in #1704, which explored task lifecycle management and background execution patterns that shaped the architectural direction here.


libs/deepagents/deepagents/graph.py
```diff
@@ -78,6 +78,30 @@ def get_default_model() -> ChatAnthropic:
...
+    Args:
+        middleware: Middleware passed to `create_deep_agent()`.
+
+    Returns:
+        Two lists: `(prepended, trailing)`.
+    """
+    prepended: list[AgentMiddleware] = []
+    trailing: list[AgentMiddleware] = []
+    for item in middleware:
+        if getattr(item, "_deepagents_prepend", False):
```
mdrxy: eugene is gonna hate this

PR #1760, opened 2026-03-10
feat(cli): support `~/.deepagents/.env` as global API key fallback

lmamww:
This pull request updates the environment variable loading logic in the CLI configuration to ensure that a global environment file (`~/.deepagents/.env`) is loaded before the local `.env` file, and that the local file can override the global settings. It also adds and updates tests to verify this behavior and the order of loading.

**Environment Variable Loading Improvements:**

* Changed `dotenv.load_dotenv` calls in `config.py` to first load the global `~/.deepagents/.env` file, followed by the local `.env` file with `override=True`, ensuring local settings take precedence. [[1]](diffhunk://#diff-ae79bb00939843f72b249990644f08fc857b2652279f4ddbeaecdef10ebe4f35L27-R28) [[2]](diffhunk://#diff-ae79bb00939843f72b249990644f08fc857b2652279f4ddbeaecdef10ebe4f35R561)

**Testing Enhancements:**

* Updated the test `test_calls_dotenv_load` to verify that both the global and local `.env` files are loaded in the correct order and with the correct arguments.
* Added a new test `test_global_env_loaded_before_cwd` to explicitly check that the global environment file is loaded before the local one.

### DIFF
--- a/libs/cli/deepagents_cli/config.py
+++ b/libs/cli/deepagents_cli/config.py
@@ -24,7 +24,8 @@ from deepagents_cli._version import __version__
 
 logger = logging.getLogger(__name__)
 
-dotenv.load_dotenv()
+dotenv.load_dotenv(Path.home() / ".deepagents" / ".env")
+dotenv.load_dotenv(override=True)
 
 # CRITICAL: Override LANGSMITH_PROJECT to route agent traces to separate project
 # LangSmith reads LANGSMITH_PROJECT at invocation time, so we override it here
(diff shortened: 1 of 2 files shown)


mdrxy:
Thanks for the contribution @lmamww! We had the same idea — #1909 covers this with a few additions (centralized `_GLOBAL_DOTENV_PATH` constant, `OSError` handling for broken/unreadable paths, and routing both bootstrap + reload through the existing `_load_dotenv` abstraction). Closing in favor of that PR, but appreciate you flagging this as a gap.

PR #1767, opened 2026-03-10
fix(sdk): prevent commands from hanging waiting for stdin

### LINKED ISSUE #1766: Commands hang waiting for stdin when executed via `LocalShellBackend`
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

_No response_

```python
1. Open `deepagents-cli`
2. Enter the prompt `run python`
3. The command hangs until timeout (e.g., 120 seconds) instead of exiting immediately

(Check out the screenshot in the "Errror Message")
```

Image

#### Problem

When running commands that require stdin input (e.g., python, cat, read, etc.) through the LocalShellBackend.execute() method without providing stdin, these commands hang indefinitely waiting for input. This causes the command to wait until the timeout is reached, which is inefficient and inconsistent with the expected behavior.

This issue manifests in `deepagents-cli` when executing shell commands through the local shell backend.

#### Expected behavior

Similar to Claude Code, when a command is started non-interactively in the background without a terminal, it should exit immediately if no input is provided, rather than hanging until timeout.

#### Proposed fix

Add `stdin=subprocess.DEVNULL` to the `subprocess.run()` call in `LocalShellBackend.execute()` to prevent the command from waiting for stdin input:

```
  result = subprocess.run(
      command,
      check=False,
      shell=True,
      capture_output=True,
      stdin=subprocess.DEVNULL,  # No stdin interaction
      text=True,
      timeout=effective_timeout,
      env=self._env,
      cwd=str(self.cwd),
  )
```

This ensures that interactive commands exit immediately instead of waiting for stdin, making the behavior consistent with Claude Code and other non-interactive shell execution environments.

### Environment / System Info

_No response_

RussellLuo:
Set `stdin=subprocess.DEVNULL` in LocalShellBackend.execute() so commands like python (without arguments) exit immediately instead of waiting for stdin until timeout.

Fixes #1766.

After the fix, the python command will exit immediately:

image

### DIFF
--- a/libs/deepagents/deepagents/backends/local_shell.py
+++ b/libs/deepagents/deepagents/backends/local_shell.py
@@ -301,6 +301,7 @@ class LocalShellBackend(FilesystemBackend, SandboxBackendProtocol):
                 check=False,
                 shell=True,  # Intentional: designed for LLM-controlled shell execution
                 capture_output=True,
+                stdin=subprocess.DEVNULL,  # No stdin interaction
                 text=True,
                 timeout=effective_timeout,
                 env=self._env,
--- a/libs/deepagents/tests/unit_tests/test_local_shell.py
+++ b/libs/deepagents/tests/unit_tests/test_local_shell.py
@@ -1,6 +1,7 @@
 """Unit tests for LocalShellBackend per-command timeout features."""
 
 import subprocess
+import time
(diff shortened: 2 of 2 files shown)


mdrxy:
Done in #2427

PR #1774, opened 2026-03-10
feat(sdk): add configurable artifact paths

eyurtsev:
Adds an  parameter to  so SDK-managed files can live under a caller-controlled parent directory instead of hardcoded root paths. Large tool result eviction and summarization history now derive their paths from that setting, and the CLI routes its own SDK-managed artifacts under  to keep the behavior consistent.

Also adds focused unit coverage for the new path plumbing in filesystem and summarization middleware factories.

Created with [Deep Agents CLI](https://docs.langchain.com/oss/python/deepagents/cli/overview).


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -461,6 +461,7 @@ def __init__(
         system_prompt: str | None = None,
         custom_tool_descriptions: dict[str, str] | None = None,
         tool_token_limit_before_evict: int | None = 20000,
+        large_tool_results_path_prefix: str = "/large_tool_results",
```
sydney-runkle, line 448: do we anticipate any more eviction settings? should we nest?

libs/deepagents/deepagents/graph.py
```diff
@@ -113,6 +113,7 @@ def create_deep_agent(  # noqa: C901, PLR0912  # Complex graph assembly logic wi
     subagents: list[SubAgent | CompiledSubAgent] | None = None,
     skills: list[str] | None = None,
     memory: list[str] | None = None,
+    artifacts: str | None = None,
```
sydney-runkle, line 92: why not `path` in the arg name? seems inconsistent?

sydney-runkle: I don't think we should allow this as a top level setting on `create_deep_agent`

makes more sense on the backend?

sydney-runkle (commented):
one high level question -- are the artifacts right now actually usable at all?
specifically the conversation history? I recall this being a pain point...

libs/deepagents/deepagents/graph.py
eyurtsev: skills/memory are paths as welll -- so a bit more consistent with those param names

the choice of keeping the names simple was to be able to promote to a union type to specify additional configuration if necessary

libs/deepagents/deepagents/middleware/filesystem.py
ccurme: there would be with https://github.com/langchain-ai/deepagents/pull/2183 (though I personally don't mind organizing them flatly)

ccurme (dismissed):

PR #1781, opened 2026-03-11
ci: consolidate pr labeling workflows to eliminate race condition

mdrxy:
Four separate workflows (`pr_size_labeler.yml`, `pr_labeler_file.yml`, `pr_labeler_title.yml`, and the PR-handling half of `tag-external-contributions.yml`) fired concurrently on `pull_request_target`, all mutating labels on the same PR. The `bcoe/conventional-release-labels` action in the title labeler internally calls `setLabels`, which *replaces* all labels rather than appending — clobbering labels that other workflows had already applied within the same second. Consolidates all PR labeling into a single sequential workflow to eliminate the race.

## Changes

- **Single sequential workflow** (`pr_labeler.yml`): size, file-based, title-based, internal/external classification, and contributor tier labels are now computed in one `actions/github-script` step and applied via a single batch `addLabels` call — no more concurrent `setLabels` clobber
- **Replace `bcoe/conventional-release-labels`** with inline title parsing using `/^(\w+)(?:\(([^)]+)\))?(\!)?:/` — preserves the same `typeToLabel` and `scopeToLabel` maps, adds support for the `\!` breaking-change indicator, and removes stale type labels explicitly instead of relying on `setLabels` side effects
- **File-based labeling inlined**: the `actions/labeler@v6` + `.github/pr-file-labeler.yml` config is replaced by equivalent `startsWith`/`endsWith`/regex checks in the same script step, eliminating the external config file and a separate workflow run
- **`ensureLabel` before batch `addLabels`**: GitHub API returns 422 if any label in a batch doesn't exist, which would prevent *all* labels from being applied — each label is now verified/created before the batch call
- **`tag-external-contributions.yml` → `tag-external-issues.yml`**: stripped of `pull_request_target` trigger and all PR-related steps, scoped to issues only, `pull-requests: write` permission removed
- **`external` label still uses App token** for PR labeling (separate step at L228-244) so the `labeled` event propagates to `require_issue_link.yml`; `internal` uses `GITHUB_TOKEN` since no downstream workflow reacts to it
- **Backfill improved**: PR backfill now covers file-based and title-based labels (old backfill only did size + contributor classification); issue backfill split into its own `workflow_dispatch` in `tag-external-issues.yml`
- **`require_issue_link.yml`**: comment-only update — dependency references changed from `tag-external-contributions.yml` to `pr_labeler.yml`


.github/workflows/pr_labeler.yml
```diff
@@ -0,0 +1,470 @@
+# Unified PR labeler — applies size, file-based, title-based, and
```
eyurtsev, line 1: Could you run a security review on this prior to merging?

PR #1788, opened 2026-03-11
feat(sdk): expose `summary_prompt` parameter in `create_deep_agent`

Saxxon-Usagi:
## Summary

- Surfaces the `summary_prompt` parameter through `create_deep_agent()`, allowing users to customize how conversation history is summarized when the context window fills up
- The custom prompt is forwarded to all three internal `create_summarization_middleware` calls: main agent, general-purpose subagent, and user-provided subagents
- Zero-cost for existing users — defaults are unchanged

## Why `summary_prompt` instead of user-constructed `SummarizationMiddleware`

`create_deep_agent` internally constructs **3** `SummarizationMiddleware` instances (main agent, GP subagent, custom subagents). A user who wants to customize summarization currently has two options, both problematic:

1. **Pass their own via `middleware=`** — this *adds* a 4th instance on top of the 3 internal ones, causing double summarization. There's no way to replace the internal ones from outside.
2. **Skip `create_deep_agent` entirely** — build the full middleware stack manually with `create_agent()`, duplicating ~100 lines of assembly logic.

The `summary_prompt` parameter solves this by letting users customize all internal instances without replacing them or replicating the assembly logic.

## Test plan

- [x] New unit test verifies `summary_prompt` is forwarded to all `create_summarization_middleware` calls
- [x] All existing smoke tests pass (5/5, snapshot tests unchanged)
- [x] Full unit test suite passes (727 passed)
- [x] Ruff lint and `ty` type check pass

## Review notes

- The `summary_prompt` parameter is added to both `create_summarization_middleware` (factory) and `create_deep_agent` (public API)
- Uses keyword-only argument (`*`) to avoid breaking positional callers
- When `None` (default), resolves to `DEFAULT_SUMMARY_PROMPT` — existing behavior is unchanged

---

> This contribution was developed with AI assistance.

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -119,6 +119,7 @@ def create_deep_agent(  # noqa: C901, PLR0912  # Complex graph assembly logic wi
     store: BaseStore | None = None,
     backend: BackendProtocol | BackendFactory | None = None,
     interrupt_on: dict[str, bool | InterruptOnConfig] | None = None,
+    summary_prompt: str | None = None,
     debug: bool = False,
     name: str | None = None,
     cache: BaseCache | None = None,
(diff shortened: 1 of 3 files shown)


eyurtsev:
@Saxxon-Usagi thank you for the PR, please open an issue first for discussion before making big API changes. This isn't an API change that we'll make -- i.e,. we do not want summarization to be configurable at all, as that will prevent us from improving it for users

Saxxon-Usagi:
@eyurtsev thank you for letting me know. I understand perfectly. It's my first PR, I will open issue first to discuss in the future. Love to contribute more to this amazing repo.

PR #1789, opened 2026-03-11
style(cli): remove unnecessary noqa comments in `ui`

JasonBuildAI:
## Summary

This PR removes a few `# noqa: E501` comments in `libs/cli/deepagents_cli/ui.py` 
by wrapping long help text lines using Python's implicit string concatenation. 
The rendered help output remains exactly the same; only the source formatting changes.

## Changes

- Wrapped the `-r, --resume` option help line across two string literals.
- Wrapped the `--model-params JSON` option help line across two string literals.
- Wrapped the non-interactive example 
  `deepagents -n 'List files' --shell-allow-list recommended  # Use safe commands` 
  across two string literals.
- Removed the corresponding `# noqa: E501` comments now that the lines fit within 
  the configured line length.

## Why

Reducing unnecessary `noqa` comments makes the codebase a bit cleaner and keeps 
lint rules effective. Python's implicit string concatenation lets us keep these 
help messages readable in the source without suppressing line-length checks.

## Testing

- [x] `uv run ruff check deepagents_cli/ui.py` passes locally
- This change only adjusts string literals; the help text content is intended 
  to be identical to the previous version.
- I am relying on the existing CLI/UI tests in this package; CI will validate 
  the full test suite.

## AI Usage

This change was drafted with AI assistance (Cursor / GPT) and then reviewed and 
submitted by me.

### DIFF
--- a/libs/cli/deepagents_cli/ui.py
+++ b/libs/cli/deepagents_cli/ui.py
@@ -77,14 +77,12 @@ def show_help() -> None:
 
     console.print("[bold]Options:[/bold]", style=COLORS["primary"])
     console.print(
-        "  -r, --resume [ID]          Resume thread: -r for most recent, "
-        "-r ID for specific"
+        "  -r, --resume [ID]          Resume thread: -r for most recent, -r ID for specific"  # noqa: E501
     )
     console.print("  -a, --agent NAME           Agent to use (e.g., coder, researcher)")
     console.print("  -M, --model MODEL          Model to use (e.g., gpt-4o)")
     console.print(
-        "  --model-params JSON        Extra model kwargs "
-        "(e.g., '{\"temperature\": 0.7}')"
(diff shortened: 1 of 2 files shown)


libs/cli/deepagents_cli/ui.py
```diff
@@ -127,9 +112,6 @@ def show_help() -> None:
     console.print(
         "  --no-stream                Buffer full response instead of streaming"
     )
```
mdrxy, line 129: please undo changes to this file, would prefer to keep these statements single line. noqa is fine here.

JasonBuildAI: Got it, thanks for clarifying — I’ve reverted the changes in `ui.py` and it looks like this PR no longer differs from `main`, so I’ll go ahead and close it. Appreciate the quick review!

mdrxy: all good, thanks!

PR #1801, opened 2026-03-11
chore(cli): update default nvidia model to nemotron-3-super-120b

soumilinandi:
Updates the default NVIDIA model from `nvidia/nemotron-3-nano-30b-a3b` to `nvidia/nemotron-3-super-120b-a12b` when NVIDIA_API_KEY is configured.

## Verification

Tested by setting only `NVIDIA_API_KEY` and verifying deepagents uses the new default model.


mdrxy:
thanks!

PR #1807, opened 2026-03-11
feat(sdk): add MemoryAgentBench evaluation suite

maahir30:
## MemoryAgentBench Context?

MemoryAgentBench is a benchmark from ICLR 2026 designed specifically for evaluating **memory in LLM agents** (not just chat assistants). It tests whether an agent can absorb a large body of facts delivered incrementally over many conversation turns and later answer questions about that information accurately.

The benchmark uses an "inject once, query multiple" protocol: the agent is given thousands of facts across multiple turns, then asked questions that require it to recall, reason over, or reconcile those facts. The key insight is that this mirrors how real agents operate -- information arrives over time, not all at once.

The benchmark tests four core memory competencies:

- **Conflict Resolution** -- facts change or contradict earlier ones (e.g. "Alice lives in Paris" then later "Alice moved to London"). The agent must use the most recent information. Includes single-hop (direct lookup) and multi-hop (derived reasoning) variants.
- **Test-Time Learning** -- the agent is given classification rules or patterns in the context (e.g. thousands of labeled examples from Banking77 or TREC) and must apply those rules to classify new, unseen inputs. This tests whether the agent can learn from context, not just retrieve.
- **Accurate Retrieval** -- straightforward factual recall from large contexts (6k to 262k tokens). Can the agent find a specific needle in a large haystack of information?
- **Long-Range Understanding** -- tasks that require synthesizing information spread across the entire context, like summarization or answering questions about a detective novel.

Each competency has multiple sub-datasets at different context sizes (6k, 32k, 64k, 128k, 262k tokens), sourced from the `ai-hyz/MemoryAgentBench` HuggingFace dataset.

## Implementation Summary

We adapted the benchmark to run against `deepagents` using our existing eval runner (`run_agent`). Each test works as follows:

1. **Data loading** -- a `BenchmarkSample` is fetched from HuggingFace containing a long context, a list of questions, and ground-truth answers.
2. **Chunking** -- the context is split into ~4096-token pieces using NLTK sentence tokenization and tiktoken token counting, so chunk boundaries respect sentence boundaries.
3. **Memorization phase** -- each chunk is sent to the agent as a separate conversation turn on the same `thread_id` (with `MemorySaver` checkpointer for state persistence). The agent processes and "memorizes" each chunk.
4. **Query phase** -- each question is sent as a new turn on the same thread. The agent's text response becomes the prediction.
5. **Scoring** -- each prediction is compared against ground-truth answer(s) using three normalized metrics:
   - **Exact Match** -- do the normalized strings match exactly?
   - **F1** -- token-level precision/recall overlap between prediction and ground truth.
   - **Substring Match** -- is the ground truth contained somewhere in the prediction?
   
   When multiple acceptable answers exist, the score is the max across all of them.
6. **Pass/fail** -- the test uses **F1 as the gate metric** and requires **>50% of questions** to score above zero. Graduated correctness (`passed / total`) is logged to LangSmith for trend tracking, while the binary pass/fail gates CI.


libs/deepagents/tests/evals/memory_agent_bench/data_utils.py
```diff
@@ -0,0 +1,194 @@
...
+            matches this value exactly (e.g. `factconsolidation_sh_6k`).
+        max_samples: Cap the number of returned samples.
+
+    Returns:
+        List of `BenchmarkSample` instances ready for evaluation.
+
+    Raises:
+        ValueError: If the split name is not recognized.
+    """
+    from datasets import load_dataset  # noqa: PLC0415
```
eyurtsev, line 69: could you float the imports to the global namespace?

libs/deepagents/tests/evals/memory_agent_bench/eval_utils.py
```diff
@@ -0,0 +1,133 @@
+"""Evaluation utilities for MemoryAgentBench integration.
```
eyurtsev, line 1: This is the scorer

libs/deepagents/tests/evals/memory_agent_bench/data_utils.py
```diff
@@ -0,0 +1,194 @@
...
+            if ok:
+                _NLTK_TOKENIZER_READY = True
+                break
+
+
+def _sent_tokenize(nltk_module: object, text: str) -> list[str]:
+    """Sentence-tokenize text with a robust fallback for constrained environments."""
+    try:
+        return list(nltk_module.sent_tokenize(text))  # type: ignore[reportUnknownMemberType,attr-defined]
+    except Exception:  # noqa: BLE001
```
eyurtsev: could we let the exception bubble so we don't get different results?

libs/deepagents/tests/evals/memory_agent_bench/data_utils.py
```diff
@@ -0,0 +1,194 @@
...
+        The value wrapped in a list, or the original list.
+    """
+    if isinstance(value, list):
+        return value
+    if value:
+        return [value]
+    return []
+
+
+def _ensure_nltk_tokenizer(nltk_module: object) -> None:
```
eyurtsev, line 165: let the exception bubble up?

libs/deepagents/tests/evals/memory_agent_bench/data_utils.py
```diff
@@ -0,0 +1,194 @@
...
+    limits per chunk.
+
+    Args:
+        text: The document to split.
+        chunk_size: Maximum number of tokens per chunk.
+
+    Returns:
+        List of text chunks.
+    """
+    import nltk  # noqa: PLC0415
```
eyurtsev, line 116: global namespace?

libs/deepagents/tests/evals/memory_agent_bench/data_utils.py
```diff
@@ -0,0 +1,194 @@
+"""Data loading utilities for MemoryAgentBench integration.
```
eyurtsev, line 1: This helps define the datasets that we're benchmarking

libs/deepagents/tests/evals/memory_agent_bench/test_memory_agent_bench.py
```diff
@@ -0,0 +1,289 @@
...
+    "Give a very short answer without extra explanation.\n\n"
+)
+
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+
+def _run_benchmark_sample(
```
eyurtsev, line 104: This is an implementation -- this part is identical to DeepAgentWrapper that we have in harbor

libs/deepagents/tests/evals/memory_agent_bench/test_memory_agent_bench.py
```diff
@@ -0,0 +1,289 @@
...
+        result = {
+            **metrics,
+            "prediction": prediction,
+            "answer": ground_truths,
+            "question": question,
+            "qa_pair_id": qa_pair_id,
+        }
+        results.append(result)
+
+        _log_feedback(key=f"q{idx}_exact_match", value=metrics["exact_match"])
```
eyurtsev: This is part of the scorer likely (i.e., it should be decoupled from the implementation)

libs/deepagents/tests/evals/memory_agent_bench/data_utils.py
maahir30: this is intentional. datasets, nltk, and tiktoken all live in the evals optional dependency group not in the core dependencies. they are only installed when someone explicitly opts in with that group for evals

libs/deepagents/tests/evals/memory_agent_bench/data_utils.py
maahir30: same reason as above

libs/deepagents/tests/evals/memory_agent_bench/test_memory_agent_bench.py
maahir30: right. this is a little specific to this benchmark because of the diff phases (1) memorization (2) query

PR #1834, opened 2026-03-12
chore(sdk): add evals `README`, ensure tracing

mdrxy:


libs/deepagents/tests/evals/conftest.py
```diff
@@ -17,6 +17,23 @@
...
+def pytest_configure(config: pytest.Config) -> None:  # noqa: ARG001  # pytest hook signature
+    """Fail fast if LangSmith tracing is not enabled.
+
+    All eval tests require `@pytest.mark.langsmith` and
+    `LANGSMITH_TRACING=true`. Detect this early so the entire suite is skipped
+    with a clear message instead of failing one-by-one.
+    """
+    if os.environ.get("LANGSMITH_TRACING", "").lower() != "true":
+        pytest.exit(
+            "Aborting: LANGSMITH_TRACING is not set to 'true'. "
```
eyurtsev, line 29: Instead should we change the decorator behavior so it's a noop with a warning?

mdrxy: The decorator-as-noop approach would emit the same warning once per test (noisy) and warnings are easily hidden by `-q`/`--disable-warnings`/CI log noise. `pytest.exit()` gives a single loud failure with a clear fix — you can't accidentally get a green suite without tracing enabled, which is the whole point.

PR #1836, opened 2026-03-12
feat(deepagents): update return type for backend.read

ccurme:
Part of 0.5 batch of changes.

**Flips `file_format` version flag introduced in https://github.com/langchain-ai/deepagents/pull/1398 to "v2".**

Defines `ReadResult` and updates backend read methods to return it:
```python
@dataclass
class ReadResult:
    """Result from backend read operations.
    Attributes:
        error: Error message on failure, None on success.
        file_data: FileData dict on success, None on failure.
    """

    error: str | None = None
    file_data: FileData | None = None
```

Updates `(a)read_file` tools to generate standard content blocks from base64 data, using file extension to determine MIME type. Enables multi-modal reads:

Screenshot 2026-03-11 at 12 58 02 PM

Also moves the line number formatting out of backends and into FileSystemMiddleware tools.


libs/deepagents/deepagents/backends/composite.py
```diff
@@ -235,7 +236,7 @@ def read(
             limit: Maximum number of lines to read.
 
         Returns:
-            Formatted file content with line numbers, or error message.
+            ReadResult
```
sydney-runkle, line 239: could use improved docs

libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -304,34 +308,40 @@ def read(
...
 
             if start_idx >= len(lines):
-                return f"Error: Line offset {offset} exceeds file length ({len(lines)} lines)"
+                return self._version_read_result(ReadResult(error=f"Line offset {offset} exceeds file length ({len(lines)} lines)"))
 
             selected_lines = lines[start_idx:end_idx]
-            return format_content_with_line_numbers(selected_lines, start_line=start_idx + 1)
-        except (OSError, UnicodeDecodeError) as e:
-            return f"Error reading file '{file_path}': {e}"
+            formatted = format_content_with_line_numbers(selected_lines, start_line=start_idx + 1)
```
sydney-runkle: This should move to the `read_file` level right? do we want to do this in a diff PR?
I think should be associated w/ this refactor

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -137,6 +137,19 @@ class FileData(TypedDict):
...
+@dataclass
+class ReadResult:
+    """Result from backend read operations.
+
+    Attributes:
+        error: Error message on failure, None on success.
+        file_data: FileData dict on success, None on failure.
+    """
+
+    error: str | None = None
```
sydney-runkle, line 149: Should errors be exceptions that we can then stringify?

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -211,6 +224,32 @@ class BackendProtocol(abc.ABC):  # noqa: B024
...
+
+        When `file_format="v1"`, returns the formatted content string (or
+        error string) cast to `ReadResult`. The type annotation is wrong on
+        purpose — v1 callers expect `str`.
+
+        When `file_format="v2"`, returns the ReadResult unchanged.
+        """
+        if self.file_format != "v1":
+            return result
+        if result.error is not None:
```
eyurtsev: I'd add an inline doc above the type ignore that explains why we're doing a type ignore

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -211,6 +224,32 @@ class BackendProtocol(abc.ABC):  # noqa: B024
...
+    file_format: FileFormat = "v2"
+    """Storage and read format version.
+
+    - `"v1"`: Legacy — `read()` returns a plain `str` (formatted text with
+      line numbers). Storage uses `list[str]` content without `encoding`.
+    - `"v2"`: Current — `read()` returns a `ReadResult` dataclass. Storage
+      uses plain `str` content with an `encoding` field.
+    """
+
+    def _version_read_result(self, result: ReadResult) -> ReadResult:
```
sydney-runkle: Would this be more simple if it was the responsibility of the caller to handle `str` | `ReadResult`, not the burden of the backend to produce either or? 

So all of our file tools would have these if conditions, rather than the backend themselves?

libs/deepagents/deepagents/backends/state.py
```diff
@@ -45,26 +47,27 @@ def __init__(
         self,
         runtime: "ToolRuntime",
         *,
-        file_format: FileFormat = "v1",
+        file_format: FileFormat = "v2",
```
eyurtsev, line 50: are we already changing the default?

eyurtsev (commented):
should we default to v1 to make sure everything works and only then flip to v2?

libs/deepagents/deepagents/backends/utils.py
```diff
@@ -18,6 +18,49 @@
...
+    ".ppt": "file",
+    ".pptx": "file",
+}
+"""Extension-to-type mapping for non-text files.
+
+Derived from Google's multimodal API supported formats:
+
+- Images: https://ai.google.dev/gemini-api/docs/image-understanding
+- Video: https://ai.google.dev/gemini-api/docs/video-understanding
+- Audio: https://ai.google.dev/gemini-api/docs/audio
```
sydney-runkle, line 61: shouldn't this be derived from the multimodal content blocks we support?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -548,6 +540,43 @@ def _create_read_file_tool(self) -> BaseTool:  # noqa: C901
...
+            lines = content.splitlines(keepends=True)
+            if len(lines) > limit:
+                lines = lines[:limit]
+                content = "".join(lines)
+
+            if token_limit and len(content) >= NUM_CHARS_PER_TOKEN * token_limit:
+                truncation_msg = READ_FILE_TRUNCATION_MSG.format(file_path=validated_path)
+                max_content_length = NUM_CHARS_PER_TOKEN * token_limit - len(truncation_msg)
+                content = content[:max_content_length]
+                content += truncation_msg
```
sydney-runkle, line 589: I think the "formatting w/ line numbers" should be here, if we want to keep it

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -610,41 +606,8 @@ async def async_read_file(
...
-            if token_limit and len(result) >= NUM_CHARS_PER_TOKEN * token_limit:
-                # Calculate truncation message length to ensure final result stays under threshold
-                truncation_msg = READ_FILE_TRUNCATION_MSG.format(file_path=validated_path)
-                max_content_length = NUM_CHARS_PER_TOKEN * token_limit - len(truncation_msg)
-                result = result[:max_content_length]
-                result += truncation_msg
-
-            return result
+            read_result = await resolved_backend.aread(validated_path, offset=offset, limit=limit)
+            return _handle_read_result(read_result, validated_path, runtime.tool_call_id, limit)
```
sydney-runkle: doesn't this only handle `ReadResult`? It needs to handle `str | ReadResult`?

sydney-runkle (commented):
Main question in my mind is can we go ahead and have **all** of our backends return `ReadResult`?

I think at runtime we can just support handling `str | ReadResult` at the tool level.

We should definitely have deprecation warnings for any returns of `str` with an expected removal timeline (e.g. 0.6)

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -548,6 +540,43 @@ def _create_read_file_tool(self) -> BaseTool:  # noqa: C901
...
+                    content_blocks=cast("list[ContentBlock]", [{"type": file_type, "base64": content, "mime_type": mime_type}]),
+                    name="read_file",
+                    tool_call_id=tool_call_id,
+                    additional_kwargs={"read_file_path": validated_path, "read_file_media_type": mime_type},
+                )
+
+            lines = content.splitlines(keepends=True)
+            if len(lines) > limit:
+                lines = lines[:limit]
+                content = "".join(lines)
```
sydney-runkle, line 583: this should be handled in `read` implementations bc of `limit` passed to that function

libs/deepagents/deepagents/backends/filesystem.py
colifran: I agree. I think this is a presentation level concern leaking into storage level concerns.

libs/deepagents/deepagents/backends/state.py
colifran: I personally think a v2 default is the right long term choice.

libs/deepagents/deepagents/backends/protocol.py
sydney-runkle: @eyurtsev had an opinion here, I can't remember why he was team `str`, just want to clarify
I'm team `Exception` in the short term bc then we're bubbling more info up that could be surfaced / custom handled

libs/deepagents/deepagents/backends/state.py
sydney-runkle: Agreed we should change in V0.5

libs/deepagents/deepagents/backends/protocol.py
eyurtsev: what kind of handling do you expect to do for an exception?

libs/deepagents/deepagents/backends/state.py
eyurtsev: will we test to make sure everything passes with v1 for backwards compatibility?

libs/deepagents/deepagents/backends/protocol.py
sydney-runkle: actually raise instead of return to the model? interrupt?

libs/deepagents/deepagents/backends/utils.py
ccurme: we don't maintain a mapping of mime types to content block types. google has the most thorough one, so I took those. the comment is basically there for posterity to see where these came from.

sydney-runkle: excellent

libs/deepagents/deepagents/backends/protocol.py
sydney-runkle: let's leave as str for now

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -548,6 +540,45 @@ def _create_read_file_tool(self) -> BaseTool:  # noqa: C901
...
+            tool_call_id: str | None,
+            limit: int,
+        ) -> ToolMessage | str:
+            if isinstance(read_result, str):
+                return read_result
+            if read_result.error:
+                return f"Error: {read_result.error}"
+
+            if read_result.file_data is None:
+                return f"Error: no data returned for '{validated_path}'"
```
sydney-runkle, line 564: ToolMessage with `status='error'`?

libs/deepagents/deepagents/backends/state.py
ccurme: I've flipped to "v2" and parametrized tests to test both versions.

libs/deepagents/deepagents/backends/filesystem.py
ccurme: Moved this to the filesystem middleware.

libs/deepagents/deepagents/middleware/filesystem.py
ccurme: done

libs/deepagents/deepagents/middleware/filesystem.py
ccurme: I think we need to limit twice (as we were doing before):
1. backend.read applies offset / limit to raw content
2. we format content, adding line numbers and continuation lines
3. because of the continuation lines, the result can have (many) more lines than specified by the tool. enforce the line limit here as well.

libs/deepagents/deepagents/backends/utils.py
```diff
@@ -200,13 +256,51 @@ def update_file_data(file_data: FileData, content: str) -> FileData:
...
+    empty_msg = check_empty_content(content)
+    if empty_msg:
+        return empty_msg
+
+    lines = content.splitlines()
+    start_idx = offset
+    end_idx = min(start_idx + limit, len(lines))
+
+    if start_idx >= len(lines):
+        return f"Error: Line offset {offset} exceeds file length ({len(lines)} lines)"
```
sydney-runkle: this should be surfaced on an error on `ReadResult` instead of swallowed into file content

libs/deepagents/deepagents/backends/utils.py
```diff
@@ -200,13 +256,51 @@ def update_file_data(file_data: FileData, content: str) -> FileData:
...
     file_data: FileData,
     offset: int,
     limit: int,
 ) -> str:
     """Format file data for read response with line numbers.
 
+    .. deprecated::
+        Use `slice_read_response` and apply
+        `format_content_with_line_numbers` separately.
+
```
sydney-runkle, line 304: let's actually deprecate this (w/ a decorator)

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -548,6 +541,55 @@ def _create_read_file_tool(self) -> BaseTool:  # noqa: C901
...
+            limit: int,
+        ) -> ToolMessage | str:
+            if isinstance(read_result, str):
+                warnings.warn(
+                    "Returning a plain `str` from `backend.read()` is deprecated. "
+                    "Return a `ReadResult` instead. Returning `str` will not be "
+                    "supported in a future version.",
+                    DeprecationWarning,
+                    stacklevel=2,
+                )
```
sydney-runkle, line 558: Let's say in 0.6

libs/deepagents/deepagents/middleware/filesystem.py
sydney-runkle: this needs a comment explaining why we do it here

sydney-runkle (approved):
Approving w/ the caveat that
* I'd like to do a full review of the v0.5 branch and add more robust deprecations
* I think we might want to consider `Exception` on the result types
* We should wrap all of our errors in `ToolMessage` with `status='error'`

PR #1870, opened 2026-03-13
feat(sdk): update backend return types for ls, glob, grep

ccurme:


eyurtsev (approved):

libs/deepagents/deepagents/backends/utils.py
```diff
@@ -664,19 +664,19 @@ def grep_matches_from_files(
...
-    Returns a list of GrepMatch on success, or a string for invalid inputs.
+    Returns a GrepResult with matches on success.
     We deliberately do not raise here to keep backends non-throwing in tool
     contexts and preserve user-facing error messages.
     """
     try:
         normalized_path = _normalize_path(path)
     except ValueError:
-        return []
+        return GrepResult(matches=[])
```
eyurtsev, line 679: Should this be an error?

eyurtsev: cc @ccurme could you scan the code base for any other occurrences like this where we should actually be returning an error message?

PR #1873, opened 2026-03-13
fix(sdk): simplify memory agent bench evals

maahir30:
- Add `max_questions` to `DatasetConfig` to cap questions per sample; CI configs use 10 instead of the full ~100 from the dataset.
- Replace F1 and exact-match scoring with substring match only. Correctness is all-or-nothing (1.0 if every question passes).
- Simplify LangSmith feedback to three columns: `questions_passed`, `questions_total`, `correctness`. Remove per-question columns and redundant token stats.
- Remove unused `AgentTrajectory` tracking from the benchmark runner.


eyurtsev:
Look good @maahir30 thank you!

PR #1901, opened 2026-03-16
fix(cli): use counter to close history-recall autocomplete race

### LINKED ISSUE #1628: History recall does not fully suppress `@` file completions
## Description

History recall (`Up` arrow) is supposed to suppress `@` file autocomplete suggestions, but a race condition allows them to appear.

## Root cause

In `ChatTextArea.set_text_from_history()` (`libs/cli/deepagents_cli/widgets/chat_input.py:593-605`):

```python
self._navigating_history = True
self.text = text          # fires on_text_changed
# ...
self._navigating_history = False  # reset before handler runs
```

The guard in `ChatInput._on_text_changed` (line 821) checks `_navigating_history`, but by the time the handler fires asynchronously, the flag is already `False`. Completions slip through.

## Why it was hidden

No non-dotfile `.json` in the repo previously fuzzy-matched `package.json` above `_MIN_FUZZY_SCORE = 15`. Adding `compact_tool_seed_messages.json` in #1627 scored 15.3 and exposed it. Worked around by renaming the fixture, but the underlying race remains.

## Repro

1. Launch the CLI in a git repo with non-dotfile `.json` files tracked
2. Submit a message containing an `@` file mention, e.g. `tell me about @package.json`
3. Press `Up` to recall that history entry
4. File completion suggestions may appear when they shouldn't

Or run the unit test with a git-tracked file that fuzzy-matches a recalled `@`-mention above score 15 — `_current_suggestions` will be non-empty after recall.

## Suggested fix

Keep `_navigating_history = True` until after the text change handler has fully processed, or use a separate guard that isn't reset synchronously (e.g., skip the next `on_text_changed` via a counter).

mohitejaikumar:
Fixes #1628

History recall (`Up` arrow) was supposed to suppress autocomplete suggestions, but a race condition allowed `@`-file completions to slip through. The boolean `_navigating_history` flag was reset synchronously in `set_text_from_history` *before* the async `TextArea.Changed` handler ran, so the guard was already `False` by the time `on_text_area_changed` checked it.

## Changes

- Replace the boolean `_navigating_history` flag with an integer counter `_skip_history_change_events` on `ChatTextArea`. The producer (`set_text_from_history`) increments before setting text; the consumer (`on_text_area_changed`) decrements when processing the resulting event. This eliminates the race because the guard isn't cleared until the handler actually fires — unlike the boolean, which was reset synchronously before the async handler ran.
- Remove the synchronous `_navigating_history = False` reset at the end of `set_text_from_history`; the handler now owns the decrement, and `clear_text` resets the counter to 0
- Add a defensive check for negative counter values with a warning log to catch any future bookkeeping bugs

### DIFF
--- a/libs/cli/deepagents_cli/widgets/chat_input.py
+++ b/libs/cli/deepagents_cli/widgets/chat_input.py
@@ -586,7 +586,6 @@ class ChatTextArea(TextArea):
             if navigate:
                 event.prevent_default()
                 event.stop()
-                self._skip_history_change_events += 1
                 if event.key == "up":
                     self.post_message(self.HistoryPrevious(self.text))
                 else:
@@ -691,6 +690,7 @@ class ChatTextArea(TextArea):
     def clear_text(self) -> None:
         """Clear the text area."""
         self._in_history = False
+        self._skip_history_change_events = 0
         self._paste_burst_buffer = ""
         self._paste_burst_last_char_time = None
         self._cancel_paste_burst_timer()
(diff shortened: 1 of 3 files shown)


mohitejaikumar:
@mdrxy Thanks for the changes.

PR #1937, opened 2026-03-17
fix(sdk): timeout python grep fallback

mdrxy:
Stopgap whilst waiting for #1935

---

> [!WARNING]
> **Behavior change to the `grep` tool.** This PR alters grep semantics in two ways:
>
> 1. **The Python fallback can no longer hang indefinitely.** Previously only the ripgrep path enforced a 30s timeout; the pure-Python fallback (used when `rg` is not on `PATH`) walked the tree unbounded. It now aborts after `DEFAULT_GREP_TIMEOUT` (30s) and returns the matches collected so far plus a partial-error message.
> 2. **Partial matches are now surfaced instead of discarded.** Previously any backend error caused the tool to return only the error string with `status="error"`, throwing away any matches already found. The tool now returns those partial matches alongside the error (`status="error"`, content includes a `Partial matches:` section). The clean, fully-successful path is unchanged (`status="success"`, identical content).

`FilesystemBackend._python_search` had no timeout — on large repos, if ripgrep timed out or wasn't installed, the Python fallback would walk the entire file tree via `rglob("*")` indefinitely, blocking the calling thread forever. This was observed in production: a subagent's grep tool call hung permanently, stalling an entire ACP session.


libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -531,26 +553,45 @@ def _ripgrep_search(self, pattern: str, base_full: Path, include_glob: str | Non
...
         """
-        # Compile escaped pattern once for efficiency (used in loop)
+        deadline = time.monotonic() + timeout
         regex = re.compile(pattern)
 
         results: dict[str, list[tuple[int, str]]] = {}
         root = base_full if base_full.is_dir() else base_full.parent
 
         for fp in root.rglob("*"):
+            if time.monotonic() > deadline:
```
eyurtsev: How fast is time.monotonic() vs. a single loop iteration?

mdrxy: Measured it: `time.monotonic()` is ~31 ns/call here. A single iteration of that loop is dominated by syscalls — `fp.is_file()` alone is a `stat` (~1–5 µs), and matching iterations add `fp.stat()` for the size check plus `fp.read_text()` (open + read). So one iteration costs at least 1–5 µs, two-to-three orders of magnitude more than the deadline check.

The check is also per-file (once per `rglob` entry), not per-line, so it's in the noise — no need to amortize it across N iterations. Per-line work stays bounded because each file is capped at `max_file_size_bytes`.

PR #1938, opened 2026-03-17
fix(sdk): pass along check checkpoint for reconciliation

christian-bromann:
Ports [langchain-ai/deepagentsjs#312](https://github.com/langchain-ai/deepagentsjs/pull/312) to Python.

## Why

When LangGraph executes the `task` tool, it injects a UUID-based `checkpoint_ns` (e.g. `"tools:a1b2c3d4-..."`) into the tool's `RunnableConfig`. This namespace identifies where the subagent's checkpoints are written. Without forwarding it to the resulting `ToolMessage`, the LangGraph SDK has no way to reconstruct a subagent's message history after a page reload — it only has the LLM provider's tool call ID, which doesn't map to a checkpoint namespace.

By storing `subgraph_checkpoint_ns` in `ToolMessage.additional_kwargs`, the SDK can query `getHistory` with the correct namespace.

## Review notes

- `runtime.config.get("configurable") or {}` is used (rather than `runtime.config.get("configurable", {})`) to safely handle a `None` value for `configurable` in addition to a missing key — this matches the defensive style of the JS `config?.configurable?.checkpoint_ns` optional chaining.
- The `additional_kwargs` key name (`subgraph_checkpoint_ns`) must stay in sync with the JS SDK, as the CLI reads this field to restore subagent history.

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -399,7 +399,11 @@ def _build_task_tool(  # noqa: C901
     else:
         description = task_description
 
-    def _return_command_with_state_update(result: dict, tool_call_id: str) -> Command:
+    def _return_command_with_state_update(
+        result: dict,
+        tool_call_id: str,
+        subgraph_checkpoint_ns: str | None = None,
+    ) -> Command:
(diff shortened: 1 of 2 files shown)


christian-bromann:
Closing due to incorrect approach.

PR #1951, opened 2026-03-17
feat(cli): truncate `write_file` and `edit_file` approval content to 10 lines with expand hotkey

jkennedyvz:
Truncates the file content and diff displayed in the `write_file` and `edit_file` HITL approval widgets to 10 lines by default. Users can press `e` to expand the full content, matching the existing behavior for long shell commands.


mdrxy:
Should just do for all tools

Screenshot 5

PR #1980, opened 2026-03-17
ci(cli): add concurrency control to deepagents workflow

mdrxy:
Add concurrency control to the `@deepagents` agent workflow. Without this, duplicate comment triggers pile up instead of cancelling the superseded run.


mdrxy:
cc @jkennedyvz fyi

PR #1973, opened 2026-03-17
fix(examples): replace `FilesystemBackend` with `upload_files` in nvidia example

mdrxy:
Closes #1862

---

Remove `FilesystemBackend` and `CompositeBackend` from the nvidia deep agent example. The old pattern routed skill/memory reads to the local filesystem, which isn't representative of production — skills don't originate from the host machine in a real deployment. The backend now uploads skill and memory files directly into the Modal sandbox on creation via `upload_files`, so the agent operates entirely within the sandbox.

PR #1982, opened 2026-03-17
fix(examples): swap nvidia example from FilesystemBackend to upload_files

vtrivedy:
Removes CompositeBackend/FilesystemBackend, uploads memory and skills into the Modal sandbox via upload_files instead.


mdrxy:
#1973

PR #1983, opened 2026-03-17
feat(sdk): port `LangSmithSandbox` from CLI to SDK

ccurme:
Move `LangSmithBackend` from the CLI into the SDK as `LangSmithSandbox`, making it available to any consumer of the `deepagents` package. The CLI-specific provider/lifecycle logic (`_LangSmithProvider`) moves into `sandbox_factory.py` alongside the other provider implementations, removing the standalone `langsmith.py` module.


libs/deepagents/pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
 ]
 dependencies = [
     "langchain-core>=1.2.19,<2.0.0",
+    "langsmith>=0.3.0",
```
ccurme, line 24: sandbox appears [added in 0.6.5](https://github.com/langchain-ai/langsmith-sdk/releases/tag/v0.6.5), not sure if we want to restrict the `langsmith` version for all usage of deepagents by that much though.

libs/cli/deepagents_cli/integrations/langsmith.py
```diff
@@ -1,4 +1,4 @@
-"""LangSmith sandbox backend implementation."""
+"""LangSmith sandbox provider for the CLI."""
```
mdrxy: We shouldn't need this file anymore, its contents can be squashed into the factory

ccurme: Done, isolated changes in 0fb7fb9fea5f703bddd7f2c65e4f35a173c59a0b.

mdrxy (approved):

PR #1993, opened 2026-03-17
test(sdk): gather `MemoryAgentBench` CI subset

maahir30:
## Summary

- Replaced the ad-hoc 3-config CI subset with a deliberate 8-config selection (2 per benchmark category), biased toward the hardest and most structurally distinct variants from the MemoryAgentBench benchmark (ICLR 2026).
- Capped all configs at 25 questions (up from unlimited for main configs, 10 for CI) as a cost/signal tradeoff -- each question replays the full conversation history, so cost grows super-linearly with question count.
- Added a guard unit test that locks the exact CI config list, category balance, and question cap so it can't drift silently.

## Dataset selection rationale

The benchmark has 23 configs across 4 memory competencies. Most configs within a category are variants of the same task at different context lengths or label counts, so running all of them yields diminishing signal. We picked 2 per category optimizing for:

1. **Maximum difficulty** — harder variants give better hill-climbing signal for regressions.
2. **Task diversity** — within each category, the two picks test structurally different skills rather than being scale variants of the same test.
3. **Cost awareness** — avoided the largest context lengths where the same task is already unsolved at shorter lengths.

| Category | Config | Why this one |
|---|---|---|
| Conflict Resolution | `CR_SH_64K` | Single-hop at 64K tokens. ~60% accuracy range in the paper gives a measurable gradient for improvements. |
| Conflict Resolution | `CR_MH_6K` | Multi-hop is near-zero accuracy even at 6K (the cheapest length). Serves as an aspirational canary without burning tokens at 262K. |
| Test-Time Learning | `TTL_CLINIC150` | Hardest multi-class classification (151 labels). Subsumes signal from BANKING77/NLU/TREC variants. |
| Test-Time Learning | `TTL_RECSYS` | Only non-classification TTL task (movie recommendation from dialogue). Structurally irreplaceable. |
| Accurate Retrieval | `AR_RULER_QA2` | Multi-hop document QA (harder than single-hop QA1). |
| Accurate Retrieval | `AR_EVENTQA_FULL` | Only temporal-reasoning retrieval task (novel event chains). Different skill from RULER/LongMemEval. |
| Long Range Understanding | `LRU_INFBENCH_SUM` | Novel summarization — tests global comprehension (generation). |
| Long Range Understanding | `LRU_DETECTIVE_QA` | Long-range reasoning QA over detective novels. Different from summarization. |

Dropped configs are either easier variants of a kept config (e.g., `CR_SH_6K` is strictly easier than `CR_SH_64K`), structurally redundant (e.g., all 5 MCC classification datasets test the same skill at different label counts), or more expensive versions of an already-unsolved task (e.g., `CR_MH_262K` when `CR_MH_6K` is already ~0%).

### DIFF
--- a/libs/deepagents/tests/evals/memory_agent_bench/configs.py
+++ b/libs/deepagents/tests/evals/memory_agent_bench/configs.py
@@ -37,21 +37,25 @@ CR_SH_6K = DatasetConfig(
     split="Conflict_Resolution",
     source="factconsolidation_sh_6k",
     max_samples=1,
+    max_questions=25,
 )
 CR_SH_32K = DatasetConfig(
     split="Conflict_Resolution",
     source="factconsolidation_sh_32k",
     max_samples=1,
+    max_questions=25,
 )
 CR_SH_64K = DatasetConfig(
     split="Conflict_Resolution",
     source="factconsolidation_sh_64k",
     max_samples=1,
+    max_questions=25,
 )
 CR_SH_262K = DatasetConfig(
     split="Conflict_Resolution",
     source="factconsolidation_sh_262k",
     max_samples=1,
+    max_questions=25,
 )
 
 # -- Conflict Resolution (multi-hop) -----------------------------------------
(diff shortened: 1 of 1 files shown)


starkelizabeth433-sys:
**Kind of an amazing project🤝**

mdrxy (approved):

PR #2015, opened 2026-03-18
fix(sdk): support bare store keys in root listing

### LINKED ISSUE #1655: `ls_info()` in backend fails to list direct files when operating in the root directory
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

_No response_

```python
from langgraph.runtime import Runtime


class MockRuntime:
    def __init__(self, store):
        self.store = store
        self.config = {"metadata": {"assistant_id": "test-assistant"}}

class MockStore:
    def search(self, namespace, query=None, filter=None, limit=100, offset=0):
        
        from langgraph.store.base import Item
        
        # 根目录下的文件
        item1 = Item(
            key="test.md",
            value={
                "content": ["# Test File"],
                "created_at": "2023-01-01T00:00:00Z",
                "modified_at": "2023-01-01T00:00:00Z"
            }
        )
        
        
        item2 = Item(
            key="subdir/file.txt",
            value={
                "content": ["File in subdirectory"],
                "created_at": "2023-01-01T00:00:00Z",
                "modified_at": "2023-01-01T00:00:00Z"
            }
        )
        
        return [item1, item2]


mock_store = MockStore()
mock_runtime = MockRuntime(mock_store)


backend = StoreBackend(runtime=mock_runtime)

#  ls_info("/")
print("Testing ls_info('/'):")
result = backend.ls_info("/")
for item in result:
    print(f"  - {item['path']} (is_dir: {item.get('is_dir', False)})")

#  "test.md" out
print("\nExpected:")
print("  - test.md (is_dir: False)")
print("  - subdir/ (is_dir: True)")
```

```shell

```

## Problem Description
In the deepagents/backends/store.py file, the ls_info method fails to list files directly in the root directory ("/") such as "test.md". This is because the current filtering logic requires file paths to start with the normalized path ("/"), but direct files like "test.md" don't meet this condition.

## Reproduction Steps
1. Create a file directly in the root directory, e.g., "test.md"
2. Call the ls_info("/") method
3. Observe the returned result, noting that "test.md" is not in the result list
## Expected Behavior
ls_info("/") should return all files and directories in the root directory, including files directly in the root directory (like "test.md").

## Actual Behavior
ls_info("/") only returns subdirectories in the root directory, not files directly in the root directory.

## Root Cause
In the ls_info method, the filtering logic at lines 314-316 is:

### Environment / System Info

_No response_

LRRuan:
Fixes #1655

Normalize bare store keys inside `StoreBackend.ls()` so root listings include direct files even when the underlying store omits a leading slash. Added a focused regression test covering bare root keys and preserving nested directory behavior.

How I verified this works:
- `PATH="$PWD/.condaenv/bin:$PATH" PYTHONPATH="$PWD/libs/deepagents:$PYTHONPATH" ./.condaenv/bin/python -m pytest libs/deepagents/tests/unit_tests/backends/test_store_backend.py -q`

AI assistance disclaimer: I used AI assistance to help draft and implement this small patch, then reviewed the resulting changes and test coverage before opening the PR.

### DIFF
--- a/libs/deepagents/deepagents/backends/store.py
+++ b/libs/deepagents/deepagents/backends/store.py
@@ -352,12 +352,16 @@ class StoreBackend(BackendProtocol):
         normalized_path = path if path.endswith("/") else path + "/"
 
         for item in items:
+            normalized_key = str(item.key)
+            if not normalized_key.startswith("/"):
+                normalized_key = "/" + normalized_key
+
             # Check if file is in the specified directory or a subdirectory
-            if not str(item.key).startswith(normalized_path):
+            if not normalized_key.startswith(normalized_path):
                 continue
 
             # Get the relative path after the directory
-            relative = str(item.key)[len(normalized_path) :]
+            relative = normalized_key[len(normalized_path) :]
 
             # If relative path contains '/', it's in a subdirectory
             if "/" in relative:
@@ -376,7 +380,7 @@ class StoreBackend(BackendProtocol):
             size = len("\n".join(raw)) if isinstance(raw, list) else len(raw)
             infos.append(
                 {
-                    "path": item.key,
+                    "path": normalized_key,
                     "is_dir": False,
                     "size": int(size),
                     "modified_at": fd.get("modified_at", ""),
--- a/libs/deepagents/tests/unit_tests/backends/test_store_backend.py
+++ b/libs/deepagents/tests/unit_tests/backends/test_store_backend.py
@@ -128,6 +128,36 @@ def test_store_backend_ls_trailing_slash():
     assert [fi["path"] for fi in listing1] == [fi["path"] for fi in listing2]
 
 
+def test_store_backend_ls_root_handles_bare_store_keys() -> None:
+    rt = make_runtime()
+    be = StoreBackend(rt, namespace=lambda _ctx: ("filesystem",))
+
+    bare_items = {
+        "test.md": "root file",
+        "notes/todo.md": "nested file",
+    }
+
+    for key, content in bare_items.items():
+        rt.store.put(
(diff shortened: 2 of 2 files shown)


lonexreb (commented):
Clean fix for a real edge case. When the store implementation returns bare keys (no leading `/`), the prefix matching in `ls()` silently drops those entries — so files exist but are invisible.

The normalization approach is correct: prepend `/` if missing, then use the normalized key consistently through the rest of the loop.

One minor thought: the `als()` method (async version) likely has the same bare-key issue. I checked `store.py` and it has a separate `async def als()` implementation — might be worth applying the same normalization there for consistency. Not a blocker for this PR though.

Test is well-written — covers both root-level files and nested directory grouping with bare keys.

LGTM.

mdrxy:
Thanks for picking this up — much cleaner approach than #1657, and the normalize-the-key strategy doesn't break the existing root-listing behavior the way that PR did.

That said, I'm not convinced this is solving a real bug. The reproducer in #1655 (and the new test here) only triggers it by bypassing `write()` and `store.put`-ing a bare key directly. Every code path that goes through `StoreBackend.write()` / `awrite()` passes the caller's absolute path straight through to `store.put()`, so item keys produced by deepagents always start with `/`. There's no public API that lands you in the broken state.

So this is defensive normalization for non-conforming stores rather than a user-facing fix. Not opposed to landing it on those grounds, but I'd want the PR description and the test docstring to be honest about that — otherwise it reads as fixing a bug that isn't reachable through the SDK's own surface. Curious if anyone has actually hit this in the wild via a real `BaseStore` implementation.

PR #2042, opened 2026-03-19
fix(sdk): filesystem prompt

vanner-young:
Hello author,

Today, while using the `LocalShellBackend` module within the backend component, I encountered an issue involving recursive loops triggered by the tool the model uses to execute commands.

Root Cause: Simply put, the issue arises because the model does not know whether the path passed during command execution should be a relative path or an absolute path.

Here is a demo:
```python
from pathlib import Path
from deepagents.backends import LocalShellBackend
from langgraph.graph.state import CompiledStateGraph

from model.Ark import ArkModel

class DeepBackend:
    agent: CompiledStateGraph
    workspace_path: str = Path(Path.cwd(), 'deep_agent_workspace')
    
    def created_agents(self, **kwargs):
        self.agent = create_deep_agent(
        model=ArkModel().model,
        **kwargs
        )
        return self.agent
    
    @property
    def local_shell(self):
        return LocalShellBackend(
            root_dir=str(self.workspace_path),
            inherit_env=True,
            virtual_mode=True
        )
        
    def start(self):
        agent = self.created_agents(
        backend = self.local_shell
    )
        result = agent.invoke({ 'messages': [{ 'role': 'user', 'content': 'Execute scripts/index.js' }] })
        print(result['messages'][-1].content)

DeepBackend().start()

```

Code Run result:

Because the filesystem prompt is not strictly defined, when executing the local_shell module, the execute function will directly retrieve the root_dir from FilesystemBackend in the self.cwd field.

image

Execution Results:

Due to a lack of rigor in the `filesystem` prompt design, when execution reached the `local_shell` module, the `self.cwd` variable within the `execute` function directly retrieved the `root_dir` value from the `FilesystemBackend`.

Consequently, the command passed by the Large Language Model (LLM) became: `node /scripts/index.js`.
As a result, the `subprocess.run` call consistently failed. Subsequently, the LLM entered a continuous retry loop,
executing the following command:

```sh
## Since /scripts is not the root directory, this command continued to fail.
cd /scripts/index.js && node index.js
...
```
There was an additional issue here: the `self.cwd` variable consistently retrieved the value of the `root_dir` parameter.
And so, this process looped *n* times, leading to a continuous depletion of my tokens.

Therefore, this Pull Request (PR) aims to resolve this specific issue.

### DIFF
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -982,7 +982,10 @@ class FilesystemMiddleware(AgentMiddleware[FilesystemState, ContextT, ResponseT]
         tool_description = self._custom_tool_descriptions.get("execute") or EXECUTE_TOOL_DESCRIPTION
 
         def sync_execute(  # noqa: PLR0911 - early returns for distinct error conditions
-            command: Annotated[str, "Shell command to execute in the sandbox environment."],
+            command: Annotated[
+                str,
+                "Shell command to execute in the sandbox environment. If the command includes a path parameter, please prefer using use relative path, such as 'xx/xxx'.",
+            ],
             runtime: ToolRuntime[None, FilesystemState],
             timeout: Annotated[
                 int | None,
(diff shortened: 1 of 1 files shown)


eyurtsev:
@vanner-young i believe that the new annotation isn't much more helpful -- you could try to improve the annotation with another pass potentially? (i.e., I would find the new annotation to be difficult to understand if I were an LLM)

Have you looked how we set up the LocalContextMiddleware in the CLI? (it has a lot of additional information in the prompt that's added at runtime to actually help with coding related tasks?)

PR #2117, opened 2026-03-20
fix(sdk): use file transfer instead of command strings for sandbox write/edit

mdrxy:
Sandbox `write()` and `edit()` embedded arbitrarily large base64-encoded payloads directly in shell command strings, which hit OS `ARG_MAX` limits when executed through `docker exec` or similar sandbox backends — surfacing as `OSError: [Errno 7] Argument list too long` during eval runs with large conversation histories. The fix separates validation (small shell command) from data transfer (`upload_files()`/`download_files()`), so file content never touches the command line.

## Changes

- `BaseSandbox.write()` now runs a small existence-check via `_WRITE_CHECK_TEMPLATE` + `execute()`, then transfers content through `upload_files()` — replacing the old `_WRITE_COMMAND_TEMPLATE` that embedded the full base64 payload in the command string
- `BaseSandbox.edit()` splits into two paths based on `_EDIT_INLINE_MAX_BYTES` (50KB): small payloads use server-side `_EDIT_COMMAND_TEMPLATE` via `execute()` (single round-trip); large payloads upload old/new strings as temp files via `upload_files()` and run `_EDIT_TMPFILE_TEMPLATE` server-side — the source file never leaves the sandbox
- `BaseSandbox.read()` switches from heredoc-stdin payload to direct string interpolation of base64-encoded path + pagination params in `_READ_COMMAND_TEMPLATE` — server-side pagination preserved, only the requested page flows over the wire
- `HarborSandbox.awrite()` and `aedit()` use Harbor's native `environment.upload_file()`/`download_file()` instead of shell commands, with new `aupload_files()`/`adownload_files()` implementations
- Add `map_file_operation_error()` in `protocol.py` to centralize exception-to-`FileOperationError` mapping — replaces duplicated `except` chains in `FilesystemMiddleware` and allows `HarborSandbox` to normalize provider errors
- Widen `FileDownloadResponse.error` and `FileUploadResponse.error` from `FileOperationError | None` to `FileTransferError | None`, allowing backends to pass through provider-specific error strings for partial-success batched transfers
- Remove stale `parent_not_found` error code from `FileOperationError` docs and integration test assertions

## Breaking change risk

**Not breaking for public API consumers** — `write()`, `edit()`, `read()`, `awrite()`, `aedit()` signatures and return types are unchanged.

**Not breaking for `BaseSandbox` subclasses** — `upload_files()` and `download_files()` now have execute-based default implementations, so subclasses that only implement `execute()` continue to work. All in-tree subclasses already override both methods with native transfer APIs.

---

Related: #1402


libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -48,21 +48,55 @@
 """
 
 
+def map_file_operation_error(exc: Exception) -> FileOperationError | None:  # noqa: PLR0911
```
eyurtsev, line 51: This doesn't look correct -> these exceptions aren't going to be returned from most SDKs

libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -690,17 +691,11 @@ def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadRespons
...
-                responses.append(FileUploadResponse(path=path, error="permission_denied"))
-            except (ValueError, OSError) as e:
-                # ValueError from _resolve_path for path traversal, OSError for other file errors
-                if isinstance(e, ValueError) or "invalid" in str(e).lower():
-                    responses.append(FileUploadResponse(path=path, error="invalid_path"))
-                else:
-                    # Generic error fallback
-                    responses.append(FileUploadResponse(path=path, error="invalid_path"))
+            except Exception as exc:
+                error = map_file_operation_error(exc)
```
eyurtsev, line 695: `map_file_operation_error` should be only in the file system backend -- it's a local utility and cannot be re-used across implementations

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -73,23 +107,23 @@ class FileDownloadResponse:
 
     path: str
     content: bytes | None = None
-    error: FileOperationError | None = None
+    error: str | None = None
```
eyurtsev, line 110: The error in the API is intentionally not open-ended as there are only a few errors which are LLM recoverable, most other errors cannot be recovered by an LLM so they shouldn't be retried by the LLM either by the LLM (i.e., would be better to have them raised)

mdrxy: Reverted here #2168

libs/deepagents/deepagents/backends/protocol.py
mdrxy: Reverted in #2168

PR #2119, opened 2026-03-20
feat(sdk): add completion notifier middleware for async subagents

hntrl:
## Summary

Adds a pre-bundled `CompletionNotifierMiddleware` that async subagents can use to proactively notify their supervisor when they complete or error -- closing the gap where the supervisor only learns about completion when someone calls `check_async_task`.

## Why

The async subagent protocol is fire-and-forget: the supervisor launches a task and only discovers its outcome when it (or the user) polls via `check_async_task`. This means the supervisor can't proactively relay results -- the user has to ask. The completion notifier solves this by having the subagent push a notification to the supervisor's thread when it finishes.

This is an opt-in middleware added to the **subagent's** stack, not the supervisor's. If the subagent doesn't use it, the extra input keys are silently ignored by LangGraph (verified against both in-process `invoke()` and `langgraph dev` via SDK).

## Architecture

```
Supervisor                    Subagent
    |                            |
    |--- start_async_task -----> |
    |<-- task_id (immediately) - |
    |                            |  (working...)
    |                            |  (done!)
    |                            |
    |<-- runs.create(            |
    |      supervisor_thread,    |
    |      "completed: ...")     |
    |                            |
    |  (wakes up, sees result)   |
```

**Parent context propagation:** The supervisor's `start_async_task` tool now includes `parent_thread_id`, `parent_assistant_id`, and `task_id` in the subagent's input state. The notifier reads these from state (not config), which means they survive thread interrupts and `update_async_task` calls. If the subagent doesn't have the middleware, these keys are silently dropped.

**Notification format:** Includes `task_id` so the supervisor can correlate notifications back to specific tracked jobs when multiple tasks of the same subagent type run concurrently:
```
[task_id=thread_abc][subagent=researcher] Completed. Result: <summary>
[task_id=thread_abc][subagent=researcher] Error: <error message>
```

### New: `CompletionNotifierMiddleware` (`deepagents/middleware/completion_notifier.py`)
- `aafter_agent` hook sends a completion notification with result summary (truncated to 500 chars)
- `awrap_model_call` hook catches errors and sends an error notification before re-raising
- `CompletionNotifierState` adds `parent_thread_id`, `parent_assistant_id`, and `task_id` to the subagent's state schema
- Notifies only once per run (guards against duplicates)
- Silently no-ops if parent context is missing (subagent launched without a supervisor)
- Uses `get_client()` with no URL (ASGI transport for same-deployment)

### Modified: `async_subagents.py`
- Added `_extract_parent_context()` to read supervisor's thread_id and assistant_id from tool runtime config
- `start_async_task` (sync + async) now includes `task_id`, `parent_thread_id`, `parent_assistant_id` in the subagent's input

### Modified: `pyproject.toml`
- Added `langgraph-sdk>=0.1.51` as a required dependency (was already imported unconditionally)

### Exports
- `CompletionNotifierMiddleware` exported from `deepagents.middleware` and top-level `deepagents`

## Open questions

- **`assistant_id` availability:** Need to validate that `config.metadata.assistant_id` is always present in deployed contexts. If there are cases where it's missing, the notifier silently no-ops (safe), but the notification won't fire.
- **Divergence from JS:** The JS reference implementation passes parent IDs as constructor args at graph factory time (from config). The Python implementation reads them from state, which is a deliberate improvement (survives interrupts, no dynamic factory needed) but a divergence.

## How to review

The core middleware is in `completion_notifier.py` -- start there. The `async_subagents.py` changes are mechanical (extract parent context, spread into input). Tests cover all the hooks, edge cases, and the parent context propagation.

---

> [!NOTE]
> This PR was developed with assistance from an AI coding agent.

### DIFF
--- a/libs/cli/uv.lock
+++ b/libs/cli/uv.lock
@@ -974,6 +974,7 @@ dependencies = [
     { name = "langchain-anthropic" },
     { name = "langchain-core" },
     { name = "langchain-google-genai" },
+    { name = "langgraph-sdk" },
     { name = "langsmith" },
     { name = "wcmatch" },
 ]
@@ -984,6 +985,7 @@ requires-dist = [
     { name = "langchain-anthropic", specifier = ">=1.4.0,<2.0.0" },
     { name = "langchain-core", specifier = ">=1.2.19,<2.0.0" },
     { name = "langchain-google-genai", specifier = ">=4.2.0,<5.0.0" },
+    { name = "langgraph-sdk", specifier = ">=0.1.51" },
     { name = "langsmith", specifier = ">=0.3.0" },
     { name = "wcmatch" },
 ]
--- a/libs/deepagents/deepagents/__init__.py
+++ b/libs/deepagents/deepagents/__init__.py
@@ -3,7 +3,6 @@
 from deepagents._version import __version__
 from deepagents.graph import create_deep_agent
 from deepagents.middleware.async_subagents import AsyncSubAgent, AsyncSubAgentMiddleware
-from deepagents.middleware.completion_notifier import CompletionNotifierMiddleware
 from deepagents.middleware.filesystem import FilesystemMiddleware
 from deepagents.middleware.memory import MemoryMiddleware
 from deepagents.middleware.subagents import CompiledSubAgent, SubAgent, SubAgentMiddleware
@@ -12,7 +11,6 @@ __all__ = [
     "AsyncSubAgent",
     "AsyncSubAgentMiddleware",
     "CompiledSubAgent",
-    "CompletionNotifierMiddleware",
     "FilesystemMiddleware",
     "MemoryMiddleware",
     "SubAgent",
--- a/libs/deepagents/deepagents/__init__.py
+++ b/libs/deepagents/deepagents/__init__.py
@@ -3,6 +3,7 @@
 from deepagents._version import __version__
 from deepagents.graph import create_deep_agent
 from deepagents.middleware.async_subagents import AsyncSubAgent, AsyncSubAgentMiddleware
+from deepagents.middleware.completion_notifier import CompletionNotifierMiddleware
 from deepagents.middleware.filesystem import FilesystemMiddleware
 from deepagents.middleware.memory import MemoryMiddleware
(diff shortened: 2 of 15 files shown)


lonexreb (commented):
This is a well-designed addition to the async subagent protocol. The push-based notification pattern closes a real gap — polling is fine for batch workflows but painful for interactive sessions where the user expects results proactively.

### Architecture observations

1. **State over config for parent context is the right call.** Surviving thread interrupts and `update_async_task` is important for long-running subagents. The JS divergence is worth it.

2. **Silent no-op when callback_thread_id is missing** makes adoption smooth — existing subagents that don't opt in are unaffected.

3. **The error notification in `awrap_model_call`** is smart — catching model errors and notifying the parent before re-raising means the supervisor learns about failures without polling.

### Questions / suggestions

1. **Duplicate notification guard**: The PR description mentions "notifies only once per run" but I don't see an explicit guard in the code. If `aafter_agent` is called once per run that may be sufficient, but is there a scenario where a retry or checkpoint resume could trigger it twice?

2. **500-char truncation**: The truncated message appends guidance to use `check_async_task` — nice touch. But the combined string (truncated content + guidance) could exceed the 500-char target. Minor, but `_MAX_MESSAGE_LENGTH` could account for the suffix length.

3. **`from __future__ import annotations`**: This file uses it (line 79) while the rest of the SDK doesn't. Not a problem functionally, but a style inconsistency worth flagging.

4. **Broader pattern**: This push notification pattern could generalize well beyond async subagents — for example, a `PeerNotificationMiddleware` where sibling agents notify each other. The `callback_graph_id` + `callback_thread_id` mechanism is essentially an addressable inbox. Curious if you've thought about making this more generic in the future.

Clean implementation with good documentation and test coverage. LGTM.

mtene:
@hntrl I'd be curious to hear what the plans are for this work. Interstingly, the JS equivalent of it has been merged https://github.com/langchain-ai/deepagentsjs/pull/361

I've been using this PR and your demo repo to implement the automated notification middleware ad-hoc https://github.com/langchain-ai/async-deep-agents. I think it is essential for background tasks / async agents, as otherwise we are leaving completion follow-ups to chance (the caller agent has to be explicitly prompted to probe and continue).

PR #2170, opened 2026-03-24
fix(cli): disable markup parsing for blocked-link notifications

mlucifer27:
Prevents Rich markup parsing in the suspicious-URL warning toast within `open_style_link`, avoiding notification rendering errors when URLs/details contain markup-like characters (for example `[` and `]`). It also adds focused unit tests for both safe-link behavior (opens browser) and suspicious-link behavior (blocks and notifies with `markup=False`).


mdrxy:
How did you encounter this?

PR #2181, opened 2026-03-24
feat(cli): inline argument hints for slash commands

### LINKED ISSUE #1609: Helper `<text>` arg on `/remember` & other skills
### Privileged issue

### Issue Content

`/remember` currently supports optional trailing text (for example, `/remember only keep GitHub workflow preferences`), but this is not discoverable in UI/help. In long conversations, users worry the command may infer and store arbitrary context.

Current behavior in CLI:
- Parser accepts `/remember <text>` and appends it as `Additional context from user` to the remember prompt.
- `/remember` descriptions in autocomplete/help do not mention argument support.
- Base remember prompt asks the model to broadly "review our conversation," which can over-collect when user intent is narrow.

Why this matters:
- Memory writes should be explicit and predictable.
- Users need confidence that only intended preferences/workflows are persisted.
- Low discoverability causes underuse/misuse of an existing capability.

### Proposed solution

1. **Improve discoverability via `argument-hint` frontmatter**
   - Claude Code skills support an `argument-hint` field in SKILL.md frontmatter that surfaces placeholder text during autocomplete (e.g. `[context]`).
   - Add `argument-hint: "[context]"` to the `/remember` SKILL.md frontmatter. This makes the optional arg visible in autocomplete without requiring custom help-screen work.
   - Audit other skills that accept `$ARGUMENTS` and add appropriate `argument-hint` values.
2. **Tighten scoping semantics when arg is supplied**
   - Keep `/remember` (no arg) as broad review behavior.
   - When `/remember <context>` is used, add explicit instructions to limit extraction to that context and avoid unrelated inference.

km322:
Closes #1609 

---

- Add argument hints to built-in skill slash commands
- Refactor slash command registry to use `CommandEntry` `NamedTuple` replacing raw tuples
- `/remember` guard requiring conversation messages

May follow up to allow arbitrary arg hints if/when support in the SDK lands w/ skill loading.

---

## Social handles (optional)
LinkedIn: https://linkedin.com/in/ketan-mittal322/


lonexreb (commented):
Welcome to open source! This is a well-structured first PR — clean separation of SDK and CLI changes, thorough test updates, and good questions in the description.

A few observations:

### What works well

- The SDK change is minimal and correct: `argument_hint` as an optional `str | None` field in `SkillMetadata`, parsed from frontmatter with proper null handling.
- The scoping addition to the `remember` skill is a nice improvement — having the skill limit its scope when given an argument makes it much more predictable.
- Test coverage is thorough — every `SkillMetadata` literal in tests was updated, which is tedious but necessary.

### Minor observations

1. **The 4-tuple pattern is getting dense.** With `(name, description, hidden_keywords, argument_hint)`, it's starting to push readability. Not a problem for this PR, but if more fields are added later, a `NamedTuple` might be cleaner. Just flagging for the maintainers.

2. **The hint is plumbed but not rendered** — you noted this already. The `_` in `for cmd, desc, kw, _ in self._commands` confirms it's wired through but not yet surfaced in the autocomplete UI. A follow-up PR to render it would be a nice companion.

3. **Re: your question about cross-library changes** — since `SkillMetadata` is defined in the SDK and consumed by the CLI, touching both is the correct approach here. The alternative (adding the field only in CLI) would create a divergence.

4. **Re: updating the `libs/cli/examples/skills/skill-creator` copy** — I'd say yes, keep them in sync to avoid confusion, but defer to the maintainers on that.

Looks good overall. Nice first contribution!

libs/cli/deepagents_cli/built_in_skills/remember/SKILL.md
```diff
@@ -1,12 +1,15 @@
...
 name: remember
 description: "Review the current conversation and capture valuable knowledge — best practices, coding conventions, architecture decisions, workflows, and user feedback — into persistent memory (AGENTS.md) or reusable skills. Use when the user says: (1) remember this, (2) save what we learned, (3) update memory, (4) capture learnings."
+argument-hint: "[context]"
 license: MIT
 compatibility: designed for deepagents-cli
 ---
 
 Review our conversation and capture valuable knowledge. Focus especially on **best practices** we discussed or discovered—these are the most important things to preserve.
 
+**Scoping:** If a **User request** is provided below, limit your review strictly to that context. Only extract and persist knowledge directly related to the stated scope — do not infer or capture unrelated information from the broader conversation.
```
mdrxy: removing; this should be a seperate PR i believe

mdrxy:
Thank you @km322 for getting this started on the right track! I reworked it a bit to limit the scope, as for now we don't want to make any SDK changes for this, though we can rework down the road if/when it becomes available.

Great work!

PR #2183, opened 2026-03-24
feat(sdk): evict large HumanMessages

ccurme:
- Implemented via `wrap_model_call` in FileSystemMiddleware
- If most recent message is HumanMessage and long, evict and mark with `"lc_evicted_to": "<path>"` in `additional_kwargs`
- Truncate all messages marked with `"lc_evicted_to"` in model request.
- Original message contents preserved in state, only `additional_kwargs` is updated.

### DIFF
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -1523,7 +1523,7 @@ class FilesystemMiddleware(AgentMiddleware[FilesystemState, ContextT, ResponseT]
         file_path: str | None = None
         if new_eviction_needed:
             backend = self._get_backend_from_runtime(request.state, request.runtime)
-            file_path = f"/large_messages/{uuid.uuid4().hex[:12]}"
+            file_path = f"/conversation_history/{uuid.uuid4().hex[:12]}"
             write_result = backend.write(file_path, _extract_text_from_message(messages[-1]))
 
         return self._apply_eviction_and_truncate(messages, write_result, file_path)
@@ -1549,7 +1549,7 @@ class FilesystemMiddleware(AgentMiddleware[FilesystemState, ContextT, ResponseT]
         file_path: str | None = None
         if new_eviction_needed:
             backend = self._get_backend_from_runtime(request.state, request.runtime)
-            file_path = f"/large_messages/{uuid.uuid4().hex[:12]}"
+            file_path = f"/conversation_history/{uuid.uuid4().hex[:12]}"
             write_result = await backend.awrite(file_path, _extract_text_from_message(messages[-1]))
 
         return self._apply_eviction_and_truncate(messages, write_result, file_path)
--- a/libs/deepagents/tests/unit_tests/test_end_to_end.py
+++ b/libs/deepagents/tests/unit_tests/test_end_to_end.py
@@ -1365,14 +1365,14 @@ class TestLargeHumanMessageEviction:
         assert msg.content == large_content
         evicted_to = msg.additional_kwargs.get("lc_evicted_to")
         assert evicted_to is not None
-        assert evicted_to.startswith("/large_messages/")
+        assert evicted_to.startswith("/conversation_history/")
 
         assert len(fake_model.call_history) == 1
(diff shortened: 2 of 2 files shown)


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1300,6 +1413,127 @@ async def _aprocess_large_message(
...
+        if not self._tool_token_limit_before_evict:
+            return message
+
+        content_str = _extract_text_from_message(message)
+
+        if len(content_str) <= NUM_CHARS_PER_TOKEN * self._tool_token_limit_before_evict:
+            return message
+
+        msg_id = message.id or "unknown"
+        file_path = f"/large_messages/{msg_id}"
```
eyurtsev: This is a breaking change -- we need to standardize around artifacts path

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1097,6 +1135,81 @@ async def async_execute(  # noqa: PLR0911 - early returns for distinct error con
...
+        """Async version of `before_agent`.
+
+        Args:
+            state: The current agent state.
+            runtime: The runtime context.
+
+        Returns:
+            A state update replacing the oversized message, or `None`.
+        """
+        if not self._tool_token_limit_before_evict:
```
eyurtsev: there's a tiny amount of code duplication, could you see if a refactor could shrink it here enough?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1300,6 +1413,127 @@ async def _aprocess_large_message(
...
+            return message
+
+        content_str = _extract_text_from_message(message)
+
+        if len(content_str) <= NUM_CHARS_PER_TOKEN * self._tool_token_limit_before_evict:
+            return message
+
+        msg_id = message.id or "unknown"
```
eyurtsev: it should be randomly generated string if an ID is not available (unknown would cause collisions)

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1300,6 +1413,127 @@ async def _aprocess_large_message(
...
+            content_sample=content_sample,
+        )
+
+        evicted = _build_evicted_human_content(message, replacement_text)
+        return HumanMessage(
+            content=cast("str | list[str | dict]", evicted),
+            id=message.id,
+            additional_kwargs=dict(message.additional_kwargs),
+            response_metadata=dict(message.response_metadata),
+        )
```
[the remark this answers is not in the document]
ccurme: Should be resolved now that IDs are always generated.

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1097,6 +1136,81 @@ async def async_execute(  # noqa: PLR0911 - early returns for distinct error con
...
 
+    def before_agent(
+        self,
+        state: FilesystemState,
+        runtime: Runtime[ContextT],
+    ) -> dict[str, Any] | None:
+        """Evict an oversized HumanMessage to the filesystem before the agent loop.
+
+        Checks the most recent message in `state["messages"]`. If it is a
+        HumanMessage whose text content exceeds `tool_token_limit_before_evict`,
```
sydney-runkle: unintuitive to me that we're using `tool_token_limit_before_evict` as the threshold for a human message

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1300,6 +1414,127 @@ async def _aprocess_large_message(
...
+    def _get_backend_from_runtime(
+        self,
+        state: AgentState[Any],
+        runtime: Runtime[ContextT],
+    ) -> BackendProtocol:
+        """Resolve the backend from a bare `Runtime`.
+
+        Constructs a `ToolRuntime` from the `Runtime` to satisfy the backend
+        factory interface. Used by hooks like `before_agent` that receive
+        `Runtime` rather than `ToolRuntime`.
```
sydney-runkle, line 1413: sigh :/ it's fine i'm fine

PR #2186, opened 2026-03-24
chore(sdk): minor clean up

eyurtsev:
exception handling code for local file system should live where implementation of local file system is


libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -354,14 +328,6 @@ def read(
             starting at line 1. Lines longer than 2000 characters are truncated.
 
             Returns an error string if the file doesn't exist or can't be read.
-
-        !!! note
```
eyurtsev, line 358: we don't want this doc-string (should've probably been a separate PR)

PR #2193, opened 2026-03-24
feat(evals): add more evals with more tools

mdrxy:
Adds back #1849 / #2192 when we can fix it


mdrxy:
@deepagents what's up

Issue #1698, opened 2026-03-07
`SubAgentMiddleware` does not propagate `recursion_limit` to subagent graphs — subagents silently use default limit of 25


marcosmarf27:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

<html>
<body>
<html><head></head><body><h1>SubAgentMiddleware does not propagate <code>recursion_limit</code> to subagent graphs — subagents silently use default limit of 25</h1>
<h2>Checked other resources</h2>
<ul>
<li>[x] This is a bug, not a usage question.</li>
<li>[x] I added a clear and descriptive title.</li>
<li>[x] I searched existing issues and didn't find this.</li>
<li>[x] I can reproduce this with the latest released version.</li>
<li>[x] I included a minimal reproducible example.</li>
</ul>
<h2>Bug description</h2>
<p>When invoking a deep agent with a custom <code>recursion_limit</code> (e.g., 150 or 300), the limit is <strong>not propagated</strong> to subagents created via the <code>SubAgentMiddleware</code>. Subagents always run with LangGraph's default <code>recursion_limit=25</code>, regardless of the parent's configuration.</p>
<p>This causes subagents performing multi-step tasks (e.g., 13+ tool calls) to silently hit the 25-step limit, raise a <code>GraphRecursionError</code>, which then propagates as <code>asyncio.CancelledError</code> through the parent agent — often leaving the entire execution stuck or crashed without a clear error message.</p>
<h2>Root cause</h2>
<p>In <a href="https://github.com/langchain-ai/deepagents/blob/master/libs/deepagents/deepagents/middleware/subagents.py"><code>libs/deepagents/deepagents/middleware/subagents.py</code></a>, the subagent is invoked without passing <code>config</code>:</p>
<pre><code class="language-python">result = await subagent.ainvoke(subagent_state)  # ← no config passed
</code></pre>
<p>Since no <code>config</code> is provided, LangGraph uses its default <code>recursion_limit=25</code>. The parent's <code>recursion_limit</code> (set via <code>invoke()</code> config or <code>.with_config()</code>) is never forwarded.</p>
[...]

PR #2194, opened 2026-03-24
fix(sdk): bump recursion limit to 10_000 in deepagents and set agent name

eyurtsev:
* bump default recursion limit to 10_000 in deepagents
* set agent name

Users can control recursion limit for subagents using:

```python
agent = create_agent(..) or create_deep_agent(..)

agent = agent.with_config({
   'recursion_limit': 100
})
```

Closes https://github.com/langchain-ai/deepagents/issues/1698

### DIFF
--- a/libs/deepagents/tests/unit_tests/test_end_to_end.py
+++ b/libs/deepagents/tests/unit_tests/test_end_to_end.py
@@ -232,7 +232,6 @@ class TestDeepAgentEndToEnd:
             config={
                 "configurable": {"thread_id": "test_tool_runtime_metadata"},
                 "tags": ["tool-tag", "tool-session-456"],
-                "metadata": {"request_id": "req-tool-123", "tenant": "acme-tool"},
             },
         )
 


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -442,7 +442,12 @@ def task(
             value_error_msg = "Tool call ID is required for subagent invocation"
             raise ValueError(value_error_msg)
         subagent, subagent_state = _validate_and_prepare_state(subagent_type, description, runtime)
-        result = subagent.invoke(subagent_state)
+        result = subagent.invoke(
+            subagent_state,
+            {
+                "recursion_limit": 10_000,
```
ccurme: why is this larger than the recursion limit for the main agent?

eyurtsev: I'll inherit from parent for now... i kind of think that we shouldn't have any recursion limits and instead potentially opt-in for default middleware that uses agent steps (which is much easier to reason about)

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -442,7 +442,13 @@ def task(
...
             raise ValueError(value_error_msg)
         subagent, subagent_state = _validate_and_prepare_state(subagent_type, description, runtime)
-        result = subagent.invoke(subagent_state)
+        recursion_limit = runtime.config["recursion_limit"]
+        result = subagent.invoke(
+            subagent_state,
+            {
+                "recursion_limit": recursion_limit,
+            },
+        )
```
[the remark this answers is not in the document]
eyurtsev: It doesn't actually drop it. Semantics are a bit odd, but recursion_limit patches the existing config rather than dropping all of it. there's merge logic happening under the hood

PR #2223, opened 2026-03-25
ci: add minimum workflow permissions and SHA-pin third-party actions

jkennedyvz:
## Summary

- Add top-level `permissions: contents: read` to three workflow files that were missing it (`auto-label-by-package.yml`, `_benchmark.yml`, `release-please.yml`), preventing GITHUB_TOKEN from inheriting broad org/repo defaults
- SHA-pin all 6 unpinned third-party actions to protect against tag-hijacking supply chain attacks:
  - `pypa/gh-action-pypi-publish@release/v1` → `ed0c5393` (release.yml ×2)
  - `googleapis/release-please-action@v4` → `c3fc4de0` (release-please.yml)
  - `astral-sh/setup-uv@v7` → `94527f2e` (release-please.yml)
  - `dorny/paths-filter@v4` → `fbd0ab8f` (ci.yml)
  - `CodSpeedHQ/action@v4` → `dfaf2584` (_benchmark.yml)

## Test plan

- [x] Verify CI passes on this PR
- [x] Confirm `_benchmark.yml` reusable workflow still functions when called from `ci.yml`
- [x] Confirm `release-please.yml` jobs still have correct effective permissions (job-level declarations are unchanged)
- [x] Confirm PyPI publish jobs in `release.yml` still work with the SHA-pinned action

### DIFF
--- a/.github/workflows/_benchmark.yml
+++ b/.github/workflows/_benchmark.yml
@@ -24,7 +24,6 @@ on:
 
 permissions:
   contents: read
-  id-token: write
 
 env:
   UV_NO_SYNC: "true"


.github/workflows/_benchmark.yml
```diff
@@ -22,6 +22,10 @@ on:
         # https://github.com/CodSpeedHQ/pytest-codspeed/issues/106
         default: "3.13.11"
 
+permissions:
+  contents: read
+  id-token: write
```
eyurtsev: why is id token top level?

eyurtsev: (I generally prefer to scope them down to where they are used and not declare at global level -- even if the current benchmarking yml job only has a single step)

.github/workflows/release-please.yml
```diff
@@ -14,6 +14,9 @@ on:
     branches:
       - main
 
+permissions:
```
mdrxy, line 17: This will break the workflow AFAIK

mdrxy: release-please needs write permissions to create/edit/push to the PRs it generates

mdrxy: Ah I see there is job-level permissions -- what does this add then?

jkennedyvz: The job-level permissions on 24/25 should provide what the release-please job needs.

PR #2266, opened 2026-03-27
fix(sdk): fix offloading for state backend

ccurme:


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1606,7 +1609,7 @@ def _evict_and_truncate_messages(
         file_path: str | None = None
         if new_eviction_needed:
             backend = self._get_backend_from_runtime(request.state, request.runtime)
-            file_path = f"/conversation_history/{uuid.uuid4().hex[:12]}"
+            file_path = f"/conversation_history/{uuid.uuid4()}.md"
```
sydney-runkle, line 1612: huh?

ccurme: This isn't needed from a technical point of view. File names can be exposed to users so adding the extension seemed friendlier. and existing files in `conversation_history` are uuids so this makes them look more uniform.

PR #2308, opened 2026-03-28
fix(acp): block dangerous shell patterns in command auto-approve

jkennedyvz:
## Summary
- Adds `contains_dangerous_patterns()` check to the ACP server's command auto-approve path, preventing shell metacharacters (`$()`, backticks, `;`, redirects, `&`, variable expansion) from bypassing the allowlist
- Extends `extract_command_types()` to split on `||` and `;` in addition to `&&`, closing a compound-command bypass
- Adds dedicated test suite for dangerous pattern detection

## Test plan
- [ ] Run `pytest libs/acp/tests/test_dangerous_patterns.py` to verify new pattern detection tests pass
- [ ] Run `pytest libs/acp/tests/test_command_allowlist.py` to verify existing allowlist tests pass with updated logic
- [ ] Verify that commands containing `$(...)`, backticks, `;`, redirects, or variable expansion are no longer auto-approved even when the base command type is allowed

### DIFF
--- a/libs/acp/deepagents_acp/server.py
+++ b/libs/acp/deepagents_acp/server.py
@@ -63,6 +63,7 @@ if TYPE_CHECKING:
     from langchain_core.runnables import RunnableConfig
 
 from deepagents_acp.utils import (
+    contains_dangerous_patterns,
     convert_audio_block_to_content_blocks,
     convert_embedded_resource_block_to_content_blocks,
     convert_image_block_to_content_blocks,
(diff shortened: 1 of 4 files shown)


lonexreb (commented):
Nice work on porting the CLI's dangerous pattern detection to ACP. The defense-in-depth approach (check patterns before allowlist lookup) is the right call — it prevents any metacharacter-based bypass regardless of what `extract_command_types` returns.

A few observations:

### 1. `re.split(r"&&|\|\||;", command)` can split inside quoted strings

The new compound-operator split (line 287) doesn't account for quoting. For example:

```
echo "hello; world" && ls
```

would incorrectly split into `["echo \"hello", " world\" ", " ls"]`. The existing `&&` splitter had the same limitation, so this isn't a regression — but worth noting. Since `contains_dangerous_patterns` runs first and `;` inside the raw string would trigger it, this is effectively guarded in the auto-approve path. The risk is only if `extract_command_types` is called independently elsewhere without the guard.

### 2. `>` false positives in quoted arguments

Commands like `python -c "if x > 0: print('yes')"` will be flagged as dangerous because `>` appears in the raw string. This is the right trade-off for auto-approve (false positive → user gets prompted, which is safe), but it's worth a brief comment in the docstring noting that the check is intentionally conservative.

### 3. Pattern list matches CLI exactly — good

Verified that `DANGEROUS_SHELL_PATTERNS` and the regex checks (`$VAR`, standalone `&`) are identical to `deepagents_cli.config`, so the two paths stay in sync. Might be worth a future refactoring to share this as a common utility, but that's out of scope here.

### 4. Test coverage is solid

The test matrix covers all pattern categories and the `test_double_ampersand_is_safe` case is a nice edge case to have explicitly. The updated `test_security_python_pytest_vs_pip` test correctly reflects the new reality where `contains_dangerous_patterns` guards against the semicolon case at a higher level.

Overall LGTM — small, well-scoped security fix with good test coverage.

PR #2319, opened 2026-03-30
fix(sdk): catch `UnicodeDecodeError` in `FilesystemBackend.read`

### LINKED ISSUE #2318: `FilesystemBackend.read` raises `UnicodeDecodeError` for non-UTF-8 text files
## Checked other resources

## Area (Required)

- [x] deepagents (SDK)

## Reproduction Steps / Example Code (Python)

```python
import tempfile, os

# Create a file encoded in GBK (common Chinese encoding, not valid UTF-8)
with tempfile.TemporaryDirectory() as tmpdir:
    fpath = os.path.join(tmpdir, "test.txt")
    with open(fpath, "wb") as f:
        # 0x87 is invalid UTF-8 (valid GBK byte)
        f.write("中文内容".encode("gbk"))

    backend = FilesystemBackend(root=tmpdir)
    result = backend.read("test.txt")
    # Raises UnicodeDecodeError instead of returning an error result
    print(result)
```

## Error Message and Stack Trace (if applicable)

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x87 in position X: invalid start byte
```

This surfaces to the caller as:

```
Agent error: {'error': 'UnicodeDecodeError', 'message': "'utf-8' codec can't decode byte 0x87 ..."}
```

## Description

`FilesystemBackend.read()` (in `deepagents/backends/filesystem.py`) opens text files with `encoding="utf-8"` but only catches `OSError` — it does not catch `UnicodeDecodeError`. When a file that passes the `_get_file_type()` "text" check actually contains non-UTF-8 bytes (e.g. GBK/GB2312-encoded Chinese files, Windows-1252 files, or other single-byte encodings), Python raises an unhandled `UnicodeDecodeError` that propagates all the way up and crashes the agent.

Root cause (filesystem.py ~line 330):

```python
with os.fdopen(fd, "r", encoding="utf-8") as f:
    content = f.read()
# ...
except OSError as e:   # <-- UnicodeDecodeError is NOT caught here
    return ReadResult(error=f"Error reading file '{file_path}': {e}")
```

By contrast, the `grep()` method already handles this correctly:

```python
except (UnicodeDecodeError, PermissionError, OSError):
    continue
```

**Fix**: add `UnicodeDecodeError` to the `except` clause in `read()`, or re-open with `errors='replace'` to surface file content with replacement characters instead of crashing.

## Environment / System Info

OS: macOS / Linux (reproducible on any platform with non-UTF-8 files)
Python: 3.11+
deepagents: latest main

BANG404:
Fixes #2318

`FilesystemBackend.read()` raises an unhandled `UnicodeDecodeError` when reading a file whose extension passes the `_get_file_type()` "text" check but whose bytes are not valid UTF-8 (e.g. GBK/GB2312-encoded Chinese files, Windows-1252 files). The exception propagates all the way up and surfaces as an agent-level crash.

The fix adds `UnicodeDecodeError` to the existing `except OSError` clause so the method returns a structured `ReadResult(error=...)` instead of raising — consistent with how all other error paths in this method already behave, and consistent with how `grep()` already handles the same error at line 580.

**Changes**
- `libs/deepagents/deepagents/backends/filesystem.py`: extend `except OSError` → `except (OSError, UnicodeDecodeError)` in `read()`
- `libs/deepagents/tests/unit_tests/backends/test_filesystem_backend.py`: add `test_filesystem_backend_read_non_utf8_file` that writes a GBK-encoded file and asserts the result is a `ReadResult` with a non-`None` error (not a raised exception)

> 🤖 This PR was created with assistance from Claude Code (claude-sonnet-4-6).

### DIFF
--- a/libs/deepagents/tests/unit_tests/backends/test_filesystem_backend.py
+++ b/libs/deepagents/tests/unit_tests/backends/test_filesystem_backend.py
@@ -206,6 +206,21 @@ def test_filesystem_backend_ls_trailing_slash(tmp_path: Path):
     assert empty.entries == []
 
 
+def test_filesystem_backend_read_non_utf8_file(tmp_path: Path):
+    """FilesystemBackend.read should return an error result, not raise, for non-UTF-8 text files."""
+    root = tmp_path
+    # Write a file with GBK-encoded bytes that are invalid UTF-8 (e.g. 0x87)
+    gbk_file = root / "chinese.txt"
+    gbk_file.write_bytes("中文内容".encode("gbk"))
+
+    be = FilesystemBackend(root_dir=str(root), virtual_mode=False)
+    result = be.read(str(gbk_file))
+
+    assert isinstance(result, ReadResult)
+    assert result.error is not None
+    assert "chinese.txt" in result.error
+
+
 def test_filesystem_backend_intercept_large_tool_result(tmp_path: Path):
     """Test that FilesystemBackend properly handles large tool result interception."""
     root = tmp_path
--- a/libs/deepagents/deepagents/backends/filesystem.py
+++ b/libs/deepagents/deepagents/backends/filesystem.py
@@ -343,7 +343,7 @@ class FilesystemBackend(BackendProtocol):
 
             selected_lines = lines[start_idx:end_idx]
             return ReadResult(file_data=FileData(content="\n".join(selected_lines), encoding="utf-8"))
-        except OSError as e:
+        except (OSError, UnicodeDecodeError) as e:
             return ReadResult(error=f"Error reading file '{file_path}': {e}")
 
     def write(


libs/deepagents/tests/unit_tests/backends/test_filesystem_backend.py
```diff
@@ -206,6 +206,21 @@ def test_filesystem_backend_ls_trailing_slash(tmp_path: Path):
     assert empty.entries == []
 
 
+def test_filesystem_backend_read_non_utf8_file(tmp_path: Path):
+    """FilesystemBackend.read should return an error result, not raise, for non-UTF-8 text files."""
+    root = tmp_path
+    # Write a file with GBK-encoded bytes that are invalid UTF-8 (e.g. 0x87)
+    gbk_file = root / "chinese.txt"
+    gbk_file.write_bytes("中文内容".encode("gbk"))
```
eyurtsev, line 214: 谢谢！🙏

PR #2334, opened 2026-03-30
fix(sdk): inherit parent `interrupt_on` for subagents

hntrl:
This change fixes inconsistent HITL behavior across subagents by making declarative custom subagents inherit the parent agent `interrupt_on` config when they do not define their own. Previously, parent `interrupt_on` was wired into the default general-purpose subagent but not propagated as a default to custom `SubAgent` specs, leading to different approval behavior for equivalent tool calls.

## Changes

- `libs/deepagents/deepagents/graph.py`
  - Stops attaching `HumanInTheLoopMiddleware` directly to the general-purpose subagent middleware stack.
  - Propagates parent `interrupt_on` into the generated general-purpose subagent spec.
  - Propagates parent `interrupt_on` as a default for each declarative custom subagent spec, while preserving per-subagent override semantics.
- `libs/evals/tests/evals/test_hitl.py`
  - Adds a LangSmith HITL integration test that validates a custom subagent without explicit `interrupt_on` inherits parent interrupt policy and interrupts the expected tools.

### DIFF
--- a/libs/deepagents/tests/unit_tests/test_subagents.py
+++ b/libs/deepagents/tests/unit_tests/test_subagents.py
@@ -802,7 +802,6 @@ class TestSubAgents:
         assert len(tool_messages) == 1
         assert tool_messages[0].content == "tool completed"
 
-
     @pytest.mark.xfail(
         reason="callbacks in parent config are not forwarded to subagent invocations (see #2315)",
         strict=True,


eyurtsev:
Should we expose also an ability to update interrupt_on  (since subagent spec supports custom tools?)

PR #2346, opened 2026-03-31
feat(cli): support root/MDM installs

mdrxy:
Support running the install script as root, targeting macOS MDM tools (Kandji, Jamf, etc.) that execute scripts as root in a minimal environment. When `id -u` is 0, the script resolves the real console user's `HOME` and `chown`s all created files back to them. Non-root installs are completely unaffected — all new code paths either short-circuit on the root check or define `fix_owner` as a no-op.


jkennedyvz (approved):
Works on my machine!

PR #2360, opened 2026-03-31
fix: deprecate backend factories

sydney-runkle:
## Summary

1. Deprecate the backend factory pattern in favor of backend instances

before
```py
agent=create_deep_agent(backend=lambda rt: StateBackend(rt))
```

after

```py
agent=create_deep_agent(backend=StateBackend())
```

2. State backend writes without using `files_update` artifacts on `EditResult` and `WriteResult`

before
```py
def my_node():
    # responsibility of the caller to update state, backend.write doesn't actually write to state
    result = backend.write(...)
    if (update := result.files_update):
        return Command(update=update)
```

after
```py
def my_node():
    # backend.write writes to state
    result = backend.write(...)
```

## Core Changes

- **StateBackend**: Uses `get_config()` + `CONFIG_KEY_READ`/`CONFIG_KEY_SEND` instead of `runtime.state`. `runtime` arg deprecated with warning. Clear error when used outside graph context.
- **StoreBackend**: Uses `get_store()`/`get_runtime()` instead of `runtime.store`. New `store=` kwarg for explicit store injection. `runtime` arg deprecated.
- **WriteResult/EditResult**: `files_update` field removed — state writes go directly through `CONFIG_KEY_SEND`.
- **Middleware**: Write/edit tools return plain strings (no `Command` wrapping). Eviction writes go through `CONFIG_KEY_SEND`. `contextvars.copy_context()` fix for glob's `ThreadPoolExecutor`.
- **CompositeBackend**: Removed `files_update` sync code.
- **Summarization**: `_offload_to_backend` returns `str | None` instead of tuple.
- **Factory pattern**: `_get_backend` callable path deprecated with warning. Default backend is `StateBackend()` instance.

## Tests

- New `TestStateBackendConfigKeys` class with write/read/edit/ls, backward compat lambda, deprecation warnings, explicit store, `get_store` fallback, and factory deprecation tests.
- All existing tests updated to use `state_config_context()` helper or `create_deep_agent` with fake models.


libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -298,7 +282,7 @@ def ls(self, path: str) -> "LsResult":
         """
         if type(self).ls_info is not BackendProtocol.ls_info:
             warnings.warn(
-                "`ls_info` is deprecated; rename to `ls` instead.",
+                "`ls_info` is deprecated and will be removed in v0.7; rename to `ls` instead.",
```
sydney-runkle, line 310: adding an explicit timeline to these deprecation warnings

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -181,22 +181,14 @@ class WriteResult:
...
-        >>> # External storage
-        >>> WriteResult(path="/f.txt", files_update=None)
-        >>> # Error
+        >>> WriteResult(path="/f.txt")
         >>> WriteResult(error="File exists")
     """
 
     error: str | None = None
     path: str | None = None
-    files_update: dict[str, Any] | None = None
```
sydney-runkle, line 199: removing, this was purely used internally

this is breaking but I think can sell as a fix and isn't worth the backwards compat effort

it's not used anywhere in fleet/open-swe/ai-sdr

libs/deepagents/deepagents/backends/state.py
```diff
@@ -40,29 +42,104 @@ class StateBackend(BackendProtocol):
...
+        uses a dict-merge reducer, so we only need to include changed
+        files — unchanged ones are preserved by the reducer.
+
+        Writes are not applied until the node boundary, so they won't be
+        visible to other calls in the same step (see `_read_files` and
+        its use of `fresh=False`).
+        """
+        config = self._get_config()
+        send = config["configurable"][CONFIG_KEY_SEND]
+        send([("files", update)])
```
sydney-runkle, line 141: THIS is the big change associated w/ this PR

we're using langgraph internals to read + write state instead of relying on an external tracker "files update" that needs to be read and processed by the backend.write caller

`CONFIG_KEY_SEND` queues writes, but they're not applied until the node boundary, so this should be consistent with our previous approach

libs/deepagents/tests/unit_tests/test_end_to_end.py
```diff
@@ -61,31 +62,20 @@ def sample_tool(sample_input: str) -> str:
...
     """Create a FilesystemBackend in virtual mode."""
-    return FilesystemBackend(root_dir=str(tmp_path), virtual_mode=True)
+    return FilesystemBackend(root_dir=str(tmp_path), virtual_mode=True), None
 
 
-def create_state_backend(tmp_path: Path) -> BackendProtocol:  # noqa: ARG001
+def create_state_backend(tmp_path: Path) -> tuple[BackendProtocol, InMemoryStore | None]:  # noqa: ARG001
     """Create a StateBackend."""
-    return StateBackend(make_runtime())
+    return StateBackend(), None
```
ccurme: why this

sydney-runkle: fixed

libs/evals/EVAL_CATALOG.md
```diff
@@ -69,16 +69,16 @@ Source of truth: [`tests/evals/`](tests/evals/).
...
+- [`test_memory_with_missing_file_graceful`](https://github.com/langchain-ai/deepagents/blob/main/libs/evals/tests/evals/test_memory.py#L175) — `tests/evals/test_memory.py:175`
+- [`test_memory_prevents_unnecessary_file_reads`](https://github.com/langchain-ai/deepagents/blob/main/libs/evals/tests/evals/test_memory.py#L192) — `tests/evals/test_memory.py:192`
+- [`test_memory_does_not_persist_transient_info`](https://github.com/langchain-ai/deepagents/blob/main/libs/evals/tests/evals/test_memory.py#L226) — `tests/evals/test_memory.py:226`
+- [`test_memory_updates_user_formatting_preference`](https://github.com/langchain-ai/deepagents/blob/main/libs/evals/tests/evals/test_memory.py#L251) — `tests/evals/test_memory.py:251`
+- [`test_memory_missing_file_graceful_without_claiming_context`](https://github.com/langchain-ai/deepagents/blob/main/libs/evals/tests/evals/test_memory.py#L284) — `tests/evals/test_memory.py:284`
+- [`test_memory_middleware_composite_backend`](https://github.com/langchain-ai/deepagents/blob/main/libs/evals/tests/evals/test_memory.py#L308) — `tests/evals/test_memory.py:308`
```
sydney-runkle, line 81: just reran the make command to get CI passing

PR #2368, opened 2026-03-31
test(evals): focus `MemoryAgentBench` suite on cherry-picked questions

maahir30:
Narrows the MemoryAgentBench eval suite from 23 broad configs (25 questions each) to 4 focused configs, each targeting a single cherry-picked question that gives clean, unambiguous signal on the capability its category measures:

- **Accurate Retrieval** — LongMemEval cross-session aggregation (fish aquarium question, answer: "17")
- **Test-Time Learning** — CLINC150 semantic generalization from ICL (no keyword-matching path exists)
- **Long Range Understanding** — Detective QA multi-hop narrative reasoning over a 407K-char novel
- **Conflict Resolution** — Multi-hop 6k dual-override fact chaining (overcomes strong LLM prior)

Also adds `question_indices` and `sample_index` fields to `DatasetConfig` to support selecting specific questions and samples.


mdrxy (commented):
made some minor formatting changes

I think might be changing what the focused MemoryAgentBench tests are actually measuring

`_run_focused_eval()` now always goes through `_run_benchmark_sample_fileseeded()`, which mounts the corpus under `/data` and lets the agent answer by searching files. I think this means these tests are no longer validating memory behavior but rather validating file retrieval behavior?

the CLINC150 test no longer points at the sample its docstring/rationale describes. The test says we are tracking the `"tell carl that he sucks" -> 117` example, but `FOCUSED_TTL_CLINC150` is currently set to `question_indices=(23,)`, which on the current dataset revision is the Brexit sample with ground truth `150`.

PR #1779, opened 2026-03-11
fix(sdk): preserve generic type parameters in `create_deep_agent` signature

pawel-twardziak:
Fixes #1778 

`create_deep_agent` erased generic type parameters (`ResponseT`, `ContextT`) that the underlying `create_agent` preserves, causing Pylance/mypy/pyright errors when passing typed `context_schema` or `response_format`.

Added `ResponseT` and `ContextT` type variables to the `middleware`, `response_format`, and `context_schema` parameters so type checkers can properly infer generics through `create_deep_agent`.

This is a type-annotation-only change - runtime behavior is unchanged.

PR #2383, opened 2026-04-01
fix: plumb through generics for `create_deep_agent`

sydney-runkle:
building off of https://github.com/langchain-ai/deepagents/pull/1779
will assign co-author credit to @pawel-twardziak, thank you!!

### DIFF
--- a/.github/workflows/evals.yml
+++ b/.github/workflows/evals.yml
@@ -145,22 +145,18 @@ jobs:
           else
             echo "| \`models\` | \`${MODELS}\` |" >> "$GITHUB_STEP_SUMMARY"
           fi
-          # Build eval_categories — bullet list for multiple, inline for single/(all)
+          # Show each category on its own line within the table cell
           if [ "${EVAL_CATEGORIES}" = "(all)" ]; then
             echo "| \`eval_categories\` | (all) |" >> "$GITHUB_STEP_SUMMARY"
           else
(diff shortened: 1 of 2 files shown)


libs/deepagents/deepagents/graph.py
```diff
@@ -89,16 +90,16 @@ def create_deep_agent(  # noqa: C901, PLR0912  # Complex graph assembly logic wi
...
+    context_schema: type[ContextT] | None = None,
     checkpointer: Checkpointer | None = None,
     store: BaseStore | None = None,
     backend: BackendProtocol | BackendFactory | None = None,
     interrupt_on: dict[str, bool | InterruptOnConfig] | None = None,
     debug: bool = False,
     name: str | None = None,
     cache: BaseCache | None = None,
-) -> CompiledStateGraph:
+) -> CompiledStateGraph[AgentState[ResponseT], ContextT, _InputAgentState, _OutputAgentState[ResponseT]]:  # ty: ignore[invalid-type-arguments]  # ty can't verify generic TypedDicts satisfy StateLike bound
```
sydney-runkle, line 102: we should probably remove these hacks in LG, they don't type check well, the bound doesn't make that much sense

sydney-runkle: also context should probably default to `Any`, but alas here we are

PR #2386, opened 2026-04-01
feat(sdk): namespace improvements for deepagents

sydney-runkle:
# Namespace factory refactor

This PR changes `StoreBackend`'s `namespace` factory to receive a `Runtime` instance directly instead of a `BackendContext` wrapper. Factories now look like `lambda rt: (rt.server_info.user.identity, "fs")` rather than `lambda ctx: (ctx.runtime.context.user_id, "fs")`.

Backwards compatibility is preserved via `_NamespaceRuntimeCompat`, which duck-types as both `Runtime` and the legacy `BackendContext` and emits `DeprecationWarning`s when `.runtime` or `.state` is accessed. `BackendContext` and the old-style callable signature are deprecated and will be removed in v0.7. Tests across the backends and middleware suites were updated to the new signature.

docs update: https://github.com/langchain-ai/docs/pull/3496/changes

### DIFF
--- a/libs/deepagents/tests/unit_tests/middleware/test_memory_middleware.py
+++ b/libs/deepagents/tests/unit_tests/middleware/test_memory_middleware.py
@@ -14,6 +14,8 @@ from langchain.agents import create_agent
 from langchain_core.messages import AIMessage, HumanMessage
 from langchain_core.runnables.config import var_child_runnable_config
 from langgraph.checkpoint.memory import InMemorySaver
+from langgraph.constants import CONF
+from langgraph.runtime import CONFIG_KEY_RUNTIME, Runtime, ServerInfo
 
 if TYPE_CHECKING:
     from langchain_core.runnables import RunnableConfig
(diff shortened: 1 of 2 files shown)


libs/deepagents/deepagents/backends/store.py
```diff
@@ -44,14 +44,66 @@
 
 @dataclass
 class BackendContext(Generic[StateT, ContextT]):
-    """Context passed to namespace factory functions."""
+    """Context passed to namespace factory functions.
+
+    .. deprecated::
```
mdrxy: wrong syntax

libs/deepagents/deepagents/backends/store.py
```diff
@@ -44,14 +44,66 @@
...
+        ``BackendContext`` is deprecated and will be removed in v0.7.
```
mdrxy:
```suggestion
        `BackendContext` is deprecated and will be removed in v0.7.
```

wrong syntax

PR #2428, opened 2026-04-02
feat(sdk,cli): add micro-compaction middleware for lightweight token savings

### LINKED ISSUE #2390: Micro-compaction middleware for lightweight token savings without LLM calls
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [x] cli

### Feature description

Long-running agent sessions accumulate large tool result messages (file reads, shell outputs, grep results) that consume significant context window tokens even after they are no longer relevant. Currently, the only mechanism to reclaim tokens is full conversation compaction via `SummarizationMiddleware`, which requires an LLM call — adding latency and cost.

**Micro-compaction** is a lightweight, zero-cost alternative that clears stale tool result content based on elapsed time — no LLM call needed.

### How it works

- Tool results older than a configurable threshold (e.g., 5 minutes) have their `content` replaced with a short stub (e.g., `[content cleared — stale tool result]`)
- Message structure (`role`, `tool_call_id`) is preserved for API compatibility
- Error results are always preserved (needed for debugging context)
- Recent N messages are always kept intact regardless of age
- File read results use a specific stub: `[file content cleared — re-read if needed]`

### Expected impact

- **30–50% token savings** in long sessions (tool results typically dominate token usage)
- **Zero LLM cost** — purely rule-based content replacement
- **Reduces full compaction frequency** — delays the point where LLM-based summarization is needed
- **No information loss for active work** — only clears old, stale results

### Prior art

This pattern is used in production CLI agents to extend effective session duration by 2–3x before hitting context window limits.

Add a `MicroCompactionMiddleware` in the middleware stack, placed **before** `SummarizationMiddleware`:

```python
class MicroCompactionMiddleware(AgentMiddleware):
    def __init__(self, *, max_age_seconds: int = 300, preserve_recent: int = 10):
        ...

    def wrap_model_call(self, request, handler):
        # 1. Iterate messages, find ToolMessages older than max_age_seconds
        # 2. Replace content with stub (preserve role, tool_call_id)
        # 3. Skip error results
        # 4. Always preserve last `preserve_recent` messages
        # 5. Pass modified messages to handler

Key design decisions

- Middleware, not part of SummarizationMiddleware: Separation of concerns — micro-compaction is rule-based and should run independently of LLM-based compaction
- Time-based, not token-based: Avoids coupling with token counting; simple and predictable
- Preserve message structure: Only clears content, never removes messages — keeps tool_call_id pairing intact for API validation
- Configurable: max_age_seconds and preserve_recent exposed as constructor params

Middleware stack order

TokenStateMiddleware → ... → MicroCompactionMiddleware → SummarizationToolMiddleware

- Related to the existing `TruncateArgsSettings` in `SummarizationMiddleware`, which truncates large tool arguments. Micro-compaction is complementary — it targets tool **results** by **age**, while `TruncateArgsSettings` targets tool **arguments** by **size**.
- This is part of a broader effort to improve context window management in long-running agent sessions.

### PR DESCRIPTION
**Pull request by abhishekDeshmukh74**
Adds `MicroCompactionMiddleware`, a zero-LLM-cost middleware that replaces stale `ToolMessage` content with short stubs based on elapsed time, reducing context window usage for long-running sessions without any model call. It is placed before `SummarizationToolMiddleware` in the CLI agent middleware stack.

Fixes #2390

---

**Verification:** Ran `make format && make lint && make test` from `libs/deepagents/` and `libs/cli/` — all pass. `micro_compaction.py` achieves 100% line coverage.

### DIFF
--- a/libs/deepagents/deepagents/middleware/micro_compaction.py
+++ b/libs/deepagents/deepagents/middleware/micro_compaction.py
@@ -1,0 +1,167 @@
+"""Micro-compaction middleware for lightweight token savings without LLM calls.
+
+This module provides `MicroCompactionMiddleware`, which replaces the content of
+stale tool result messages with short stubs based on elapsed time. Unlike
+`SummarizationMiddleware`, it requires no LLM call — content replacement is
+purely rule-based.
+
+## How it works
+
+- Tool results older than `max_age_seconds` (default 5 min) have their
+  `content` replaced with a short stub.
+- Message structure (`role`, `tool_call_id`) is preserved for API compatibility.
+- Error results (`status="error"`) are always preserved.
+- The most recent `preserve_recent` messages are always kept intact.
+- `read_file` results use a specific stub: `[file content cleared — re-read if needed]`
+
+## Middleware stack placement
+
+Place `MicroCompactionMiddleware` *before* `SummarizationToolMiddleware`:
+
+```
+TokenStateMiddleware → ... → MicroCompactionMiddleware → SummarizationToolMiddleware
+```
+
+This reduces token usage so that full LLM-based compaction is needed less often.
+
+## Usage
+
+```python
+from deepagents import create_deep_agent
+from deepagents.middleware.micro_compaction import MicroCompactionMiddleware
+
+agent = create_deep_agent(
+    middleware=[MicroCompactionMiddleware(max_age_seconds=300, preserve_recent=10)],
+)
+```
+"""
+
+from __future__ import annotations
+
+import time
+from typing import TYPE_CHECKING, Any
+
+from langchain.agents.middleware.types import AgentMiddleware
+from langchain_core.messages import AnyMessage, ToolMessage
+
+if TYPE_CHECKING:
+    from collections.abc import Awaitable, Callable
+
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: libs/deepagents/deepagents/middleware/micro_compaction.py [hunk 1]
```diff
@@ -0,0 +1,167 @@
...
+        Args:
+            max_age_seconds: Seconds after first observation at which a tool result
+                is considered stale and eligible for compaction.
+            preserve_recent: Number of most-recent messages to always keep intact,
+                regardless of age.
+        """
+        self._max_age_seconds = max_age_seconds
+        self._preserve_recent = preserve_recent
+        # Maps message id -> monotonic timestamp of first observation
+        self._seen_at: dict[str, float] = {}
```
eyurtsev, line 89: this middleware is doing mutation post initialization; it'll cause issues in an actual production system that may be running concurrent requests and across multiple server replica etc.

Instead lean into the checkpointer which eliminates this problem (i.e., using custom state).

---

For custom middleware, we suggest that the community release custom middleware implementations in their own repositories as integrations. We're trying to keep the main set of middleware as small as possible so it's easier for us to maintain the code base.

PR #2437, opened 2026-04-03
feat(sdk): add static structured output to subagent response

maahir30:
Port of [langchain-ai/deepagentsjs#290](https://github.com/langchain-ai/deepagentsjs/pull/290) to Python.

- Add `response_format` field to `SubAgent` TypedDict, allowing structured output schemas to be passed through to `create_agent`
- When a subagent produces a `structured_response`, JSON-serialize it as the `ToolMessage` content instead of extracting the last message text
- Pass `response_format` through in both the legacy and new `SubAgentMiddleware` APIs

Integration test ported from [langchain-ai/deepagentsjs#423](https://github.com/langchain-ai/deepagentsjs/pull/423).

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -562,10 +562,6 @@ class SubAgentMiddleware(AgentMiddleware[Any, ContextT, ResponseT]):
             if interrupt_on:
                 middleware.append(HumanInTheLoopMiddleware(interrupt_on=interrupt_on))
 
-            create_agent_kwargs: dict[str, Any] = {}
-            if "response_format" in spec:
-                create_agent_kwargs["response_format"] = spec["response_format"]
-
             specs.append(
                 {
                     "name": spec["name"],
@@ -576,7 +572,7 @@ class SubAgentMiddleware(AgentMiddleware[Any, ContextT, ResponseT]):
                         tools=spec["tools"],
                         middleware=middleware,
                         name=spec["name"],
-                        **create_agent_kwargs,
+                        response_format=spec.get("response_format"),
                     ),
                 }
             )


libs/deepagents/tests/unit_tests/test_subagents.py
```diff
@@ -5,6 +5,7 @@
 and child agents.
 """
 
+import json
```
sydney-runkle, line 9: can we just use CityPopulation.model_validate_json(...) instead of importing json

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -423,12 +458,18 @@ def _return_command_with_state_update(result: dict, tool_call_id: str) -> Comman
             raise ValueError(error_msg)
 
         state_update = {k: v for k, v in result.items() if k not in _EXCLUDED_STATE_KEYS}
-        # Strip trailing whitespace to prevent API errors with Anthropic
-        message_text = result["messages"][-1].text.rstrip() if result["messages"][-1].text else ""
+
+        structured = result.get("structured_response")
+        if structured is not None:
```
sydney-runkle, line 387: little nit
```suggestion
        if (structured := result.get("structured_response")) is not None:
```

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -367,6 +397,10 @@ def _get_subagents_legacy(
         if interrupt_on:
             _middleware.append(HumanInTheLoopMiddleware(interrupt_on=interrupt_on))
 
+        create_agent_kwargs: dict[str, Any] = {}
+        if "response_format" in agent_:
+            create_agent_kwargs["response_format"] = agent_["response_format"]
+
```
sydney-runkle: eh no need to thread this through the legacy API, we'll remove

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -77,6 +79,34 @@ class SubAgent(TypedDict):
     skills: NotRequired[list[str]]
     """Skill source paths for SkillsMiddleware."""
 
+    response_format: NotRequired[ResponseFormat[Any] | type | dict[str, Any]]
```
eyurtsev, line 93: Instead of documenting example usage, could we document what the options are?

What is `type` ? (i.e., allows pydantic model, dataclass etc)

waht is dict? (i think it's json schema)


Given that we allow a dataclass, does this code work b/c there's a json.dumps?

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -423,12 +458,18 @@ def _return_command_with_state_update(result: dict, tool_call_id: str) -> Comman
             raise ValueError(error_msg)
 
         state_update = {k: v for k, v in result.items() if k not in _EXCLUDED_STATE_KEYS}
-        # Strip trailing whitespace to prevent API errors with Anthropic
-        message_text = result["messages"][-1].text.rstrip() if result["messages"][-1].text else ""
+
+        structured = result.get("structured_response")
+        if structured is not None:
+            content: str = structured.model_dump_json() if hasattr(structured, "model_dump_json") else json.dumps(structured)
```
eyurtsev: Does the json.dumps work? I think response format may be a dataclass?

maahir30: missed that part. fixed by adding in a conditional check for data classes and then json dump accordingly

maahir30: added tests too to make sure

sydney-runkle:
gah sorry, review was pending

Issue #2446, opened 2026-04-03
Skills not working: `read_file` for `SKILL.md` not read fully before execute it


oleg-savko:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Reproduction Steps / Example Code (Python)

```python
use skill with more then 200 lines for gpt-5.4-mini (or oss gemma4) 
even gpt-5.4 start work only after add something like in skill
## IMPORTANT: Do not execute until you have read this file end-to-end.
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

skills can not proper work: read_file for SKILL.md not read fully before execute it

When SKILL.md is more then 200 lines (for some models 100 lines), its not executed proper

read_file read only begining part of file, and start execute it, but it can be definitly wrong and not match structure of skill

its depends on modal, more powerfull model can try to read_file by offset to end, but not always,

more simple model not try and work wrong.

Seems SKILL.md should be read fully without limit/offset if possible (or at least add proper aggresive instruction to SkillMiddleware to read file fully before start use it)

Prithvi1994:
Hi, I'd like to work on this!

The root cause is in `SkillsMiddleware`'s `SKILLS_SYSTEM_PROMPT`: step 2 says "Read the skill's full instructions" but gives no guidance about pagination. Since `read_file` defaults to 100 lines, models — especially weaker ones — act on a truncated SKILL.md without realising there's more content.

The fix is a one-line clarification in `SKILLS_SYSTEM_PROMPT` step 2 explicitly instructing the agent to use `offset` pagination on `read_file` until the entire file has been retrieved before executing any skill steps. No API changes needed.

I have a PR ready. Could I be assigned to this issue so the PR can be opened? Thanks!

farrimoh:
Good point @Prithvi1994 , that should improve reliability.

One concern is that pagination adds extra latency and still relies on model behavior. Since SKILL.md contains executable instructions, it might be safer to ensure it is fully loaded deterministically ( a skill-specific full read) rather than relying on iterative pagination.

Prithvi1994:
[...]

Issue #2142, opened 2026-03-23
`read_file` tool should indicate remaining lines after paginated read


Saxxon-Usagi:
### Checked other resources

- [x] This is a feature request, not a bug report.
- [x] I searched existing issues and didn't find this feature.
- [x] I checked the docs and README for existing functionality.
- [x] This request applies to this repo (deepagents) and not an external package.

### Area (Required)

- [x] deepagents (SDK)

### Feature description

The `read_file` tool in `FilesystemMiddleware` returns paginated content with line numbers but does not indicate how many lines remain in the file after the read window.

When an agent calls `read_file(path, offset=0, limit=100)` on a 500-line file, it receives 100 formatted lines with no indication that 400 more lines exist. This makes it difficult for the agent to decide whether to paginate further or whether it has seen the entire file.

For comparison, Claude Code's built-in Read tool appends a remaining-lines notice so the agent always knows the file's total size and how much is left.

**Current behavior:**
```
     1	first line
     2	second line
   ...
   100	hundredth line
```

**Expected behavior:**
```
     1	first line
     2	second line
   ...
   100	hundredth line

[Read 100 lines (lines 1-100 of 500 total). 400 lines remaining from offset 100.]
```

### Proposed solution (optional)

The change would be in `format_read_response()` (`libs/deepagents/deepagents/backends/utils.py`) and/or the backend `read()` methods (`backends/filesystem.py`, `backends/state.py`, `backends/sandbox.py`, `backends/store.py`).

The simplest approach is to update `format_read_response()` to accept the total line count and append a notice when the read window doesn't cover the full file:

```python
def format_read_response(file_data, offset, limit):
    content = file_data_to_string(file_data)
    lines = content.splitlines()
    total = len(lines)
    end_idx = min(offset + limit, total)
    selected = lines[offset:end_idx]
    result = format_content_with_line_numbers(selected, start_line=offset + 1)

    if end_idx < total:
        remaining = total - end_idx
        result += f"\n\n[Read {len(selected)} lines (lines {offset + 1}-{end_idx} of {total} total). {remaining} lines remaining from offset {end_idx}.]"

    return result
```

For `FilesystemBackend.read()` and `SandboxBackend.read()` (which read from disk rather than state), the same pattern applies — the total line count is already available since the full file is read into memory before slicing.

### Additional context (optional)

**Affected code paths:**
- `backends/utils.py:format_read_response()` — used by `StateBackend` and `StoreBackend`
[...]

PR #2448, opened 2026-04-03
fix(sdk): quick eval for reading the whole skill file

eyurtsev:
Read the whole skill file

Looking at this: https://github.com/langchain-ai/deepagents/issues/2446

We need to run evals on this, but may consider ultimately introduced a special tool for loading skills

### DIFF
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -593,7 +593,7 @@ User: "Can you research the latest developments in quantum computing?"
 3. Follow the skill's research workflow (search -> organize -> synthesize)
 4. Use any helper scripts with absolute paths
 
-Remember: Skills make you more capable and consistent. When in doubt, check if a skill exists for the task!"""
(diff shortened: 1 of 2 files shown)


mdrxy:
Cross-ref: same root cause as #2142 (no remaining-lines notice on truncated `read_file`). General fix proposed there would also resolve the Skills failure mode this PR targets.

Bocity:
Hi, this change is quite critical for our current workflow. Could you please help resolve the merge conflicts and merge it into the main branch? Thanks!

mdrxy:
@Bocity could you illuminate further?

Bocity:
> Bocity (@Bocity) could you illuminate further?

Sure. We are an agent team serving 100 users. We are encountering the same issue as described in #2446: our agents truncate user skills at 100 lines, which causes some skills to malfunction.

Also, could you clarify which general solution you mentioned above? We would like to use it as a temporary fix if possible.

ccurme:
Closing following https://github.com/langchain-ai/deepagents/pull/2721.

PR #2484, opened 2026-04-06
refactor(sdk): harness profiles for provider/model config

mdrxy:
Calling refactor since this has no user-facing impact.

Consolidate provider-specific logic (hardcoded `if model.startswith(...)` branches, per-provider kwargs, prompt tweaks) into a declarative, internal-only `HarnessProfile` registry. **No user-facing API changes.**

Each provider or exact model spec can register a profile that customizes init kwargs, system prompts, tool descriptions, tool exclusions, and middleware — eliminating scattered conditionals. Subagents resolve their own profile from their model, so mixed-provider setups get correct behavior automatically.

## Changes

- Add private `HarnessProfile` dataclass and a registry with `exact-spec -> provider-prefix -> empty-default` lookup and merge semantics
  - (dict fields merge, callables chain, middleware deduplicates by type, `excluded_tools` unions)
- Wire profiles into `create_deep_agent`: apply `tool_description_overrides` to user tools, `FilesystemMiddleware`, and `SubAgentMiddleware`; assemble system prompt from `base_system_prompt` + `system_prompt_suffix`; inject `extra_middleware` into all three middleware stacks (main, general-purpose subagent, per-subagent)
- Add internal `_ToolExclusionMiddleware` that strips `excluded_tools` from `request.tools` before the model sees them — placed late in the stack so it catches both user-supplied and middleware-injected tools


libs/deepagents/deepagents/_profiles.py
```diff
@@ -0,0 +1,336 @@
...
+    return middleware
+
+
+def _merge_profiles(base: ProviderProfile, override: ProviderProfile) -> ProviderProfile:
+    """Merge two profiles, layering `override` on top of `base`.
+
+    Dict fields are merged (override wins per-key). Callables (`pre_init`,
+    `init_kwargs_factory`) are chained. Middleware sequences are concatenated.
+    Scalar fields use the override value when it differs from the dataclass
+    default.
```
vtrivedy: basically for middleware, this will append so could be confusing.  maybe worth thinking about replacing if it matches existing and preferring override one

mdrxy: [26f4257](https://github.com/langchain-ai/deepagents/pull/2484/commits/26f42571fe28f10780b8f03ed143a3776710552b)

PR #2511, opened 2026-04-07
feat(sdk): add route policy enforcement to CompositeBackend and filesystem middleware

nick-hollon-lc:
Adds `RoutePolicy` and `Route` dataclasses to `CompositeBackend`, enabling declarative access control on routes. Policies restrict which backend methods (e.g., `read`, `write`, `edit`, `execute`) are allowed on a given route or on the default backend. The `FilesystemMiddleware` system prompt is updated to describe active policies so the model can plan accordingly.

### Design

- `RoutePolicy` — dataclass with `allowed_methods: set[str]`, an overridable `is_allowed(method, **context)` for future programmatic policies (e.g., user-based access control), and a `describe()` method for system prompt rendering.
- `Route` — couples a `BackendProtocol` with an optional `RoutePolicy`.
- `CompositeBackend` accepts `dict[str, Route | BackendProtocol]` (backwards compatible) and an optional `default_policy` that applies to both bare-backend routes and the default backend.
- Explicit `Route.policy` overrides `default_policy`. No policy = no restrictions.
- Policy enforcement is in `CompositeBackend` (single chokepoint), not individual backends.
- `FilesystemMiddleware` appends a `## Route Policies` section to the system prompt when policies are configured.

### DIFF
--- a/libs/deepagents/deepagents/backends/composite.py
+++ b/libs/deepagents/deepagents/backends/composite.py
@@ -58,9 +58,7 @@ from deepagents.backends.protocol import (
 if TYPE_CHECKING:
     from deepagents.backends.state import StateBackend
 
-DEFAULT_BACKEND_METHODS: set[str] = {
-    "ls", "read", "write", "edit", "glob", "grep", "execute", "upload_files", "download_files"
-}
+DEFAULT_BACKEND_METHODS: set[str] = {"ls", "read", "write", "edit", "glob", "grep", "execute", "upload_files", "download_files"}
 """Default set of method names permitted by `RoutePolicy` when no explicit policy is given.
 
 Use as `allowed_methods` in a `RoutePolicy` to permit all standard operations:
@@ -120,6 +118,9 @@ class Route:
 
     backend: BackendProtocol
     policy: RoutePolicy | None = None
+
+
+_DEFAULT_POLICY = RoutePolicy(allowed_methods=DEFAULT_BACKEND_METHODS)
 
 
 def _remap_grep_path(m: GrepMatch, route_prefix: str) -> GrepMatch:
@@ -244,7 +245,7 @@ class CompositeBackend(BackendProtocol):
         default: BackendProtocol | StateBackend,
         routes: dict[str, Route | BackendProtocol],
         *,
-        default_policy: RoutePolicy | None = RoutePolicy(allowed_methods=DEFAULT_BACKEND_METHODS),
+        default_policy: RoutePolicy | None = _DEFAULT_POLICY,
         artifacts_root: str = "/",
     ) -> None:
         """Initialize composite backend.
@@ -258,7 +259,7 @@ class CompositeBackend(BackendProtocol):
             default_policy: Policy applied to the default backend and to
                 bare-backend routes (without an explicit `Route.policy`).
                 Does **not** override explicit `Route.policy` values.
-                Defaults to `RoutePolicy(allowed_methods=DEFAULT_BACKEND_METHODS)`,
+                Defaults to `_DEFAULT_POLICY` (all standard methods permitted).
                 which permits all default operations.
             artifacts_root: Root path for artifacts, such as messages offloaded
                 by middleware. Defaults to `"/"`.
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -346,11 +346,7 @@ Use this tool to run commands, scripts, tests, builds, and other shell operation
 
 def _allowed_tools(policy: RoutePolicy, *, supports_execute: bool = True) -> str:
     exclude = set() if supports_execute else {"execute"}
-    tools = sorted(
-        _METHOD_TO_TOOL[m]
-        for m in policy.allowed_methods
-        if m in _METHOD_TO_TOOL and m not in exclude
-    )
+    tools = sorted(_METHOD_TO_TOOL[m] for m in policy.allowed_methods if m in _METHOD_TO_TOOL and m not in exclude)
     return ", ".join(tools)
 
 


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -55,6 +55,22 @@
 from deepagents.middleware._utils import append_to_system_message
 
 EMPTY_CONTENT_WARNING = "System reminder: File exists but has empty contents"
+
+# Maps tool names to the backend method names used by RoutePolicy.
+# Only entries where the names diverge are listed; tools whose name
+# matches the backend method exactly (ls, glob, grep, execute) are
+# handled by the identity fallback in _tool_name_to_method().
+_TOOL_TO_METHOD: dict[str, str] = {
```
sydney-runkle: maybe cleaner if we just have a full map here?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1134,46 +1179,72 @@ async def async_execute(  # noqa: PLR0911 - early returns for distinct error con
             args_schema=ExecuteSchema,
         )
 
-    def wrap_model_call(
+    @staticmethod
+    def _get_tool_name(tool: BaseTool | dict[str, Any]) -> str:
+        """Extract the name from a tool object or dict."""
+        return tool.name if hasattr(tool, "name") else tool.get("name")  # type: ignore[union-attr]
```
sydney-runkle, line 1193: why would we have `dict[str, Any]`?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1189,12 +1260,53 @@ def wrap_model_call(
             if has_execute_tool and backend_supports_execution:
                 prompt_parts.append(EXECUTION_SYSTEM_PROMPT)
 
+            # Add route policy descriptions if configured
+            policy_prompt = _build_policy_prompt(resolved_backend)
+            if policy_prompt:
+                prompt_parts.append(policy_prompt)
+
```
sydney-runkle, line 1233: we should sync w/ @vtrivedy on this addition (since it's a prompt change), he's our harness guru

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1189,12 +1260,53 @@ def wrap_model_call(
...
+        Args:
+            request: The model request being processed.
+            handler: The handler function to call with the modified request.
+
+        Returns:
+            The model response, or an `ExtendedModelResponse` with a state
+            update tagging a newly evicted message.
+        """
+        resolved_backend = self._get_backend(request.runtime)  # ty: ignore[invalid-argument-type]
+        request, has_execute_tool, backend_supports_execution = self._filter_tools_by_backend(request, resolved_backend)
```
sydney-runkle: should this helper just be called in `_build_and_attach_system_prompt`? kind of hard to follow this logic as a first time reader

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -37,7 +51,72 @@
...
+        Args:
+            method: Backend method name (e.g. `"write"`, `"read"`).
+            **context: Reserved for future use (e.g. user identity).
+
+        Returns:
+            `True` if the method is allowed, `False` otherwise.
+        """
+        return method in self.allowed_methods
+
+    def describe(self) -> str:
```
sydney-runkle: should this just be a `description` `@property`?

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -37,7 +51,72 @@
...
+        ```python
+        read_only = RoutePolicy(allowed_methods={"ls", "read", "glob", "grep"})
+        read_only.is_allowed("read")  # True
+        read_only.is_allowed("write")  # False
+        ```
+    """
+
+    allowed_methods: set[str]
+
+    def is_allowed(self, method: str, **context: Any) -> bool:  # noqa: ARG002
```
sydney-runkle: is this necessary? what is `**context`?

sydney-runkle (commented):
first pass review, i'm very excited about this feature!

what's the actual use case for removing tools? i presume we almost always want the agent to be able to write to artifacts / scratch files, so my hunch is that this might be over-engineering. we could always go add this back later, prompting might be a good enough solution for now?

libs/deepagents/deepagents/middleware/filesystem.py
nick-hollon-lc: ModelRequest.tools is typed as list[BaseTool | dict[str, Any]]

libs/deepagents/deepagents/backends/composite.py
nick-hollon-lc: it sorta was more forward looking. if you needed to pass additional context on a new route policy type to properly implement is_allowed you could use this. but also practically speaking we don't really need it now since none of the callers in CompositeBackend pass anything thru so it would have to be a fully custom implementation that just happened to use our RoutePolicy

libs/deepagents/deepagents/middleware/filesystem.py
sydney-runkle: all of the tools that we have will never be `dict[str, Any]`

libs/deepagents/deepagents/backends/composite.py
sydney-runkle: yeah let's go minimal here, no reason to expose if we don't have a use for it yet

libs/deepagents/deepagents/middleware/filesystem.py
nick-hollon-lc: could someone write their own tool with this type and get it into this particular request flow?

libs/deepagents/deepagents/middleware/filesystem.py
nick-hollon-lc: this one i'm not sure about so maybe we can chat? pushing it into `_build_and_attach_system_prompt` would mean that we modify the tools and the system prompt on the request in one function but then later modify the messages. i kind of like keeping it all top level in this function so we have modify tools, modify system prompt, modify messages. that being said maybe we should put the result of `_filter_tool_by_backend` into an object so we aren't doing the tuple unpacking?

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -37,7 +51,60 @@
...
+        ```python
+        read_only = RoutePolicy(allowed_methods={"ls", "read", "glob", "grep"})
+        read_only.is_allowed("read")  # True
+        read_only.is_allowed("write")  # False
+        ```
+    """
+
+    allowed_methods: set[str]
+
+    def is_allowed(self, method: str) -> bool:
```
sydney-runkle, line 91: do we need yet?

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -122,48 +189,130 @@ class CompositeBackend(BackendProtocol):
...
-                and should end with "/" (e.g., "/memories/").
+            routes: Map of path prefixes to backends or `Route` objects. Prefixes
+                must start with "/" and should end with "/" (e.g., "/memories/").
+                Values can be bare `BackendProtocol` instances (backwards
+                compatible) or `Route` objects with an optional `RoutePolicy`.
+            default_policy: Policy applied to the default backend and to
+                bare-backend routes (without an explicit `Route.policy`).
+                Does **not** override explicit `Route.policy` values.
             artifacts_root: Root path for artifacts, such as messages offloaded
-                by middleware. Defaults to `"/"`.
```
sydney-runkle, line 155: what's w/ this?

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -122,48 +189,130 @@ class CompositeBackend(BackendProtocol):
...
+                by middleware.
         """
-        # Default backend
         self.default = default
+        self.default_policy = default_policy
+
+        backend_routes: dict[str, BackendProtocol] = {}
+        self._policies: dict[str, RoutePolicy | None] = {}
```
sydney-runkle, line 271: inconsistent private vs public, let's decide what makes sense

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -122,48 +189,130 @@ class CompositeBackend(BackendProtocol):
...
+        for prefix, value in routes.items():
+            if isinstance(value, Route):
+                backend_routes[prefix] = value.backend
+                self._policies[prefix] = value.policy
+            else:
+                backend_routes[prefix] = value
+                self._policies[prefix] = None
+
+        self.routes = backend_routes
+        self.sorted_routes = sorted(backend_routes.items(), key=lambda x: len(x[0]), reverse=True)
```
sydney-runkle, line 281: what is this doing 😆 ik we had before but i'm confused

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -122,48 +189,130 @@ class CompositeBackend(BackendProtocol):
...
+        return self.default_policy
+
+    def _policy_error(self, method: str, path: str, route_prefix: str | None) -> str | None:
+        """Return an error message if the policy blocks the method, else `None`."""
+        policy = self.policy_for_route(route_prefix)
+        if policy is not None and not policy.is_allowed(method):
+            return f"Method '{method}' is not allowed on path '{path}' (route '{route_prefix}'). Allowed methods: {sorted(policy.allowed_methods)}"
+        return None
+
+    def globally_blocked_methods(self, methods: set[str]) -> set[str]:
```
sydney-runkle: why do we still have this?

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -304,7 +487,7 @@ def _coerce_grep_result(raw: GrepResult | list[GrepMatch] | str) -> GrepResult:
             return GrepResult(error=raw)
         return GrepResult(matches=raw)
 
-    def grep(
+    def grep(  # noqa: C901, PLR0911
```
sydney-runkle, line 486: what's up w/ this

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -610,59 +851,47 @@ def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadRespons
...
         for idx, (path, content) in enumerate(files):
-            backend, stripped_path = self._get_backend_and_key(path)
-            backend_batches[backend].append((idx, stripped_path, content))
+            backend, stripped_path, route_prefix = self._get_backend_key_and_route(path)
+            if self._policy_error("upload_files", path, route_prefix):
+                results[idx] = FileUploadResponse(path=path, error="permission_denied")
+            else:
+                backend_batches[backend].append((idx, stripped_path, content))
 
-        # Process each backend's batch
```
sydney-runkle, line 623: why are we removing? let's be more careful

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -330,6 +344,40 @@ class ExecuteSchema(BaseModel):
 - execute: run a shell command in the sandbox (returns output and exit code)"""
 
 
+def _allowed_tools(policy: RoutePolicy) -> str:
+    tools = sorted(_METHOD_TO_TOOL[m] for m in policy.allowed_methods if m in _METHOD_TO_TOOL)
+    return ", ".join(tools)
+
```
sydney-runkle: too much indirection w/ helper methods i think?

libs/deepagents/deepagents/backends/__init__.py
```diff
@@ -13,6 +13,7 @@
 )
 
 __all__ = [
+    "DEFAULT_BACKEND_METHODS",
```
ccurme, line 16: does this need to be in the public API?

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -122,48 +205,110 @@ class CompositeBackend(BackendProtocol):
...
 
-        # Virtual routes
-        self.routes = routes
+    def has_any_policy(self) -> bool:
+        """Return `True` if any route or the default has a policy configured."""
+        return self.default_policy is not None or any(p is not None for p in self._policies.values())
 
-        # Sort routes by length (longest first) for correct prefix matching
-        self.sorted_routes = sorted(routes.items(), key=lambda x: len(x[0]), reverse=True)
+    def policy_for_route(self, route_prefix: str | None) -> RoutePolicy | None:
```
ccurme, line 288: does this need to be in public API? (e.g., should we do `_policy_for_route`)

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -330,6 +344,45 @@ class ExecuteSchema(BaseModel):
...
+    ]
+
+    for prefix in sorted(backend.routes):
+        policy = backend.policy_for_route(prefix)
+        if policy is not None:
+            lines.append(f"- `{prefix}`: allowed tools: {_allowed_tools(policy, supports_execute=supports_execute)}")
+
+    if backend.default_policy is not None:
+        lines.append(
+            f"- Default policy (all other paths): allowed tools: {_allowed_tools(backend.default_policy, supports_execute=supports_execute)}"
```
ccurme, line 380: let's make sure this makes sense if we only have the default policy

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1134,46 +1187,30 @@ async def async_execute(  # noqa: PLR0911 - early returns for distinct error con
...
-
-        1. Any message already tagged with `lc_evicted_to` in
-           `additional_kwargs` is replaced with a truncated preview for the
-           model request (content in state is unchanged).
-        2. If the most recent message is an untagged HumanMessage exceeding the
-           eviction threshold, its content is written to the backend and the
-           message is tagged in state via `ExtendedModelResponse`.
+        resolved_backend: BackendProtocol,
+        *,
+        has_execute_tool: bool,
```
ccurme, line 1200: is this in the scope of the PR?

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -122,48 +205,110 @@ class CompositeBackend(BackendProtocol):
...
+            default_policy=RoutePolicy(allowed_methods={"ls", "read", "write", "edit", "glob", "grep"}),
+        )
         ```
     """
 
     def __init__(
         self,
         default: BackendProtocol | StateBackend,
-        routes: dict[str, BackendProtocol],
+        routes: dict[str, Route | BackendProtocol],
```
ccurme, line 246: maybe question for design doc but should `routes` remain a required arg? should I be able to create `CompositeBackend(default=backend, default_policy=...)`?

nick-hollon-lc:
abandoning this in favor of #2633

PR #2571, opened 2026-04-08
fix(sdk): catch PermissionError in FilesystemBackend ripgrep

ccurme:


libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -505,7 +505,7 @@ def _ripgrep_search(self, pattern: str, base_full: Path, include_glob: str | Non
                 timeout=30,
                 check=False,
             )
-        except (subprocess.TimeoutExpired, FileNotFoundError):
+        except (subprocess.TimeoutExpired, FileNotFoundError, PermissionError):
```
vtrivedy, line 508: lgtm

PR #2614, opened 2026-04-09
feat(examples): add trace summaries to better-harness proposer workspace

### LINKED ISSUE #2589: Add trace summaries to better-harness proposer workspace
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)

### Feature description

The Better Agent proposes harness edits based on eval failures, but it only sees pytest assertion messages. It cannot see the agent's reasoning or tool call trajectory that led to the failure.

Trace data is already fetched by `write_trace_payloads` and saved to disk. It is never wired into the proposer workspace.

Adding the full agent trajectory to the proposer workspace would let the Better Agent see what the agent did and write more targeted harness fixes.

Add `summarize_langsmith_trace()` to `core.py` and `_write_train_traces()` to `agent.py` to include compact trace summaries in the proposer workspace as `train_traces.json`.

I have an implementation ready with 7 tests, tested against real evals. Happy to open a PR if this is welcome.

Ref: "Better Harness" blog post, future work section on deriving errors automatically from traces.

johanity:
Fixes #2589

## What this does

The Better Agent currently sees failure messages like:

> "Expected slack_send_dm but agent called github_create_issue"

It knows what went wrong but not why. This change adds the full agent
trajectory to the proposer workspace so it can see the reasoning and
tool calls that led to the failure.

## How it works

Trace data is already fetched by `write_trace_payloads` and saved to
disk. This reads those files and writes compact summaries to
`train_traces.json` in the proposer workspace.

- `core.py`: `summarize_langsmith_trace()` — extracts messages from
  LangSmith traces, supports both LC constructor and OpenAI chat
  formats, truncates tool output at 2000 chars
- `agent.py`: `_write_train_traces()` — called from
  `build_proposer_workspace`, skips cases without trace artifacts
- 7 new tests, both message formats

Tested against real evals (`test_tool_selection`, `test_followup_quality`)
with traces from gpt-4.1-mini.

### DIFF
--- a/examples/better-harness/better_harness/core.py
+++ b/examples/better-harness/better_harness/core.py
@@ -878,8 +878,7 @@ def summarize_langsmith_trace(
 ) -> TraceSummary:
     """Summarize a LangSmith trace into a compact structure.
 
-    Supports both LangChain constructor format (`kwargs.type`) and
-    OpenAI chat format (`role`/`content`) as returned by the LangSmith
+    Handles the `role`/`content` format returned by the LangSmith
     `include_messages=true` API.
 
     Args:
@@ -894,22 +893,14 @@ def summarize_langsmith_trace(
     messages = trace_data.get("messages") or []
     steps: list[TraceStep] = []
     for msg in messages:
-        if "kwargs" in msg:
-            kwargs = msg["kwargs"]
-            msg_type = kwargs.get("type", "unknown")
-            content = kwargs.get("content")
-            raw_tool_calls = kwargs.get("tool_calls")
-            name = kwargs.get("name")
-            status = kwargs.get("status")
-        else:
-            role = msg.get("role", "unknown")
-            # user/assistant/tool/system all pass through; normalize
-            # the two that differ from LC naming.
-            msg_type = {"user": "human", "assistant": "ai"}.get(role, role)
-            content = msg.get("content")
-            raw_tool_calls = msg.get("tool_calls")
-            name = msg.get("name")
-            status = msg.get("status")
+        role = msg.get("role", "unknown")
+        # user/assistant/tool/system all pass through; normalize
+        # the two that differ from LC naming.
+        msg_type = {"user": "human", "assistant": "ai"}.get(role, role)
+        content = msg.get("content")
+        raw_tool_calls = msg.get("tool_calls")
+        name = msg.get("name")
+        status = msg.get("status")
 
         if content is None:
             content = ""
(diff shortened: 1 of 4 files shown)


johanity:
note: `train_traces.json` will be empty in the normal pipeline right now.

the pytest reporter outputs experiment comparison URLs where the last UUID is a session ID, not a run ID. `extract_langsmith_trace_id` returns that session ID, `fetch_langsmith_trace` tries `GET /runs/{session_id}`, gets a 404, and the trace files never make it to disk.

the runs do exist inside that session though. i verified the summarizer works by fetching real traces using the correct run IDs from actual eval runs (gpt-4.1-mini on `test_tool_selection` and `test_followup_quality`). full trajectories come through correctly.

left the fetch pipeline untouched since that felt like your call to make. the summarizer picks up traces automatically once they land on disk, and the tests cover that path.

examples/better-harness/better_harness/core.py
```diff
@@ -846,6 +847,118 @@ def fetch_langsmith_trace(*, endpoint: str, api_key: str, trace_id: str) -> dict
...
+
+    Returns:
+        A `TraceSummary` with extracted steps. Returns a summary with
+        empty steps if the trace cannot be parsed.
+
+    """
+    messages = trace_data.get("messages") or []
+    steps: list[TraceStep] = []
+    for msg in messages:
+        if "kwargs" in msg:
```
eyurtsev: this conditional doesn't look right to me -- why does the code need to handle 2 code paths?

johanity: Hey! Nice to "meet" you @eyurtsev! 😊
What type of tea do you drink?
I love your authenticity, so refreshing 🐈

You're right! The API always returns role/content (via include_messages=true), kwargs was unnecessary. I dropped it and cleaned up the fixtures, pushed :)

mdrxy:
Thanks for putting this together — really appreciate the effort here.

We've decided to stop accepting new examples into the monorepo. Each example we merge becomes ours to maintain: keeping it building against new SDK versions, patching dependencies, responding to issues filed against it, and reviewing its security surface (especially examples that wrap shell execution, browser automation, or third-party APIs with credentials). That overhead has grown faster than our capacity to absorb it, and rather than letting examples bit-rot in-tree we'd rather not take them in the first place.

That said, we'd love for this to live somewhere people can find it. A couple of options that work well:

- Post it in [Discussions](https://github.com/langchain-ai/deepagents/discussions) under the "Show and tell" category, with a link to a repo or gist on your own account.
- Keep it on your own GitHub — we're happy to surface community examples from the docs or README once a curation story exists.

Closing this for now, but please don't read it as a rejection of the work itself — it's a scoping decision about the repo, not the contribution. Thanks again.

PR #2619, opened 2026-04-09
fix(runloop): catch per-file exceptions in download_files/upload_files

### LINKED ISSUE #2507: `RunloopSandbox.download_files`/`upload_files` failure crashes entire batch
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)

### Related Issues / PRs

_No response_

```python
from langchain_runloop import RunloopSandbox

client = Runloop()
devbox = client.devboxes.create()
sandbox = RunloopSandbox(devbox=devbox)

# Attempt to download two files where the first doesn't exist.
# Expected: returns [FileDownloadResponse(error="file_not_found"), FileDownloadResponse(content=...)]
# Actual: raises an unhandled exception from the Runloop SDK on the first file;
#         the second file is never attempted.
responses = sandbox.download_files(["/nonexistent.txt", "/etc/hostname"])

# Same problem with upload_files:
sandbox.upload_files([
    ("/proc/1/root", b"impossible"),  # will fail (not a writable path)
    ("/tmp/valid.txt", b"hello"),      # never reached
])
```

```
Traceback (most recent call last):
  File "...", line X, in download_files
    content = self._devbox.file.download(path=path)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  <Runloop SDK exception for missing file>

The above exception is unhandled — no try/except wraps the per-file call.
```

### Description

`RunloopSandbox.download_files()` and `upload_files()` in `libs/partners/runloop/langchain_runloop/sandbox.py` do not catch per-file exceptions. If any single file operation fails (e.g., file not found, permission denied), the uncaught exception aborts the entire batch — no subsequent files are processed, and the caller receives a raw SDK exception instead of a structured `FileDownloadResponse`/`FileUploadResponse` with an `error` field.

This violates the `BaseSandbox` protocol contract defined in `libs/deepagents/deepagents/backends/sandbox.py`:

> **Line 729** (upload): *"Implementations must support partial success - catch exceptions per-file and return errors in `FileUploadResponse` objects rather than raising."*
>
> **Line 740** (download): *"Implementations must support partial success - catch exceptions per-file and return errors in `FileDownloadResponse` objects rather than raising."*

**Current code** (`libs/partners/runloop/langchain_runloop/sandbox.py:62-78`):

```python
def download_files(self, paths: list[str]) -> list[FileDownloadResponse]:
    responses: list[FileDownloadResponse] = []
    for path in paths:
        content = self._devbox.file.download(path=path)  # no try/except
        responses.append(FileDownloadResponse(path=path, content=content, error=None))
    return responses

def upload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadResponse]:
    responses: list[FileUploadResponse] = []
    for path, content in files:
        self._devbox.file.upload(path=path, file=content)  # no try/except
        responses.append(FileUploadResponse(path=path, error=None))
    return responses
```

**Every other sandbox provider handles this correctly.** For example, `ModalSandbox` (`libs/partners/modal/langchain_modal/sandbox.py:24-51`):

```python
def _read_file(self, path: str) -> FileDownloadResponse:
    if not path.startswith("/"):
        return FileDownloadResponse(path=path, content=None, error="invalid_path")
    try:
        f = self._sandbox.open(path, "rb")
        # ...
    except FileNotFoundError:
        error = "file_not_found"
    except modal.exception.FilesystemExecutionError as e:
        msg = str(e).lower()
        error = "is_directory" if "is a directory" in msg else "file_not_found"
    return FileDownloadResponse(path=path, content=content_bytes, error=error)
```

`LangSmithSandbox` also correctly implements partial-success semantics. `RunloopSandbox` is the only provider that violates the protocol.

**Impact:** The `FilesystemMiddleware` calls `download_files` during skill loading, file reads, and other batch operations. A single missing file in a batch causes the entire operation to fail with an unhandled exception, which is particularly problematic in multi-subagent scenarios where concurrent file operations are common.

### Environment / System Info

```
OS: macOS (Darwin 24.5.0)
Python: 3.12.7
deepagents: 0.4.5
langchain: 1.2.10
langchain-core: 1.2.17
```

kuncevichandrew2:
## Summary
- Wrap per-file SDK calls in `download_files()` and `upload_files()` with try/except to return structured error responses (`file_not_found`, `permission_denied`, `is_directory`) instead of propagating raw exceptions
- Aligns `RunloopSandbox` with the `BaseSandbox` partial-success contract that all other providers already follow
- Add unit tests covering success, individual error types, and partial-success batches

Closes #2507

## Test plan
- [x] Unit tests for `download_files`: success, file_not_found, permission_denied, is_directory, partial success
- [x] Unit tests for `upload_files`: success, file_not_found, permission_denied, partial success
- [x] All 9 new tests passing locally


libs/partners/runloop/langchain_runloop/sandbox.py
```diff
@@ -63,16 +63,49 @@ def download_files(self, paths: list[str]) -> list[FileDownloadResponse]:
...
-            content = self._devbox.file.download(path=path)
-            responses.append(
-                FileDownloadResponse(path=path, content=content, error=None)
-            )
+            try:
+                content = self._devbox.file.download(path=path)
+                responses.append(
+                    FileDownloadResponse(path=path, content=content, error=None)
+                )
+            except FileNotFoundError:
```
eyurtsev, line 71: Does runloop return FileNotFoundError?

ccurme: +1, do we need to catch https://github.com/runloopai/api-client-python/blob/main/src/runloop_api_client/_exceptions.py?

mdrxy:
Thanks for the contribution!

Thanks again! 🙏

PR #2638, opened 2026-04-09
fix(sdk): add sandbox blob read thresholds

eyurtsev:
Adjusts `BaseSandbox` read behavior to keep text reads capped at 500 KiB while handling blobs in three tiers: inline base64 up to 500 KiB, download_files up to 10 MiB, and a clear threshold error above that. This also adds unit coverage for the download fallback, oversize blobs, and preserving non-threshold blob errors.


mdrxy:
Will re-open as needed

PR #2643, opened 2026-04-09
fix(evals): use per-trial langsmith templates to avoid concurrent teardown races

maahir30:
- LangSmith sandbox templates were named solely by container image, so concurrent trials sharing the same image raced on the same template — one trial's teardown would delete a template still in use by another, causing `ResourceInUseError` and failing the trial.
- Each trial now creates its own template (keyed by `session_id`), giving it exclusive ownership of the full sandbox + template lifecycle. This eliminates the race by construction and removes ~55 lines of shared-template reuse/recreation logic (`_ensure_template`).


vtrivedy (commented):
lgtm

eyurtsev (approved):

PR #2659, opened 2026-04-10
feat(sdk): scope permissions to routes for composite backends with sandbox default

nick-hollon-lc:
## Summary

Previously, `_PermissionMiddleware` unconditionally rejected any backend that supported execution (`SandboxBackendProtocol`), even when permissions only targeted paths under composite backend routes that don't involve execution at all.

This PR adds a check (`_all_paths_scoped_to_routes`) so that when a `CompositeBackend` has a sandbox default, permissions are allowed as long as every permission path is scoped under a known route prefix. Permissions that cover paths outside routes (hitting the sandbox default) still raise `NotImplementedError`.

### Examples

```python
from deepagents.backends import CompositeBackend
from deepagents.middleware.permissions import FilesystemPermission

sandbox = SandboxBackend(...)  # supports execution
memories = StoreBackend(...)

composite = CompositeBackend(
    default=sandbox,
    routes={"/memories/": memories},
)

# Previously raised NotImplementedError — now works because
# permissions are scoped entirely to the /memories/ route
agent = create_deep_agent(
    model=model,
    backend=composite,
    permissions=[
        FilesystemPermission(operations=["write"], paths=["/memories/**"], mode="deny"),
    ],
)

# Permissions outside any route still raise — they'd hit the sandbox default
agent = create_deep_agent(
    model=model,
    backend=composite,
    permissions=[
        FilesystemPermission(operations=["write"], paths=["/workspace/**"], mode="deny"),
    ],
)
# -> NotImplementedError

# Wildcard paths also raise — /** covers both routes and the default
agent = create_deep_agent(
    model=model,
    backend=composite,
    permissions=[
        FilesystemPermission(operations=["read"], paths=["/**"], mode="deny"),
    ],
)
# -> NotImplementedError
```

### Changes

- **`permissions.py`**: Added `_all_paths_scoped_to_routes` helper and refined the `__init__` guard to allow route-scoped permissions with sandbox defaults
- **`test_end_to_end.py`**: Added `TestCompositeBackendPermissionsEndToEnd` with 6 tests covering route-scoped allow/deny, wildcard rejection, mixed-path rejection, and multi-route scenarios

### DIFF
--- a/libs/deepagents/deepagents/middleware/permissions.py
+++ b/libs/deepagents/deepagents/middleware/permissions.py
@@ -18,6 +18,7 @@ from langchain.tools.tool_node import ToolCallRequest
 from langchain_core.messages import ToolMessage
 from langgraph.types import Command
 
+from deepagents.backends.composite import CompositeBackend
 from deepagents.backends.protocol import BACKEND_TYPES, BackendProtocol, GlobResult, GrepResult, LsResult
 from deepagents.backends.utils import (
     format_grep_matches,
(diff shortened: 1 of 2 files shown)


libs/deepagents/tests/unit_tests/test_end_to_end.py
```diff
@@ -1567,6 +1567,215 @@ async def test_filesystem_permission_deny_write_async(self) -> None:
...
+                FilesystemPermission(operations=["write"], paths=["/memories/**"], mode="allow"),
+            ],
+        )
+        result = agent.invoke({"messages": [HumanMessage(content="Write a note")]})
+
+        tool_messages = [msg for msg in result["messages"] if msg.type == "tool"]
+        assert len(tool_messages) == 1
+        assert "permission denied" not in tool_messages[0].content
+
+    def test_permissions_outside_routes_still_raises_with_sandbox_default(self) -> None:
```
sydney-runkle, line 1678: NICE

sydney-runkle (approved):
BOOM

libs/deepagents/deepagents/middleware/permissions.py
```diff
@@ -143,6 +144,31 @@ def _filter_paths_by_permission(
...
+    backend: BackendProtocol,
+) -> bool:
+    """Check if every permission path is scoped under a CompositeBackend route.
+
+    Returns ``True`` only when *backend* is a ``CompositeBackend`` and every
+    path pattern in *rules* starts with one of its route prefixes. This means
+    the permissions only govern file operations on route-specific backends and
+    never touch the (sandbox-capable) default backend.
+    """
+    if not isinstance(backend, CompositeBackend):
```
eyurtsev, line 158: This should raise an assertion error rather than a false based on how it's used int the code below. hitting this if statement is an internal logical mistake (Assertion)

eyurtsev (approved):

PR #2665, opened 2026-04-10
fix(sdk): raise `ValueError` for permission paths without leading slash

nick-hollon-lc:
Relative paths in FilesystemPermission silently never match because input paths are canonicalized to start with '/'.  Also prevent path traversal (..) and home directory expansion (~). Validate at construction time to catch misconfiguration early.


eyurtsev (approved):

eyurtsev (dismissed):
Let's add relative paths and `~` expansion as well as forbidden paths?

PR #2708, opened 2026-04-13
feat(cli): user scoped memory

sydney-runkle:
# Per-User Memory for Deep Agents Deploy

## What is it?

When you deploy an agent with `deepagents deploy`, it can now remember things
about each user across conversations. The agent gets a writable `AGENTS.md`
file per user — it can read and update this file to store preferences, context,
and learnings that persist between sessions.

Each user's memory is completely isolated. User A's preferences never leak into
User B's experience.

## How to use it

1. Create a `user/` directory in your project (next to `deepagents.toml`)
2. Optionally add a `user/AGENTS.md` with starter content (template for new users)
3. Deploy — that's it

```txt
my-agent/
├── deepagents.toml
├── AGENTS.md          # shared agent instructions (read-only)
└── user/
    └── AGENTS.md      # per-user memory template (writable at runtime)
```

If you create `user/` but don't add an `AGENTS.md`, an empty one is seeded
for each user automatically.

## How it works

- At deploy time, `user/AGENTS.md` is bundled into the seed payload
- The first time a user interacts, their own copy is created in the store
- The agent sees it in context every conversation and can update it via `edit_file`
- Redeployments never overwrite existing user data
- If no `user/` directory exists, user memory is disabled entirely

## Technical details

- User memory lives at `/memories/user/AGENTS.md` in the VFS
- Namespaced per `(assistant_id, user_id)` — fully isolated per user
- `user_id` comes from `runtime.user.identity` (custom auth)
- Gracefully skipped (with a warning) if no `user_id` is provided
- Shared `AGENTS.md` and skills remain read-only; only user memory is writable


shanta4100:
Thanks for your support

PR #2713, opened 2026-04-14
perf(sdk): add cache breakpoint to `MemoryMiddleware`

kevinbfrank:
Add `add_cache_control` param to `MemoryMiddleware` that tags the memory content block with `cache_control: {"type": "ephemeral"}` when using Anthropic models. This creates a second cache breakpoint on the memory block (the first is placed by `AnthropicPromptCachingMiddleware` on the static system prompt).

## Why

With a single breakpoint (current main), turn 2 of a multi-turn conversation only achieves ~60% cache hit rate because memory content falls outside the cache boundary and gets re-written every turn. With the dual breakpoint, turn 2 hits ~99.8%.

## Eval results

10-turn checkpointed conversation, cold cache, `claude-sonnet-4-6`:

| Metric | Baseline (main) | Dual-breakpoint | Change |
|---|---|---|---|
| Total cost | $0.0649 | $0.0542 | **-16.5%** |
| Cache hit rate | 85.8% | 89.8% | +4.0pp |

Turn 2 comparison:

| Turn 2 | Baseline | Dual-breakpoint |
|---|---|---|
| cache_read | 4,797 (60%) | 7,982 (99.8%) |
| cache_creation | 3,201 | 16 |

## Parity with deepagentsjs

Matches the approach in deepagentsjs where `createMemoryMiddleware({ addCacheControl: anthropicModel })` tags the memory block with `cache_control` ([source](https://github.com/langchain-ai/deepagentsjs/blob/main/libs/deepagents/src/middleware/memory.ts)).


libs/deepagents/deepagents/middleware/memory.py
```diff
@@ -187,9 +189,13 @@ def __init__(
                      Display names are automatically derived from the paths.
 
                      Sources are loaded in order.
+            add_cache_control: Whether to tag the memory content block with
+                     ``cache_control: {"type": "ephemeral"}`` for Anthropic
```
eyurtsev, line 193: Is it only for anthropic?

kevinbfrank: only anthropic - ported from js https://github.com/langchain-ai/deepagentsjs/pull/285/commits

sydney-runkle: i think we should come up w/ a more model agnostic solution for this? and rethink caching in general for deepagents w/ this in mind?

mdrxy: Agree a provider-agnostic caching story is the right long-term direction, but I don't think this PR should block on it. This PR follows existing patterns for `AnthropicPromptCachingMiddleware`.

mdrxy (approved):

PR #2721, opened 2026-04-14
fix(sdk): skill loading should default to 1000 lines

sydney-runkle:


libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -591,7 +591,7 @@ async def _alist_skills(backend: BackendProtocol, source_path: str) -> list[Skil
 User: "Can you research the latest developments in quantum computing?"
 
 1. Check available skills -> See "web-research" skill with its path
-2. Read the skill using the path shown
+2. Read the full skill file: `read_file(path, limit=1000)` to load the complete instructions
```
eyurtsev: This line will mislead the agent? The skill may have more than 1000 lines, so the instructions may not load completely

eyurtsev (approved):

libs/deepagents/deepagents/middleware/skills.py
shanta4100: {
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "@/*": ["./*"]
    }
  }
}

PR #2730, opened 2026-04-15
feat(sdk): anthropic profile

maahir30:
Adds harness profiles for Anthropic Claude models, grounded in Anthropic's published prompting best practices and per-model migration guides.

## What

- **`_anthropic.py`** — provider-level suffix applied to every Claude model. Covers parallel tool calls, grounded responses, post-tool reflection, and decisive execution (from Anthropic's "latest Claude models" guidance).
- **`_anthropic_opus.py`** — Opus 4.6 overlay. Adds minimal-changes, subagent discipline, and focused-exploration sections to curb Opus 4.6's documented overeagerness / overthinking / subagent-overuse tendencies.
- **`_anthropic_opus47.py`** — Opus 4.7 overlay. Inverts the 4.6 approach because the 4.7 migration guide documents *reduced* tool and subagent use by default; encourages active tool use and appropriate subagent spawning.
- **`_anthropic_sonnet.py` / `_anthropic_haiku.py`** — intentionally empty registrations. Anthropic docs contain no Sonnet/Haiku-specific behavioral prompt recommendations, so these inherit the provider suffix unchanged. The modules exist as audit anchors with the rationale inline, plus a stable hook for future tests and migrations.

Every overlay cites the exact Anthropic doc section it derives from, so future edits require new evidence rather than vibes.

### DIFF
--- a/libs/deepagents/deepagents/profiles/_anthropic_sonnet.py
+++ b/libs/deepagents/deepagents/profiles/_anthropic_sonnet.py
@@ -1,0 +1,64 @@
+"""Anthropic Sonnet 4.6 model harness profile.
+
+!!! warning
+
+    This is an internal API subject to change without deprecation. It is not
+    intended for external use or consumption.
+
+Registers an intentionally empty model-level profile for
+``anthropic:claude-sonnet-4-6``.  The provider-level Anthropic profile
+(:data:`~deepagents.profiles._anthropic._ANTHROPIC_SYSTEM_PROMPT_SUFFIX`)
+already covers every prompt-suffix recommendation that Anthropic's
+`prompting best practices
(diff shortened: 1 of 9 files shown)


vtrivedy:
can we link the eval run here too, the new prompts match the guides well so looks like better defaults

libs/deepagents/deepagents/profiles/_anthropic_sonnet.py
```diff
@@ -0,0 +1,64 @@
+"""Anthropic Sonnet 4.6 model harness profile.
```
vtrivedy, line 1: the file labeling is confusing here.  we should do 1 file per family or 1 file per model.  right now its mixed between opus, opus47, sonnet

mdrxy:
Superseded by #2935

PR #2805, opened 2026-04-17
chore(sdk): perf optimization for patching tool calls

sydney-runkle:
* O(n) rather than O(n^2)
* early exit if available


eyurtsev (approved):

eyurtsev:
Let's add @AdemBoukhris457 as a co-author he spotted this performance optimization a while back

PR #2812, opened 2026-04-18
fix(evals): various eval fixes

maahir30:
## Summary

- Add `openai:gpt-5.3-codex` to the model registry (`set0`, `set1`, `openai` groups) and wire it into the `evals` + `harbor` workflow dropdowns.
- Fix missing `verifier_result` in terminal-bench runs on LangSmith/Daytona sandboxes.

## Context — verifier fix

Harbor's `Verifier.verify()` calls `environment.exec(...)` without a cwd. LangSmith's `sandbox.run(cwd=None)` spawns commands from `/` (not the image's `WORKDIR`), so terminal-bench `test.sh` scripts hit their `if [ "$PWD" = "/" ]; then exit 1; fi` guard and abort before writing `/logs/verifier/reward.txt`. The reward file never materializes, verifier raises `RewardFileNotFoundError`, harbor's outer `except Exception` swallows it, and `result.json` ends up with `verifier_result: null` → `add-feedback` logs "no verifier_result" and falls back to 0.0.

`LangSmithEnvironment` now probes `readlink /proc/1/cwd` at `start()` to resolve the container's runtime `WORKDIR` (PID 1 inherits it from the image) and uses that as the default cwd for subsequent `exec()` calls. Falls back to `/app` (terminal-bench convention) if the probe is inconclusive. Explicit `cwd=...` arguments from callers still win.

### DIFF
--- a/libs/evals/deepagents_harbor/langsmith_environment.py
+++ b/libs/evals/deepagents_harbor/langsmith_environment.py
@@ -382,7 +382,7 @@ class LangSmithEnvironment(BaseEnvironment):
 
         When `cwd` is not provided, defaults to the container's
         `WORKDIR` (resolved at `start()`) rather than LangSmith's
-        dataplane default of ``/``. This preserves the semantics that
+        dataplane default of `/`. This preserves the semantics that
         terminal-bench verifier scripts assume when they probe ``$PWD``.
 
         Args:
--- a/libs/evals/deepagents_harbor/langsmith_environment.py
+++ b/libs/evals/deepagents_harbor/langsmith_environment.py
@@ -71,6 +71,7 @@ class LangSmithEnvironment(BaseEnvironment):
         self._sandbox: AsyncSandbox | None = None
         self._client: AsyncSandboxClient | None = None
         self._template_name: str | None = None
+        self._default_cwd: str | None = None
         super().__init__(
             environment_dir,
             environment_name,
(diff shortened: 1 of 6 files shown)


libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -274,6 +275,52 @@ async def start(self, force_build: bool) -> None:
...
+            "LangSmith sandbox '%s' default cwd resolved to '%s'",
+            sandbox.name,
+            self._default_cwd,
+        )
+
+    @staticmethod
+    async def _detect_workdir(sandbox: AsyncSandbox) -> str:
+        """Resolve the container's Dockerfile ``WORKDIR`` at runtime.
+
+        LangSmith's ``sandbox.run(cwd=None)`` spawns each command from the
```
mdrxy:
```suggestion
        LangSmith's `sandbox.run(cwd=None)` spawns each command from the
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -274,6 +275,52 @@ async def start(self, force_build: bool) -> None:
...
+        dataplane daemon's cwd, not the image's ``WORKDIR``. Terminal-bench
```
mdrxy:
```suggestion
        dataplane daemon's cwd, not the image's `WORKDIR`. Terminal-bench
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -274,6 +275,52 @@ async def start(self, force_build: bool) -> None:
...
+        verifier scripts rely on ``WORKDIR`` (e.g. ``/app``) — many include
```
mdrxy:
```suggestion
        verifier scripts rely on `WORKDIR` (e.g. `/app`) — many include
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -274,6 +275,52 @@ async def start(self, force_build: bool) -> None:
...
+        ``if [ "$PWD" = "/" ]; then exit 1; fi`` as a guard and abort without
```
mdrxy:
```suggestion
        `if [ "$PWD" = "/" ]; then exit 1; fi` as a guard and abort without
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -274,6 +275,52 @@ async def start(self, force_build: bool) -> None:
...
+        writing ``/logs/verifier/reward.txt`` when this assumption is violated.
```
mdrxy:
```suggestion
        writing `/logs/verifier/reward.txt` when this assumption is violated.
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -274,6 +275,52 @@ async def start(self, force_build: bool) -> None:
...
+
+        PID 1 (the container entrypoint) inherits the image's ``WORKDIR`` as
```
mdrxy:
```suggestion
        PID 1 (the container entrypoint) inherits the image's `WORKDIR` as
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -274,6 +275,52 @@ async def start(self, force_build: bool) -> None:
...
+        its cwd, so ``readlink /proc/1/cwd`` yields the correct directory
```
mdrxy:
```suggestion
        its cwd, so `readlink /proc/1/cwd` yields the correct directory
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -274,6 +275,52 @@ async def start(self, force_build: bool) -> None:
...
+        without requiring access to the image metadata. Falls back to ``/app``
```
mdrxy:
```suggestion
        without requiring access to the image metadata. Falls back to `/app`
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -332,9 +380,15 @@ async def exec(
     ) -> ExecResult:
         """Execute a command inside the LangSmith sandbox.
 
+        When ``cwd`` is not provided, defaults to the container's
```
mdrxy:
```suggestion
        When `cwd` is not provided, defaults to the container's
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -332,9 +380,15 @@ async def exec(
...
+        ``WORKDIR`` (resolved at ``start()``) rather than LangSmith's
```
mdrxy:
```suggestion
        `WORKDIR` (resolved at `start()`) rather than LangSmith's
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -332,9 +380,15 @@ async def exec(
...
+        dataplane default of ``/``. This preserves the semantics that
```
mdrxy:
```suggestion
        dataplane default of `/`. This preserves the semantics that
```

libs/evals/deepagents_harbor/langsmith_environment.py
```diff
@@ -332,9 +380,15 @@ async def exec(
...
+        terminal-bench verifier scripts assume when they probe ``$PWD``.
```
mdrxy:
```suggestion
        terminal-bench verifier scripts assume when they probe `$PWD`.
```

PR #2833, opened 2026-04-19
feat(cli): MCP OAuth auth for http/sse servers

jkennedyvz:
## Summary

Adds OAuth support for MCP servers in the `deepagents` CLI, plus env-var interpolation for static headers. This is a slimmed-down successor to #2822 — scoped to the core OAuth flow, the auto-load path, and the `deepagents mcp login <server>` subcommand. TUI `/mcp login` integration is not included here; the `/mcp` viewer still lists configured servers and surfaces `unauthenticated` status with the login hint.

- **Env-var headers** — `resolve_headers` substitutes `${VAR}` references in `headers` entries, validated at config-load and applied at connection time.
- **`auth: oauth` config field** — validated at load time; rejected on stdio, rejected alongside a static `Authorization` header.
- **`FileTokenStorage`** — `mcp.client.auth.TokenStorage` backed by JSON under `~/.deepagents/mcp-tokens/<server>.json` (atomic write, 0o600 tokens, 0o700 dir).
- **Paste-back OAuth provider** — `build_oauth_provider` returns an `OAuthClientProvider` with redirect + callback handlers suitable for CLI use.
- **`deepagents mcp login <server>`** — new subcommand that drives the OAuth flow for an http/sse server from `.mcp.json` and persists the tokens.
- **Agent auto-load path** — `get_mcp_tools` attaches the OAuth provider when `auth: oauth` is set and fails fast with a helpful message if no tokens are cached.
- **Cross-loop session safety** — the MCP session manager survives loop switches so metadata discovery (in a throwaway loop) doesn't leave the runtime loop holding dead streams.
- **Per-server failures non-fatal** — a single dark or unauthenticated server no longer blocks startup; it shows up in the `/mcp` viewer with status + remediation.
- **Runtime re-auth as `ToolException`** — mid-session token-refresh failures surface as `"Run: deepagents mcp login <server>"` instead of blocking on `input()` in the event loop.
- **Slack and GitHub MCP** — pre-seeded `client_info` (Slack, DCR-forbidden) and OAuth Device Flow (GitHub, requires `client_secret` for code exchange).
- **Server-name path-safety** — `_validate_server_config` rejects `mcpServers` keys that don't match `^[A-Za-z0-9_-]+$`, so malicious or typo'd names can't escape the tokens directory in `FileTokenStorage._path`.

## Test plan

- [x] `make test` in `libs/cli` passes (139 new MCP unit tests, plus the existing suite)
- [x] `make lint` in `libs/cli` is clean
- [x] `deepagents mcp login notion` completes the OAuth flow end-to-end and tokens are written to `~/.deepagents/mcp-tokens/notion.json`
- [x] Running an agent against an `auth: oauth` server with no cached tokens shows the `Run: deepagents mcp login ...` hint in `/mcp`
- [x] Running the same agent after login succeeds and tools are discovered
- [x] A `.mcp.json` with a path-traversal key (`"../evil"`) is rejected at config load

### DIFF
--- a/libs/cli/deepagents_cli/main.py
+++ b/libs/cli/deepagents_cli/main.py
@@ -265,6 +265,7 @@ def parse_args() -> argparse.Namespace:
     from deepagents_cli.deploy import setup_deploy_parsers
     from deepagents_cli.output import add_json_output_arg
     from deepagents_cli.skills import setup_skills_parser
+    from deepagents_cli.mcp_commands import setup_mcp_parsers
 
     # Factory that builds an argparse Action whose __call__ invokes the
     # supplied *help_fn* instead of argparse's default help text.  Each
@@ -391,6 +392,8 @@ def parse_args() -> argparse.Namespace:
         subparsers,
         make_help_action=_make_help_action,
     )
+
+    setup_mcp_parsers(subparsers, make_help_action=_make_help_action)
 
     threads_parser = subparsers.add_parser(
         "threads",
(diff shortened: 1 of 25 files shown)


mdrxy:
Superseded by #2906

PR #2838, opened 2026-04-20
feat(sdk): opt-in `delete_file` tool to filesystem middleware

nick-hollon-lc:
## Summary
- Adds a `delete_file` tool to the filesystem middleware, opt-in via the `additional_tools` parameter
- Implements delete across backend protocol, filesystem, composite, sandbox, state, and store backends
- Catches `NotImplementedError` in `delete_file` for backwards compatibility with backends that don't implement it


mdrxy:
#3066

PR #2907, opened 2026-04-23
feat(sdk): add fork mode to subagents for prompt-cache reuse

ramon-langchain:
Adds opt-in forked subagents for prompt-cache reuse.

When `SubAgent(fork=True)` is used, the subagent inherits the parent agent's model, system prompt, and message prefix so the provider can reuse the parent's prompt-cache entry. The fork's own `system_prompt` is intentionally not placed in the system slot; it is injected as a preamble into the trailing task `HumanMessage`, keeping the inherited prefix byte-identical to what the parent already sent.

Forks inherit only the cache-relevant message prefix. The current `task` tool-call turn is trimmed out before seeding the fork, private/runtime state such as skills metadata, memory contents, and CLI local context is not copied, and only the fork's final answer returns to the parent as a `ToolMessage`.

The invalid model state is now unrepresentable: forked subagents cannot declare `model` at all and always use the parent model. `CompiledSubAgent(fork=True)` is rejected because compiled subagents own their graph and system prompt.

The `task` tool description now teaches the parent agent how much context to pass. Mixed fork/non-fork setups annotate only forked subagents; all-fork setups use default guidance that subagents inherit context. Fork calls are tagged as `ls_agent_type="fork-subagent"` for trace filtering.

This also adds `DEEPAGENTS_CLI_FORK_SUBAGENT=1` for the CLI. When set, CLI-loaded custom subagents are created as forks, subagent-local model frontmatter is ignored, and the CLI passes an explicit forked `general-purpose` subagent when needed.

Implementation notes:

1. `graph.py` still owns high-level agent assembly: parent prompt construction, model resolution, defaulting, and middleware stack assembly.
2. Subagent middleware owns the fork-specific prompt-padding helpers and final fork spec preparation, so sync/async subagent prompt text has one source of truth.
3. Fork middleware mirrors the parent system-prompt append order without giving the fork the recursive `task` tool.

Areas worth careful review:

1. Fork cache correctness depends on the fork seeing the same system-message bytes as the parent after middleware system-prompt padding.
2. Fork state seeding is intentionally include-list based; only messages are inherited today.
3. CLI `DEEPAGENTS_CLI_FORK_SUBAGENT=1` intentionally makes fork mode an all-subagents mode for CLI-loaded agents.

Verification:

1. Unit coverage includes parent/fork system prompt equality, inherited message seeding, current task-call trimming, parallel task-call trimming, private state exclusion, memory, async subagents, skills, custom system middleware, multiple subagents, no recursive `task` tool, `SystemMessage` content blocks, explicit fork model rejection, compiled fork rejection, output isolation, and CLI env wiring.
2. Live Anthropic test uses `claude-haiku-4-5-20251001` and asserts fork cache reads cover the inherited large message, not just the system prompt.
3. Live OpenAI test uses `gpt-5.4` and asserts the same inherited-message cache behavior through raw `prompt_tokens_details.cached_tokens`.
4. Latest local checks run: `ruff check`, `ty check`, `tests/unit_tests/test_fork_subagents.py`, `tests/unit_tests/test_agent.py`, and the live Anthropic/OpenAI fork cache integration tests.

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -11,7 +11,7 @@ from langchain.agents.middleware.types import AgentMiddleware, ContextT, ModelRe
 from langchain.agents.structured_output import ResponseFormat
 from langchain.tools import BaseTool, ToolRuntime
 from langchain_core.language_models import BaseChatModel
-from langchain_core.messages import HumanMessage, ToolMessage
+from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
 from langchain_core.runnables import Runnable, RunnableConfig
 from langchain_core.tools import StructuredTool
 from langgraph.types import Command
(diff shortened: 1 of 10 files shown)


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -343,9 +372,179 @@ class _SubagentSpec(TypedDict):
...
+            return copied[:idx]
+    return copied
+
+
+FORKED_SUBAGENT_MARKER = "[forked — inherits full conversation context]"
+"""Annotation rendered next to a forked subagent's name in the task tool's
+available-agents list. Also scanned by tests to verify the list rendering."""
+
+
+FORK_USAGE_GUIDANCE = (
```
ramon-langchain, line 533: this is too long, its probably enough to just say "subagents marked with ... inherit the full conversation context so only a minimal specific instruction is needed"

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -402,13 +613,42 @@ def _return_command_with_state_update(result: dict, tool_call_id: str) -> Comman
...
+            final_content = f"{preamble}\n\n{description}" if preamble else description
+            subagent_state["messages"] = [*parent_messages, HumanMessage(content=final_content)]
+        else:
+            subagent_state["messages"] = [HumanMessage(content=description)]
+        return subagent, subagent_state, is_fork
+
+    def _build_subagent_config(runtime: ToolRuntime, *, is_fork: bool) -> RunnableConfig:
+        """Build the RunnableConfig for a subagent invocation.
+
+        Forked subagents get ``ls_agent_type="fork-subagent"`` so LangSmith
```
ramon-langchain, line 652: excessive comment

PR #2914, opened 2026-04-23
feat(repl): enable sync invocation for PTC

ccurme:
```python
from langchain_core.tools import tool
from langchain_quickjs import REPLMiddleware

@tool
def get_user(user_id: str) -> int:
    """Square n."""
    return "Name: Alice"

agent = create_agent(
    "anthropic:claude-sonnet-4-6",
    middleware=[REPLMiddleware(ptc=[get_user])],
)

input_message = {
    "role": "user",
    "content": "Can you look up the name for user_id abc123?",
}
result = agent.invoke({"messages": [input_message]})
for message in result["messages"]:
    message.pretty_print()
```
this currently raises
> PanicException: assertion `left == right` failed: _engine::context::QjsContext is unsendable, but sent to another thread
>   left: ThreadId(2)
>  right: ThreadId(1)

`quickjs_rs.Runtime` and `Context` are !Send and panic if used from a thread other than the one that created them. The sync tool handler ran ctx.eval on whichever executor thread LangGraph picked.

Here, each REPLMiddleware owns a dedicated worker thread running its own asyncio loop; Runtime/Context creation, tool/skill install, and every eval marshal onto that thread, with sync callers blocking on .result() and async callers awaiting a wrapped future.

With these changes all repl evals (can) pass in `deepagents`.

### DIFF
--- a/libs/partners/quickjs/langchain_quickjs/_repl.py
+++ b/libs/partners/quickjs/langchain_quickjs/_repl.py
@@ -303,39 +303,122 @@ def _inject_tool_args_for_ptc(
     return enriched
 
 
+class _QuickjsWorker:
+    """Dedicated OS thread with its own asyncio event loop.
+
+    ``quickjs_rs.Runtime`` and ``quickjs_rs.Context`` are ``!Send`` — they
+    panic if used from a thread other than the one that created them.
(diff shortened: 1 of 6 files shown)


ccurme:
Went with https://github.com/langchain-ai/deepagents/pull/2926.

PR #2924, opened 2026-04-24
feat(cli): bundled chat frontend for deepagent deploy

victorm-lc:
Adds an opt-in `[frontend]` block to `deepagent deploy` that ships a prebuilt chat UI with the generated LangGraph deployment. The bundle (React + assistant-ui + LangGraph SDK) mounts at `/app` on the deployment and talks to the same backend — no separate hosting.

## Usage

```toml
[frontend]
enabled = true
auth_provider = "supabase"  # or "clerk"
app_name = "My Deep Agent"
```

Supabase picks up `SUPABASE_URL` / `SUPABASE_ANON_KEY` from `.env`; Clerk picks up `CLERK_PUBLISHABLE_KEY`. The CLI rejects deploys where the required vars are missing.

## What ships

- React + assistant-ui chat UI, built once per CLI release.
- Pluggable auth — `supabase` or `clerk`, dynamically imported at runtime via `auth/loader.tsx`.
- Thread picker backed by the LangGraph SDK, with auto-generated titles from the first user message.
- Light/dark theme, Deep Agents branding (logo, favicons).
- Supabase-only: password-reset flow via the official `@supabase/auth-ui-react` component.

## How it deploys

The Vite bundle is packaged as Python package data under `deepagents_cli/deploy/frontend_dist/`. At deploy time the bundler copies the dist into the tarball and writes a `window.__DEEPAGENTS_CONFIG__` script tag into `index.html` with the user's auth provider, keys, and app name — so there is no per-deploy rebuild.

The generated `app.py` (Starlette, since FastAPI isn't in the langgraph-api runtime) mounts the frontend at `/app/`, adds a `/app` → `/app/` 308 redirect and `/healthz`, and the generated `auth.py` exempts `/app`, `/app/*`, `/healthz`, `/favicon.ico`, and `/.well-known/*` from the auth middleware so the static shell can load before the user signs in.

Clerk JWTs expire at 60s, so the Clerk adapter refreshes the token every 45s via `getToken({ skipCache: true })` to keep long-idle tabs from handing the SDK a stale token.

## How verified

Deployed to real LangSmith Deployments with both providers end-to-end. Verified sign-in, sign-out, thread creation, thread switching with history hydration, theme toggle, and the Supabase password-reset email → new-password flow. Also ran locally via `uv run --directory libs/cli deepagent dev`.
                                                                                                                                                                                                                              
Docs PR: [https://github.com/langchain-ai/docs/pull/3703](https://github.com/langchain-ai/docs/pull/3703)


libs/cli/frontend/src/types.ts
```diff
@@ -0,0 +1,7 @@
+export type TodoStatus = "pending" | "in_progress" | "completed";
+
+export interface TodoItem {
```
christian-bromann, line 3: We export this from `langchain` and can just import it here.

libs/cli/frontend/src/RuntimeProvider.tsx
```diff
@@ -0,0 +1,229 @@
...
+  useRef,
+  useState,
+  type ReactNode,
+} from "react";
+import { AssistantRuntimeProvider } from "@assistant-ui/react";
+import {
+  useLangGraphRuntime,
+  type LangChainMessage,
+} from "@assistant-ui/react-langgraph";
+import { Client } from "@langchain/langgraph-sdk";
```
christian-bromann, line 16: I would suggest instead of using this, just use `useStream`, there is [an example](https://github.com/langchain-ai/langgraphjs/blob/main/examples/assistant-ui-claude/src/components/my-assistant.tsx) how you can stitch this together.

libs/cli/frontend/src/RuntimeProvider.tsx
```diff
@@ -0,0 +1,229 @@
...
+          raw = content
+            .map((p) => {
+              if (typeof p === "string") return p;
+              const pp = p as { text?: string; type?: string };
+              return pp.type === "text" && typeof pp.text === "string"
+                ? pp.text
+                : "";
+            })
+            .join(" ");
+        }
```
christian-bromann, line 105: With useStream, this should fold into
```suggestion
        const firstUser = messages.find(HumanMessage.isInstance);
        const content = firstUser?.text
```

victorm-lc:
Closing this in favor of a rewrite on `useStream` from `@langchain/react` instead of assistant-ui — simpler code, drops a large dep, and it resolves the thread-switching issues we were hacking around. Will open a fresh PR with the new direction.

PR #2940, opened 2026-04-25
feat(cli): bundled chat frontend for deepagent deploy

victorm-lc:
Adds an opt-in `[frontend]` block to `deepagent deploy` that ships a prebuilt chat UI alongside the generated LangGraph deployment. The bundle mounts at `/app` on the deployment, streams via `useStream` from `@langchain/react`, and talks to the same backend — no separate hosting.

The frontend always uses one of three authentication providers — **Clerk**, **Supabase**, or **anonymous**.

## Usage

```toml
# Per-user auth via Supabase or Clerk
[frontend]
enabled = true
app_name = "My Deep Agent"

[auth]
provider = "supabase"   # or "clerk"
```

```toml
# Anonymous mode — no [auth] section, API open to anyone with the URL
[frontend]
enabled = true
app_name = "My Deep Agent"
```

Supabase picks up `SUPABASE_URL` / `SUPABASE_PUBLISHABLE_DEFAULT_KEY` from `.env`; Clerk picks up `CLERK_SECRET_KEY` and `CLERK_PUBLISHABLE_KEY`. Anonymous mode needs no extra env vars. The CLI rejects deploys where the required vars are missing, and prompts for interactive `[y/N]` confirmation before pushing an anonymous deploy.

## What ships

- React chat UI built on `useStream` from `@langchain/react` — deep-agent–aware (`stream.subagents`, `stream.getSubagentsByMessage`, `filterSubagentMessages`), streaming through Streamdown for per-word markdown animation.
- Pluggable auth — `supabase`, `clerk`, or `anonymous`, dynamically imported at runtime via `auth/loader.tsx`.
- Thread picker backed by the LangGraph SDK, with auto-generated titles from the first user message.
- Per-browser thread scoping in anonymous mode — frontend mints a UUID into a `dap_anon_id` cookie, writes it as thread metadata at creation, filters `threads.search` by it on read (UX-only scoping, not security).
- Subagent cards with collapsible task description, vertically resizable output, and sticky-only-when-at-bottom scroll.
- Tool-call cards dispatched by name with per-tool renderers (files, search, think, todos).
- Light/dark theme, Deep Agents branding (logo, favicons).
- Uses Clerk and Supabase auth UI components.

## How it deploys

The Vite bundle is packaged as Python package data under `deepagents_cli/deploy/frontend_dist/`. At deploy time the bundler copies the dist into the tarball and writes a `window.__DEEPAGENTS_CONFIG__` script tag into `index.html` with the user's auth provider, keys, and app name — so there is no per-deploy rebuild.

The generated `app.py` (Starlette, since FastAPI isn't in the langgraph-api runtime) mounts the frontend at `/app/`, adds a `/app` → `/app/` 308 redirect and `/healthz`. The generated `auth.py` (one of three flavors: Supabase, Clerk, or a permissive anonymous block) exempts `/app`, `/app/*`, `/healthz`, `/favicon.ico`, and `/.well-known/*` from the auth middleware so the static shell can load before the user signs in (or — in anonymous mode — before any cookie is minted).

The anonymous `auth.py` overrides LangSmith Cloud's default `x-api-key` requirement (otherwise the bundled frontend can't reach `/threads`) and tags every request with a constant `"anonymous"` identity. Per-browser thread isolation is enforced client-side via the `dap_anon_id` metadata filter.

Clerk JWTs expire at 60s, so the Clerk adapter refreshes the token every 45s via `getToken({ skipCache: true })` to keep long-idle tabs from handing the SDK a stale token.

The deploy summary prints one of three auth labels: `Auth: <provider>` (configured), `Auth: anonymous (API open to anyone)`, or `Auth: default (LangSmith API key required)`. Anonymous deploys also print a yellow ⚠ warning and require interactive `[y/N]` confirmation before pushing.

## How verified

Smoke-tested locally against `langgraph dev` with all three modes:

- **Supabase and Clerk**: sign-in, sign-out, thread creation, thread switching with history hydration, auto-titling from the first message, theme toggle, tool-call rendering with live status + expand/collapse, subagent cards with live streaming output and resize, and the Supabase password-reset email → new-password flow.
- **Anonymous**: cookie minted on first load, per-browser thread isolation verified across two browser profiles (each profile sees only its own threads), no sign-in/sign-out UI in header, deploy-time confirmation prompt fires before push.
- **LangSmith Cloud deploy** (anonymous): end-to-end deploy of the anonymous bundle confirmed working — `POST /threads` returns 200 (permissive `auth.py` overrides default key requirement); thread picker filters per browser.

`make format && make lint && make test` pass in `libs/cli/`.

Docs for the new `[frontend]` config section live in a companion PR on the docs repo: [https://github.com/langchain-ai/docs/pull/3703](https://github.com/langchain-ai/docs/pull/3703)


libs/cli/frontend/src/components/MessageList.tsx
```diff
@@ -0,0 +1,180 @@
+import type { FC, RefObject } from "react";
+import { Streamdown } from "streamdown";
+import { APP_NAME } from "../constants";
+import {
+  getErrorMessage,
+  getImageBlocks,
+} from "../lib/stream";
+import type { AgentStream } from "../types";
+import type { AIMessage } from "@langchain/langgraph-sdk";
```
christian-bromann:
```suggestion
import type { AIMessage } from "@langchain/core/messages";
```

libs/cli/frontend/src/components/MessageList.tsx
```diff
@@ -0,0 +1,180 @@
...
+            return (
+              <div key={message.id ?? index} className="anim-msg flex justify-end">
+                <div className="max-w-[90%] rounded-2xl rounded-br-sm bg-[var(--primary)] px-4 py-2.5 text-sm leading-relaxed text-[var(--primary-foreground)]">
+                  {content}
+                </div>
+              </div>
+            );
+          }
+
+          if (message.getType() === "ai") {
```
christian-bromann:
```suggestion
          if (AIMessage.isInstance(message)) {
```

libs/cli/frontend/src/components/SubagentActivity.tsx
```diff
@@ -0,0 +1,269 @@
...
+    typeof lastAIContent === "string"
+      ? lastAIContent
+      : Array.isArray(lastAIContent)
+        ? lastAIContent
+            .filter((b): b is { type: "text"; text: string } =>
+              typeof b === "object" && b !== null && "type" in b && b.type === "text" && "text" in b,
+            )
+            .map((b) => b.text)
+            .join("")
+        : "";
```
christian-bromann:
```suggestion
  const lastAIText = lastAIContent.text;
```

libs/cli/frontend/src/components/SubagentActivity.tsx
```diff
@@ -0,0 +1,269 @@
...
+    typeof subagent.toolCall.args.subagent_type === "string"
+      ? subagent.toolCall.args.subagent_type
+      : null;
+  const description =
+    typeof subagent.toolCall.args.description === "string"
+      ? subagent.toolCall.args.description
+      : "";
+  const title = typeName ?? `Agent ${subagent.id.slice(0, 8)}`;
+  const elapsed = getElapsedTime(subagent.startedAt, subagent.completedAt);
+  const lastAIMessage = subagent.messages.filter((m) => m.type === "ai").at(-1);
```
christian-bromann:
```suggestion
  const lastAIMessage = subagent.messages.filter(AIMessage.isInstance).at(-1);
```

libs/cli/frontend/src/types.ts
```diff
@@ -0,0 +1,21 @@
+import type { UseStream, DefaultToolCall, SubagentStreamInterface } from "@langchain/react";
+import type { Message, Thread, ToolCallWithResult as SdkToolCallWithResult } from "@langchain/langgraph-sdk";
```
christian-bromann: Let's not use any message types from SDK anymore.
```suggestion
import type { BaseMessage } from "@langchain/core/messages";
import type { Thread, ToolCallWithResult as SdkToolCallWithResult } from "@langchain/langgraph-sdk";
```

libs/cli/frontend/src/App.tsx
```diff
@@ -0,0 +1,334 @@
...
+    filterSubagentMessages: true,
+    // Coalesce token deltas so Streamdown's word-level fade animation has
+    // time to play between updates. Dropping this to 0/false re-renders
+    // per token (jumpy); pushing above ~150 makes the stream feel laggy.
+    throttle: 100,
+    defaultHeaders,
+  });
+
+  const { messages, isLoading, error, values } = stream;
+  const files = (values.files ?? {}) as Record<string, unknown>;
```
christian-bromann: casting not needed
```suggestion
  const files = values.files ?? {};
```

libs/cli/frontend/src/runtimeConfig.ts
```diff
@@ -0,0 +1,71 @@
...
+    __DEEPAGENTS_CONFIG__?: Partial<RuntimeConfig> & { __PLACEHOLDER__?: boolean };
+  }
+}
+
+export function getRuntimeConfig(): RuntimeConfig {
+  const cfg = window.__DEEPAGENTS_CONFIG__;
+  if (!cfg || cfg.__PLACEHOLDER__) {
+    throw new Error(
+      "window.__DEEPAGENTS_CONFIG__ not injected. Run through `deepagent deploy` or `deepagent dev`.",
+    );
```
christian-bromann, line 39: nit: since we run this on module load, there is no way for us to render this error properly so in case this error throws we just see a blank page. Probably not worth optimizing for it but a cleaner way would be to put this check into a `RuntimeConfigProvider` and wrap the app around it.

christian-bromann (commented):
Some minor suggestions, overall LGTM 👍

libs/cli/frontend/src/runtimeConfig.ts
victorm-lc: makes sense, I agree we don't need to optimize on this at the moment. We'll need to separate out into a separate repo soon, so can optimize for that then.

hari-dhanushkodi (dismissed):

libs/cli/deepagents_cli/deploy/config.py
```diff
@@ -194,6 +220,16 @@ def validate(self, project_root: Path) -> list[str]:
         if self.auth is not None:
             errors.extend(_validate_auth_credentials(self.auth.provider))
 
+        if self.frontend is not None and self.frontend.enabled:
+            if self.auth is None:
+                errors.append(
+                    "[frontend].enabled requires [auth] to be configured. "
```
hari-dhanushkodi, line 232: this seems to contradict the docstring at the top, is auth required for frontend? docstring says ``[auth]` is
    optional; omitting it ships in anonymous mode`

victorm-lc: ok found other contradictions, just pushed a new commit to address!

libs/cli/pyproject.toml
```diff
@@ -151,6 +151,12 @@ test = [
 
 [tool.hatch.build.targets.wheel]
 packages = ["deepagents_cli"]
+include = [
+    "deepagents_cli/**/*.py",
+    "deepagents_cli/**/*.md",
+    "deepagents_cli/**/*.tcss",
+    "deepagents_cli/deploy/frontend_dist/**/*",
+]
```
mdrxy, line 159: why is this needed

PR #2980, opened 2026-04-28
fix(sdk): treat boundary-truncated UTF-8 in `read()` prefix check as text

### LINKED ISSUE #2983: `BaseSandbox.read()` returns base64 instead of text for valid UTF-8 files when a multi-byte character is split at the prefix boundary
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] repl
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

### Related Issues / PRs

#2980 

```python
import base64
import json
import pathlib
import subprocess
import sys

from deepagents.backends.sandbox import _READ_COMMAND_TEMPLATE


def reproduce(tmp_path: pathlib.Path) -> None:
    # Build a UTF-8 text file just over 8192 bytes whose last character lands
    # across the prefix boundary used by the read detection script.
    target = tmp_path / "korean.md"
    payload = ("a" * 8190).encode("utf-8") + "가나다".encode("utf-8")
    target.write_bytes(payload)

    cmd = _READ_COMMAND_TEMPLATE.format(
        path_b64=base64.b64encode(str(target).encode("utf-8")).decode("ascii"),
        file_type="text",
        offset=0,
        limit=2000,
    )
    # Run the embedded detection script directly to isolate behavior from
    # any specific sandbox provider.
    _, sep, tail = cmd.partition('python3 -c "')
    assert sep
    script, sep, _ = tail.rpartition('" 2>&1')
    assert sep

    proc = subprocess.run(
        [sys.executable, "-c", script],
        capture_output=True,
        text=True,
        check=True,
    )
    data = json.loads(proc.stdout.strip())
    print("encoding:", data.get("encoding"))   # expected: 'utf-8'; actual: 'base64'
    print("content head:", data.get("content")[:40])
```

```shell
No traceback. The bug is silent: read() returns the file as base64-encoded
bytes instead of UTF-8 text, so the LLM sees garbled content rather than
the original file.

Reproduction output on main:
  encoding: base64
  content head: YWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFhYWFh...
```

### Description

* I'm using `BaseSandbox.read()` to read a UTF-8 text file (a `.md`
  document containing Korean text) larger than 8192 bytes.
* I expect `read()` to return the content as UTF-8 text, since `file -i`
  reports the file as `text/plain; charset=utf-8`.
* Instead, `read()` returns the entire file base64-encoded under
  `encoding='base64'`, as if it were a binary file.

The detection script embedded in `_READ_COMMAND_TEMPLATE` reads the first
8192 bytes and attempts a UTF-8 decode to identify binary content. When a
multi-byte UTF-8 character (3 bytes for CJK) is split at byte 8192, the
decoder raises `UnicodeDecodeError` with `reason='unexpected end of data'`
even though the file is valid UTF-8. The script then routes the file to
the binary path and base64-encodes the entire contents.

This affects any text file containing CJK characters (Korean, Japanese,
Chinese), emoji, or other 3 to 4 byte UTF-8 characters once the file
exceeds 8192 bytes. Probability of triggering is roughly 2/3 for a file
that is mostly Korean.

### Environment / System Info

OS: Linux
Python: 3.11
deepagents: 0.5.3

SeongBeomLEE:
Fixes #2983 

`BaseSandbox.read()` decodes the first 8192 bytes of a file to detect binary content.

When a multi-byte UTF-8 character is truncated at that boundary — common with CJK text where each character is 3 bytes — the decoder raises `UnicodeDecodeError` even though the file is valid UTF-8, so the script base64-encodes the entire file.

The fix narrows the binary path to real invalid bytes by skipping `UnicodeDecodeError` cases caused by a truncated trailing character.

### Trigger

```python
# A CJK text file just over 8192 bytes whose last character lands across
# the prefix boundary is misclassified as binary on `main` and returned
# as base64. With this patch it is correctly returned as UTF-8 text.
payload = ("a" * 8190).encode("utf-8") + "가나다".encode("utf-8")
```

### Suggested test

A unit test was deliberately omitted to keep the diff minimal.

The following sketch reproduces the bug end-to-end by extracting the script from `_READ_COMMAND_TEMPLATE` and running it directly:

```python
import base64
import json
import pathlib
import subprocess
import sys

from deepagents.backends.sandbox import _READ_COMMAND_TEMPLATE


def test_read_script_handles_truncated_utf8_prefix(tmp_path: pathlib.Path) -> None:
    target = tmp_path / "korean.md"
    payload = ("a" * 8190).encode("utf-8") + "가나다".encode("utf-8")
    target.write_bytes(payload)

    cmd = _READ_COMMAND_TEMPLATE.format(
        path_b64=base64.b64encode(str(target).encode("utf-8")).decode("ascii"),
        file_type="text",
        offset=0,
        limit=2000,
    )
    _, sep, tail = cmd.partition('python3 -c "')
    assert sep
    script, sep, _ = tail.rpartition('" 2>&1')
    assert sep

    proc = subprocess.run(
        [sys.executable, "-c", script],
        capture_output=True,
        text=True,
        check=True,
    )
    data = json.loads(proc.stdout.strip())
    assert data.get("encoding") == "utf-8"
    assert "가나다" in data.get("content", "")
```

I'm happy to add this as `test_read_script_handles_truncated_utf8_prefix` in `tests/unit_tests/backends/test_sandbox_backend.py` if maintainers prefer to land the test together with the fix.

### AI disclaimer

Generated with assistance from Claude (Anthropic). 

All design decisions and the final wording were reviewed by the author.


SeongBeomLEE:
@eyurtsev - Would you have a moment to take a look at this PR?

The change is small and scoped to the prefix detection inside `_READ_COMMAND_TEMPLATE`, and the linked issue includes a minimal reproduction.

Happy to address any review feedback or add a unit test in the same PR if preferred.

Thanks!

PR #2994, opened 2026-04-29
feat(quickjs): add `max_ptc_calls` budget for ptc calls

hntrl:
This change adds a per-eval budget for `tools.*` host bridge calls in `langchain-quickjs` to prevent prompt-driven host-call flooding.

It introduces `max_ptc_calls` on `REPLMiddleware` (default `256`), enforces the budget inside the bridge before tool invocation, and resets mutable PTC state for each eval. This keeps normal tool use unchanged while bounding worst-case bridge amplification.

## Changes

- Add `max_ptc_calls` middleware/config plumbing and validation (`>= 1` or `None`)
- Track per-eval PTC state (`remaining_calls`, buffered commands) and reset it at eval boundaries
- Raise and surface `PTCCallBudgetExceeded` for uncaught overflows while still allowing JS `try/catch` to observe generic `HostError`
- Document DoS risk when `max_ptc_calls=None` via docstring warning and README updates
- Add unit and end-to-end coverage for overflow, reset semantics, `None` escape hatch, and invalid config

## Review Focus

- Error-surface behavior for caught vs uncaught overflow paths
- Per-eval lifecycle of PTC state and command buffering
- Middleware/docs wording around the `None` escape hatch and risk

### DIFF
--- a/libs/partners/quickjs/langchain_quickjs/_repl.py
+++ b/libs/partners/quickjs/langchain_quickjs/_repl.py
@@ -13,7 +13,7 @@ import json
 import logging
 import threading
 import uuid
-from dataclasses import dataclass, field
+from dataclasses import dataclass, field, replace
 from typing import TYPE_CHECKING, Any, cast
 
 from langgraph.types import Command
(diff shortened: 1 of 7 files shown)


libs/partners/quickjs/langchain_quickjs/_repl.py
```diff
@@ -77,6 +77,53 @@ class EvalOutcome:
...
+
+
+@dataclass
+class _PTCState:
+    """Mutable per-eval PTC state (reset on each eval call)."""
+
+    remaining_calls: int | None
+    command_buffer: list[Command] = field(default_factory=list)
+
+    @classmethod
```
eyurtsev: This is unusual python -> usually we use 

`@dataclass(frozen=True, slots=True)`

and then

rely on `dataclasses.replace` api (it can be introduced with as a method on the instance rather than a classmethod

libs/partners/quickjs/langchain_quickjs/_repl.py
```diff
@@ -77,6 +77,53 @@ class EvalOutcome:
...
 
 
+class _PTCCallBudgetExceededError(RuntimeError):
+    """Raised when one eval exceeds its configured PTC call budget."""
+
+    def __init__(self, *, limit: int, attempted: int, function_name: str) -> None:
+        self.limit = limit
+        self.attempted = attempted
+        self.function_name = function_name
+        super().__init__(self.render_message())
```
eyurtsev: style: inline the message instead of relying on `self.render_message()` in the __init__ 

It's usually based to avoid calling methods from __init__ as it creates a bit of sphaghetti

PR #2999, opened 2026-04-29
fix(quickjs): bound console buffering at capture time

hntrl:
## Summary

This PR fixes unbounded console buffering in `langchain-quickjs` by enforcing the stdout cap at capture time instead of only truncating at final formatting. This keeps model-visible behavior consistent while preventing memory growth from repeated `console.log` spam in a single eval.

### `langchain-quickjs`

- Bound `_ConsoleBuffer` by `max_result_chars` so console output is clipped as logs are appended, not after the fact.
- Added `stdout_truncated_chars` to `EvalOutcome` so formatting can surface how much console output was dropped.
- Threaded the stdout budget through `REPLMiddleware -> _Registry -> _ThreadREPL -> _ConsoleBuffer`.
- Updated stdout formatting to append a truncation marker after captured output when drops occurred, while keeping result/error truncation behavior unchanged.
- Extended unit tests to cover bounded capture, overflow prefix behavior, zero-budget marker behavior, reset semantics across evals, and constructor callsites that now pass the stdout cap.

### DIFF
--- a/libs/partners/quickjs/langchain_quickjs/_repl.py
+++ b/libs/partners/quickjs/langchain_quickjs/_repl.py
@@ -69,6 +69,7 @@ class EvalOutcome:
     """
 
     stdout: str = ""
+    stdout_truncated_chars: int = 0
     result: str | None = None
     result_kind: str | None = None  # "handle" when marshaling fell back
     error_type: str | None = None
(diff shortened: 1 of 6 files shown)


libs/partners/quickjs/langchain_quickjs/_repl.py
```diff
@@ -69,6 +69,7 @@ class EvalOutcome:
     """
 
     stdout: str = ""
+    stdout_truncated_chars: int = 0
```
eyurtsev, line 72: Do we need something for stderr?

hntrl: stdout == stderr here, console.log + console.warn + console.error mean the same thing (we could probably fix this later if we think its important)

PR #3002, opened 2026-04-29
feat(sdk): add forked subagent context

vtrivedy:
## Summary

This PR adds forked context support for synchronous subagents without touching the CLI.

The current implementation supports two ways to opt in:

- Per call: the `task` tool accepts `fork_context=true` or `fork_context=false`.
- Per subagent default: `SubAgent` and `CompiledSubAgent` can set `fork_context=True`, while individual task calls can still override that default.

When a task runs with forked context, the child starts from the parent state at the fork point:

- copies the concrete parent `messages` from state,
- removes only the active parent `task` call bookkeeping turn,
- appends `HumanMessage(content=description)` as the child task instruction,
- copies normal parent state such as files/custom state,
- excludes private/lifecycle state (`todos`, `structured_response`, `skills_metadata`, `memory_contents`, `_summarization_event`),
- derives a child `thread_id` from the parent thread id and task `tool_call_id` so checkpointed fork runs do not overwrite parent thread state.

Default behavior remains unchanged: if neither the subagent default nor the task call opts into forked context, the child receives only `HumanMessage(description)` as before.

## Why

Some delegation tasks need the current working context rather than a fresh isolated prompt. Today the agent must restate that context into the task description, which is expensive, incomplete, and easy to get wrong. This lets a subagent start from the parent conversation and working state, then do independent work and return a single result to the parent.

This is especially useful for the `general-purpose` subagent, which is already close to the parent agent's broad capabilities and is the natural target for a parent-context fork.

## Current Semantics

A forked subagent is a context/state fork into the selected subagent runnable. It is not a byte-for-byte clone of the parent runnable.

That means:

- `general-purpose` is the most parent-like target and the best fit for forked context.
- Specialized subagents still use their own configured prompt/tools/model/middleware, but receive the parent messages/state as input.
- The implementation does not apply summarization events before copying messages; it copies concrete state messages. `_summarization_event` is excluded so the child starts with its own lifecycle state.

## Cache-Hit Follow-Up Design

The goal for a stronger follow-up is a full prompt-cache hit when forking parent-like agents.

The current implementation preserves the parent message prefix, but it does not guarantee that the full model request is byte-identical to the parent request. A selected subagent may have a different system prompt, tools, middleware prompt additions, or model. Any of those can reduce or break provider prompt-cache reuse.

A stricter cache-oriented design would treat fork as an exact parent clone rather than as copied state passed into an existing subagent runnable:

```text
parent agent at task call
        |
        v
clone parent execution context
- same model
- same system prompt
- same middleware-generated prompt additions
- same tool surface, or an explicit cache-aware tool policy
- same concrete state/messages up to before current task-call bookkeeping
        |
        v
append HumanMessage(task description)
        |
        v
run fork independently with a distinct child thread id
        |
        v
return only final child answer as parent ToolMessage
```

That design would likely be restricted to `general-purpose` or parent-like subagents, because specialized subagents intentionally differ from the parent. For arbitrary specialized subagents, the current context-copy semantics are useful, but should not be described as guaranteed prompt-cache reuse.

Open design questions for that follow-up:

- Should true cache-preserving fork be limited to `general-purpose`?
- Should the forked parent clone keep the `task` tool and allow recursive forks, or should recursion be guarded by depth/config?
- Should the API distinguish `fork_context` from a stricter `fork_parent` / exact-clone mode?
- Should LangSmith metadata distinguish context forks from exact parent-clone forks?

## Validation

From `libs/deepagents`:

- `uv run --group test ruff check deepagents/middleware/subagents.py tests/unit_tests/test_subagents.py`
- `uv run --group test ruff format --check deepagents/middleware/subagents.py tests/unit_tests/test_subagents.py`
- `uv run --group test ty check deepagents/middleware/subagents.py`
- `uv run --group test pytest tests/unit_tests/test_subagents.py tests/unit_tests/middleware/test_subagent_middleware_init.py`
- `uv run --group test pytest tests/unit_tests/smoke_tests/test_system_prompt.py`
- `uv run --group test pytest tests/unit_tests`

Latest full SDK unit result: `1189 passed, 84 skipped, 4 xfailed`.


mdrxy:
related #2907

sydney-runkle:
closing as stale for now, will pick back up later

PR #3006, opened 2026-04-29
test(quickjs): add benchmarks

hntrl:
Adds a memory benchmark mode to CI, adds a benchmark suite for quickjs, and wires it into CI so benchmark regressions

## Changes

- Add QuickJS benchmark test modules under `tests/benchmarks` using fake chat models:
  - memory-focused scenarios (`console.log` and PTC tool usage) with parameterized concurrency
  - throughput-focused scenarios including repeated single-thread runs and multi-turn same-context persistence on a shared `thread_id`
- Add a `benchmark` make target for QuickJS package benchmark execution.
- Register benchmark-related pytest markers in QuickJS test config and add benchmark dependencies (`pytest-benchmark`, `pytest-codspeed`).
- Extend reusable CodSpeed workflow (`_benchmark.yml`) with optional memory-mode execution behind `has-memory-benchmarks`.
- Add a QuickJS benchmark job in `ci.yml` that:
  - runs on QuickJS package changes
  - enables memory benchmark mode for QuickJS
  - is included in the `ci_success` required job aggregation
- Update QuickJS change detection filters so `_benchmark.yml` changes also trigger QuickJS benchmark CI.

### DIFF
--- a/libs/partners/quickjs/tests/benchmarks/_common.py
+++ b/libs/partners/quickjs/tests/benchmarks/_common.py
@@ -1,0 +1,115 @@
+"""Shared helpers for QuickJS CodSpeed benchmark suites."""
+
+from __future__ import annotations
+
+from collections.abc import (
+    Iterator,  # noqa: TC003  # pydantic resolves this annotation at runtime
+    Sequence,  # noqa: TC003  # used in runtime-accessed annotations
+)
+from typing import TYPE_CHECKING, Any
+
+from deepagents import create_deep_agent
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
+from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
+from langchain_core.tools import tool
+from pydantic import Field
+
+if TYPE_CHECKING:
+    from langchain_quickjs import REPLMiddleware
+
+CONSOLE_LOG_CODE = "for (let i = 0; i < 200; i += 1) {  console.log(`line-${i}`);}'ok';"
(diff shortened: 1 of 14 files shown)


libs/partners/quickjs/tests/benchmarks/_common.py
```diff
@@ -0,0 +1,115 @@
...
+THROUGHPUT_ITERATIONS = 200
+
+
+class FakeChatModel(GenericFakeChatModel):
+    """Generic fake chat model whose ``bind_tools`` keeps scripted messages."""
+
+    messages: Iterator[AIMessage | str] = Field(exclude=True)
+
+    def bind_tools(self, tools: Sequence[Any], **_: Any) -> FakeChatModel:
+        del tools
```
eyurtsev:
```suggestion
```

libs/partners/quickjs/tests/benchmarks/test_quickjs_memory.py
```diff
@@ -0,0 +1,91 @@
...
+    """Benchmarks that compare Python heap pressure under common REPL workloads."""
+
+    def _run_concurrent_memory_workload(
+        self,
+        *,
+        thread_count: int,
+        code: str,
+        use_ptc: bool,
+    ) -> None:
+        start_barrier = threading.Barrier(thread_count)
```
eyurtsev: threading barrier isn't needed here if thread pool executor already limits the max number of works to be equal to thread_count

PR #3049, opened 2026-04-30
feat(quickjs): surface tool exceptions as the original error

ccurme:
If a tool raises an exception and is invoked in the REPL, we now generally propagate that exception (consistent with ToolNode for vanilla tools):
```python
from langchain.tools import tool
from langchain_quickjs import REPLMiddleware

@tool
def my_tool(x: int) -> int:
    """Test tool."""
    raise ValueError()  # <-- now raises


agent = create_deep_agent(
    "openai:gpt-5.4",
    middleware=[REPLMiddleware(ptc=[my_tool])],
)

input_message = {
    "role": "user",
    "content": "Use the quickjs REPL to invoke my_tool with value 3.",
}

for step in agent.stream({"messages": [input_message]}, stream_mode="values"):
    step["messages"][-1].pretty_print()
```
Two exceptions to this:

1. If the `eval` code wraps the tool invocation in a `try / catch`, we substitute error message content with `"Host function failed"` as before to guard against error-channel information disclosure as described in the quickjs-rs threat model [here](https://github.com/langchain-ai/quickjs-rs/blob/main/.github/THREAT_MODEL.md#t3-error-channel-information-disclosure).
2. If the tool is invoked with improper args, that exception is bubbled up out of `eval`, but is caught and handled by ToolNode in the usual way.

Requires https://github.com/langchain-ai/quickjs-rs/pull/17.


libs/partners/quickjs/langchain_quickjs/_repl.py
```diff
@@ -620,25 +630,12 @@ async def _aeval_async(  # noqa: C901
             outcome.error_type = "OutOfMemory"
             outcome.error_message = str(e)
         finally:
-            self._ptc_state = None
+            self._ptc_state = prev_ptc_state
```
ccurme, line 651: there's a race condition that the previous HostError masking was hiding. When two evals overlap, the one that loses the race used to clear `_ptc_state` in its finally block while the winner was still running. So the winner's next call would hit the "PTC bridge called outside active eval" check and raise. Here we save the previous one and put it back, so the winner's state survives the loser's cleanup.

libs/partners/quickjs/pyproject.toml
```diff
@@ -55,6 +55,7 @@ test = [
 
 [tool.uv.sources]
 deepagents = { path = "../../deepagents", editable = true }
+quickjs-rs = { git = "https://github.com/langchain-ai/quickjs-rs", branch = "cc/error_handling_2" }
```
eyurtsev: do we want this?

ccurme: this is temporary, will not land with this, will update once we release quickjs-rs

hntrl (dismissed):

PR #3063, opened 2026-05-01
feat(cli): reconnect logic using metadata for `deepagents deploy`

emanbareket:
Fixes #

---

Previously, deploy-generated graphs only cached sandbox backend objects in the process-local _SANDBOXES dictionary. If the graph was reloaded, imported in another module instance, or served by another process, the cache could miss and create a new remote sandbox for the same thread or assistant. That is especially problematic for assistant-scoped sandboxes and for any code outside the graph that needs to resolve the same sandbox.

This PR makes deployed sandbox resolution more robust by persisting sandbox IDs in LangGraph metadata and reconnecting to existing sandboxes on cache miss. This builds on reconnect behavior that already exists in the CLI sandbox factory. Note that the existing reconnect behavior in CLI does not support reconnecting for daytona, so this currently does not either, but that should be investigated further.


libs/cli/deepagents_cli/deploy/templates.py
eyurtsev, line 31: _SANDBOXES at the global namespace doesn't look correct. Do we know why this is happening?

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -103,12 +136,51 @@ def _get_or_create_sandbox(cache_key):
         ) from e
     backend = LangSmithSandbox(sandbox)
     _SANDBOXES[cache_key] = backend
-    logger.info(
```
eyurtsev, line 106: why is the logging statement removed?

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -153,6 +251,38 @@ def _get_or_create_sandbox(cache_key):
...
+    if record.get("provider") != "modal" or not isinstance(sandbox_id, str):
+        return None
+
+    try:
+        import modal
+
+        sb = modal.Sandbox.from_id(sandbox_id)
+        logger.info("Reconnected Modal sandbox %s for key %s", sandbox_id, cache_key)
+        return ModalSandbox(sandbox=sb)
+    except Exception:
```
eyurtsev, line 380: Could we wrap only the lines:

```python
import modal
sb = modal.Sandbox.from_id(sandbox_id)
```

So the try/except is as narrow as possible? Ideally

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
+    *, scope: str, thread_id: str | None, assistant_id: str | None
+) -> str:
+    if scope == "assistant":
+        return f"assistant:{{assistant_id}}"
+    return f"thread:{{thread_id or 'local'}}"
 
 
-def _build_backend_factory(assistant_id: str):
+# Validate that a sandbox record belongs to this generated deployment and scope.
+def _valid_sandbox_record(record: object, *, cache_key: str) -> dict | None:
```
eyurtsev: Let's use a TypedDict and explicitly declare what a "sandbox_record" is. But also if we can rename it so we don't use the word "record". instead "SandboxInfo" or something like that is a bit more typical for this use case.

Validation can be done with pydantic type adapter to make sure that the input type (which can be typed as Any if we want) gets converted into SandboxInfo


```python
class SandboxInfo(TypedDict):
   ...
```

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
+    if record.get("image") != SANDBOX_IMAGE:
+        return None
+    if not isinstance(record.get("sandbox_id"), str):
+        return None
+    return record
+
+
+# Convert a resolved sandbox backend into the canonical persisted record shape.
+def _sandbox_record_for_backend(cache_key: str, backend) -> dict:
+    sandbox_id = getattr(backend, "id", None)
```
eyurtsev: Let's avoid using `getattr`. if the backend is a sandbox type then `id` always exists on it. Perhaps we could add type annotations for backend that are appropriate?

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
+
+
+# Read the persisted sandbox record from thread or assistant metadata.
+async def _get_sandbox_record_from_metadata(
+    cache_key: str,
+    *,
+    scope: str,
+    thread_id: str | None,
+    assistant_id: str | None,
+) -> dict | None:
```
eyurtsev: Let's type the dict as SandboxInfo to be explicit about what the output of the function should be

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
+                return None
+            target = await client.assistants.get(assistant_id)
+        else:
+            return None
+        metadata = target.get("metadata") or {{}}
+        return _valid_sandbox_record(
+            metadata.get(_SANDBOX_RECORD_KEY),
+            cache_key=cache_key,
+        )
+    except Exception:
```
eyurtsev: Why not let this crash? (we can provide a good error message, but also not return a None?)

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
+async def _get_sandbox_record_from_metadata(
+    cache_key: str,
+    *,
+    scope: str,
+    thread_id: str | None,
+    assistant_id: str | None,
+) -> dict | None:
+    """Read the durable sandbox record for this thread/assistant scope."""
+    try:
+        from langgraph_sdk import get_client
```
eyurtsev: Let's move this import into global scope. local scope imports are reserved only for special cases in python (e.g., avoiding circular imports)

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
+            "Failed to read sandbox metadata for key %s",
+            cache_key,
+            exc_info=True,
+        )
+        return None
+
+
+# Persist the canonical sandbox record back to thread or assistant metadata.
+async def _persist_sandbox_record_if_possible(
+    record: dict,
```
eyurtsev: let's type record appropriately. this is probably sandbox_info: SandboxInfo at this point?

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
+            scope=scope,
+            thread_id=thread_id,
+            assistant_id=assistant_id,
+        )
+    return sandbox_backend, canonical_record
+
+
+def _build_backend_factory(
+    assistant_id: str,
+    initial_sandbox_record: dict | None = None,
```
eyurtsev: `initial_sandbox_record` -> `sandbox_info: Any` ? if it's not validated yet

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -103,12 +136,51 @@ def _get_or_create_sandbox(cache_key):
...
+# Reconnect to an existing LangSmith sandbox from a validated sandbox record.
+def _reconnect_sandbox(
+    cache_key,
+    *,
+    scope=None,  # noqa: ARG001
+    thread_id=None,  # noqa: ARG001
+    assistant_id=None,  # noqa: ARG001
+    sandbox_record=None,
+):
+    record = sandbox_record
```
eyurtsev: * Could we avoid locally renaming variables?
* Could we add a type annotation to sandbox_record? (sandbox_info?)

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -103,12 +136,51 @@ def _get_or_create_sandbox(cache_key):
...
+    if not record:
+        return None
+    sandbox_id = record.get("sandbox_id")
+    if record.get("provider") != "langsmith" or not isinstance(sandbox_id, str):
```
eyurtsev: Could we raise an exception here instead of returning None? If we're trying to connect to a non langsmith provider using langsmith code that's an internal logic error in the server and shouldn't be possible

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
+            target = await client.assistants.get(assistant_id)
+            metadata = dict(target.get("metadata") or {{}})
+            metadata[_SANDBOX_RECORD_KEY] = record
+            await client.assistants.update(assistant_id, metadata=metadata)
+    except Exception:
+        logger.warning(
+            "Failed to persist sandbox metadata for key %s",
+            record.get("cache_key"),
+            exc_info=True,
+        )
```
eyurtsev: re-raise exception?

libs/cli/deepagents_cli/deploy/templates.py
```diff
@@ -821,20 +1011,192 @@ def _factory(rt):
...
-            cache_key = f"thread:{{thread_id}}"
-        sandbox_backend = _get_or_create_sandbox(cache_key)
+        configurable = get_config().get("configurable", {{}})
+        thread_id = configurable.get("thread_id")
+        cache_key = _sandbox_cache_key(
+            scope=SANDBOX_SCOPE,
+            thread_id=thread_id,
+            assistant_id=assistant_id,
+        )
+        sandbox_record = _valid_sandbox_record(
```
eyurtsev: is the simpler variation correct? (i.e., or on the sandbox record, not on the _valid_sandbox_record)

```python
sandbox_record = _valid_sandbox_record(
   configurable.get(_SANDBOX_RECORD_KEY) or initial_sandbox_record,
   cache_key=cache_key
)
```

libs/cli/deepagents_cli/deploy/templates.py
emanbareket: It's the process local cache for sandboxes, this used to be the only source of truth before this pr. Not sure what the initial reasoning was.

libs/cli/deepagents_cli/deploy/templates.py
emanbareket: should not have been, added it back

libs/cli/deepagents_cli/deploy/templates.py
emanbareket: the difference is whether an existing but invalid sandbox record in the configurable can fall back to initial_sandbox_record or not. With the old version we can, with the simpler variation we can't. I would say we probably shouldn't fall back in this case so i'll keep your version.

PR #3066, opened 2026-05-01
feat(sdk): add `delete_file` tool to filesystem middleware

sydney-runkle:
Adds a `delete_file` tool to `FilesystemMiddleware`.

## Backwards compatibility

- **Existing backends are unaffected.** `BackendProtocol.delete` raises `NotImplementedError` by default; custom backends that don't override it simply won't see the tool.
- Follows the same pattern as `execute`: `supports_delete(backend)` checks whether the backend class overrides `delete`. If not, `delete_file` is filtered from `request.tools` and omitted from the system prompt before the model call — the LLM never sees a tool it can't use.
- All three built-in backends (`StateBackend`, `FilesystemBackend`, `CompositeBackend`) implement `delete`/`adelete`.
- `StateBackend` reuses the existing `None`-valued reducer logic (`_file_data_reducer` already treats `None` as a deletion marker).
- Permissions: `delete_file` is treated as a `"write"` operation, consistent with the existing permission model.

### DIFF
--- a/libs/deepagents/deepagents/backends/state.py
+++ b/libs/deepagents/deepagents/backends/state.py
@@ -10,6 +10,7 @@ from langgraph.config import get_config
 from deepagents._api.deprecation import warn_deprecated
 from deepagents.backends.protocol import (
     BackendProtocol,
+    DeleteResult,
     EditResult,
     FileData,
     FileDownloadResponse,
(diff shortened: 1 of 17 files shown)


nick-hollon-lc (dismissed):

eyurtsev:
likely need to discuss how to introduce this in a backwards compatible way?

libs/deepagents/deepagents/backends/state.py
```diff
@@ -290,6 +291,18 @@ def edit(
         self._send_files_update({file_path: self._prepare_for_storage(new_file_data)})
         return EditResult(path=file_path, occurrences=int(occurrences))
 
+    def delete(self, file_path: str) -> DeleteResult:
```
colifran, line 294: Will this delete files eagerly so that we maintain the read-your-write semantics with state backends? I think it should but wanted to confirm.

colifran (approved):
Nice!

colifran:
@eyurtsev I'm probably missing something but isn't this strictly additive? What needs to be backwards compatible?

sydney-runkle:
@colifran we want to make sure this doesn't break for backend implementations (community) that haven't implemented yet

Issue #3121, opened 2026-05-04
Non-interactive `deepagents auth` subcommands


mdrxy:
### Privileged issue

- [x] I am a Deep Agents maintainer.

### Issue Content

The new in-TUI `/auth` flow lands credentials in `~/.deepagents/.state/auth.json` via the modal, but there is no scriptable way to manage credentials from outside the interactive session. Add CLI subcommands so dotfile bootstrap, CI bootstrap, and "set this on a remote box without launching the TUI" all work.

Proposed surface (mirrors the modal verbs):

```
deepagents auth list                         # one row per provider with status
deepagents auth set <provider>               # read key from stdin (preferred)
deepagents auth set <provider> --from-env VAR   # copy from process env
deepagents auth remove <provider>            # alias: rm/delete
deepagents auth status [<provider>]          # print resolution source for one or all
deepagents auth path                         # print resolved auth.json path (incl. `PI_AGENT_DIR`-style override if added later)
```

Notes:

- Reuse `auth_store.set_stored_key`, `delete_stored_key`, and `load_credentials`. `set` should call the same `WriteOutcome` path so chmod warnings surface on stderr.
- Default to **stdin** for the key value so it never lands in shell history or argv. Reject TTY stdin without `--from-env` to avoid hanging on accidental invocations.
- `status` should print the same precedence source the TUI shows (`stored`, `env: VAR`, `missing`) so debugging "why isn't my key picked up?" is one command.
- Hook into `command_registry` if it ever grows a non-slash side, but the initial PR can register under `main.parse_args` next to `mcp` / `agents` / `threads`. Update `ui.show_help` and the help-screen drift test in lockstep.
- Coverage: round-trip via subprocess in `tests/unit_tests/`, including a "stdin redirected from a file" case and a "--from-env on an unset var" failure case.
- Out of scope for this issue: OAuth subscription login (`deepagents auth login <provider>`) — file the OAuth follow-up separately when that work begins.

Context: this came up while reviewing the in-TUI auth PR — see the conversation around the `AuthManagerScreen` description copy ("Lists installed providers… Install more via `pip install deepagents-cli[<provider>]`. Docs."). The docs link points to https://docs.langchain.com/oss/python/deepagents/cli/providers; the CLI subcommands should be documented there too once they ship.

### Area

- [x] code

PR #3123, opened 2026-05-04
feat(cli): in-TUI API key entry via `/auth`

mdrxy:
Add an in-TUI API key entry flow to `deepagents-cli` so users can paste credentials directly into the prompt instead of exporting environment variables before launch.

## Changes

- **Storage layer** (`auth_store.py`): new module persists credentials at `~/.deepagents/.state/auth.json` (mode 0600, parent 0700) using the same atomic `O_EXCL | 0o600` temp+replace pattern as `mcp_auth.FileTokenStorage`. `StoredCredential = ApiKeyCredential | OAuthCredential` is a discriminated union (OAuth stub today) so the future variant lands additively. `set_stored_key` returns a `WriteOutcome` whose `warnings` tuple surfaces chmod failures back to the UI — a `logger.warning` alone is invisible inside a Textual session.
- **Resolver unification** (`model_config.py`): `resolve_provider_credential(provider)` returns the credential value with priority *stored → env var* (matching the modal's "I just typed this in" affordance). `apply_stored_credentials` exports a stored key into `os.environ` before LangChain's chat-model factory reads it. `get_provider_auth_status` now reports `"stored credential"` when present so selectors and the manager all share one source of truth.
- **Modal screens** (`widgets/auth.py`): `AuthPromptScreen[AuthResult]` captures one provider's key with `password=True`, masking the value and never logging it. Dismissal is a three-arm `AuthResult` (`SAVED`/`DELETED`/`CANCELLED`) so the recovery path retries only on `SAVED` — `DELETED` would otherwise loop into the same missing-credentials error. `Ctrl+D` opens a `DeleteCredentialConfirmScreen` (red-bordered, patterned on `DeleteThreadConfirmScreen`).
- **Manager UX** (`AuthManagerScreen`): lists `PROVIDER_API_KEY_ENV ∩ get_available_models()` (so uninstalled providers are hidden) plus any stored or `config.toml`-defined providers (so stale credentials remain reachable for cleanup). Env badge resolves through `_resolving_env_var` to show `DEEPAGENTS_CLI_*` when the prefixed variant is what's actually set. Description includes a clickable `Docs` link to the providers reference. `1fr` flex on the option list keeps the footer pinned on short terminals.
- **Recovery flow** (`app.py`): `/auth` slash command registered (IMMEDIATE_UI tier) and dispatched to `_show_auth_manager`. `App.action_quit_app` and `action_toggle_auto_approve` delegate to the auth screens for `Ctrl+D` and `Shift+Tab` so the App-level priority bindings don't intercept.
- **Model selector** (`widgets/model_selector.py`): `_select_with_auth_check` opens the prompt instead of dismissing when the highlighted provider has `blocks_start` auth. On `SAVED` the original selection completes; on `CANCELLED`/`DELETED` the selector stays open and refreshes its credential indicator.
- **Corruption handling**: a malformed `auth.json` is swallowed in resolution paths (so the model selector keeps working) but surfaced visibly when `/auth` opens — the manager renders an inline warning banner and the prompt shows an inline notice if construction reads a broken file.

Follow-up tracked in #3121: non-interactive `deepagents auth` subcommands for dotfile/CI bootstrap.


mdrxy:
Capturing two recovery-UX follow-ups surfaced during review — not needed for this PR, filing here so they're not lost.

When the user selects a missing-credential provider via `/model`'s selector, `_select_with_auth_check` routes them straight into `AuthPromptScreen` already targeted at the right provider. Two paths into a `MissingCredentialsError` *don't* get that affordance and currently land the user in `/auth` with no context:

1. **Server startup failed because the default model's key is missing.**
2. **A direct `/model <spec>` command failed because that provider's key is missing.**

The error message gives them the spec to copy back, but they still have to find the provider in `/auth`'s list to add the key. Two ways to close the gap, in increasing scope:

- **Auto-open `AuthPromptScreen` on `MissingCredentialsError`.** Instead of (or alongside) the chat error message, push the prompt for the failing provider with the original spec stashed for retry on `AuthResult.SAVED`. Same UX as the model-selector path. Reuses everything that's already wired.
- **Clickable "Add key →" link in the error message.** Use `Style(link=...)` (same pattern as the docs link in `AuthManagerScreen`'s description) to make the error itself actionable. Lighter-weight — no auto-push, just a one-click route to the prompt — but requires `ErrorMessage` to grow click-handling, which it doesn't have today.

I'd lean toward option 1; option 2 ends up doing roughly the same work via a longer path. Either is post-merge cleanup.

PR #3135, opened 2026-05-04
test(quickjs): add dockerized postgres integration coverage for PTC loop affinity

hntrl:
This PR adds Docker-backed integration coverage for the QuickJS PTC async path when LangGraph is using a Postgres checkpointer.

The goal is to exercise the same loop-affinity scenario seen in production, but in a self-contained test that does not require preconfigured `POSTGRES_URL` state.

## Changes

- Add `tests/integration_tests/test_postgres.py` in `langchain-quickjs`.
- Start Postgres via `testcontainers` and create an `AsyncPostgresSaver` checkpointer from the container DSN.
- Assert the `tools.task` subagent async execution stays on the outer loop and returns successfully through the QuickJS bridge.
- Normalize the DSN emitted by `testcontainers` (`postgresql+psycopg2://`) to a psycopg-compatible scheme for `AsyncPostgresSaver`.
- Add test-only dependencies required for this path (`langgraph-checkpoint-postgres`, `testcontainers[postgres]`, and `psycopg[binary]`) and refresh `uv.lock`.

### DIFF
--- a/libs/partners/quickjs/tests/integration_tests/test_postgres.py
+++ b/libs/partners/quickjs/tests/integration_tests/test_postgres.py
@@ -1,0 +1,156 @@
+from __future__ import annotations
+
+import uuid
+from collections.abc import (
+    Iterator,  # noqa: TC003 — pydantic resolves annotations at runtime
+)
+from typing import Any
+
+import pytest
+from deepagents.middleware.subagents import SubAgentMiddleware
+from langchain.agents import create_agent
+from langchain_core.language_models.fake_chat_models import GenericFakeChatModel
+from langchain_core.messages import AIMessage, HumanMessage, ToolMessage
+from pydantic import Field
+
+from langchain_quickjs import REPLMiddleware
+
+try:
+    from langgraph.checkpoint.postgres.aio import AsyncPostgresSaver
+except ImportError:  # pragma: no cover - env-dependent optional dependency
+    AsyncPostgresSaver = None  # type: ignore[assignment]
+
+try:
+    from docker.errors import DockerException
+    from testcontainers.postgres import PostgresContainer
+except ImportError:  # pragma: no cover - env-dependent optional dependency
+    DockerException = RuntimeError  # type: ignore[assignment]
+    PostgresContainer = None  # type: ignore[assignment]
+
+
+pytestmark = [
+    pytest.mark.skipif(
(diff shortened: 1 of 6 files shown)


libs/partners/quickjs/tests/integration_tests/test_postgres.py
```diff
@@ -0,0 +1,162 @@
...
+        reason="testcontainers[postgres] not installed",
+    ),
+]
+
+
+class _FakeChatModel(GenericFakeChatModel):
+    messages: Iterator[AIMessage | str] = Field(exclude=True)
+
+    def bind_tools(self, tools: list[Any], **_: Any) -> _FakeChatModel:
+        del tools
```
eyurtsev: why is there `del` here?

libs/partners/quickjs/tests/integration_tests/test_postgres.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+async def test_ptc_task_with_postgres_checkpointer_keeps_loop_affinity(
+    postgres_url: str,
+) -> None:
+    owner_loop = asyncio.get_running_loop()
+    subagent_loop_ids: list[int] = []
+
+    def _subagent_sync(state: dict[str, Any], config: Any) -> dict[str, Any]:
+        del state, config
```
eyurtsev: why are there `del` here

libs/partners/quickjs/tests/integration_tests/test_postgres.py
```diff
@@ -0,0 +1,152 @@
...
+
+    The assertion here is intentionally behavioral: the nested `task` call
+    should complete successfully and return the subagent result.
+    """
+    # AsyncPostgresSaver.from_conn_string returns an async iterator/context manager.
+    assert AsyncPostgresSaver is not None  # narrowed by pytest skip
+    async with AsyncPostgresSaver.from_conn_string(postgres_url) as checkpointer:
+        await checkpointer.setup()
+        thread_id = f"qjs-pg-{uuid.uuid4().hex[:8]}"
+        agent = create_agent(
```
ccurme, line 115: should we use `create_deep_agent` so we don't have to populate `SubAgentMiddleware` below?

ccurme (dismissed):

PR #3182, opened 2026-05-05
feat(evals): add `open-fireworks` model preset

mdrxy:
Add an exclude filter to the eval runner so specific categories can be skipped without having to enumerate every other category. Previously, running "all categories except memory" required listing every category individually in the include filter.


mdrxy:
Will fix harbor tests later

PR #3210, opened 2026-05-06
feat(quickjs): propagate return types

ccurme:
separate commits here:
1. update the prompt to encourage more efficient use of repl (this is independent / extraneous)
2. deliver raw types from foreign functions, bypass ToolNode stringification (also adds support for injected tool IDs along the way)
3. inspect return types from tool callables, translate to JS and propagate to the prompt.

### DIFF
--- a/libs/partners/quickjs/langchain_quickjs/_prompt.py
+++ b/libs/partners/quickjs/langchain_quickjs/_prompt.py
@@ -84,8 +84,10 @@ def render_ptc_prompt(tools: Sequence[BaseTool], *, tool_name: str = "eval") -> 
         "### API Reference — `tools` namespace\n\n"
         "The agent tools listed below are exposed on the global object at "
         "`globalThis.tools` (also reachable as `tools`). Each takes a single "
-        "object argument and returns a Promise that resolves to a string "
-        "(structured return values are JSON-serialized).\n\n"
+        "object argument and returns a Promise that resolves to the tool's "
+        "native value: strings as strings, numbers as numbers, lists as "
+        "arrays, dicts as objects, and `None` as `null`. You do NOT need to "
+        "`JSON.parse` results — they are already typed.\n\n"
         "Invocation pattern: `await tools.<name>({ ... })`.\n\n"
         "- Use `await` to get tool results; combine with `Promise.all` for "
         "independent calls so they run concurrently.\n"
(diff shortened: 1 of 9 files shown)


libs/partners/quickjs/langchain_quickjs/_prompt.py
```diff
@@ -122,7 +171,162 @@ def _render_signature(fn_name: str, schema: dict[str, Any] | None) -> str:
...
+        return target
+    return None
+
+
+def _get_return_annotation(target: Callable[..., Any]) -> Any:
+    """Resolve the return annotation for a callable, ``Signature.empty`` if absent."""
+    with contextlib.suppress(TypeError, ValueError, NameError):
+        signature = inspect.signature(target)
+        resolved = get_type_hints(target)
+        return resolved.get("return", signature.return_annotation)
```
hntrl: won't block on this, but I do think we should consider exposing some hook for an output schema on BaseTool. Even if it is just coercing a pydantic model/TypedDict to something we can access on the tool instance

hntrl: That way we can just use the same rendering method that we use for input schemas

libs/partners/quickjs/langchain_quickjs/_repl.py
```diff
@@ -449,13 +497,28 @@ async def _ainvoke_tool_on_outer_loop(
...
+        Uses ``BaseTool.arun(args, tool_call_id=...)`` rather than
+        ``ainvoke(envelope)`` so the result is the tool's native return
+        value rather than a string-coerced ``ToolMessage``. We only pass
+        ``tool_call_id`` when the tool declares ``InjectedToolCallId`` —
+        otherwise ``_format_output`` would wrap the result anyway.
+        """
+        args = tool_call["args"]
+        tool_call_id = (
+            tool_call.get("id") if _tool_uses_injected_tool_call_id(tool) else None
+        )
```
hntrl, line 511: does this break our assumptions about how tools with this injected field works? E.g. if I specify a tool with `InjectedToolCallId` do I know that it can populate with `None`?

hntrl (dismissed):

libs/partners/quickjs/langchain_quickjs/_repl.py
ccurme: double-checked this and the tool should never see None as the injected tool call ID (let me know if that wasn't your concern)

libs/partners/quickjs/langchain_quickjs/_prompt.py
ccurme: refactored to use the same `_json_schema_to_js` helper that we use for inputs. I had originally stolen the implementation from `langchain-repl` in `libs/repl`. but this way we just have one path we have to manage.

libs/partners/quickjs/langchain_quickjs/_repl.py
hntrl: ok yeah I read this code wrong then

hntrl: basically I was worried that I might have a tool that uses InjectedToolCallId but it can be invoked with None

hntrl (approved):

PR #3232, opened 2026-05-08
fix: skip skills loading only when skills_metadata is non-empty

### LINKED ISSUE #3233: SkillsMiddleware skips loading on first turn when skills_metadata=[] (PrivateStateAttr default)
## Description

`SkillsMiddleware.before_agent` (and `abefore_agent`) skip loading skills when `"skills_metadata" in state` — but `PrivateStateAttr` initialises this field to `[]` by default in the graph state. This means the initial load is always skipped on the first turn.

## Steps to reproduce

1. Create an agent with `skills=["/skills/"]` pointing to a valid SKILL.md with YAML frontmatter
2. Invoke the agent
3. The system prompt shows "(No skills available yet)" even though valid skills exist

## Root cause

In `before_agent` (and `abefore_agent`):

```python
if "skills_metadata" in state:
    return None
```

The key exists with value `[]` due to `PrivateStateAttr` default, so the guard fires and `_list_skills_with_errors` is never called.

## Proposed fix

Change the guard to also check the list is non-empty:

```python
if "skills_metadata" in state and state["skills_metadata"]:
    return None
```

## Environment

- deepagents 0.5.7
- Reproduced with FilesystemBackend and CompositeBackend

BI4O:
Fixes #3233

## Problem

`SkillsMiddleware.before_agent` (and `abefore_agent`) skip loading skills when `"skills_metadata" in state` — but `PrivateStateAttr` initialises this field to `[]` by default in the graph state. This means the initial load is **always skipped on the first turn**, and the system prompt shows "(No skills available yet)" even when valid skills exist.

## Root cause

```python
# before_agent, ~line 1017
if "skills_metadata" in state:
    return None
```

The key exists (value is `[]`), so the guard fires and the middleware never reads the skill files.

## Fix

Check that the metadata list is non-empty before skipping:

```python
if "skills_metadata" in state and state["skills_metadata"]:
    return None
```

Applied to both `before_agent` and `abefore_agent`.

### DIFF
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -1013,8 +1013,11 @@ class SkillsMiddleware(AgentMiddleware[SkillsState, ContextT, ResponseT]):
         Returns:
             State update with `skills_metadata` populated, or `None` if already present.
         """
-        # Skip if skills_metadata is already present in state (even if empty)
-        if "skills_metadata" in state:
+        # Skip if skills_metadata is already loaded (non-empty list).
+        # PrivateStateAttr initialises the key to [] by default, so we
+        # must also check the list is non-empty to avoid skipping the
+        # initial load on the first turn.
+        if "skills_metadata" in state and state["skills_metadata"]:
             return None
 
         # Resolve backend (supports both direct instances and factory functions)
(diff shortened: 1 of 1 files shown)


BI4O:
This PR fixes #3233

PR #3252, opened 2026-05-08
feat(cli): incognito shell command mode

### LINKED ISSUE #2091: Incognito shell commands
### Privileged issue

### Issue Content

Don't show as context to the LLM

`!!`

### Area (Required)

- [ ] deepagents (SDK)
- [x] cli

miguel-mongiardim:
Closes #2091

---

Adds an incognito local shell mode for `!!` in the interactive CLI. Existing `!` commands still follow the normal shell path, while `!!` commands run locally and render their command and output as app-level transcript entries without adding user or assistant records to the model-visible conversation.

The input parser now detects the longest mode prefix so `!!` survives alongside `!`, history recall restores the right mode, queued and sent messages render the correct shell glyph, and shell-mode typing can be upgraded to incognito by entering a second `!`.

The UI distinguishes incognito shell mode with a dedicated theme color, input border title, message border, and `SHELL` status badge styling. Help text and startup tips document the new prefix, and focused unit coverage exercises prefix parsing, input and history behavior, queue draining, shell routing, local-only output, rendering, status, theme variables, help text, and tips.

### DIFF
--- a/libs/cli/deepagents_cli/app.py
+++ b/libs/cli/deepagents_cli/app.py
@@ -396,7 +396,7 @@ def _format_model_params(extra_kwargs: dict[str, Any] | None) -> str:
     return f" with model params {json.dumps(extra_kwargs, sort_keys=True)}"
 
 
-InputMode = Literal["normal", "shell", "command"]
+InputMode = Literal["normal", "shell", "shell_incognito", "command"]
 
 _TYPING_IDLE_THRESHOLD_SECONDS: float = 2.0
 """Seconds since the last keystroke after which the user is considered idle and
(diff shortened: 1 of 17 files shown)


mdrxy:
Hi, this doesn't appear to work on my end. Can you share a screen recording?

miguel-mongiardim:
Thanks, I found the issue and pushed a fix.

The problem happened when typing `!!` as separate keypresses. The first `!` immediately entered regular shell mode and stripped the visible prefix, so the second `!` was treated as another regular shell trigger.

I updated the input handling so pressing `!` while already in shell mode promotes the input to incognito shell mode. I also added a regression test for pressing `!`, then `!`, then submitting a command.

I tested it locally and confirmed it works now.

PR #3290, opened 2026-05-10
feat(cli): filter threads in switcher by directory

### LINKED ISSUE #1990: Filter threads to cwd
### Privileged issue

### Issue Content

Unsure if we should do this automatically or just make it a toggle option. CC has both but defaults to current dir.

### Area (Required)

- [ ] deepagents (SDK)
- [x] cli

Chesars:
Closes #1990.

## What this does

The `/threads` picker (and `deepagents threads list`) now scopes to the current working directory by default, so you only see conversations that started in this folder. The Options side-panel exposes a `Show threads from:` dropdown to widen the view to every project.

## How to use it

**Inside a session:**
- Open the picker with `/threads`.
- You'll see only threads from the current cwd. In the right-hand **Options** panel, the `Show threads from:` `Select` shows the active scope.
- Open the dropdown and choose **`All directories`** to see threads from any cwd. Choose **`Current directory`** to scope back.

**From the shell:**
- `deepagents threads list` → unchanged, lists everything (kept for script compatibility).
- `deepagents threads list --cwd` → only threads from the current directory.
- `deepagents threads list --cwd /some/path` → only threads from that path.

## Tests

All existing tests pass. Test fixtures that pass `initial_threads` opt out of the default cwd scoping with `filter_cwd=None`, since their mock rows don't carry a matching `cwd`. The Tab-focus tests were updated to account for the new `Select` sitting between the search input and the sort checkbox.

## Docs

Couldn't find any user-facing docs file that lists `/threads` controls or `deepagents threads list` flags beyond the in-CLI help text (which I did update in `ui.py`). If there's an external docs page that should mention the new dropdown/flag, happy to update it — just point me to it.


mdrxy:
I think in this case I'd rather this exist as an option alongside the existing `/thread` screen options as opposed to a key binding (https://textual.textualize.io/widget_gallery/#select)

Chesars:
Make sense, updated in 95d60f0

PR #3338, opened 2026-05-11
feat(sdk): add `ContextHubBackend` backend type

vishnu-ssuresh:
This PR introduces Context Hub as a first-class SDK backend through `ContextHubBackend`.

Why:
- Context Hub support existed as a deploy-only one-off in the CLI, which created duplication and long-term drift risk.
- Making it an SDK backend gives Deep Agents a canonical, reusable implementation behind `BackendProtocol`.
- This is the foundation for the follow-up PR that will make deploy import the SDK backend directly.

How:
- Added `ContextHubBackend` to the SDK backends package and exported it publicly.
- Moved the existing Context Hub behavior into the SDK backend while preserving semantics:
  - lazy tree load from Hub
  - missing repo (`pull_agent` 404) treated as empty and created on first write
  - in-memory file cache plus linked-entry tracking via `get_linked_entries()`
  - commit hash tracking with optimistic commit chaining via `parent_commit`
  - cache invalidation on mutating failures so subsequent reads re-pull fresh state
  - batch upload as a single commit for valid UTF-8 files with per-file success/error reporting
- Aligned implementation with SDK backend conventions:
  - use standardized backend error constants where applicable
  - replaced cache return type-ignore with an explicit runtime guard
- Added SDK tests for this backend (unit + integration) so behavior is enforced in the SDK itself.

Testing:
- `uv run --project libs/deepagents --group test pytest -n auto -vvv --disable-socket --allow-unix-socket tests/unit_tests/backends/test_context_hub_backend.py`
- `uv run --project libs/cli --group test pytest -n auto -vvv --disable-socket --allow-unix-socket tests/unit_tests/deploy/test_context_hub.py`
- Live Hub validation with `LANGSMITH_API_KEY` set:
  - `uv run --project libs/deepagents --group test pytest -n auto -vvv tests/integration_tests/test_context_hub_backend.py` (`7 passed`)
  - `uv run --project libs/cli --group test pytest -n auto -vvv tests/integration_tests/test_context_hub.py` (`7 passed`)


libs/deepagents/deepagents/backends/context_hub.py
```diff
@@ -0,0 +1,337 @@
...
+    ) -> None:
+        """Initialize ContextHubBackend.
+
+        Args:
+            identifier: Hub agent repo, as ``"owner/name"`` or ``"-/name"``.
+            client: LangSmith client. Defaults to ``Client()``.
+        """
+        self._identifier = identifier
+        self._client = client if client is not None else Client()
+        self._cache: dict[str, str] | None = None
```
ccurme, line 62: does the cache grow unbounded? any concerns around that?

libs/deepagents/deepagents/backends/context_hub.py
```diff
@@ -0,0 +1,337 @@
...
+            offset: 0-indexed starting line.
+            limit: Maximum number of lines.
+
+        Returns:
+            ReadResult with raw (unformatted) content.
+        """
+        hub_path = self._strip_prefix(file_path)
+        try:
+            cache = self._ensure_cache()
+        except Exception as exc:
```
ccurme: ok if this is broad? (e.g., would mask SDK errors)

libs/deepagents/deepagents/backends/context_hub.py
vishnu-ssuresh: Bounded by repo size, but I don't think we will huge repos to start. Can add a cache limit in follow up PR

libs/deepagents/deepagents/backends/context_hub.py
vishnu-ssuresh: 3db04264bf63a084a08c386cde43e8b8cfb2b30f

sydney-runkle (dismissed):
can we remove the "backend standard suite", not sure what's up w/ that

PR #3357, opened 2026-05-12
fix(sdk): remove ID assignment from _messages_delta_reducer

sydney-runkle:
## Problem

`_messages_delta_reducer` was assigning random UUIDs to ID-less messages (e.g. a `HumanMessage` created without an explicit `id`). The intent was to give the HumanMessage eviction mechanism stable IDs it could reference in a follow-up `Command` update. But this approach is fundamentally broken for `DeltaChannel`.

### Why it doesn't work

LangGraph checkpointers (`InMemorySaver`, Postgres, etc.) **serialize pending writes before `DeltaChannel.update()` runs**. The call order inside Pregel is:

1. Task produces output write (e.g. `HumanMessage(id=None)`)
2. `put_writes(...)` → serializes and stores the write — **still `id=None` at this point**
3. After all tasks complete: `DeltaChannel.update([write])` → reducer assigns `id=<uuid>` in-memory only

On the next `invoke` (e.g. turn 2 in a multi-turn conversation), `channels_from_checkpoint` finds no `_DeltaSnapshot` (none taken yet at `snapshot_frequency=50`) and calls `replay_writes` with the stored writes. The stored write has `id=None`, so the reducer assigns a **fresh random UUID** — different from the one baked into the eviction `Command`'s tagged message. Both the original and the tagged version co-exist as separate messages in the replayed state, surfacing as duplicate `HumanMessage`s in the model call.

### Why position-based IDs (`__auto_N__`) also don't work

A deterministic position-based scheme was explored as an alternative. It breaks in LangGraph's parallel execution model: two tool results are processed in **separate `update()` calls that both start from the same base state** (LangGraph does intermediate per-task updates before the final batched update). Both compute `non_none_count = N` and get `id=__auto_N__` — the second overwrites the first.

## Fix

Remove all ID assignment from the reducer. ID-less messages are simply appended without deduplication — exactly what LangGraph's own `_messages_delta_reducer` does, and what the docstring of `DeltaChannel` requires for batching invariance:

> `reducer(reducer(state, xs), ys) == reducer(state, xs + ys)`

Random UUID assignment violates this invariant because step-by-step and batched replay assign different UUIDs to the same ID-less message.

## Root cause and real fix

The real fix is **upstream in langchain-core**: auto-assign stable IDs at message creation time, so every message carries its ID before it ever reaches a channel or checkpointer. A separate PR against langchain-core is being opened for this.

Once langchain-core auto-assigns IDs, the eviction mechanism will work correctly: the `HumanMessage` will have a stable ID from birth, the eviction `Command` will reference that same ID, and the reducer will correctly update-in-place on replay.

## Changes

- `_messages_delta_reducer`: remove UUID assignment; ID-less messages are appended (matches LangGraph upstream)
- `TestLargeHumanMessageEviction` + two `TestArtifactsRoot` eviction tests: marked `xfail(strict=True)` with a detailed reason pointing to the langchain-core fix

## Test result

```
1566 passed, 84 skipped, 8 xfailed
```

### DIFF
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -2001,9 +2001,8 @@ class FilesystemMiddleware(AgentMiddleware[FilesystemState, ContextT, ResponseT]
                     },
                 }
             )
-            stable: list[AnyMessage] = [(m.model_copy(update={"id": str(uuid.uuid4())}) if m.id is None else m) for m in messages[:-1]] + [tagged]
-            state_command = Command(update={"messages": Overwrite(stable)})
-            messages = stable
+            state_command = Command(update={"messages": Overwrite([*messages[:-1], tagged])})
+            messages = [*messages[:-1], tagged]
 
         processed: list[AnyMessage] = []
         for msg in messages:
--- a/libs/deepagents/deepagents/middleware/filesystem.py
+++ b/libs/deepagents/deepagents/middleware/filesystem.py
@@ -27,10 +27,10 @@ from langchain.agents.middleware.types import (
 from langchain.tools import ToolRuntime
 from langchain.tools.tool_node import ToolCallRequest
 from langchain_core.messages import AnyMessage, BaseMessage, HumanMessage, RemoveMessage, ToolMessage
-from langgraph.graph.message import REMOVE_ALL_MESSAGES
 from langchain_core.messages.content import ContentBlock
 from langchain_core.tools import BaseTool, StructuredTool
 from langgraph.channels.delta import DeltaChannel
+from langgraph.graph.message import REMOVE_ALL_MESSAGES
 from langgraph.runtime import Runtime
 from langgraph.types import Command
 from pydantic import BaseModel, Field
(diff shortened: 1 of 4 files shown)


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1973,9 +1974,24 @@ def _apply_eviction_and_truncate(
...
+        producing a duplicate. The REMOVE_ALL_MESSAGES sentinel in the rewrite
+        discards everything before it in the replay sequence, then restores
+        all messages with stable IDs — eliminating the duplicate.
+
+        Cost: O(N) writes only when a new eviction fires, not every turn.
+
         Args:
             messages: The message list (may be modified if write succeeded).
-            write_result: Result of the backend write, or `None` if no new
+            write_result: Result of the backend write, or ``None`` if no new
```
sydney-runkle:
```suggestion
            write_result: Result of the backend write, or `None` if no new
```

ccurme (approved):

libs/deepagents/pyproject.toml
```diff
@@ -35,8 +35,7 @@ dependencies = [
 # replacements ship.
 prerelease = "allow"
 
-[project.optional-dependencies]
-quickjs = ["langchain-quickjs"]
+
```
sydney-runkle:
```suggestion
```

Issue #3424, opened 2026-05-16
Bug: `eval` interpreter tool does not await async IIFE return values


evan-rash:
### Submission checklist

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [x] quickjs
- [ ] runloop
- [ ] langsmith-sandbox
- [ ] Other / not sure / general

### Case 1 — top‑level await (works as expected)

Code passed to the interpreter:

```js
const v1 = await Promise.resolve(123);
v1;
```

Observed result (good):

<result>123</result>

### Case 2 — async IIFE with await (same logic, different shape)

Code passed to the interpreter:

```js
(async () => {
  const v2 = await Promise.resolve(456);
  return v2;
})();
```

Observed result (problematic):

<result kind="handle">[object]</result>

i.e., the interpreter is returning a handle / unresolved Promise instead of the resolved value `456`.
```

### Description

The JavaScript interpreter tool supports top‑level `await`, but **does not await Promises returned by an async IIFE** when that IIFE is the final expression. This leads to different behavior for semantically equivalent code:

- `const x = await Promise.resolve(123); x;` → returns `123` as the result.
- `(async () => { const x = await Promise.resolve(123); return x; })();` → returns a Promise/handle instead of `123`.

Either the interpreter should detect that the final expression is a Promise and await it, or the middleware prompt/docs should explicitly warn users not to return Promises (e.g. from async IIFEs) and require them to `await` such calls at top level.

## Why does this matter?

It seems like many models lean towards IIFEs by default, so the middleware should either handle this likely common scenario or explicitly instruct models not to use this pattern

### Environment / System Info

_No response_

lost-particles:
[...]

PR #3429, opened 2026-05-16
fix(quickjs): warn against bare async IIFE final expression in REPL prompt

lost-particles:
## Summary
- Adds a bullet to the QuickJS `eval` REPL system prompt explaining that the final expression must not be an unresolved Promise — a bare `(async () => { ... })();` surfaces as `[object]` instead of the resolved value.
- Regenerates the two prompt snapshots via `make update_snapshots`.

**Superseded by #3430** (engine-side fix). Closing this in favor of the proper auto-await fix.

Refs #3424


lost-particles:
Superseded by #3430, which fixes the bug at the engine layer (auto-awaits a final-expression Promise via `Handle.await_promise`) rather than relying on a prompt-side warning.

PR #3454, opened 2026-05-18
fix(sdk): anchor ripgrep glob to search root

### LINKED ISSUE #2732: ripgrep `glob` with directory components silently matches nothing
## Checked other resources

- [x] I searched existing issues and didn't find a similar one

## Description

`FilesystemBackend._ripgrep_search` passes the resolved absolute path directly as the search root to ripgrep:

```python
cmd.extend(["--", pattern, str(base_full)])
```

ripgrep's `--glob` flag matches against paths relative to the **process cwd**, not the search root. When `base_full` is absolute and the process cwd differs, glob patterns with directory components (e.g. `subdir/*.md`) silently match nothing — the search returns zero results with no error.

Simple globs without directory components (e.g. `*.md`) are unaffected because ripgrep auto-expands them to `**/*.md`.

## Reproduction

```bash
# Setup
mkdir -p /tmp/test-rg/docs
echo "hello world" > /tmp/test-rg/docs/guide.md

# cwd != search root → glob with directory component fails
cd /tmp
rg --glob "docs/*.md" -l -F "hello" /tmp/test-rg
# exit 1, no output

# cwd == search root → same glob works
cd /tmp/test-rg
rg --glob "docs/*.md" -l -F "hello" /tmp/test-rg
# /tmp/test-rg/docs/guide.md
```

Tested with ripgrep 15.1.0 on macOS (Darwin 25.4.0).

## Suggested fix

Pass `.` as the search path with `cwd=base_full` so ripgrep resolves globs against the intended directory:

```diff
- cmd.extend(["--", pattern, str(base_full)])
+ cmd.extend(["--", pattern, "."])

  proc = subprocess.run(
      cmd, capture_output=True, text=True, timeout=30, check=False,
+     cwd=str(base_full),
  )

- p = Path(ftext)
+ p = (base_full / ftext).resolve()
```

I have a PR ready: #2731

## System Info

- deepagents 0.4.12
- ripgrep 15.1.0
- Python 3.13.9
- macOS (Darwin 25.4.0)

nick-hollon-lc:
Fixes #2732.

`FilesystemBackend._ripgrep_search` passed an absolute search root to ripgrep, but `--glob` patterns are resolved against the process cwd. Globs with a directory component (e.g. `docs/*.md`) silently matched nothing when the agent's cwd differed from the search root.

Run rg with `cwd=base_full` and `.` as the search path so globs anchor correctly, and reanchor emitted relative paths back to absolute so virtual-mode handling is unchanged.

### DIFF
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -110,6 +110,16 @@ jobs:
         shell: bash
         run: uv sync --group test
 
+      - name: "🔍 Install ripgrep (Linux)"
+        if: runner.os == 'Linux'
+        shell: bash
+        run: sudo apt-get update && sudo apt-get install -y ripgrep
+
+      - name: "🔍 Install ripgrep (macOS)"
+        if: runner.os == 'macOS'
+        shell: bash
+        run: brew install ripgrep
+
       - name: "🧪 Run Unit Tests"
         if: runner.os != 'Windows'
         shell: bash
--- a/.github/workflows/_test.yml
+++ b/.github/workflows/_test.yml
@@ -110,15 +110,10 @@ jobs:
         shell: bash
         run: uv sync --group test
 
-      - name: "🔍 Install ripgrep (Linux)"
+      - name: "🔍 Install ripgrep"
         if: runner.os == 'Linux'
         shell: bash
         run: sudo apt-get update && sudo apt-get install -y ripgrep
-
-      - name: "🔍 Install ripgrep (macOS)"
-        if: runner.os == 'macOS'
-        shell: bash
-        run: brew install ripgrep
 
       - name: "🧪 Run Unit Tests"
         if: runner.os != 'Windows'
--- a/libs/deepagents/deepagents/backends/filesystem.py
+++ b/libs/deepagents/deepagents/backends/filesystem.py
@@ -570,7 +570,11 @@ class FilesystemBackend(BackendProtocol):
         cmd = ["rg", "--json", "-F"]  # -F enables fixed-string (literal) mode
         if include_glob:
             cmd.extend(["--glob", include_glob])
-        cmd.extend(["--", pattern, str(base_full)])
+        # ripgrep resolves --glob patterns against the process cwd, not the
+        # search root, so directory-component globs (e.g. "docs/*.md") fail
+        # silently when cwd != base_full. Run rg from base_full with "." as
+        # the search path so globs anchor correctly. (#2732)
+        cmd.extend(["--", pattern, "."])
 
         try:
             proc = subprocess.run(  # noqa: S603
@@ -579,6 +583,7 @@ class FilesystemBackend(BackendProtocol):
                 text=True,
                 timeout=30,
                 check=False,
+                cwd=str(base_full),
             )
         except (subprocess.TimeoutExpired, FileNotFoundError, PermissionError):
             return None
@@ -595,7 +600,11 @@ class FilesystemBackend(BackendProtocol):
             ftext = pdata.get("path", {}).get("text")
             if not ftext:
                 continue
-            p = Path(ftext)
+            # rg now runs with cwd=base_full and search path ".", so it emits
+            # paths relative to base_full. Re-anchor to absolute so virtual /
+            # normal mode path handling below sees the same shape as before.
+            raw = Path(ftext)
+            p = raw if raw.is_absolute() else (base_full / raw).resolve()
             if self.virtual_mode:
                 try:
                     virt = self._to_virtual_path(p)
--- a/libs/deepagents/deepagents/backends/filesystem.py
+++ b/libs/deepagents/deepagents/backends/filesystem.py
@@ -555,7 +555,7 @@ class FilesystemBackend(BackendProtocol):
                 matches.append({"path": fpath, "line": int(line_num), "text": line_text})
         return GrepResult(matches=matches)
 
-    def _ripgrep_search(self, pattern: str, base_full: Path, include_glob: str | None) -> dict[str, list[tuple[int, str]]] | None:  # noqa: C901  # Split except clauses for logging
+    def _ripgrep_search(self, pattern: str, base_full: Path, include_glob: str | None) -> dict[str, list[tuple[int, str]]] | None:  # noqa: C901, PLR0912  # Split except clauses for logging; dir/file + containment branches
         """Search using ripgrep with fixed-string (literal) mode.
 
         Args:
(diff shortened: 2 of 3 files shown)


mdrxy (changes requested):
## Critical

### `NotADirectoryError` unhandled — regression introduced by this PR

`filesystem.py`, `subprocess.run` call + `except` clause

Before this PR, `rg` received `str(base_full)` as a search argument; if `base_full` was a file, ripgrep searched it directly. Now `cwd=str(base_full)` causes Python to `chdir()` in the child before exec. If `base_full` is a file, the kernel returns `ENOTDIR` → Python raises `NotADirectoryError` — a subclass of `OSError`, but **not** a subclass of `FileNotFoundError` or `PermissionError`. The existing `except` clause misses it, so it propagates unhandled. Previously working single-file greps (e.g. `grep(pattern, path="pyproject.toml")`) now crash instead of falling back to `_python_search` (which handles `base_full.is_file()` correctly at the `root = base_full if base_full.is_dir() else base_full.parent` line).

Two complementary fixes:

```python
# Minimal: add NotADirectoryError to the except clause
except (subprocess.TimeoutExpired, FileNotFoundError, PermissionError, NotADirectoryError):
    return None
```

```python
# More correct: skip the cwd= trick when base_full is a file
if base_full.is_dir():
    cmd.extend(["--", pattern, "."])
    subprocess_kwargs["cwd"] = str(base_full)
else:
    cmd.extend(["--", pattern, str(base_full)])
    # glob is irrelevant for a single named file; no cwd override needed
```

---

## Important

### `.resolve()` silently changes symlink semantics in non-virtual mode

`filesystem.py`, path re-anchoring line (`p = raw if raw.is_absolute() else (base_full / raw).resolve()`)

The PR description says this preserves "the same shape as before." Pre-fix, `rg` emitted absolute paths and the code did `p = Path(ftext)` — no resolution. Post-fix, non-virtual mode now calls `.resolve()`, which collapses symlinks. Users who grep a directory tree containing symlinks will see resolved targets instead of the symlinked paths they searched under. Easy fix: use `base_full / raw` (already absolute) without `.resolve()` in the non-virtual branch.

---

### `virtual_mode=True` path is not covered by the new test

`test_filesystem_backend.py`

The existing virtual-mode test stubs out `_ripgrep_search` entirely, so the re-anchoring code (`(base_full / raw).resolve()` → `_to_virtual_path`) is exercised by zero tests. A regression there would silently drop results via the `except ValueError: continue` guard inside `_to_virtual_path`. Here's a ready-to-add variant:

```python
def test_grep_ripgrep_glob_virtual_mode_with_directory_component(
    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
) -> None:
    if shutil.which("rg") is None:
        pytest.skip("ripgrep not installed")

    root = tmp_path / "project"
    (root / "docs").mkdir(parents=True)
    (root / "docs" / "guide.md").write_text("hello world\n")
    (root / "notes.md").write_text("hello world\n")

    other_cwd = tmp_path / "elsewhere"
    other_cwd.mkdir()
    monkeypatch.chdir(other_cwd)

    be = FilesystemBackend(root_dir=str(root), virtual_mode=True)
    matches = be.grep("hello", path="/", glob="docs/*.md").matches
    assert matches is not None
    matched_paths = [m["path"] for m in matches]
    assert any(p == "/docs/guide.md" for p in matched_paths)
    assert not any("notes" in p for p in matched_paths)
```

---

### No path containment check after re-anchoring (non-virtual mode)

`filesystem.py`, path re-anchoring line

Pre-fix, `rg` emitted absolute paths beginning with `base_full`, so containment was implicit. Post-fix, paths are relative and re-anchored manually. Symlinks inside `base_full` that point outside it cause `.resolve()` to produce an out-of-root path that flows into the result set unchecked in `virtual_mode=False`. (`virtual_mode=True` is safe — `_to_virtual_path` raises `ValueError` which is caught.) Suggested addition after resolving:

```python
try:
    p.relative_to(base_full.resolve())
except ValueError:
    continue  # escaped the search root via symlink; skip
```

---

## Suggestions

**Comment phrasing** (`filesystem.py`, first comment block): "resolves `--glob` patterns against the process cwd" is slightly imprecise — ripgrep matches globs against paths as passed; the cwd only matters because it determines what `.` resolves to. More accurate phrasing: *"When rg is given an absolute search path, directory-component globs silently match nothing if the process cwd ≠ search root (#2732). Fix: set `cwd=base_full` and use `.` as the search path."*

**`noqa: S603` missing justification** (pre-existing, line 580): CLAUDE.md requires a comment on every `noqa`. Suggested: `# noqa: S603  # cmd is a list literal; "--" guards against flag injection from pattern`

**Confirm CI has `rg` installed**: the new regression test skips silently when `rg` is absent. If CI doesn't have ripgrep, the exact bug this PR fixes is untested in CI.

**Broad `except` swallows failures without logging**: `PermissionError` and `TimeoutExpired` are silently swallowed with no log, making fallback activation invisible. Consider splitting into per-exception branches with a `logger.debug` / `logger.warning` before `return None`.

nick-hollon-lc:
@mdrxy responses inline on each item — addressed in 7301e2fc.

**Critical**

- **`NotADirectoryError` unhandled.** Verified: `subprocess.run(cwd=<file>)` raises `NotADirectoryError`, which is not a subclass of `FileNotFoundError`/`PermissionError`. Took your "more correct" variant — only set `cwd=base_full` when `base_full.is_dir()`; for single-file paths we keep the absolute search arg. Also added `NotADirectoryError` to the except clause as belt-and-suspenders. New `test_grep_on_single_file_path` covers this.

**Important**

- **`.resolve()` symlink semantics.** Agreed — the join now uses `(base_full / raw)` without `.resolve()` so display paths keep symlink form. `.resolve()` is only invoked for the containment comparison (next item). New `test_grep_preserves_symlink_path_in_results` covers it (symlinked dir).
- **Virtual mode untested.** Added `test_grep_ripgrep_glob_virtual_mode` mirroring the existing test but with `virtual_mode=True` and a virtual-path assertion.
- **Containment check.** Added `p.resolve().relative_to(base_full.resolve())` guard after re-anchoring; skips with a `logger.debug` if the result resolves outside the root. No new regression test for this one since rg doesn't follow symlinks by default and I couldn't construct an observable escape — the check is defense-in-depth.

**Suggestions**

- **Comment phrasing** — rewrote the comment block to match your wording (search-path vs cwd) rather than the original "rg resolves globs against cwd" framing.
- **`noqa: S603` justification** — left as-is. Pre-existing on the line I touched; happy to address in a follow-up if you'd like a separate PR.
- **CI lacks rg** — confirmed: my first test was skipping silently on Ubuntu runners. Added `apt-get install ripgrep` / `brew install ripgrep` steps to `_test.yml` for Linux and macOS.
- **Broad `except` swallows failures without logging** — left as-is, pre-existing pattern. Same offer for a separate PR.

mdrxy (approved):

PR #3471, opened 2026-05-19
fix(sdk): surface backend errors in `SandboxBackend.grep`

### LINKED ISSUE #3441: `SandboxBackend.grep` crashes with `ValueError` when container exec fails
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
import docker
from deepagents.backends.sandbox import SandboxBackend, ExecuteResult


class DockerSandbox(SandboxBackend):
    """Minimal SandboxBackend that runs each exec against a real container."""

    def __init__(self, container, work_dir: str) -> None:
        self._container = container
        self._work_dir = work_dir

    @property
    def id(self) -> str:
        return self._container.id

    def execute(self, cmd: str, **_kwargs) -> ExecuteResult:
        exit_code, output = self._container.exec_run(
            ["sh", "-c", cmd], workdir=self._work_dir,
        )
        return ExecuteResult(
            output=output.decode("utf-8", errors="replace"),
            exit_code=exit_code,
        )


client = docker.from_env()
container = client.containers.run(
    "alpine:3.20", command=["sleep", "60"], detach=True, auto_remove=True,
)
try:
    backend = DockerSandbox(container, work_dir="/does-not-exist")
    backend.grep("TODO", path="/")
finally:
    container.stop(timeout=5)

# ValueError: invalid literal for int() with base 10: ' exec failed'
```

```shell
File ".../deepagents/backends/sandbox.py", line 807, in grep
    "line": int(parts[1]),
ValueError: invalid literal for int() with base 10: ' exec failed'
```

### Description

I'm running `grep` via a `SandboxBackend` against a Docker container whose `work_dir` doesn't exist inside the container.

* **Expected**: `GrepResult(error="...", matches=None)`, matching how other backend methods surface exec failures (e.g. `glob` returns `GlobResult(matches=None, error=...)` when the search path is unreachable).
* **Actual**: the tool node crashes with `ValueError`, propagating out of the agent loop and terminating the run.

### Environment / System Info

OS: Linux (Arch 7.0.3-arch1-2)
Python: 3.12
deepagents: 0.6.1 (latest released)
langgraph: 1.2.0
langchain: 1.3.1
docker (host): native Linux

Nik-Reddy:
Closes #3441

---

`grep()` masks its own non-zero exit codes with `2>/dev/null || true`, but when the backend (e.g., a container exec) fails *before* the shell starts, that redirection never takes effect: the backend's error text lands in `result.output` with a non-zero `exit_code`, and the match parser then crashes with `ValueError: invalid literal for int() with base 10: ' exec failed'` while reading the second `:`-delimited field as a line number.

This mirrors the convention already used by `ls`/`read`/`edit`/`glob` (after #3359): check `exit_code` first and return a structured `GrepResult(error=...)` instead of raising.

As defense in depth, the match parser also rejects any line that doesn't fit the `path:line:text` shape — whether the line-number field isn't an integer or the line has no colons at all — surfacing a structured error instead of silently dropping the line (which would mask the failure as a no-match). When this happens, any matches accumulated before the malformed line are discarded, matching `glob()`'s preference for a clean error over ambiguous partial output.

`LocalSubprocessSandbox` is intentionally unaffected: its shell always runs, so `|| true` continues to force exit 0 on grep-side failures (no behavior change for the no-match-on-missing-path case there). The fix targets exactly the failure surface described in the issue — backend exec failures that produce output before the shell is reached.

Note: @pierrel mentioned on the issue that they had a fix ready. Happy to defer if their work is further along — closing this PR if so. Posting in parallel given the issue is two weeks old and the bug is currently blocking grep against any Docker-style sandbox with an unreachable path.


Nik-Reddy:
Closing this in favor of #3637, which landed on main with the same exit_code-first plus structured GrepResult(error=...) approach. The additional grep -Z NUL-delimited filename change in #3637 is a nice orthogonal hardening for paths containing colons.

Thanks for picking up the design and for the in-branch refinements on 7480fd6. The original root-cause writeup and test plan from 2026-05-19 lives on issue #3441 for the audit trail.

Nothing left to merge here.

PR #3472, opened 2026-05-19
feat(quickjs): add swarm task tool

colifran:
Port of https://github.com/langchain-ai/deepagentsjs/pull/500

### DIFF
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -353,7 +353,7 @@ def _validate_skill_name(name: str, directory_name: str) -> tuple[bool, str]:
 def _parse_allowed_tools(raw_tools: object, skill_path: str) -> list[str]:
     """Parse the `allowed-tools` frontmatter value into a list of tool names."""
     if isinstance(raw_tools, str):
-        return [t.strip(",") for t in raw_tools.split() if t.strip(",")]
+        return [t.strip(",").strip() for t in raw_tools.split() if t.strip(",").strip()]
     if raw_tools is not None:
         logger.warning(
             "Ignoring non-string 'allowed-tools' in %s (got %s)",
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -363,7 +363,7 @@ def _parse_allowed_tools(raw_tools: object, skill_path: str) -> list[str]:
     return []
 
 
-def _parse_skill_metadata(  # noqa: C901
+def _parse_skill_metadata(
     content: str,
     skill_path: str,
     directory_name: str,
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -363,14 +363,6 @@ def _parse_allowed_tools(raw_tools: object, skill_path: str) -> list[str]:
     return []
 
 
-def _parse_required_ptc_tools(metadata_obj: Any) -> list[str]:  # noqa: ANN401
-    """Extract `required-ptc-tools` from the metadata mapping."""
-    if not isinstance(metadata_obj, dict):
-        return []
-    raw = metadata_obj.get("required-ptc-tools")
-    return str(raw).split() if raw else []
-
-
 def _parse_skill_metadata(  # noqa: C901
     content: str,
     skill_path: str,
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -104,7 +104,7 @@ import json
 import logging
 import re
 from pathlib import PurePosixPath
-from typing import TYPE_CHECKING, Annotated
+from typing import TYPE_CHECKING, Annotated, Any
 
 import yaml
 from langchain.agents.middleware.types import PrivateStateAttr
@@ -363,7 +363,7 @@ def _parse_allowed_tools(raw_tools: object, skill_path: str) -> list[str]:
     return []
 
 
-def _parse_required_ptc_tools(metadata_obj: object) -> list[str]:
+def _parse_required_ptc_tools(metadata_obj: Any) -> list[str]:  # noqa: ANN401
     """Extract ``required-ptc-tools`` from the metadata mapping."""
     if not isinstance(metadata_obj, dict):
         return []
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -288,30 +288,6 @@ class SkillMetadata(TypedDict):
     - Space-delimited list of tool names
     """
 
-<<<<<<< HEAD
-=======
-    module: NotRequired[str | None]
-    """Path to a JS/TS entrypoint file for a QuickJS REPL module, relative to the skill directory.
-
-    Warning: this is experimental.
-
-    When present, consumers of this metadata (notably `langchain-quickjs`'s
-    `CodeInterpreterMiddleware`) may install the skill as a dynamic-importable ES
-    module. The string is a POSIX path like `./index.ts` pointing at a
-    file inside the skill dir. This middleware only parses and validates
-    the field — it does not load or execute any JavaScript.
-    """
-
-    required_ptc_tools: NotRequired[list[str]]
-    """PTC tool names that must be present in the QuickJS middleware's ``ptc`` configuration
-    for this skill to function.
-
-    The middleware validates these at skill-load time and returns a descriptive error
-    if any are missing. Tool names should match the agent tool names as configured
-    in ``ptc`` (e.g. ``read_file``, not ``readFile``).
-    """
-
->>>>>>> 46df55e1 (add required ptc tools to skill metadata)
 
 class SkillsState(AgentState):
     """State for the skills middleware."""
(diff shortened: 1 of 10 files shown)


libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -300,6 +300,15 @@ class SkillMetadata(TypedDict):
     the field — it does not load or execute any JavaScript.
     """
 
+    required_ptc_tools: NotRequired[list[str]]
+    """PTC tool names that must be present in the QuickJS middleware's ``ptc`` configuration
```
ccurme: (throughout) use single backticks as we don't use sphinx for docs

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
```diff
@@ -0,0 +1,363 @@
...
+    from langchain_core.runnables import Runnable
+
+logger = logging.getLogger(__name__)
+
+SwarmTaskMode = Literal["agent", "invoke"]
+
+_VARIANT_TTL_S = 60.0
+
+
+class SwarmSubAgent(BaseModel):
```
ccurme: I think prefer typeddict or dataclass unless there's a reason to use BaseModel here

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
```diff
@@ -0,0 +1,363 @@
...
+    model_config = {"arbitrary_types_allowed": True}
+
+
+def _is_schema_node(value: object) -> bool:
+    """Check whether a value is a plain dict suitable for JSON Schema traversal."""
+    return isinstance(value, dict)
+
+
+def normalize_schema(schema: dict[str, Any]) -> dict[str, Any]:
+    """Recursively add ``additionalProperties: false`` to every object-typed node."""
```
ccurme: I don't think this fix should live here, we should be able to pass json schema directly to subagents. if they require setting additionalProperties False, they will manage that.

libs/partners/quickjs/langchain_quickjs/_skills.py
```diff
@@ -230,6 +230,14 @@ def _build_scope_modules(
     installed_index = _pick_installed_index_name(entry_rel)
     entry_installed = False
 
+    # When the entrypoint lives in a subdirectory (e.g. "scripts/index.ts"),
+    # it gets flattened to root-level "index.<ext>" for quickjs-rs bare-
+    # specifier resolution.  All sibling files must be relocated the same
+    # way so that relative imports (e.g. `import "./table.js"` from the
+    # now-root-level index) still resolve.
+    entry_parent = str(PurePosixPath(entry_rel).parent)
+    strip_prefix = entry_parent + "/" if entry_parent != "." else ""
```
ccurme, line 239: is this kosher security-wise

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
```diff
@@ -0,0 +1,363 @@
...
+        return len(self._entries)
+
+    def _sweep(self) -> None:
+        now = time.monotonic()
+        expired = [k for k, (_, ts) in self._entries.items() if now - ts > self._ttl_s]
+        for k in expired:
+            del self._entries[k]
+
+
+def _content_to_string(content: Any) -> str:
```
ccurme: how about `message.text`?

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
```diff
@@ -0,0 +1,363 @@
...
+        ],
+        tool_choice=tool_name,
+    )
+    response = await bound.ainvoke(messages)
+
+    if not isinstance(response, AIMessage) or not response.tool_calls:
+        msg = "invoke mode with response_schema: model did not return structured output"
+        raise ValueError(msg)
+
+    return json.dumps(response.tool_calls[0]["args"])
```
ccurme: I think it's best to use `model.with_structured_output` for this. this will let the model use the right strategy (whether tool calling or dedicated structured output features). No reason to hand-roll something using tool calling here.

libs/deepagents/deepagents/middleware/skills.py
colifran: Updated!

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
colifran: Updated to use dataclass.

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
colifran: Updated. I think there might actually be a gap here where Python covers this but JS doesn't. I'll do a deeper dive on the JS side.

libs/partners/quickjs/langchain_quickjs/_skills.py
colifran: Yeah this should be safe. `entry_rel` is validated upstream in `_validate_module_path` which rejects absolute paths and `../` traversals, and `_relative()` ensures all file paths are under `skill_dir`. But I added an additional check on the relocated result that raises `SkillInstallError` if it would escape the scope.

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
colifran: Updated.

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
colifran: Updated.

libs/partners/quickjs/langchain_quickjs/_swarm_task.py
```diff
@@ -0,0 +1,363 @@
...
+    response_schema: dict[str, Any] | None,
+    variant_cache: VariantCache,
+) -> str:
+    """Full agentic invocation with optional schema-constrained variant caching."""
+    agent = entry.agent
+
+    if response_schema is not None:
+        normalized = _normalize_response_schema(response_schema)
+        cache_key = f"{entry.spec.name}::{json.dumps(normalized, sort_keys=True)}"
+        agent = variant_cache.get_or_create(
```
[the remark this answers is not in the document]
colifran: This appears to be a valid vulnerability. I just pushed a fix for this - I added schema size/depth/property limits (4 KB, 5 levels, 32 properties) validated at the _run entry point, and a max entries cap (64) with LRU eviction on VariantCache

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -300,6 +300,15 @@ class SkillMetadata(TypedDict):
...
 
+    required_ptc_tools: NotRequired[list[str]]
+    """PTC tool names that must be present in the QuickJS middleware's `ptc` configuration
+    for this skill to function.
+
+    The middleware validates these at skill-load time and returns a descriptive error
+    if any are missing. Tool names should match the agent tool names as configured
+    in `ptc` (e.g. `read_file`, not `readFile`).
+    """
+
```
hntrl: is this not something we're gleaning from metadata? feels awk to me to have this at top level

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -362,6 +371,27 @@ def _validate_skill_name(name: str, directory_name: str) -> tuple[bool, str]:
     return True, ""
 
 
+def _parse_allowed_tools(raw_tools: object, skill_path: str) -> list[str]:
+    """Parse the `allowed-tools` frontmatter value into a list of tool names."""
+    if isinstance(raw_tools, str):
+        return [t.strip(",") for t in raw_tools.split() if t.strip(",")]
```
hntrl: maybe we want to trim entries too to get things like `foo, bar, baz`

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -431,21 +461,7 @@ def _parse_skill_metadata(  # noqa: C901
...
-        ]
-    else:
-        if raw_tools is not None:
-            logger.warning(
-                "Ignoring non-string 'allowed-tools' in %s (got %s)",
-                skill_path,
-                type(raw_tools).__name__,
-            )
-        allowed_tools = []
+    allowed_tools = _parse_allowed_tools(frontmatter_data.get("allowed-tools"), skill_path)
```
hntrl, line 435: maybe we should fix while we're here: allowed tools can also be space separated: https://agentskills.io/specification#allowed-tools-field

libs/deepagents/tests/unit_tests/middleware/test_skills_middleware.py
```diff
@@ -2183,3 +2183,119 @@ def test_modify_request_uses_custom_template() -> None:
...
+# --- required_ptc_tools parsing -------------------------------------------
+
+
+def test_parse_skill_metadata_required_ptc_tools_space_delimited() -> None:
+    """Parse space-delimited required-ptc-tools from nested metadata."""
+    content = """---
+name: test-skill
+description: A test skill
+metadata:
+  required-ptc-tools: swarm_task read_file write_file glob
```
hntrl, line 2116: wait what did I just read the python code wrong

libs/deepagents/deepagents/middleware/skills.py
sydney-runkle: i don't think we should have this on our generic skills spec right? this is very quickjs specific?

colifran: Yes - agree. I'll update

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -350,7 +350,20 @@ def _validate_skill_name(name: str, directory_name: str) -> tuple[bool, str]:
     return True, ""
 
 
-def _parse_skill_metadata(  # noqa: C901
+def _parse_allowed_tools(raw_tools: object, skill_path: str) -> list[str]:
```
sydney-runkle, line 353: this is just moving the helper function right?

sydney-runkle (approved):

libs/deepagents/deepagents/middleware/skills.py
colifran: Right!

PR #3499, opened 2026-05-20
fix(quickjs): auto-await final-expression Promise in eval REPL

### LINKED ISSUE #3424: Bug: `eval` interpreter tool does not await async IIFE return values

lost-particles:
## Summary

- Fixes #3424 at the engine layer: the QuickJS `eval` REPL auto-awaits a final-expression Promise (e.g. a bare async IIFE) so the model sees the resolved value rather than `[object]`.
- Switches `_aeval_async` from `ctx.eval_async` to `ctx.eval_handle_async` + `Handle.await_promise` + `Handle.to_python`. `await_promise` is a no-op for non-Promise handles, so the existing fast paths are unchanged.
- Five regression tests in `test_repl_middleware.py`.

(Re-opening: the previous attempt at this fix was #3430, which got auto-closed by the issue-link bot before assignment and can no longer be reopened after the force-push to address review feedback. This PR carries that fix forward, rebased on current `main` and addressing @hntrl's review.)

## Root cause

`ctx.eval_async(code)` returns whatever the script's final expression evaluates to. When that expression is `(async () => { ... })()` the value is a Promise object; `quickjs_rs` cannot marshal a Promise to Python and raises `MarshalError`. The previous handler re-evaluated the code via `eval_handle_async` and formatted the handle as `"[" + handle.type_of + "]"` — which for a Promise is `"[object]"`.

Result: semantically equivalent code surfaces different values to the agent:

```js
const x = await Promise.resolve(123); x;
// → 123

(async () => { const x = await Promise.resolve(123); return x; })();
// → [object]    ← the headline bug
```

## Fix

`_aeval_async` now drives evaluation through handles:

1. `handle = await ctx.eval_handle_async(code, timeout=...)`
2. If `handle.is_promise()`, drive it via `await handle.await_promise(timeout=...)` to get a handle to the resolved value.
3. Marshal via `resolved.to_python()`; fall back to `format_handle(resolved)` on `MarshalError` (e.g. for genuinely unmarshalable values like functions). MarshalError is handled inline — no helper, no tuple return — so all QJS error handling stays co-located in one function.

### Side benefit — no more side-effect doubling

The previous `MarshalError` handler called `_describe_via_handle_async(code)`, which re-evaluated the user's program to get a handle. That meant any user code that hit the marshal-fallback path (functions, Promises) ran twice — `console.log` was double-emitted, host-bridge tools were double-invoked, etc. With the new flow the program runs exactly once and the resulting handle is reused. The unused `_describe_via_handle_async` and `_HANDLE_PLACEHOLDER` sentinel are removed.

## Response to review on #3430

- **Inline `_aeval_and_marshal` back into `_aeval_async`** (@hntrl) — done. The helper is gone; all QJS error handling lives in one function.
- **Drop the dual return** (@hntrl) — done. MarshalError writes directly to `outcome.result` / `outcome.result_kind` instead of returning `(text, kind)`.
- **Rename `test_top_level_await_still_works`** (@hntrl, "still works as of when?") — renamed to `test_top_level_await_marshals_resolved_value` with a docstring explaining what it pins.

## Test plan

Regression tests in `tests/unit_tests/test_repl_middleware.py` cover:

- [x] **Headline bug** — async IIFE returning Promise unwraps to the resolved value
- [x] **Plain `Promise.resolve(...)`** as final expression unwraps
- [x] **Top-level await marshals resolved value** — existing path unaffected
- [x] **Promise rejection** — surfaces as `JSError` with the JS error type preserved
- [x] **No side-effect doubling** — `console.log` count == 1 even when the final value is a Promise

Other verification:

- [x] `make test` — **170 tests pass** (165 existing + 5 new), 91% coverage
- [x] `make lint` — ruff check + format + ty all clean

Fixes #3424


libs/partners/quickjs/langchain_quickjs/_repl.py
```diff
@@ -739,20 +733,41 @@ async def _aeval_async(  # noqa: C901
             outer_loop=outer_loop,
         )
         try:
-            value = await ctx.eval_async(code, timeout=self._per_call_timeout)
-            outcome.result = stringify(value)
+            # Drive any final-expression Promise (e.g. a bare async
+            # IIFE) to its resolved value before marshaling. Without
+            # this the Promise object itself fails to marshal and
+            # the result surfaces as ``[object]`` rather than the
+            # awaited value. Issue #3424.
```
hntrl:
```suggestion
            # Drive any final-expression Promise (e.g. a bare async
            # IIFE) to its resolved value before marshaling. Without
            # this the Promise object itself fails to marshal and
            # the result surfaces as ``[object]`` rather than the
            # awaited value.
```

hntrl (dismissed):

hntrl (approved):

hntrl (approved):

PR #3500, opened 2026-05-20
fix(code): persist `_context_tokens` to local cache instead of graph state

mdrxy:
`_context_tokens` is dcode-client display state — neither the agent nor the deepagents SDK reads it — but it was previously pushed into LangGraph state via `aupdate_state`. For remote agents (`langgraph dev` / LangGraph Platform), every `aupdate_state` call traverses HTTP and the server emits a root-level `UpdateState` LangSmith trace on every successful turn that the client cannot suppress (no `langsmith_tracing` kwarg on `threads.update_state`, no server-side header honored, no `TracingMiddleware` installed). Moving the value to a per-thread JSON file under `~/.deepagents/.state/context_tokens/` eliminates the trace noise. Graph state remains the right long-term home — once upstream gains a way to suppress tracing on `threads.update_state` the field should move back so cross-machine resume can carry the count again.

### DIFF
--- a/libs/code/deepagents_code/app.py
+++ b/libs/code/deepagents_code/app.py
@@ -5576,7 +5576,7 @@ class DeepAgentsApp(App):
             self._on_tokens_update(result.tokens_after)
             from deepagents_code.textual_adapter import _persist_context_tokens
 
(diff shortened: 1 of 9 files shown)


mdrxy:
Going with #3496

PR #3513, opened 2026-05-20
fix(sdk): assign UUIDs to ID-less messages in _messages_delta_reducer

sydney-runkle:
## Summary

- `_messages_delta_reducer` now assigns `str(uuid.uuid4())` to messages with `id=None` inside the existing write-loop iteration (no extra pass), matching the behaviour of `add_messages`
- Version bumped `0.6.2 → 0.6.3`
- `uv lock` refreshed across all `libs/` and `examples/` packages

## Test plan

- [ ] Confirm messages written without an explicit ID arrive in state with a stable UUID
- [ ] Confirm deduplication still works for messages that already carry an ID
- [ ] Confirm `RemoveMessage` tombstoning still works correctly


libs/deepagents/deepagents/_messages_reducer.py
```diff
@@ -65,6 +64,9 @@ def _messages_delta_reducer(  # noqa: C901
     for msg in msgs:
         mid = msg.id
         if mid is None:
+            msg.id = str(uuid.uuid4())
```
ccurme, line 72: if these are completed messages, this should be fine.

if they are chunks, these IDs will append when the chunks are aggregated.

they don't append if they begin with a [magic prefix](https://github.com/langchain-ai/langchain/blob/ac41199338b00ba3900bb5ff424a98c54e521d7e/libs/core/langchain_core/utils/utils.py#L498). all chat models auto-assign IDs to message chunks with this prefix. so I suspect not relevant here!

ccurme (dismissed):

PR #3529, opened 2026-05-21
feat(sdk): `RubricMiddleware` for self-evaluated agent iteration

davibinboi:
### Fixes:
Adds RubricMiddleware, which lets a caller declare what correct outputs look like via a rubric supplied on state.

After the agent completes an iteration and has what it thinks is a result, a separate grader sub-agent spins up, evaluates the transcript, and determines if further revision is needed. If so, the main agent is looped again with the grader's feedback on what in the rubric's criteria was not satisfactory (synthetic HumanMessage tagged with name="rubric_grader"). The loop terminates on satisfied, max_iterations_reached, or grader failed. KeyboardInterrupt and asyncio.CancelledError propagate so Ctrl+C / cancellation are preserved.

### Usage:

```python
from deepagents import RubricMiddleware, create_deep_agent
from langchain.messages import HumanMessage
from langchain.tools import tool
from langgraph.checkpoint.memory import InMemorySaver
from langsmith import Client

agent = create_deep_agent(
    model="anthropic:claude-haiku-4-5",
    middleware=[
        RubricMiddleware(
            model="anthropic:claude-haiku-4-5",
            system_prompt="You are an editor grading short paragraphs against a rubric.",
            tools=[check_lipogram],
            max_iterations=5,
            on_evaluation=lambda ev: print(ev["result"], ev["explanation"]),
        ),
    ],
    checkpointer=InMemorySaver(),
)

config = {"configurable": {"thread_id": "example"}}

result = agent.invoke(
        {
            "messages": [
                HumanMessage(
                    "Write a short paragraph (3-4 sentences) about the ocean without using the letter 'e'."
                )
            ],
            "rubric": (
                "- The paragraph contains zero instances of the letter 'e' \n"
                "- The paragraph is about the ocean and includes at least two vivid sensory details"
                " (sight, sound, smell, touch, or taste).\n"
                "- The paragraph is 3-4 sentences long."
            ),
        },
        config=config,
    )
```

RubricMiddleware takes flat kwargs:

- model (required): the grader's model. Accepts a "provider:model-id" string or a BaseChatModel instance.
- system_prompt (optional): replaces the built-in grader prompt.
- tools (optional): tools the grader may call before verdicting (e.g. count_words, a test runner). Defaults to no tools (pure-LLM grader).
- max_iterations (optional, default 3, hard-capped at 20): grader iterations per rubric attempt before terminating with max_iterations_reached.
- on_evaluation (optional): callback invoked with each RubricEvaluation.

The rubric itself is supplied per-invocation on state ({"rubric": "..."}) and is sticky across follow-up turns on the same thread.

### What's included:

**`RubricMiddleware` core** — `libs/deepagents/deepagents/middleware/outcomes.py`
The middleware itself, plus all public types: `RubricResult`, `RubricState`, `RubricEvaluation`, `CriterionEval`, `GraderResponse`, and the `RUBRIC_GRADER_MESSAGE_SOURCE` tag.

**Re-exports**:
· `deepagents/__init__.py` (just the middleware) 
· `deepagents/middleware/__init__.py` (middleware + public types)

---

**Unit tests** — `tests/unit_tests/middleware/test_rubric_middleware.py`

Covers construction validation, `before_agent` rubric-change detection, grader plumbing (lazy construction, model/prompt/tools propagation, nonce-bracketed prompt-injection hardening), transcript shaping, and multi-invocation rubric stickiness.

**End-to-end tests** — `tests/unit_tests/test_end_to_end.py::TestRubricMiddlewareEndToEnd`

Drives a real `create_deep_agent` with fake chat models for both main and grader agents. Exercises all four terminal statuses (`satisfied`, `needs_revision → satisfied`, `max_iterations_reached`, `grader exception → failed`), synthetic-message tagging, `KeyboardInterrupt` propagation, and custom system prompt honoring.

---

**Evals** — `libs/evals/tests/evals/test_iterative_constraint_satisfaction.py` · `EVAL_CATALOG.md` · `utils.py`

Tests whether a model with RubricMiddleware can hit an exact word count while starting every sentence with a custom phrase. Uses a deterministic grader to verify the middleware's verdict matches ground truth.

### Breaking changes:

None - this middleware is a no-op until a rubric is supplied on state, so it's safe to install unconditionally.

### How I verified my code:

- make format, make lint, make test from libs/deepagents/
- TestRubricMiddlewareEndToEnd drives all four terminal statuses under a real create_deep_agent stack.
- Iterative-constraint-satisfaction eval runs against claude-sonnet-4-6

LangSmith Trace: https://smith.langchain.com/o/ebbaf2eb-769b-4505-aca2-d11de10372a4/projects/p/5d0c5054-c7e0-4223-a3a0-870b1ddc4613/r/20260528T232328309718Z019e70e6-66b5-7b20-ab71-cc40fdd1c315


libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,749 @@
...
+    "satisfied",
+    "needs_revision",
+    "max_iterations_reached",
+    "failed",
+    "interrupted",
+]
+"""Terminal status reported on `outcome_status` and per evaluation."""
+
+
+_DEFAULT_EVALUATOR_MODEL = "anthropic:claude-haiku-4-5"
```
ccurme: let's force users to specify this. otherwise this model will get stale and changing it will break workflows.

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,749 @@
...
+    """Caller-supplied rubric describing what `done` looks like."""
+
+    outcome_status: NotRequired[OutcomeResult]
+    """The most recent terminal status. Updated each iteration."""
+
+    outcome_iterations: NotRequired[int]
+    """Number of grader evaluations performed for the current outcome."""
+
+    outcome_evaluations: NotRequired[list[OutcomeEvaluation]]
+    """Accumulated grader evaluations across all outcomes on this thread."""
```
ccurme: is it possible to only have `rubric` in the public API? IMO callers just need to specify the rubric and then read (improved) results from the `messages` key. I don't think they need to read or write status, iterations, or the evaluations.

If we ever decide to make these public, that's easy to do. Once we make them public, it's difficult to remove them, so there's a high bar.

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,749 @@
...
+        self,
+        runtime: Runtime[Any],
+        state: OutcomeState,
+        outcome_id: str,
+        iteration: int,
+        exc: BaseException,
+    ) -> dict[str, Any]:
+        # Cancellation is special-cased to `interrupted` so callers can
+        # distinguish "the grader blew up" from "the user pulled the plug".
+        is_cancel = isinstance(exc, (KeyboardInterrupt, _CANCEL_TYPES))
```
ccurme: if we KeyboardInterrupt, don't we want execution to stop?

libs/deepagents/deepagents/middleware/rubric.py
```diff
@@ -0,0 +1,749 @@
...
+        return "user"
+    if isinstance(msg, AIMessage):
+        return "assistant"
+    if isinstance(msg, ToolMessage):
+        name = msg.name or "tool"
+        return f"tool:{name}"
+    return getattr(msg, "type", "message")
+
+
+def _coerce_text(msg: AnyMessage) -> str:
```
ccurme, line 789: for a string (just text parts), use `message.text`.

I think for this function you can do `"\n".join(message.content_blocks)`

read more on content blocks [here](https://docs.langchain.com/oss/python/langchain/messages#standard-content-blocks).

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,749 @@
...
+    if isinstance(msg, AIMessage) and msg.tool_calls:
+        for call in msg.tool_calls:
+            name = call.get("name", "tool")
+            args = call.get("args", {})
+            parts.append(f"<tool_call name={name!r} args={args!r}/>")
+    return "\n".join(parts) if parts else "(empty)"
+
+
+def _aggregate_usage(messages: list[AnyMessage]) -> dict[str, int] | None:
+    """Sum usage_metadata across grader AIMessages.
```
ccurme: why do we need this?

libs/deepagents/tests/unit_tests/middleware/test_outcome_middleware.py
```diff
@@ -0,0 +1,715 @@
+"""Unit tests for `OutcomeMiddleware`.
```
ccurme: let's favor end-to-end tests. see tests/unit_tests/test_end_to_end.py for examples.

the test(s) should use the middleware with `create_deep_agent` and a fake chat model.

we don't need to go overboard on these, but end-to-end tests are required to check that things hang together. we can also rearrange internals and still get coverage on logic, whereas the unit tests here may need to change if we refactor things. OK to have a smaller number of end-to-end tests + more edge cases in unit tests.

examples/outcome_middleware/outcome_demo.py
```diff
@@ -0,0 +1,252 @@
+"""Offline demo of `OutcomeMiddleware`.
```
ccurme: I'm not sure how useful these demos are. They don't get run in CI, so they get stale / broken / untrusted as we make changes to the middleware. I'd prioritize explanations + a demo with concise code in a new page in the [docs](https://github.com/langchain-ai/docs/tree/main/src/oss/deepagents).

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,739 @@
...
+        ValueError: If `max_iterations` is outside `[1, 20]`.
+    """
+
+    state_schema = OutcomeState
+
+    def __init__(  # noqa: D107
+        self,
+        *,
+        max_iterations: int = 3,
+        evaluator_model: str | BaseChatModel | None = None,
```
ccurme: (nit) why not just `model`? is it implicit that the model is doing the evaluating?

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,739 @@
...
+        description=_GRADER_RESULT_DESCRIPTION,
+    )
+    explanation: str = Field(description=_GRADER_EXPLANATION_DESCRIPTION)
+    criteria: list[CriterionEval] = Field(
+        default_factory=list,
+        description=_GRADER_CRITERIA_DESCRIPTION,
+    )
+
+
+class OutcomeMiddleware(AgentMiddleware[OutcomeState, Any, Any]):
```
ccurme: Can we mark this beta, similar to what's done here? https://github.com/langchain-ai/deepagents/blob/f037b140f90a1ba3725b3ef23ab385b3cafe223b/libs/partners/quickjs/langchain_quickjs/middleware.py#L93

This will emit a warning alerting users that the API may change.

libs/deepagents/deepagents/middleware/outcomes.py
davibinboi: usage is mentioned in the Outcomes API Architecture document under sections 2.5, 3.1, and 5. Can remove if not needed

ccurme:
Can we wire up the middleware to the eval we created in libs/evals?

libs/deepagents/deepagents/__init__.py
```diff
@@ -23,16 +32,23 @@
 )
 
 __all__ = [
+    "OUTCOME_GRADER_MESSAGE_SOURCE",
     "AsyncSubAgent",
     "AsyncSubAgentMiddleware",
     "AsyncSubagentRunStream",
     "CompiledSubAgent",
+    "CriterionEval",
```
ccurme: not sure we need to export all these

libs/deepagents/deepagents/middleware/__init__.py
```diff
@@ -59,12 +68,19 @@
 )
 
 __all__ = [
+    "OUTCOME_GRADER_MESSAGE_SOURCE",
```
sydney-runkle: shouldn't be exported?

libs/deepagents/deepagents/__init__.py
```diff
@@ -23,16 +32,23 @@
 )
 
 __all__ = [
+    "OUTCOME_GRADER_MESSAGE_SOURCE",
```
sydney-runkle: same, don't need to export

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,691 @@
...
+        gap: When `passed` is False, a short, actionable description of
+            what's missing or incorrect. Omitted when `passed` is True.
+    """
+
+    name: str
+    passed: bool
+    gap: str
+
+
+class OutcomeEvaluation(TypedDict, total=False):
```
sydney-runkle: let's not use total=False if we don't need it

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,691 @@
...
+        *,
+        model: str | BaseChatModel,
+        max_iterations: int = 3,
+        grader_tools: Sequence[BaseTool] = (),
+        on_evaluation: Callable[[OutcomeEvaluation], None] | None = None,
+    ) -> None:
+        # Fail-fast on configuration. Matches the FilesystemMiddleware pattern.
+        if not model:
+            msg = "OutcomeMiddleware: `model` is required. Pass a model string like 'provider:model-id' or a BaseChatModel instance."
+            raise ValueError(msg)
```
sydney-runkle, line 358: for other middlewares we just inherit from the main agent (check what summarization does)

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,691 @@
...
+            graded = await self._agrade(state, iteration)
+        except Exception as exc:  # noqa: BLE001
+            return self._handle_grader_exception(runtime, state, outcome_id, iteration, exc)
+
+        return self._finalize_evaluation(graded, state, runtime, outcome_id, iteration)
+
+    def _prepare_evaluation(
+        self,
+        state: OutcomeState,
+        runtime: Runtime[Any],
```
sydney-runkle: should type properly as ContextT i think, see other middlewares in langchain

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,691 @@
...
+        rubric = state.get("rubric")
+        if not rubric:
+            # No rubric ever supplied -> middleware is a no-op for this run.
+            return None
+        if state.get("_active_rubric") == rubric:
+            # Sticky rubric / follow-up turn on the same outcome.
+            return None
+        return {
+            "_outcome_iterations": 0,
+            "_outcome_status": None,
```
sydney-runkle: this violates type of this state field right

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,691 @@
...
+        ValueError: If `model` is missing/empty or `max_iterations` is
+            outside `[1, 20]`.
+    """
+
+    state_schema = OutcomeState
+
+    def __init__(  # noqa: D107
+        self,
+        *,
+        model: str | BaseChatModel,
```
sydney-runkle, line 350: grader_model right?

sydney-runkle (changes requested):
should you be able to configure the grader subagent further? prompt? other subagent configs? you could even support just a subagent spec right?

looking awesome, very cool feature

libs/deepagents/deepagents/middleware/outcomes.py
ccurme: unfortunately that's what it was originally (`evaluator_model`) and I recommended changing it to `model` :) https://github.com/langchain-ai/deepagents/pull/3529#discussion_r3283811558

libs/evals/tests/evals/test_iterative_constraint_satisfaction.py
```diff
@@ -27,102 +33,151 @@
     run_agent,
 )
 
-pytestmark = [pytest.mark.eval_category("conversation")]
-
-
-# All vowel characters (upper and lowercase) used to check sentence starts.
-_VOWELS = frozenset("aeiouAEIOU")
```
sydney-runkle, line 34: why the refactor here?

libs/deepagents/deepagents/graph.py
```diff
@@ -708,6 +709,16 @@ def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly
...
+    # Inject the deep agent's backend into any user-supplied RubricMiddleware
+    # instances so the grader can default to it when `skills` or
+    # `permissions` are set on the grader spec but no explicit `backend`
+    # was passed to RubricMiddleware. This is parallel to how
+    # RubricMiddleware itself captures the main agent's model via
+    # `wrap_model_call`; the model has a per-call event we can hook, but
+    # the backend does not, so the wiring happens here.
+    for mw in deepagent_middleware:
+        if isinstance(mw, RubricMiddleware):
+            mw._fallback_backend = backend
```
sydney-runkle: do we do this for any other middleware? I think it'd be fine to require that the user passes it through

sydney-runkle (commented):
why does the grader agent need full subagent capabilities? can we start more simple than that?

it's not attached by the task tool right, it's a diff kind of "subagent", one that fires deterministically? so doesn't necessarily need the same semantics

mdrxy:
We need to update pr title/description with new name

libs/evals/tests/evals/test_iterative_constraint_satisfaction.py
davibinboi: I updated the eval to check whether every sentence starts with the letter “z” instead of a vowel. The refactor also improves the sentence-splitting logic to make it more robust.

sydney-runkle (dismissed):
fantastic!

libs/deepagents/deepagents/middleware/rubric.py
sydney-runkle, line 1: can we name this _rubric.py

sydney-runkle: or rubric.py

libs/deepagents/deepagents/middleware/rubric.py
```diff
@@ -0,0 +1,711 @@
...
+    """
+
+    state_schema = RubricState
+
+    def __init__(  # noqa: D107
+        self,
+        *,
+        model: str | BaseChatModel,
+        system_prompt: str | None = None,
+        tools: Sequence[BaseTool] | None = None,
```
sydney-runkle, line 352: is there a case for more tools?

libs/deepagents/deepagents/middleware/rubric.py
mdrxy: +1

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,711 @@
+"""Rubric middleware for self-evaluated agent iteration.
+
+`RubricMiddleware` lets a caller declare *what done looks like* via a
+rubric. After each natural agent stop the middleware invokes a
+separate grader sub-agent; if the grader returns `needs_revision`, the
+deepagent is looped back to the model with the grader's feedback injected
```
mdrxy: this wording is confusing

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,711 @@
...
+as a `HumanMessage`. The loop terminates on `satisfied`, on
+`max_iterations_reached`, or on grader `failed`. `KeyboardInterrupt` and
```
mdrxy:
```suggestion
as a `HumanMessage`. The loop terminates either on `satisfied`, on
`max_iterations_reached`, or on grader `failed`. `KeyboardInterrupt` and
```

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,711 @@
...
+`asyncio.CancelledError` propagate naturally so callers retain normal
+Python interrupt / cancellation semantics.
```
mdrxy: can remove, don't think this needs to be mentioned

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,711 @@
...
+
+The rubric is supplied per invocation via the `rubric` state field; it is
+sticky across invocations on a checkpointed thread. If no rubric is ever
+supplied the middleware is a no-op, so it is safe to include
+unconditionally.
```
mdrxy: would remove also

libs/deepagents/deepagents/middleware/rubric.py
```diff
@@ -0,0 +1,711 @@
...
+RubricResult = Literal[
+    "satisfied",
+    "needs_revision",
+    "max_iterations_reached",
+    "failed",
+]
+"""Terminal status reported per evaluation."""
+
+
+_MAX_TRANSCRIPT_MESSAGES = 30
```
mdrxy, line 89: can we say *why* the upper bound is needed?

libs/deepagents/deepagents/middleware/rubric.py
```diff
@@ -0,0 +1,711 @@
...
+always kept (see `_build_grader_transcript`).
+"""
+
+_MAX_TRANSCRIPT_CHARS_PER_MESSAGE = 4_000
+"""Per-message character budget for transcript snippets."""
+
+_MAX_ITERATIONS_HARD_CAP = 20
+"""Hard upper bound for `max_iterations`."""
+
+_PAYLOAD_CLOSER_RE = re.compile(r"</(rubric|transcript)", re.IGNORECASE)
```
mdrxy, line 113: do we need the closing tag?
```suggestion
_PAYLOAD_CLOSER_RE = re.compile(r"</(rubric|transcript)>", re.IGNORECASE)
```

mdrxy: nvm i see it will end with `-XXXX`

libs/deepagents/deepagents/middleware/rubric.py
```diff
@@ -0,0 +1,711 @@
...
+on each failing criterion with a short, actionable explanation of what's \
+missing or wrong.
+- `failed`: the rubric is malformed, contradictory, or otherwise impossible \
+to evaluate against the transcript.
+
+Be conservative: every criterion you cannot positively confirm should be \
+marked failed with a `gap` describing what evidence would be needed."""
+
+
+class CriterionEval(TypedDict):
```
mdrxy: like `RubricState`, the attribute docstrings should be on the attributes themselves

libs/deepagents/deepagents/middleware/rubric.py
```diff
@@ -0,0 +1,711 @@
...
+        gap: When `passed` is False, a short, actionable description of
+            what's missing or incorrect. Omitted when `passed` is True.
+    """
+
+    name: str
+    passed: bool
+    gap: NotRequired[str]
+
+
+class RubricEvaluation(TypedDict):
```
mdrxy, line 189: https://github.com/langchain-ai/deepagents/pull/3529#discussion_r3313398794

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,711 @@
...
+    "criterion passes; 'needs_revision' when at least one criterion fails; "
+    "'failed' when the rubric cannot be evaluated."
+)
+
+_GRADER_EXPLANATION_DESCRIPTION = (
+    "One or two sentence verdict summary that will be sent back to the agent as feedback if the task needs to be reattempted."
+)
+
+_GRADER_CRITERIA_DESCRIPTION = "Per-criterion verdicts. Each criterion should appear once with `passed` True/False and a `gap` string when failing."
+
```
mdrxy: why are these constants? let's just set on `GraderResponse`

libs/deepagents/deepagents/middleware/outcomes.py
```diff
@@ -0,0 +1,711 @@
...
+    The middleware activates only when a caller passes a `rubric` on
+    invocation state. With no rubric, both `before_agent` and `after_agent`
+    return without modifying state, so the middleware is safe to include
+    unconditionally in a `create_deep_agent` stack.
+
+    Args:
+        model: Model used by the grader sub-agent. Accepts either a model
+            string like `"provider:model-id"` or a `BaseChatModel`
+            instance. Required.
```
mdrxy: we don't need to say this when it's inferred from the signature
```suggestion
            instance.
```

mdrxy: same in other places e,g, `max_iterations`

mdrxy:
`What's included:` of PR description is not fun to read, can we make it human readable

mdrxy:
On sticky rubrics: `_reset_for_new_rubric` only resets `_rubric_iterations` when the rubric string changes. That means on a checkpointed thread where the caller re-supplies the same rubric across turns, the iteration counter is **cumulative across turns**, not per-turn.

Concretely:

| Turn | Start iter | Grader outcome | End iter | Status |
|---|---|---|---|---|
| 1 | 0 | `needs_revision` → revise → `satisfied` | 2 | `satisfied` |
| 2 (same rubric, new user message) | 2 | `needs_revision` | 3 → `next_iteration >= max_iterations` | `max_iterations_reached` (no revision loop) |

So after a couple of successful turns, follow-ups effectively can't iterate at all — even though the user-facing task is new.

Is this intentional? My read of the docstring ("same rubric attempt") suggests yes — one attempt spanning the whole thread — but the most natural usage I can imagine is "every response in this conversation must satisfy this rubric," which would want a fresh budget per task.

If unintentional, smallest fix is to also reset in `_reset_for_new_rubric` when `_rubric_status` is terminal (`satisfied | max_iterations_reached | failed`). If intentional, would be worth documenting the cumulative-budget semantics on the `max_iterations` argument so callers don't get surprised.

mdrxy (commented):
^

mdrxy:
Two small style nits on `_GRADER_SYSTEM_PROMPT` to bring it in line with the other middleware in this package:

**1. Drop the leading underscore.** Every other middleware exports its default prompt as a public constant — `SUMMARIZATION_SYSTEM_PROMPT`, `MEMORY_SYSTEM_PROMPT`, `SKILLS_SYSTEM_PROMPT`, `TASK_SYSTEM_PROMPT`, `ASYNC_TASK_SYSTEM_PROMPT`, `EXECUTION_SYSTEM_PROMPT`, `FILESYSTEM_SYSTEM_PROMPT`, `DEFAULT_SUBAGENT_PROMPT`. Making the rubric grader's prompt private breaks the pattern callers rely on for inspecting, wrapping, or composing on top of the default (e.g., `system_prompt = my_prefix + SUMMARIZATION_SYSTEM_PROMPT`). Suggest `GRADER_SYSTEM_PROMPT` and re-exporting from `middleware/__init__.py` to match.

**2. Drop the `\` line continuations inside the string.** The other prompts are written as natural paragraphs with no backslash continuations, and silence ruff E501 either with a file-level `# ruff: noqa: E501` (see the top of `memory.py`) or by leaving the lines long. The backslash style is harder to read and one stray trailing space away from a syntax surprise — worth matching the existing convention.

Same applies to `_GRADER_RESULT_DESCRIPTION` / `_GRADER_EXPLANATION_DESCRIPTION` / `_GRADER_CRITERIA_DESCRIPTION` if you want them inspectable; otherwise they could just be inlined into the `Field(description=...)` calls since each is used exactly once.

sydney-runkle (dismissed):

mdrxy:
Made a handful of fixes--

**Telling apart "the rubric is bad" from "the grader crashed."** Before, both ended up labeled `failed`. So if your OpenAI key was missing or the provider timed out, it looked identical to the grader saying "your rubric is malformed." Now infrastructure failures get their own status, `grader_error`, while `failed` stays reserved for "the grader looked at the rubric and couldn't evaluate it." Much easier to debug, and re-running after a transient blip now correctly starts a fresh attempt instead of inheriting the broken one's iteration count.

**Hitting `max_iterations` is no longer silent.** The status was already there in private state, but if you weren't reading it you'd have no idea the loop gave up — the last message in the conversation would just be whatever the agent said right before the grader rejected it. Now there's a warning log when the cap fires, and the class docstring explicitly tells you where to look for non-satisfied terminations.

**The grader can't contradict itself anymore.** LLMs occasionally emit responses like `result="satisfied"` with a failing criterion attached, or `result="needs_revision"` while marking every criterion as passing. That used to slip through and silently flip downstream control flow. Now `GraderResponse` rejects those at validation time. Same idea on individual criteria: a failing criterion has to come with a `gap` explaining what's missing — no more bare "this didn't pass" with zero detail.

**The grader stops grading its own feedback.** Subtle one: the middleware injects `HumanMessage`s tagged as grader feedback to drive the next loop. In a long conversation — especially with summarization compacting things — the grader could end up picking one of *its own* injected messages as "the user's original request" and grade the agent against that. The transcript builder now skips messages it injected itself when looking for the real prompt.

**Broken import in the eval.** `test_iterative_constraint_satisfaction.py` was still importing from `deepagents.middleware.outcomes`, which got renamed to `rubric` mid-branch. The eval would have failed to collect. Fixed to import from the public package.

Tests added for all of the above, existing exception test renamed to reflect the new `grader_error` status. `make lint` and `make test` clean — 1671 passed.

mdrxy (approved):

mdrxy:
LGTM! hyped

libs/deepagents/deepagents/middleware/__init__.py
```diff
@@ -50,6 +50,19 @@
...
+    RUBRIC_GRADER_MESSAGE_SOURCE,
+    CriterionEval,
+    CriterionFail,
+    CriterionPass,
+    GraderResponse,
+    GraderVerdict,
+    RubricEvaluation,
+    RubricMiddleware,
+    RubricResult,
+    RubricState,
```
ohmcpatel, line 64: please clean up this imports

PR #3542, opened 2026-05-21
fix(sdk): keep private state out of subagent propagation

hntrl:
This change makes Deep Agents derive private state keys directly from middleware state schemas and filters them out when state is copied into subagents.

It fixes two leak paths:
- parent state fields marked with `PrivateStateAttr` no longer propagate into child agents
- private fields written by one sibling subagent no longer leak into another sibling through shared runtime state

The test coverage now exercises the declarative `SubAgent` path with fake chat models and verifies both parent-to-child and sibling-to-sibling isolation.


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -241,9 +241,6 @@ def node(state):
     "messages",
     "todos",
```
sydney-runkle, line 242: should these be private

sydney-runkle (approved):
fine for now, don't love this but workaround make sense

PR #3543, opened 2026-05-21
fix(quickjs): keep private state out of subagent propagation

hntrl:
This PR applies the matching private-state propagation fix to the QuickJS partner package.

It keeps the QuickJS interpreter state isolated by aligning the slot lookup and propagation behavior with the Deep Agents private-state rules.

The test coverage exercises a shared middleware instance across sibling agents and verifies the state does not leak between runs.

### DIFF
--- a/libs/partners/quickjs/langchain_quickjs/middleware.py
+++ b/libs/partners/quickjs/langchain_quickjs/middleware.py
@@ -5,7 +5,7 @@ import contextlib
 import logging
 import uuid
 from collections.abc import Awaitable, Callable
-from typing import TYPE_CHECKING, Annotated, Any, Literal, NotRequired
+from typing import TYPE_CHECKING, Annotated, Any, Literal, NotRequired, cast
 
 from deepagents.middleware._utils import append_to_system_message
 from langchain.agents.middleware.types import (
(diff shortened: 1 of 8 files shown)


sydney-runkle (approved):

libs/partners/quickjs/langchain_quickjs/middleware.py
```diff
@@ -412,52 +403,80 @@ def before_agent(
...
         payload = state.get("_quickjs_snapshot_payload")
         if not payload:
-            return None
-        if not self._snapshot_authenticated(payload, state):
-            return {"_quickjs_snapshot_payload": None, "_quickjs_snapshot_hmac": None}
-        thread_id = _resolve_thread_id(self._fallback_thread_id)
-        repl = self._registry.get(thread_id)
+            return update
+        if not self._snapshot_authenticated(
+            payload, cast("REPLState", {**state, "_quickjs_slot_id": slot_id})
```
hntrl: no casts pls

hntrl:
!keep-open

libs/partners/quickjs/langchain_quickjs/middleware.py
```diff
@@ -412,52 +415,88 @@ def before_agent(
...
     async def abefore_agent(
         self,
         state: REPLState,
         runtime: "Runtime[ContextT]",  # noqa: ARG002
     ) -> dict[str, Any] | None:
         """Async variant of `before_agent` snapshot restore."""
-        if self._mode != "thread":
-            return None
+        slot_id = state.get("_quickjs_slot_id")
+        is_fork = state.get("_deepagents_forked_context") is True
```
hntrl: this is a named key exported from subagents I think, we shouldn't just need the loose string?

hntrl (commented):
would it be possible to deal with forking in a separate PR?

PR #3574, opened 2026-05-25
fix(sdk): surface missing path errors in `FilesystemBackend.ls`

### LINKED ISSUE #3573: Surface missing path errors in `FilesystemBackend.ls`
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
import tempfile
from pathlib import Path

def main() -> None:
    with tempfile.TemporaryDirectory() as tmp:
        root = Path(tmp)
        (root / "empty").mkdir()

        backend = FilesystemBackend(root_dir=str(root), virtual_mode=True)

        missing = backend.ls("/missing/")
        empty = backend.ls("/empty/")

        print("ls('/missing/'):")
        print(f"  error   = {missing.error!r}")
        print(f"  entries = {missing.entries!r}")
        print()
        print("ls('/empty/'):")
        print(f"  error   = {empty.error!r}")
        print(f"  entries = {empty.entries!r}")
        print()

        if missing.error is None and missing.entries == []:
            print("BUG: missing path looks like an empty directory.")
        else:
            print("OK: missing path is reported as an error.")


if __name__ == "__main__":
    main()
```

### Description

FilesystemBackend.ls() returned LsResult(error=None, entries=[]) for missing paths and file paths, which made them indistinguishable from empty directories. Agents using the ls tool could incorrectly conclude a directory exists but is empty.

This change aligns local filesystem behavior with sandbox backends:

- Missing path → error="Path '<path>': path_not_found", entries=None
- File path → error="Path '<path>': not_a_directory", entries=None
- Empty directory → error=None, entries=[] (unchanged)

FilesystemMiddleware already checks ls_result.error before formatting results, so no middleware changes were needed. Async als() inherits the fix via BackendProtocol.als().

I already have open a fixed for it in this PR:  #3574

### Environment / System Info

_No response_

keenborder786:
Closes #3573

---

> [!WARNING]
> **Behavior change for direct backend callers.** Code calling `FilesystemBackend.ls()` / `als()` directly (not through `FilesystemMiddleware`) now receives `entries=None` instead of `entries=[]` on **every** failure path — missing path, not-a-directory, and resolver errors (e.g. symlink loops). This realigns `ls` with the long-documented `LsResult` contract (`entries` is `None` on failure), and the middleware was already null-safe (`ls_result.entries or []`), so the tool surface is unaffected. But a direct caller that assumed `entries` was always a list will now hit `None` on these paths. Treat this as a **contract correction**, not a purely additive change.

## Behavior

`FilesystemBackend.ls(path)` now tells callers *why* a listing is empty instead of collapsing every case into `entries=[]`.

| You call `ls` on… | Returns |
|---|---|
| A missing path | `error="Path '<path>': path_not_found"`, `entries=None` |
| A file (not a directory) | `error="Path '<path>': not_a_directory"`, `entries=None` |
| A path that errors while resolving (e.g. symlink loop) | `error="Cannot list '<path>': <reason>"`, `entries=None` *(was `entries=[]`)* |
| An empty directory | `error=None`, `entries=[]` *(unchanged)* |
| A populated directory | `error=None`, `entries=[...]` *(unchanged)* |

So an agent can finally distinguish "that path isn't there" from "that directory exists but is empty."

## Why

Previously the success-like cases above returned `error=None, entries=[]`. A missing path looked identical to an empty directory, so an agent calling the `ls` tool would conclude a directory exists and is empty when really the path was wrong or gone — a silent failure that sends the agent down the wrong branch.

This also brings the local backend in line with the sandbox backends, which already surface `path_not_found` / `not_a_directory` errors. Agent reasoning shouldn't change based on which backend is wired in.

## Scope

- Only `FilesystemBackend.ls()` changed; `LsResult.error` already existed and `FilesystemMiddleware` already checks it before formatting output, so no middleware changes were needed.
- `ls()` now upholds a single failure contract: **every** error path returns `entries=None` (the `LsResult` docstring's documented "`None` on failure"). This includes the pre-existing `OSError`/`RuntimeError` resolver-failure path, which previously returned `entries=[]` alongside its error.
- Async `als()` inherits the fix via `BackendProtocol.als()`.
- Tests updated in the composite and filesystem backend suites (sync + async) to assert the new `error` / `entries=None` contract, including the symlink-loop resolver-failure path.


mdrxy (approved):
TY!

PR #3575, opened 2026-05-25
feat(code): add macOS keyboard shortcuts for line navigation

nitboss:
Add three new keyboard shortcuts for improved text editing efficiency:

- CMD+DEL: Delete all characters from cursor to line start
- CMD+Right: Move cursor to end of current line
- CMD+Left: Move cursor to start of current line

These shortcuts follow standard macOS text editing conventions and provide a familiar experience for users coming from other macOS applications.

Includes comprehensive unit tests covering:
- Basic functionality of each shortcut
- Partial line deletion scenarios
- Empty text handling
- Multi-line text behavior

### DIFF
--- a/libs/code/deepagents_code/widgets/chat_input.py
+++ b/libs/code/deepagents_code/widgets/chat_input.py
@@ -340,6 +340,24 @@ class ChatTextArea(TextArea):
             "Delete left to start of word",
             show=False,
         ),
+        Binding(
+            "cmd+delete",
+            "delete_to_line_start",
+            "Delete to line start",
+            show=False,
+        ),
+        Binding(
+            "cmd+right",
+            "cursor_line_end",
+            "Go to line end",
+            show=False,
+        ),
+        Binding(
+            "cmd+left",
+            "cursor_line_start",
+            "Go to line start",
+            show=False,
+        ),
     ]
     """Key bindings for the chat text area.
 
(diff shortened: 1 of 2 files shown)


mdrxy:
Leaving context here for anyone who finds this PR in the future.

This was reverted in #3589. The short version: terminal emulators intercept `Cmd`-modified keys before they reach a TUI, so Textual bindings on `cmd+left` / `cmd+right` / `cmd+delete` never fire. The action methods were correct, but the keybindings could not be delivered by any terminal.

**What actually works**

Deep Agents Code already supports the equivalent motions via Textual's `TextArea` defaults, which are standard readline bindings:

- `Ctrl+A` → start of line
- `Ctrl+E` → end of line
- `Ctrl+U` → delete to start of line
- `Ctrl+K` → delete to end of line
- `Ctrl+W` → delete word left
- `Ctrl+Left` / `Ctrl+Right` → word-wise motion

**How macOS `Cmd+...` shortcuts get to the app**

The translation happens in the terminal, not in the app:

- **Ghostty** ships this out of the box: `Cmd+Left` → `\x01` (`Ctrl+A`), `Cmd+Right` → `\x05` (`Ctrl+E`), `Cmd+Backspace` → `\x15` (`Ctrl+U`).
- **iTerm2** users can add the same mapping under *Settings → Profiles → Keys → Key Mappings* using \`Send Text with vim special chars\`.
- **Terminal.app** has no native UI for this; use the `Ctrl` shortcuts directly.

**Why the tests passed**

The tests invoked \`text_area.action_delete_to_line_start()\` etc. directly rather than \`await pilot.press(\"cmd+left\")\`, so they verified the action handlers worked but not that the bindings could ever be triggered.

Docs updated in https://github.com/langchain-ai/docs to make the macOS conventions discoverable.

PR #3584, opened 2026-05-26
fix(code): assign stable UUID to `HumanMessage` input before graph entry

sydney-runkle:
## Summary

- Reverts the per-call-site UUID workaround (assigning `\"id\"` in the `user_msg` dict) — no longer needed
- The root fix is in LangGraph: **https://github.com/langchain-ai/langgraph/pull/7913**
- Also removes `import uuid` that was added only for the workaround

## Root cause (from investigation)

`DeltaChannel` stores raw writes and replays them through the reducer. When a reducer assigns message IDs in-place inside `apply_writes()`, a race exists: the background serialisation thread may capture `id=None` before `apply_writes()` mutates the object. Every `get_state()` call (and every resumed invocation on the same thread) then replays `id=None` through the reducer and gets a fresh UUID — the same HumanMessage appears with a different ID in each LangSmith trace.

## Fix (in LangGraph)

`ensure_message_ids()` is called in `put_writes()` **before** the background thread is submitted, assigning stable UUIDs to any `id=None` BaseMessages synchronously. The serialised bytes always reflect the stable ID.

## End-to-end verification

```
# Without fix:
Turn 1 — 'write a hello world script' id: f7acc553-...
Turn 2 — 'write a hello world script' id: 56d42628-...  ← different ID, bug

# With LangGraph fix:
Turn 1 — 'write a hello world script' id: 3843ae2c-...
Turn 2 — 'write a hello world script' id: 3843ae2c-...  ← same ID, fixed
```

## Test plan

- [ ] LangGraph PR #7913 merged and released
- [ ] `test_convert_to_messages_preserves_id_field` — pipeline preserves explicit IDs
- [ ] `test_convert_to_messages_without_id_gives_none` — baseline: no id without fix


sydney-runkle:
Closing in favour of a clean PR with only the relevant changes.

PR #3591, opened 2026-05-26
fix(sdk): stable `HumanMessage` IDs across resumed threads

sydney-runkle:
**Depends on:** langchain-ai/langgraph#7913 (merged as langgraph 1.2.2)

## Summary

- **Bumps `langchain` minimum to 1.3.2** — which pulls in `langgraph>=1.2.2` transitively; explicit `langgraph` pin removed
- **Removes `Overwrite` workaround** in summarization middleware (`bee514fd2`): now that DeltaChannel writes are stamped with stable IDs before serialisation, same-ID message updates work correctly via the reducer's in-place update path
- **Removes ID assignment from `_messages_delta_reducer`** — IDs are now assigned by `ensure_message_ids()` in LangGraph's `put_writes()` before writes reach the checkpointer, making reducer-level assignment redundant
- **Updates snapshots** — langchain 1.3.2 updated the `write_todos` tool description and system prompt wording
- **Updates all lockfiles** across libs and examples

## Root cause (from investigation)

`DeltaChannel` stores raw writes and replays them through the reducer. In the default async/sync durability modes, `put_writes()` submits checkpoint writes to a background thread before `apply_writes()` runs. If the thread serialised a message before the reducer assigned its ID, the checkpoint stored `id=None` and every `get_state()` call produced a different UUID — the same message appeared with a different ID in each LangSmith trace.

## Fix (in LangGraph 1.2.2)

`ensure_message_ids()` in `put_writes()` coerces message-like write values to typed `BaseMessage` objects with stable UUIDs before the background thread serialises them. Handles `BaseMessage` objects (single and in lists) and dicts with known `role`/`type` keys (OpenAI-style API input).

## Tests

- `test_get_state_messages_have_ids` — all messages have non-None ids after `get_state()`
- `test_dict_style_invoke_messages_have_stable_ids` — dict-style `{"role": "user"}` input also gets stable IDs (covers the API / over-the-wire path)
- `test_human_message_id_stable_across_invocations_sync` — original HumanMessage keeps its ID across resumed threads (sync)
- `test_human_message_id_stable_across_invocations_async` — same, async path


libs/deepagents/pyproject.toml
```diff
@@ -61,6 +62,9 @@ test = [
 ]
 
 
+[tool.uv.sources]
+langgraph = { git = "https://github.com/langchain-ai/langgraph", subdirectory = "libs/langgraph", branch = "fix/delta-channel-write-ordering" }
```
eyurtsev: accidental commit?

sydney-runkle: no i wanted to test

eyurtsev:
What are semantics of overwrite is there a reason not to use overwrite in this case?

eyurtsev (dismissed):

eyurtsev (approved):

PR #3193, opened 2026-05-06
feat(sdk)!: v0.6

sydney-runkle:
BEGIN_COMMIT_OVERRIDE
feat(sdk): v0.6
END_COMMIT_OVERRIDE

Issue #2362, opened 2026-03-31
`SubAgentMiddleware` regression — runtime config not passed to subagent invocations (PR #602 fix lost in #1041 refactor)


kushmadlani:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

#569, #139, #602, #1041

### Reproduction Steps / Example Code (Python)

```python
from dataclasses import dataclass
from langchain.agents import create_agent
from langchain.tools import tool, ToolRuntime
from deepagents import create_deep_agent

@dataclass
class MyContext:
    api_url: str = "https://example.com"

@tool
def check_context(query: str, runtime: ToolRuntime[MyContext]) -> str:
    """Tool that needs runtime.context."""
    return f"Context: {runtime.context.api_url}"

agent = create_deep_agent(
    tools=[check_context],
    context_schema=MyContext,
)

# This works:
agent.invoke({"messages": [...]}, context=MyContext())

# But when the agent delegates to a subagent via the task tool,
# the subagent's tools receive runtime.context = None
```

### Error Message and Stack Trace (if applicable)

```shell

```

### Description

* I'm trying to use the `deepagents` library with `context_schema` so subagent tools can access `runtime.context`.
* I expect `runtime.context` to be available in tools called by subagents (as it was after PR #602 fixed this in Dec 2025).
* Instead, `runtime.context` is `None` in subagent tools. PR #1041 (Feb 4, 2026) rewrote `subagents.py` and dropped the `runtime.config` pass-through from `_build_task_tool()`. The fix is to pass `runtime.config` to `subagent.invoke()` and `subagent.ainvoke()` — the exact one-line change from PR #602.

sydney-runkle:
Curious, what context do you want to pass through to the subagent?

mdrxy:
Possibly related? #2316

madhavc21:
Possibly resolved here? #2845

mdrxy:
Consolidating #2315 here as a duplicate — same root cause, broader symptom set worth capturing in the fix.

In addition to `runtime.context = None` in subagent tools, the missing `config` forward also drops:

[...]

Issue #2315, opened 2026-03-29
task tool does not forward config (callbacks, metadata, tags) to subagent invocations, breaking observability and Langfuse tracing


snuffles5:
### Checked other resources

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

#1698 — Similar issue for recursion_limit (partially fixed in PR #2194, but only for recursion_limit default value, not config forwarding)

### Reproduction Steps / Example Code (Python)

```python
"""Minimal reproduction of config not being forwarded to subagents."""

from langchain_core.callbacks import BaseCallbackHandler
from langchain_core.messages import HumanMessage
from langchain.agents import create_agent
from deepagents import create_deep_agent
from deepagents.middleware.subagents import CompiledSubAgent
import asyncio


class TestCallback(BaseCallbackHandler):
    """Test callback that tracks which agents it receives events from."""
    
    def __init__(self, name: str):
        self.name = name
        self.llm_starts = []
    
    def on_llm_start(self, serialized, prompts, **kwargs):
        self.llm_starts.append(f"{self.name}: {kwargs.get('run_name', 'unknown')}")
        print(f"🔔 Callback '{self.name}' received on_llm_start")


async def test_config_propagation():
    # Create a simple subagent
    subagent = create_agent(
        "openai:gpt-4o-mini",  # or any model
        name="test-subagent",
    )
    
    # Wrap it as a CompiledSubAgent
    compiled_subagent = CompiledSubAgent(
        name="test-sub",
        description="A test subagent",
        runnable=subagent,
    )
    
    # Create parent agent with the subagent
    parent_agent = create_deep_agent(
        "openai:gpt-4o-mini",
        subagents=[compiled_subagent],
    )
    
    # Create a callback to trace
    callback = TestCallback("parent-callback")
    
    # Invoke parent with callback in config
    result = await parent_agent.ainvoke(
        {"messages": [HumanMessage(content="Use the task tool to call test-sub with any request")]},
        config={
            "callbacks": [callback],
            "run_name": "parent-agent",
            "metadata": {"test_key": "test_value"},
            "tags": ["test-tag"],
        },
    )
    
    print(f"\n📊 Results:")
[...]

PR #3635, opened 2026-05-27
fix(sdk): forward parent metadata into subagent invoke config

### LINKED ISSUE #3634: Forward parent `metadata` into subagent invoke config
Follow-up to #2362 (closed by #2845 + #3193). Scoping the one remaining gap.

## Current behavior

`_build_subagent_config` in `libs/deepagents/deepagents/middleware/subagents.py` forwards `callbacks`, `tags`, and `configurable` from the parent's `runtime.config` into the subagent's `invoke` call, but explicitly drops `metadata`. The docstring justifies this:

> Passing `metadata` in the invoke config replaces the subagent's bound metadata (e.g. `lc_agent_name`).

This is true under LangChain's default merge semantics (passed-in metadata wins on key collision), but it means user-set metadata on the parent invoke never reaches subagent runs.

## Desired behavior

Forward parent `metadata` into the subagent's invoke config, but strip the LangChain reserved keys (`lc_agent_name`, `lc_agent_type`) so the subagent's bound identity is preserved.

## Repro

```python
parent.ainvoke(
    {"messages": [...]},
    config={"metadata": {"customer_id": "abc-123", "lc_agent_name": "parent-agent"}},
)
```

Expected: subagent's run has `customer_id=abc-123` and `lc_agent_name=<subagent-name>` (bound value preserved).
Actual: subagent's run has neither — parent metadata is dropped entirely.

## Proposed fix

In `_build_subagent_config`:

```python
_RESERVED_METADATA_KEYS = frozenset({"lc_agent_name", "lc_agent_type"})

parent_metadata = parent_config.get("metadata") or {}
filtered = {k: v for k, v in parent_metadata.items() if k not in _RESERVED_METADATA_KEYS}
if filtered:
    config["metadata"] = filtered
```

mdrxy:
Closes #3634

---

Follow-up to #2362 (closed by #2845 + #3193). The task tool's `_build_subagent_config` previously forwarded `callbacks`, `tags`, and `configurable` from the parent's runtime config into the subagent's invoke, but dropped `metadata` entirely. The original justification was that forwarding metadata would let LangChain's \`merge_configs\` (passed values win on key collisions) overwrite the subagent's bound \`lc_agent_name\`.

This change forwards \`metadata\` with framework-owned keys stripped, so user-set parent metadata (e.g. \`customer_id\`) reaches subagent runs in LangSmith / Langfuse while the subagent's bound identity and Pregel step context survive intact.

The stripped set covers:

- LangChain / deepagents bound identity: \`lc_agent_name\`, \`ls_integration\`, \`versions\`. Forwarding these would overwrite the subagent's own bound values under LangChain's merge semantics.
- LangGraph per-step internals: any key prefixed \`langgraph_\`, plus \`thread_id\` and \`checkpoint_ns\`. Forwarding these seeds the subagent's invoke with the parent's step context and was observed to drop the subagent's bound \`lc_agent_name\` from tool-runtime metadata.

\`_build_subagent_config\` is hoisted from a closure inside \`_build_task_tool\` to module level so the strip predicate and its companion constants live alongside it.

The existing xfail test on callbacks (#2315) is unrelated and left as-is.

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -433,6 +433,48 @@ class _SubagentSpec(TypedDict):
     runnable: Runnable
 
 
+# Metadata keys (and prefixes) set by LangChain / LangGraph / deepagents
+# framework code on `runtime.config["metadata"]` that must NOT be forwarded
+# from parent to subagent — they are framework-owned, not user-set:
+# - `lc_agent_name`, `ls_integration`, `versions`: bound by `create_agent` /
+#   `create_deep_agent`; forwarding would overwrite the subagent's own bound
+#   identity under LangChain's `merge_configs` semantics.
+# - `langgraph_*`, `thread_id`, `checkpoint_ns`: added per-step by LangGraph;
+#   forwarding seeds the subagent's invoke with the parent's step context and
+#   interferes with the subagent's own Pregel machinery (observed to drop the
+#   subagent's bound `lc_agent_name` from tool-runtime metadata).
+_RESERVED_SUBAGENT_METADATA_KEYS = frozenset({"thread_id", "checkpoint_ns", "versions"})
+_RESERVED_SUBAGENT_METADATA_PREFIXES = ("lc_", "ls_", "langgraph_")
+
+
+def _is_reserved_subagent_metadata_key(key: str) -> bool:
+    return key in _RESERVED_SUBAGENT_METADATA_KEYS or key.startswith(_RESERVED_SUBAGENT_METADATA_PREFIXES)
+
+
+def _build_subagent_config(runtime: ToolRuntime) -> RunnableConfig:
+    """Derive the subagent's `RunnableConfig` from the parent's runtime config.
+
+    Forwards `callbacks`, `tags`, `configurable`, and `metadata` — with
+    framework-reserved metadata keys stripped (see
+    `_is_reserved_subagent_metadata_key`) so the subagent's bound identity
+    and its own Pregel step context aren't clobbered when LangChain merges
+    configs at invoke time.
+
+    `recursion_limit` is intentionally not forwarded — the subagent's own
+    bound limit must take precedence.
+    """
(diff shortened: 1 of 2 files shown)


jackjin1997:
Verified the strip-set approach locally against the repro — it correctly fixes the no-user-metadata case (the `langgraph_*`/`checkpoint_ns`/`ls_*` pollution that was dropping bound `lc_agent_name`). 

One gap I hit though: once the parent **does** pass user metadata, the bound `lc_agent_name` is still lost, which I think contradicts the issue's stated expected behavior (`customer_id=abc-123` **and** `lc_agent_name=<subagent-name>`). Captured from the subagent tool runtime with `config={"metadata": {"customer_id": "abc-123", "lc_agent_name": "PARENT"}}`:

```python
{'customer_id': 'abc-123', 'thread_id': '...', 'checkpoint_ns': '...',
 'langgraph_step': 2, 'langgraph_node': 'tools', 'langgraph_triggers': (...),
 'langgraph_path': (...), 'langgraph_checkpoint_ns': '...'}
# note: no lc_agent_name at all
```

`customer_id` arrives ✅ but the subagent's bound `lc_agent_name=<subagent-name>` is gone — forwarding a non-empty `metadata` still triggers LangChain's wholesale **replace** (not merge), so the subagent's bound identity is wiped whenever any user key survives the strip.

The current test passes because it asserts `metadata.get("lc_agent_name") != "PARENT-OVERRIDE"` — but `None != "PARENT-OVERRIDE"` is also true, so a fully-dropped `lc_agent_name` slips through. A stricter assert (`== <subagent-name>`) surfaces it:

```python
# fails on this branch when user metadata is present:
assert captured_config["metadata"]["lc_agent_name"] == "general-purpose"  # KeyError
```

So this is the merge-vs-replace question from the issue thread — to actually preserve bound identity, the forwarded metadata needs to merge *under* the subagent's bound metadata rather than replace it. Happy to help if useful; I have a local repro test ready that exercises the user-metadata-present path.

PR #2452, opened 2026-04-03
fix(sdk): `read_file` pagination skipping lines after wrapping

farrimoh:
Fixes #2453  the bug in read_file where long-line wrapping combined with post-format truncation could skip source lines across pagination. Removes redundant row-based truncation while preserving size-based safeguards.

PR #3641, opened 2026-05-27
fix(sdk): `read_file` pagination skipping lines after wrapping

### LINKED ISSUE #2453: read_file pagination skips lines after wrapping due to double limit application
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Related Issues / PRs

_No response_

```python
from deepagents.utils import prepopulate_file

# Create a file with one very long line
backend = StateBackend()
file_path = "/wrapped.txt"

long_line = "x" * 15000
content = "\n".join([
    "line1",
    long_line,
    "important instruction",
    "line4",
])

files = prepopulate_file(backend, file_path, content)

agent = create_deep_agent(backend=backend)

# Read first page
result1 = agent.invoke({
    "messages": [HumanMessage(content=f"read {file_path} with offset=0 limit=3")],
    "files": files
})

# Read second page
result2 = agent.invoke({
    "messages": [HumanMessage(content=f"read {file_path} with offset=3 limit=3")]
})

print(result1)
print(result2)
```

```shell

```

### Description

read_file can skip lines when paginating a file that contains very long lines.

If a long line is wrapped into multiple display rows, only part of the formatted output may be returned, even though the backend already fetched the next line. When the next page is requested, pagination continues from a later offset, so the skipped line is never shown.

Expected: all fetched lines should always appear across pages.
Actual: some lines are skipped when wrapping occurs.

### Environment / System Info

OS: Windows
Python: 3.12.4
deepagents: 0.5.0a3

mdrxy:
Closes #2453

---

`read_file` applies row-based truncation twice: once when the backend slices raw lines by `offset`/`limit`, then again after `format_content_with_line_numbers` wraps long lines into continuation rows. When a wrapped line inflates the row count past `limit`, real source lines fall off the page — and the next page resumes from `offset + limit`, skipping them entirely.

`_truncate` now takes the row-based limit as optional. The new-style `ReadResult` path (the backend has already paginated by source lines) drops the row chop and keeps only the byte-budget cap. The legacy `str` backend path keeps the row-based behavior, since those backends still format with line numbers inline.

`limit` now bounds *source* lines, not formatted rows — a wrapped 15k-char line with `limit=3` renders all of its continuation chunks alongside the surrounding source lines instead of silently dropping the tail. The `read_file` tool description is updated to match.

Carries forward the fix originally proposed in #2452 by @farrimoh; that PR was closed and could not be reopened, so the patch was rewritten against current `main` with farrimoh credited as a co-author on the commit.


ccurme:
The application of `limit` to the rendered lines was deliberate and explained in https://github.com/langchain-ai/deepagents/pull/809. What is the point of having the line continuation mechanism in deepagents if we don't respect those lines in `read_file`? Is there any remaining benefit to it?

PR #3642, opened 2026-05-28
feat(sdk): allow passing `state_schema` in `create_deep_agent`

### LINKED ISSUE #3249: Allow to pass `state_schema` to `create_deep_agent` function
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] repl
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

### Feature description

Simply allow to pass `state_schema` to `create_deep_agent` function. 

```python
def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly logic with many conditional branches
    ...
    response_format: ResponseFormat[ResponseT] | type[ResponseT] | dict[str, Any] | None = None,
    state_schema: type[AgentState[ResponseT]] | None = None,
    context_schema: type[ContextT] | None = None,
    ...):
```

then return
```python
    return create_agent(
        model,
        system_prompt=final_system_prompt,
        tools=_tools,
        middleware=deepagent_middleware,
        response_format=response_format,
        state_schema=state_schema,
        context_schema=context_schema,
        checkpointer=checkpointer,
        store=store,
        debug=debug,
        name=name,
        cache=cache,
    ).with_config(
        {
            "recursion_limit": 9_999,
            "metadata": {
                "ls_integration": "deepagents",
                "versions": {"deepagents": __version__},
                "lc_agent_name": name,
            },
        }
    )
```

I encountered cases where I want to pass some data (e.g., metadata for the page user seeing at the moment and file urls) to subagent to do some tasks. For now, I needed to use config to pass this data.

imnishitha:
Fixes #3249

Allow callers to pass a custom `state_schema` to `create_deep_agent`, enabling them to extend the graph state with extra fields (e.g. page metadata, file URLs) while preserving the `DeltaChannel` message reducer that keeps checkpoint size O(N). Defaults to `_DeepAgentState` so existing behavior is unchanged.

## How verified

- Added two unit tests: one verifying the default `_DeepAgentState` is used when no schema is passed, another verifying a custom subclass is forwarded through to `create_agent`.
- All 85 existing tests in `test_graph.py` pass.
- Ruff lint checks pass with no violations.

---

**AI Disclosure:** This PR was authored with Claude Opus 4.7 assistance for code generation and testing.


libs/deepagents/deepagents/graph.py
```diff
@@ -437,6 +438,29 @@ def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly
             For example, `interrupt_on={"edit_file": True}` pauses before
             every edit.
         response_format: A structured output response format to use for the agent.
+        state_schema: Custom state class to use for the agent graph.
+
+            Must be a subclass of
+            [`_DeepAgentState`][deepagents.graph._DeepAgentState], which
+            itself extends [`AgentState`][langchain.agents.AgentState] and
+            adds a `DeltaChannel` reducer that keeps checkpoint size O(N)
+            rather than O(N²). Attach extra fields here to carry per-run
```
mdrxy: unsure we need to go into implementation specifics here
```suggestion
            adds a `DeltaChannel` reducer that keeps checkpoint sizes small.
            Attach extra fields here to carry per-run
```

libs/deepagents/deepagents/graph.py
```diff
@@ -437,6 +438,29 @@ def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly
             For example, `interrupt_on={"edit_file": True}` pauses before
             every edit.
         response_format: A structured output response format to use for the agent.
+        state_schema: Custom state class to use for the agent graph.
```
mdrxy: Docstring drifts a bit from upstream `create_agent`'s `state_schema` description. Two things worth carrying over:

1. **Middleware recommendation.** Upstream notes: *"Generally, it's recommended to use `state_schema` extensions via middleware to keep relevant extensions scoped to corresponding hooks / tools."* Same guidance applies here — worth including so users reach for middleware first.
2. **Middleware-merge behavior.** Upstream calls out that the schema is used as the base for merging with middleware state schemas. Since `create_deep_agent` wires up middleware internally, this interaction is relevant for users subclassing `_DeepAgentState`.

Also minor: the type annotation is `type[_DeepAgentState]` but the docstring just says "Custom state class." Could clarify it must be a `TypedDict` subclass of `_DeepAgentState` to match upstream's phrasing.

libs/deepagents/tests/unit_tests/test_graph.py
```diff
@@ -20,6 +20,7 @@
     _REQUIRED_MIDDLEWARE_CLASSES,
```
mdrxy, line 20: A few thoughts on tests and a small follow-up

**No negative-path coverage.** A non-`_DeepAgentState` subclass is accepted silently today. Either runtime-validate via `issubclass` (and test the `TypeError`), or add a comment that the constraint is typing-only.

**Test placement.** `TestCreateDeepAgentMetadata` is scoped to metadata / `ls_integration` assertions. A dedicated `TestStateSchema` class (mirroring `TestExtraMiddlewareWiring`, `TestSystemPromptAssembly`) would fit better.

**Mock-heaviness.** Idelly we avoid mocks as much as possible. `TestToolDescriptionOverrideWiring` and `TestExtraMiddlewareWiring` in the same file use real middleware and assert against the actual agent — a better template. The current approach also couples the tests to internal import paths (`deepagents.graph.SubAgentMiddleware`, etc.), so any refactor of imports breaks the tests without behavior changing.

mdrxy (changes requested):
I think we should propagate `state_schema` into declarative subagents as well.

LangChain `create_agent` treats an explicit `state_schema` as the base graph schema and merges middleware schemas into it. Since `create_deep_agent(state_schema=...)` now exposes that same concept, callers will reasonably expect built-in `task` subagents to see the same custom state fields when the task tool forwards parent state.

Today `SubAgentMiddleware` compiles declarative subagents with `create_agent(...)` but does not pass `state_schema`, so custom fields are only available on the top-level graph. That makes the new option behave inconsistently for the default delegation path.

I'd thread the resolved schema into `SubAgentMiddleware` and pass it when compiling `SubAgent` specs. I would leave `CompiledSubAgent` unchanged, since callers own those runnable schemas.

mdrxy:
Could we also make the `CompiledSubAgent` boundary explicit in the docs/docstrings?

If `state_schema` is threaded into declarative `SubAgent` compilation, users may reasonably assume the same schema applies to all subagent forms. For `CompiledSubAgent`, though, the runnable is already compiled, so `create_deep_agent(state_schema=...)` can't safely retrofit its graph schema. The caller needs to compile that runnable with a compatible schema themselves.

Suggested wording for the `state_schema` arg docstring:

```txt
state_schema: Custom state schema for the agent graph. This must be a
    `TypedDict` subclass of `_DeepAgentState`.

    Generally, prefer defining state extensions with middleware so the
    extra fields stay scoped to the hooks and tools that use them.

    When provided, this schema is used as the base graph schema and is
    merged with state schemas from middleware. It is also used when
    compiling declarative `SubAgent` specs for the `task` tool.

    `CompiledSubAgent` runnables do not inherit this schema because they
    are already compiled; compile those runnables with a compatible state
    schema if they need access to the same custom state fields. Remote
    `AsyncSubAgent` specs likewise use the schema configured on the
    remote graph.
```

And a shorter note near `CompiledSubAgent` docs/docstring:

```txt
`CompiledSubAgent` runnables are used as provided. They do not inherit
`create_deep_agent(state_schema=...)`; if the runnable needs custom state
fields, compile it with a compatible state schema.
```

mdrxy (approved):
LGTM! yay first feature!

PR #3659, opened 2026-05-29
feat(sdk): delete operation to backend protocol

imnishitha:
Adds an optional `delete`/`adelete` method to `BackendProtocol` and implements it across all file backends, so agents and host code can remove files
(not just create/edit them). Backends that don't implement mentedError` default, and a `supports_delete()` helper letscallers detect support without triggering it.

### What changed

- **Protocol:** new `DeleteResult` dataclass, `delete`/`adelete` methods (default `raise NotImplementedError`, async defaults to `to_thread`), and a
`supports_delete(backend)` helper using the same override-c`grep_raw` methods.
- **Implementations:**
  - `StateBackend` — queues `{path: None}` through the exisucer
  - `FilesystemBackend` — `Path.unlink()` with directory / not-found guards
  - `StoreBackend` — native `store.delete`/`adelete`
  - `CompositeBackend` — routes by prefix, remaps the result path back
  - `BaseSandbox` — single server-side shell command (`rm`)ssification
  - `ContextHubBackend` — commits a `None`-entry payload to `push_agent`
  - `LangSmithSandbox` and `LocalShellBackend` inherit `del

No `delete_file` LLM tool is added in this PR — this is the

### Compatibility

`BackendProtocol` intentionally avoids `@abstractmethod`, se `delete` remain instantiable and report `supports_delete()== False`. No breaking changes.

### How I verified

- Added unit tests per backend plus protocol-level tests for the `NotImplementedError` default and `supports_delete`, and an end-to-end
`StateBackend` delete inside a graph context.
- `make format`, `make lint`, `make test` pass from `libs/deepagents` (full suite: 1705 passed, 95 skipped, 4 xfailed).

> AI disclaimer: this change was implemented with the assistance of generative AI and reviewed by the author.


libs/deepagents/deepagents/backends/context_hub.py
```diff
@@ -123,6 +124,29 @@ def _commit(self, files: dict[str, str]) -> None:
             for path, content in files.items():
                 self._cache[path] = content
 
+    def _commit_deletions(self, paths: list[str]) -> None:
```
nick-hollon-lc: pardon the ignorance but why do we need a separate commit_deletions method when we already have commit?

nick-hollon-lc (commented):
so exciting!! ooc did we decide about if this was gonna be opt-in or not?

libs/deepagents/deepagents/backends/context_hub.py
sydney-runkle: why was this resolved?

imnishitha: I was able to make the modifications in `_commit` method to not have a seperate `_commit_deletions` method.

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -848,5 +889,24 @@ def execute_accepts_timeout(cls: type[SandboxBackendProtocol]) -> bool:
         return "timeout" in sig.parameters
 
 
+def supports_delete(backend: BackendProtocol) -> bool:
```
sydney-runkle: this is similar to supports_execution

eventually we could probably clean up these conditional code paths

libs/deepagents/deepagents/backends/state.py
```diff
@@ -290,6 +291,20 @@ def edit(
...
+
+        The deletion is queued via `CONFIG_KEY_SEND` as a ``None`` value, which
+        the `files` channel reducer interprets as a deletion marker.
+        """
+        files = self._read_files()
+
+        if file_path not in files:
+            return DeleteResult(error=f"Error: File '{file_path}' not found")
+
+        self._send_files_update({file_path: None})
```
sydney-runkle, line 305: do you know why this works?

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -756,6 +757,39 @@ def _map_edit_error(error: str, file_path: str, old_string: str) -> EditResult:
...
+        """
+        quoted = shlex.quote(file_path)
+        # `-e` misses dangling symlinks, so also probe `-L`; guard directories
+        # since `rm -f` won't remove them and we only delete files.
+        cmd = (
+            f"if [ ! -e {quoted} ] && [ ! -L {quoted} ]; then echo __MISSING__; "
+            f"elif [ -d {quoted} ]; then echo __ISDIR__; "
+            f"else rm -f {quoted} && echo __DELETED__ || echo __FAILED__; fi"
+        )
+        result = self.execute(cmd)
```
sydney-runkle: is this a clean way to do this?

libs/deepagents/deepagents/backends/context_hub.py
```diff
@@ -123,6 +124,29 @@ def _commit(self, files: dict[str, str]) -> None:
...
+        """Push a commit removing `paths`; update the cache on success.
+
+        A `None` entry in the push payload is the deletion marker: relative to
+        `parent_commit`, the server drops that path from the tree.
+        """
```
sydney-runkle: is this true?

libs/deepagents/deepagents/backends/context_hub.py
```diff
@@ -123,6 +124,29 @@ def _commit(self, files: dict[str, str]) -> None:
...
+        if not paths:
+            return
+
+        payload: dict[str, FileEntry | AgentEntry | SkillEntry | None] = dict.fromkeys(paths)
```
sydney-runkle: what does dict.fromkeys do? haven't seen before

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -534,6 +535,33 @@ async def aedit(
...
+
+        Raises:
+            NotImplementedError: If the routed backend does not implement
+                `delete`.
+        """
+        backend, stripped_key = self._get_backend_and_key(file_path)
+        res = backend.delete(stripped_key)
+        if res.path is not None:
+            res.path = file_path
+        return res
```
sydney-runkle, line 557: wow just realized we have so much boilerplate :(

sydney-runkle (changes requested):
looking good overall! some questions :)

libs/deepagents/deepagents/backends/state.py
imnishitha: yes, the files channel uses _file_data_delta_reducer (filesystem.py), which treats None as delete.

libs/deepagents/deepagents/backends/context_hub.py
imnishitha: It creates a dictionary from a sequence of keys, however this method was replaced by _commit method.

libs/deepagents/deepagents/backends/sandbox.py
imnishitha: I believe so, it does all checks and deletion in a single shell command rather than multiple execution round-trip to the sandbox.

libs/deepagents/deepagents/backends/composite.py
imnishitha: I can work on this sometime to clean up.

libs/deepagents/deepagents/backends/protocol.py
imnishitha: supports_execution and supports_delete can't be merged into one helper as composite.py imports protocol.py which can lead to circular import.

libs/deepagents/deepagents/backends/context_hub.py
```diff
@@ -102,13 +103,18 @@ def has_prior_commits(self) -> bool:
...
+        deletion marker: relative to `parent_commit`, the server drops that path
+        from the tree.
+        """
+        if not changes:
             return
 
         payload: dict[str, FileEntry | AgentEntry | SkillEntry | None] = {
-            path: FileEntry(type="file", content=content) for path, content in files.items()
+            path: FileEntry(type="file", content=content) if content is not None else None for path, content in changes.items()
         }
```
sydney-runkle, line 118: hmm, i don't think this works correctly? how does context hub support deletes?

libs/deepagents/deepagents/backends/protocol.py
sydney-runkle: can we make this a private util for the short term bc we anticipate potentially changing it

libs/deepagents/deepagents/backends/sandbox.py
sydney-runkle: @eyurtsev could you please review

sydney-runkle (commented):
a few more change requests, let's rebase on v0.7 also

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -569,6 +586,30 @@ async def aedit(
...
+        """Delete a file from the filesystem.
+
+        This method is optional. Backends that do not implement it inherit this
+        default, which raises `NotImplementedError`. Callers that need to support
+        a mix of backends should guard with
+        [`supports_delete`][deepagents.backends.protocol.supports_delete] before
+        calling, or catch `NotImplementedError`.
+
+        Args:
+            file_path: Absolute path to the file to delete. Must start with '/'.
```
eyurtsev: Could we document what we expect to happen if file_path is a path to a directory? 

file_path indicates that it's only supporting paths to files not directory, but I believe that the intention is to recursively support deletion of directories as well.

libs/deepagents/deepagents/backends/store.py
```diff
@@ -681,6 +682,35 @@ async def aedit(
         await store.aput(namespace, file_path, store_value)
         return EditResult(path=file_path, occurrences=int(occurrences))
 
+    def delete(self, file_path: str) -> DeleteResult:
```
eyurtsev, line 685: For store could we confirm how `*` is handled? If I'm not mistaken `*` is a wild card character

libs/deepagents/tests/unit_tests/backends/test_store_backend.py
```diff
@@ -515,3 +515,39 @@ def test_store_backend_legacy_path_rejects_malicious_assistant_id() -> None:
             warnings.simplefilter("always")
             with pytest.raises(ValueError, match="disallowed characters"):
                 be.write("/test.txt", "content")
+
+
+def test_store_backend_delete():
```
eyurtsev: nit: If possible let's explicitly documented output of test functions as -> None. I don't think we enforce this with the linter.

libs/deepagents/tests/unit_tests/test_end_to_end.py
```diff
@@ -2695,6 +2695,53 @@ def write_then_read(path: str, content: str) -> str:
         assert "buffered" in tool_msg.content
         assert "ERROR" not in tool_msg.content
 
+    def test_state_backend_delete_in_graph_context(self) -> None:
+        """delete() removes a file from state; missing paths report an error."""
+        backend = StateBackend()
+
+        @tool
```
eyurtsev, line 2702: Could we add a test on the async path as well?

libs/deepagents/deepagents/backends/context_hub.py
```diff
@@ -120,8 +126,11 @@ def _commit(self, files: dict[str, str]) -> None:
             self._commit_hash = match.group(1)
 
         if self._cache is not None:
-            for path, content in files.items():
-                self._cache[path] = content
+            for path, content in changes.items():
+                if content is None:
+                    self._cache.pop(path, None)
+                else:
+                    self._cache[path] = content
```
eyurtsev: Mutation in place

libs/deepagents/deepagents/backends/context_hub.py
imnishitha: I inspected the push_agent in langsmith.client. It explicitly preserves None entries when building the request payload.

mdrxy:
(nit) title of this PR got truncated:

`feat(sdk): delete operation to backend protocol and implem…`

we should restore where it got cut off with ellipses `...`

libs/deepagents/deepagents/backends/store.py
imnishitha: `file_path` is used as an exact store key. Wildcards are treated literally and do not expand to multiple entries.

libs/deepagents/tests/unit_tests/test_end_to_end.py
imnishitha: Added.

libs/deepagents/deepagents/backends/protocol.py
imnishitha: Added this info in the docstrings

libs/deepagents/tests/unit_tests/backends/test_store_backend.py
imnishitha: Added -> None

libs/deepagents/deepagents/backends/context_hub.py
imnishitha: Rewrote without mutation

libs/deepagents/deepagents/backends/context_hub.py
```diff
@@ -102,13 +103,18 @@ def has_prior_commits(self) -> bool:
...
+        Each value is either new file content or `None`. A `None` value is the
+        deletion marker: relative to `parent_commit`, the server drops that path
+        from the tree.
+        """
+        if not changes:
             return
 
         payload: dict[str, FileEntry | AgentEntry | SkillEntry | None] = {
-            path: FileEntry(type="file", content=content) for path, content in files.items()
+            path: FileEntry(type="file", content=content) if content is not None else None for path, content in changes.items()
```
sydney-runkle, line 117: this is a no op right?

libs/deepagents/deepagents/backends/protocol.py
sydney-runkle: this wasn't resolved?

libs/deepagents/deepagents/backends/sandbox.py
sydney-runkle: let's do a simple rm to start, this is really hard to reason about?

sydney-runkle (changes requested):
a few more change requests, looking close!

libs/deepagents/deepagents/backends/store.py
eyurtsev: Is there a test to confirm that? What are the rules for wild card expansion with store backend?

imnishitha: No, there was no test confirming how * is handled in delete. The method's docstring documents that file_path is used as an exact store key and * is treated literally (no wildcard expansion), but nothing verified that.
I've now added three tests to test_store_backend.py to verify this.

sydney-runkle (approved):

PR #3666, opened 2026-05-30
fix(sdk): align `glob` path default with `grep`

mdrxy:
> [!IMPORTANT]
> **TL;DR:** Omitting `path` in `glob` now searches each backend's own default root (e.g. `FilesystemBackend`'s `cwd`) instead of forcing a literal `"/"`, matching how `grep` already behaves. Explicit `path="/"` is unchanged.

> [!WARNING]
> **Custom backends:** `BackendProtocol.glob`/`aglob` now declare `path: str | None = None`, and the tool layer passes `None` (not `"/"`) when the caller omits `path`. Built-in backends are updated to treat `None` as their default root. A custom `BackendProtocol` subclass that assumed `path` is always a string must add a guard such as `path = path or "/"` at the top of its `glob`/`aglob` to avoid an `AttributeError` on unscoped calls.

---

`glob` and `grep` are paired filesystem search tools. Previously their schemas disagreed on the default search root: `glob.path` defaulted to `"/"` while `grep.path` defaulted to `null`. This makes omitted `glob.path` flow through as `None`, so each backend resolves its own default root instead of being handed a literal `"/"`.

**Behavior with an omitted `path`:**

- **`FilesystemBackend`** searches its configured `root_dir` / `cwd` — the agent's working directory, *not* the real OS root. Both an omitted `path` and an explicit `path="/"` resolve to `cwd`.
- **`StateBackend`**, **`StoreBackend`**, **sandbox**, **context hub**, **Harbor**, and **composite** backends search their respective default root.
- Error messages for an omitted `path` now display `<default>` rather than a misleading `'/'`.

**Behavior with an explicit `path`:**

- `path="/"` remains fully supported and behaves exactly as before, so existing callers are unaffected.
- Any other explicit `path` is resolved relative to the backend as it always has been.

**Schema change:**

- Generated tool schemas now expose `glob.path` as nullable with default `null`, matching `grep.path`. Unscoped `glob()` calls keep the same runtime behavior.

This is source-compatible for normal callers. The one compatibility risk is code that introspects the Python signature or generated tool schema and expects `glob.path.default == "/"` or a non-nullable `path` — those will now see `None` / a nullable field.


eyurtsev:
This makes sense! Could help to update the PR description a bit to focus up front on what the behavior is rather than focusing up front on the fact that the behavior is the same between grep and glob.

Basically expanding this more: 

> This aligns glob with grep by making omitted glob.path flow through as None, letting each backend choose its default search root. Explicit path="/" remains supported for compatibility.

To explain how things behave / expected to behave

eyurtsev (dismissed):

PR #3676, opened 2026-05-31
fix(code): paste clipboard images on macOS via Ctrl+V binding

### LINKED ISSUE #3502: "Copy image" macOS doesn't work
### Privileged issue

### Issue Content

from finder preview, web embedded inline

### Area (Required)

- [ ] deepagents (SDK)
- [ ] cli
- [x] code
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

NinaadRao:
Fixes #3502

---

**Root cause (updated from original diagnosis):**

When you "Copy Image" in a browser or Finder, the clipboard holds binary image data with no plain text. Terminal emulators (iTerm2, Terminal.app) only send bracketed paste sequences to the running app when there is plain text on the clipboard — with image-only clipboard data, they send **nothing**. This means no `Paste` event fires in Textual at all, so the original fix (checking `event.text == ""` in `_on_paste`) never ran. Confirmed via event logging.

**Fix:**

`get_clipboard_image()` — with its `pngpaste` / `osascript` backends — was already fully implemented in `media_utils.py` but needed a trigger the terminal actually forwards. `ctrl+v` is forwarded to the app as a key event on macOS (unlike Cmd+V which is consumed by the terminal for text paste). The fix adds a `ctrl+v` binding on `ChatTextArea` with `priority=True` that calls `get_clipboard_image()` in a background thread and inserts the `[image N]` placeholder if an image is found. No-op on non-macOS platforms.

Also adds `Ctrl+V  Paste image from clipboard (macOS)` to the `/help` keyboard shortcut reference.

**Usage:**
1. Right-click an image in a browser → Copy Image (or Finder → Copy)
2. Switch to `dcode`, press **Ctrl+V**
3. `[image 1]` placeholder appears in the input

- Manually tested: Copy Image in Safari → `ctrl+v` in running `dcode` → `[image 1]` inserted ✓
- `test_ctrl_v_with_clipboard_image_attaches_image` — action with mocked clipboard image inserts `[image 1]` and registers in tracker
- `test_ctrl_v_without_clipboard_image_is_noop` — action with no clipboard image leaves input and tracker unchanged
- `test_ctrl_v_on_non_macos_is_noop` — action is a no-op on non-macOS platforms
- All existing `TestDroppedImagePaste` tests pass unchanged

## Social handles (optional)
LinkedIn: https://linkedin.com/in/ninaadrao

### DIFF
--- a/libs/code/deepagents_code/widgets/chat_input.py
+++ b/libs/code/deepagents_code/widgets/chat_input.py
@@ -916,6 +916,23 @@ class ChatTextArea(TextArea):
         if self._paste_burst_buffer:
             await self._flush_paste_burst()
 
+        # On macOS, "Copy Image" (e.g. right-click → Copy Image in a browser)
+        # puts binary data in the clipboard that terminals cannot forward as
+        # text.  The paste event fires with empty text while the image lives in
+        # the system clipboard — read it directly via pngpaste / osascript.
+        if not event.text and self._chat_input_owner is not None:
+            from deepagents_code.media_utils import get_clipboard_image
+
+            image_data = await asyncio.to_thread(get_clipboard_image)
+            if image_data is not None:
+                tracker = self._chat_input_owner._image_tracker
+                if tracker is not None:
+                    placeholder = tracker.add_media(image_data, "image")
+                    self.insert(placeholder + " ")
+                    event.prevent_default()
+                    event.stop()
+                    return
+
         from deepagents_code.input import parse_pasted_path_payload
 
         try:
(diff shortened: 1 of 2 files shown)


NinaadRao:
Image I copied after pr
69cbaa4ce7ec0692a2d0dc06_Deep-Agents-CLI-blog
essing ctrl+v

Respon
Screenshot 2026-06-12 at 8 17 46 AM
se:

PR #3691, opened 2026-06-01
feat(sdk): `delete_file` tool

imnishitha:
Exposes the backend `delete` operation added in #3659 as an agent-facing
`delete_file` tool on the filesystem middleware, with sync/async variants,
path validation, write-permission gating, and a clear "deletion not available"
message for backends that don't implement `delete`.

This PR also merges `ContextHubBackend._commit` and `_commit_deletions` into a
single `_commit` that takes a `dict[str, str | None]` (a `None` value is the
deletion marker), removing duplicated push/commit-hash logic.

## Changes
- `delete_file` tool in the filesystem middleware (registered in the default
  tool set; `write` permission scope). Returns `Deleted file <path>` on success
  and surfaces backend errors (e.g. file-not-found) without faking success.
- Backends raising `NotImplementedError` for `delete` produce a friendly
  "Deletion not available" tool error rather than crashing.
- Refactor: `ContextHubBackend._commit` now handles both writes and deletions.
- Tests: unit tests for the tool (success, missing file, unsupported backend)
  and updated system-prompt/tool-schema snapshots.
- Evals: delete-focused trajectory evals in `libs/evals` plus a new
  `file_absent` assertion (a deleted path is removed from state entirely, so
  `FileExcludes` was too weak).

## How I verified
- `make format`, `make lint`, `make test` pass in `libs/deepagents`.
- Eval files lint and type-check clean (`libs/evals`); the evals themselves
  require a live `--model` + LangSmith and were not run in CI.

## AI disclaimer
Parts of this change (the `_commit` refactor and the eval scaffolding) were
drafted with assistance from an AI coding tool and reviewed by me.


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1190,6 +1209,113 @@ async def async_edit_file(
             args_schema=EditFileSchema,
         )
 
+    def _create_delete_file_tool(self) -> BaseTool:  # noqa: C901  # Tool wiring + permission/support handling
+        """Create the delete_file tool."""
+        tool_description = self._custom_tool_descriptions.get("delete_file") or DELETE_FILE_TOOL_DESCRIPTION
+
+        unsupported_msg = "Error: Deletion not available. This agent's backend does not implement the delete operation."
```
sydney-runkle: we shouldn't attach the tool if the backend doesn't support, i don't think?

sydney-runkle (changes requested):
let's stack on top of https://github.com/langchain-ai/deepagents/pull/3659#pullrequestreview-4405688709, can review further then

libs/deepagents/deepagents/middleware/filesystem.py
imnishitha: I've now taken care of this in the wrap_model_call to filter out the tool using `_filter_unsupported_tools_and_apply_prompt`

.github/workflows/release.yml
```diff
@@ -51,12 +51,14 @@ on:
         type: string
         default: ""
         description:
-          "SHA on main whose pyproject.toml declares the input version.
-          Typically the release-please PR's squash-merge commit. May also be a
-          pre-publish hotfix commit on top of it. Look up the release-please
-          merge with: gh pr view <release-pr-number> --json mergeCommit --jq
-          .mergeCommit.oid. Required unless dangerous-nonmain-release=true. See
-          .github/RELEASING.md > Manual Release."
+          "Exact commit (40-char SHA, not a branch ref) to build, publish, and
```
sydney-runkle, line 54: why are we changing this?

libs/deepagents/deepagents/backends/composite.py
sydney-runkle, line 542: hmm doesn't look like we've rebased properly

libs/deepagents/tests/unit_tests/smoke_tests/snapshots/system_prompt_with_execute_tools.json
```diff
@@ -144,6 +144,25 @@
     },
     "type": "function"
   },
+  {
+    "function": {
+      "description": "Deletes a file from the filesystem.\n\nUsage:\n- Permanently removes the file at the given absolute path.\n- Returns an error if the file does not exist or the path is a directory.\n- This cannot be undone, so only delete files you are sure are no longer needed.",
```
eyurtsev: We probably don't need this: ```- Returns an error if the file does not exist or the path is a directory.\n```

Part in the prompt. It's eating up tokens, but i doubt it meaningfully changes the behavior when an error is encountered

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1198,6 +1217,113 @@ async def async_edit_file(
...
+            if _check_fs_permission(self._permissions, "write", validated_path) == "deny":
+                return ToolMessage(
+                    content=f"Error: permission denied for write on {validated_path}",
+                    name="delete_file",
+                    tool_call_id=runtime.tool_call_id,
+                    status="error",
+                )
+            try:
+                res: DeleteResult = await resolved_backend.adelete(validated_path)
+            except NotImplementedError:
```
eyurtsev: Can we remove the try and execpt here? we expect adelete to always work (as there's a default fallback implementation)

imnishitha: Yes. 
I removed the catch from the middleware. The remaining try/except lives in CompositeBackend, because although CompositeBackend itself supports adelete, it may route to a sub-backend that doesn't. In that case the routed call can still raise NotImplementedError, so the composite converts it into a normal DeleteResult(error=...) rather than letting the exception escape.

sydney-runkle:
@copilot resolve the merge conflicts in this pull request

sydney-runkle (approved):
Amazing!!! nice work @imnishitha

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -39,6 +39,8 @@
 )
 from deepagents.backends.state import StateBackend
 
+_DELETE_UNSUPPORTED_ERROR = "Error: deletion is not supported for '{file_path}' (its backend does not implement delete)."
```
eyurtsev, line 42: The term "backend" is leaking an implementation detail to the LLM. I don't think the LLM is aware that there are different backends.

imnishitha: Fixed in: https://github.com/langchain-ai/deepagents/pull/3765

PR #3710, opened 2026-06-02
feat(code): show connection state in the status bar

mdrxy:
Connection and queued-message state now stays visible in the bottom status bar, even after the welcome banner scrolls out of view. Reconnect paths distinguish initial startup from mid-session restarts so users can see whether the app is connecting, reconnecting, or holding queued input.


mdrxy:
TODO after #3810 lands: please verify the connection/status-bar work here plays well with the new non-blocking `/threads` switch flow.

That PR makes thread switching run as a background worker and shows:
- status text via `_update_status("Loading thread: ...")`
- chat-area spinner via `_set_spinner("Loading thread")`

Before merging this PR, please smoke test:
1. `/threads` keyboard navigation still works
2. selecting a thread shows the loading state without freezing the UI
3. the status bar connection/queue indicator does not conflict with the thread-loading status message
4. the loading state clears after success, failure, and cwd-switch cancel paths

PR #3739, opened 2026-06-03
fix(sdk): add sandbox blob read thresholds

sydney-runkle:
Adjusts `BaseSandbox` read behavior to handle large blobs correctly. PDFs and other media files are commonly larger than 1 MB, so the inline blob threshold is raised to 10 MiB. Files above that fall back to `download_files()`, with a clear error if they exceed the download threshold too.

## Changes

- Bump `MAX_BINARY_BYTES` / `_MAX_BLOB_INLINE_BYTES` from 500 KiB → 10 MiB
- Add `_read_blob()` helper that drives the inline → download fallback path
- Unit tests covering the download fallback, oversize blobs, and non-threshold blob errors


libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -542,6 +552,62 @@ def read(
...
+            offset=0,
+            limit=0,
+        )
+        result = self.execute(cmd)
+        output = result.output.rstrip()
+
+        try:
+            data = json.loads(output)
+        except (json.JSONDecodeError, ValueError):
+            data = None
```
eyurtsev: The None path doesn't look valid

PR #3743, opened 2026-06-03
fix(sdk): make `CompositeBackend` `ls`/`als` work for both absolute and relative store keys

### LINKED ISSUE #3436: `CompositeBackend.als()` returns wrong directory when path exactly matches a registered route
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
"""Bug reproduction - Minimal reproducible example

  Bug: CompositeBackend.als() returns incorrect directory when path exactly matches a registered route
  """
  import asyncio
  from deepagents.backends import StateBackend, CompositeBackend, StoreBackend
  from deepagents.backends.utils import create_file_data


  async def reproduce_bug():
      store = InMemoryStore()
      store.put(
          namespace=("test-agent",),
          key="/memory/skills/test_skill/SKILL.md",
          value=create_file_data("---\nname: test_skill\n---\n# Test")
      )

      backend = CompositeBackend(
          default=StateBackend(),
          routes={
              "/memory/skills/": StoreBackend(
                  store=store,
                  namespace=lambda rt: ("test-agent",)
              ),
          },
      )

      # Direct call to StoreBackend - correct result
      store_backend = backend.sorted_routes[0][1]
      result1 = await store_backend.als("/memory/skills/")
      print(f"StoreBackend.als: {result1.entries}")

      # Call through CompositeBackend - incorrect result
      result2 = await backend.als("/memory/skills/")
      print(f"CompositeBackend.als: {result2.entries}")


  if __name__ == "__main__":
      asyncio.run(reproduce_bug())

  Expected output:
  StoreBackend.als: [{'path': '/memory/skills/test_skill/', 'is_dir': True, ...}]
  CompositeBackend.als: [{'path': '/memory/skills/test_skill/', 'is_dir': True, ...}]

  Actual output:
  StoreBackend.als: [{'path': '/memory/skills/test_skill/', 'is_dir': True, ...}]
  CompositeBackend.als: [{'path': '/memory/skills/memory/', 'is_dir': True, ...}]
```

```shell

```

### Description

When calling `CompositeBackend.als("/memory/skills/")` with a path that exactly matches a registered route, the
  returned directory listing is incorrect.

  **Expected:** The returned path should be `/memory/skills/test_skill/` (the actual skill directory stored in Backend).

  **Actual:** The returned path is `/memory/skills/memory/` (incorrect path).

  **Root cause:** After route matching, `_route_for_path` returns `backend_path = "/"`. StoreBackend's `ls("/")` returns
   entries with paths relative to the root (e.g., `/memory/`), but `_remap_file_info_path` incorrectly prepends the full
   route prefix to these already-absolute paths, resulting in duplicated path segments like `/memory/skills/memory/`.

  This bug affects SkillsMiddleware when loading skills from StoreBackend via CompositeBackend routes, causing skills to
   fail to load or return wrong paths.

### Environment / System Info

 - deepagents: 0.6.1
  - langchain: 1.3.1
  - langgraph: 1.2.0
  - Python version: 3.11
  - OS: Windows 11 (reproducible on all platforms - this is a pure logic bug)

imnishitha:
Fixes #3436

## Issue

CompositeBackend routes file ops by path prefix, stripping the matched prefix before delegating to the routed backend. For a routed ls/als it lists the route-stripped path and then re-prepends the prefix to each returned entry. That re-prefixing is wrong when the routed backend already stores full, route-prefixed keys (e.g. a StoreBackend seeded directly via store.put): the entry paths already contain the prefix, so re-prepending it duplicates a path segment.

The result is that listing a route returns paths like `/memory/skills/memory/ `instead of the real /`memory/skills/my-skill/`. This breaks SkillsMiddleware when skills live in a StoreBackend mounted behind a route — skills fail to load or resolve to nonexistent paths.

```py
store = InMemoryStore()
# A skill seeded directly into the store, so its key carries the full mounted path:
store.put(("agent",), "/memory/skills/my-skill/SKILL.md", create_file_data("..."))

backend = CompositeBackend(
    default=StateBackend(),
    routes={"/memory/skills/": StoreBackend(store=store, namespace=lambda _: ("agent",))},
)

# Direct backend call — correct:
await backend.routes["/memory/skills/"].als("/memory/skills/")
# -> /memory/skills/my-skill/   - RIGHT

# Through CompositeBackend — wrong:
await backend.als("/memory/skills/")
# -> /memory/skills/memory/     - WRONG (route prefix duplicated)

Expected: /memory/skills/my-skill/ (the actual skill directory in the store).
Actual: /memory/skills/memory/ (first segment of the stored key, with the route prefix re-prepended on top).

```

## Fix

CompositeBackend.ls/als duplicated a path segment when a routed backend held full route-prefixed keys. For a routed listing it now tries the requested path as-is first — a backend that stores absolute keys answers directly with correct paths — and only falls back to the route-stripped path with re-prefixing when that finds nothing (the mount-relative layout that CompositeBackend writes produce). Both key layouts now return identical, correct results, like a local ls.

A routed backend can hold keys in one of two layouts:

- Absolute — full route-prefixed keys (e.g. /memory/skills/test_skill/...), what a StoreBackend used directly holds.
- Mount-relative — route prefix stripped (e.g. /test_skill/...), what CompositeBackend writes after routing.

The absolute probe runs first and is treated as authoritative; the two listings are intentionally not merged, because re-prefixing absolute keys is exactly what produces the doubled prefix this PR fixes.

No breaking changes: backends written through the composite (mount-relatifore; only the previously-broken absolute-key case changes.

## How I verified

- Ran the issue's reproduction script verbatim — StoreBackend.als and Comturn /memory/skills/test_skill/ (previously the composite returned theduplicated /memory/skills/memory/).
- Added TestCompositeLsKeyLayouts covering both key layouts (absolute andnc ls and async als, at the route root and a nested directory.
- make format, make lint, make test all pass from libs/deepagents (full suite green; composite.py at 98% coverage).

BEGIN_COMMIT_OVERRIDE
fix(sdk): make `CompositeBackend` `ls`/`als` work for both absolute and relative store keys (#3743)

CompositeBackend.ls/als duplicated a path segment when a routed backend held full route-prefixed keys; it now probes the requested path as-is first and only falls back to the route-stripped path with re-prefixing, so both absolute and mount-relative key layouts list correctly.
END_COMMIT_OVERRIDE


sydney-runkle (commented):
is there any merit to the openswe comment?
could we add some more inline comments throughout this function to help w/ understanding? i like the minimalistic ones we already have, just a few lines

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -203,10 +203,13 @@ def ls(self, path: str) -> LsResult:
             path=path,
         )
         if route_prefix is not None:
-            ls_result = self._coerce_ls_result(backend.ls(backend_path))
-            if ls_result.error:
-                return ls_result
-            return LsResult(entries=[_remap_file_info_path(fi, route_prefix) for fi in (ls_result.entries or [])])
+            absolute = self._coerce_ls_result(backend.ls(path))
+            if not absolute.error and absolute.entries:
+                return absolute
```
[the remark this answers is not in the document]
imnishitha: The example assumes a single backend contains both:

```
/memory/skills/absolute.md
and
/relative.md
```

for the same route. In practice, a backend uses one convention or the other. A StoreBackend used directly stores absolute keys, while a StoreBackend mounted behind CompositeBackend stores mount-relative keys because the route prefix is stripped on writes.

StoreBackend.ls("/") matches all keys, including absolute route-prefixed ones. Those entries would then be remapped again, producing paths like:
`/memory/skills/memory/skills/...`

which is the original bug. The current approach intentionally treats a successful absolute-path listing as authoritative and only falls back to the stripped-path listing when no absolute entries are found. That avoids double-prefixing while still supporting both key layouts.

eyurtsev:
Could you improve the description for the PR so it's easy to understand what the problem was?

Ideally it's a very short code snippet that makes it easy to see intuitively what the expected result was vs. what the actual result is prior to this fix.

libs/deepagents/tests/unit_tests/backends/test_composite_backend.py
```diff
@@ -1403,3 +1405,67 @@ def test_edit_result_path_restored_to_full_routed_path():
...
+    keys (e.g. seeded via direct `store.put`) must not produce duplicated path
+    segments, while a store written through the composite (mount-relative keys)
+    keeps working.
+    """
+
+    def _absolute_keyed_composite(self) -> CompositeBackend:
+        store = InMemoryStore()
+        # Keys carry the FULL mounted path, as a direct store.put caller writes.
+        store.put(("agent",), "/memory/skills/test_skill/SKILL.md", create_file_data("x"))
+        store.put(("agent",), "/memory/skills/notes.md", create_file_data("y"))
```
sydney-runkle: side note, skills aren't stored in this format

sydney-runkle (dismissed):
approving, but maybe for correctness let's use proper skill format, see https://agentskills.io/home

libs/deepagents/tests/unit_tests/backends/test_composite_backend.py
imnishitha: added the right folder structure

libs/deepagents/tests/unit_tests/backends/test_composite_backend.py
```diff
@@ -1403,3 +1405,89 @@ def test_edit_result_path_restored_to_full_routed_path():
...
+    #   ├── scripts/       # executable code
+    #   ├── references/    # documentation
+    #   └── assets/        # templates, resources
+    _SKILL_FILES = (
+        "/memory/skills/my-skill/SKILL.md",
+        "/memory/skills/my-skill/scripts/run.py",
+        "/memory/skills/my-skill/references/guide.md",
+        "/memory/skills/my-skill/assets/template.txt",
+        "/memory/skills/notes.md",
+    )
```
sydney-runkle, line 1431: i don't think we need all of these, just wanted to make sure we weren't using the wrong structure in examples, let's simplify then merge?

sydney-runkle (approved):

PR #3759, opened 2026-06-04
feat(talon): add local runtime host

jkennedyvz:
Add `deepagents-talon`, a local single-operator runtime host for long-running Deep Agents over channels and scheduled jobs. The package wires a Deep Agents runtime to channel adapters, cron tools, MCP loading, optional voice transcription, LangSmith trace context, and a runnable WhatsApp Docker Compose example.

This first PR focused is on the local host, WhatsApp adapter, scheduler, and CI/release plumbing while reusing existing `deepagents-code` web and MCP utilities (where practical).

Areas that need careful review:

- Whether the generic channel abstractions should stay in the first PR or be narrowed to WhatsApp-only.
- WhatsApp exposure defaults, session storage, downloaded media retention, and open-mode acknowledgement.
- The dependency boundary with `deepagents-code`, especially MCP helpers and web tools.


examples/talon-whatsapp/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+COPY libs/talon ./libs/talon
+COPY examples/talon-whatsapp/AGENTS.md /opt/deepagents-talon/AGENTS.md
+
+RUN mkdir -p /opt/whatsapp-bridge && \
+    cp libs/talon/deepagents_talon/channels/whatsapp_bridge/package.json \
+        /opt/whatsapp-bridge/package.json && \
+    cd /opt/whatsapp-bridge && \
+    npm install --omit=dev
+
+RUN cd libs/talon && uv sync --frozen --extra speech
```
jkennedyvz, line 59: 🟡 **Speech extra breaks Docker build**

This image builds on `python3.13`, but the `uv sync --frozen --extra speech` step pulls the locked `numpy==1.26.4` via `transformers`, and that lock has no cp313 wheel. Reproducing the same install path with `uv sync --project libs/talon --python 3.13 --frozen --extra speech` fails while trying to build NumPy from sdist (`Cannot compile Python.h`), so the published WhatsApp Docker example cannot build. Either use a Python version covered by the lock for this image or refresh/constrain the speech lock so the extra installs on 3.13.

*(Refers to line 32)*

---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

libs/talon/deepagents_talon/channels/whatsapp_bridge/bridge.js
```diff
@@ -0,0 +1,544 @@
...
+  try {
+    const media = await message.downloadMedia();
+    if (!media || !media.data) {
+      return [];
+    }
+    const messageId = serializedId(message.id) || String(Date.now());
+    const extension = mediaExtension(media.mimetype, message.type);
+    const fileName = `${Date.now()}_${messageId.replace(/[^A-Za-z0-9]/g, "_")}.${extension}`;
+    const filePath = path.join(mediaDir, fileName);
+    fs.writeFileSync(filePath, Buffer.from(media.data, "base64"), { mode: 0o600 });
```
jkennedyvz, line 294: 🟡 **Inbound media can exhaust resources**

The bridge downloads the entire WhatsApp attachment and then decodes/writes it before any size checks run. The later Python media helpers skip images/documents above their limits, but an allowed (or open-mode) sender can still force the Node process to allocate the full base64 payload and persist it under `mediaDir`, which can exhaust memory or disk before those caps apply. Add a decoded-size limit in the bridge before `Buffer.from(..., "base64")`/`writeFileSync` so oversized media is rejected at the ingestion boundary.

*(Refers to lines 271-279)*

---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

libs/talon/deepagents_talon/config.py
```diff
@@ -0,0 +1,138 @@
...
+    default: str | None,
+) -> str | None:
+    for key in keys:
+        if key in env:
+            return env[key]
+    return default
+
+
+def _validate_assistant_id(assistant_id: str | None) -> None:
+    if not assistant_id or not _ASSISTANT_ID_PATTERN.fullmatch(assistant_id):
```
jkennedyvz: 🟡 **Dot assistant IDs escape home**

The validator accepts `.` and `..` as complete assistant ids, but `from_env()` uses the value directly as the per-assistant path segment (`home=root.expanduser() / assistant_id`). With `DEEPAGENTS_TALON_ASSISTANT_ID=..`, `ensure_home()` operates on the parent of the Talon root and creates/chmods `agent`, `cron`, `channels`, and media directories there instead of under the assistant namespace. Reject dot-only path segments before constructing `home`.

*(Refers to line 129)*

---
*Was this helpful? React with 👍 or 👎 to provide feedback.*

```suggestion
    if not assistant_id or assistant_id in {".", ".."} or not _ASSISTANT_ID_PATTERN.fullmatch(assistant_id):
```

jkennedyvz (commented):
**Open SWE Review** found 3 potential issues.

[View Open SWE trace](https://smith.langchain.com/o/ebbaf2eb-769b-4505-aca2-d11de10372a4/projects/p/6a5cf28f-7c41-4ee9-a11e-696c74ddb5f6/t/cc5d91bc-c18b-5056-935f-30283d7ba328)

PR #3762, opened 2026-06-04
fix(sdk): resolve virtual route paths in `CompositeBackend.execute`

### LINKED ISSUE #3050: Virtual paths from `CompositeBackend` routes don't resolve in `LocalShellBackend.execute`
Converted from #2974 (originally raised by @Smile0524).

## Problem

`CompositeBackend` routes filesystem ops (`read`/`ls`/`glob`/`grep`) by path prefix, but `execute` is not path-routable — it always delegates to the default backend, which receives the command string verbatim. When a skill references a virtual path like `python /common/skills/yyy/xxx.py`, the shell sees a literal OS path that doesn't exist, so `execute` fails. The agent then probes with `ls`/`grep`, which return more virtual paths, reinforcing the wrong mental model and producing a retry loop.

## Repro (from @Smile0524, `deepagents==0.5.2`, Python 3.11.4)

```python
default = LocalShellBackend(root_dir=specified_path, virtual_mode=True, timeout=60, inherit_env=True)
backend = CompositeBackend(
    default=default,
    routes={
        "/common/": FilesystemBackend(root_dir=common_path, virtual_mode=True),
        "/specified/": FilesystemBackend(root_dir=specified_path, virtual_mode=True),
    },
)
```

`SKILL.md` instructs the agent to run `python /common/skills/yyy/xxx.py`. File-ops resolve the path (route stripped, mapped to `{common_path}/...`). `execute` does not — `subprocess.run` runs the command literally with `cwd=specified_path`, and `/common/...` is not a real OS path there.

## Confirmed by source

- `CompositeBackend.execute` — *"Unlike file operations, execution is not path-routable — it always delegates to the default backend."*
- `LocalShellBackend.execute` — runs `subprocess.run(command, ..., cwd=str(self.cwd))` with the command string passed literally to the shell.
- `LocalShellBackend` `virtual_mode` docstring — *"This only affects filesystem operations. Shell commands executed via `execute()` are NOT restricted and can access any path."*

Documented design tension, but the namespaces (virtual FS vs. host shell) share syntax without sharing semantics, and the seam isn't visible to the model.

## Constraints (per @Smile0524)

- Cloud deployment — root path of the OS isn't known when authoring `SKILL.md`.
- Skill authors shouldn't hard-code deployment paths into `SKILL.md`.

## Possible directions

1. `CompositeBackend.execute` rewrites known virtual prefixes → backend `root_dir` in the command string before delegating.
2. `SKILL.md` resolution at load time substitutes virtual prefixes with resolved host paths.
3. An agent-facing wrapper tool that translates virtual paths before calling `execute`.
4. Docs-only — make the namespace split prominent and prescribe a translation pattern for skill authors.

(1) is most agent-transparent but mutates command strings (footguns: substring matches inside heredocs, quoted args, etc.). (2) keeps `execute` honest but pushes the route table into every skill loader. Worth design discussion before implementing.

---

Original discussion: #2974

imnishitha:
Fixes #3050

## Issue:
`CompositeBackend.execute()` always forwards the command string directly to the default backend's shell. 

So when the agent runs:
`python /common/skills/yyy/xxx.py`
the shell receives that exact string and tries to execute a real OS path named `/common/skills/yyy/xxx.py`. 

But `/common/` is only a virtual mount inside `CompositeBackend`. It does not exist on the host filesystem. The actual file lives somewhere like `/Users/.../shared/skills/yyy/xxx.py`.

The root cause is a mismatch the agent can't see: the file tools present routed **virtual** paths (`/common/x.py`), but `execute`/`aexecute` run on the backend's **host** shell, and the agent is never told the host path. Rather than
rewriting the command string which can't be done correctly for arbi(`$VAR`/`$(...)` expansion, here-docs, quoting, `cd`, dynamically built paths), this fixes it at the prompting layer: tell the model each route's hos
forms the correct command itself.

## What changed

- Added `_route_host_path_prompt(backend)` in the filesystem middleware. For a`CompositeBackend`, it builds a "Shell paths vs. virtual paths" prompt section that maps each routed virtual prefix to its host path.
  - Routes backed by a host filesystem (those exposing a `cwd` and running in `virtual_mode`) get an explicit `virtual → host` mapping the model should use
    in shell commands.
  - Routes with no host path (e.g. `StoreBackend`) are listed as shell-inaccessible, steering the model to the file tools
    (`read_file`/`write_file`/`edit_file`/`grep`/`glob`) for those instead.
- `wrap_model_call`/`awrap_model_call` append this section right after EXECUTION_SYSTEM_PROMPT`, only when the execute tool is present and the backen supports execution. Returns an empty string (no-op) for non-composite backends or backends with no routes.

## Why prompting instead of command rewriting

A regex rewrite of the command string can't be made correct: it can't see variable/command substitution, here-docs, or paths the command builds itself, and it risks corrupting legitimately literal occurrences. Steering the model at the prompt acts on whole path tokens the model itself produces, so it composes naturally with quoting, `cd`, and substitution, none of which a string rewrite handles.

## How I verified

- Added `tests/unit_tests/middleware/test_execute_route_prompt.py` covering:
 non-composite backends and route-less composites (no-op), virtual filesystem routes mapped to their resolved host path, store-backed routes marked shell-inaccessible (and never offered as a mapping), non-virtual filesystem
  routes excluded from mappings, and a mixed host/non-host route set.
- Ran `make format`, `make lint`, and `make test` from `libs/deepagents`.

Shoutout to @Smile0524 for raising this issue!


eyurtsev (changes requested):
Let's not do this -- I don't think there's a way to do this correctly with regexps.

What are other alternative solutions that might help at the prompting layer?

eyurtsev:
Would you mind adding details to description to help explain what the problem is more clearly?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -523,6 +523,59 @@ class ExecuteSchema(BaseModel):
...
+    lines = [
+        "## Shell paths vs. virtual paths",
+        "",
+        "The `execute` shell runs on the host filesystem. Some paths you see from "
+        "the file tools are virtual mounts that do NOT exist in the shell. When "
+        "running shell commands, use the host path instead of the virtual path.",
+    ]
+    if host_mappings:
+        lines.append("")
+        lines.extend(f"- `{prefix}` → use `{host}` in shell commands" for prefix, host in host_mappings)
```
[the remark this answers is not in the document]
imnishitha: Route prefixes and backend.cwd are no longer wrapped in Markdown backticks. They're now serialized with json.dumps() before being embedded in the prompt.

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -523,6 +524,57 @@ class ExecuteSchema(BaseModel):
 - execute: run a shell command in the sandbox (returns output and exit code)"""
 
 
+def _route_host_path_prompt(backend: BackendProtocol) -> str:
```
eyurtsev, line 548: Could we create a test that's a system prompt snapshot so it's easy to see what the materialized system prompt + tool descriptions are in this case?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -523,6 +524,57 @@ class ExecuteSchema(BaseModel):
...
+    lines = [
+        "## Shell paths vs. virtual paths",
+        "",
+        "The `execute` shell runs on the host filesystem. Some paths you see from "
+        "the file tools are virtual mounts that do NOT exist in the shell. When "
+        "running shell commands, use the host path instead of the virtual path.",
+    ]
+    if host_mappings:
+        lines.append("")
+        lines.extend(f"- {json.dumps(prefix)} → use {json.dumps(host)} in shell commands" for prefix, host in host_mappings)
```
eyurtsev: `→` => `->`

eyurtsev: We probably don't need json.dumps here and can just inline prefix and host. these are developer controller parameters

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -523,6 +524,57 @@ class ExecuteSchema(BaseModel):
...
+
+    Returns an empty string if there are no routes to describe.
+    """
+    if not isinstance(backend, CompositeBackend):
+        return ""
+
+    host_mappings: list[tuple[str, str]] = []
+    no_host_routes: list[str] = []
+    for route_prefix, route_backend in backend.sorted_routes:
+        host_root = getattr(route_backend, "cwd", None)
```
eyurtsev: Instead of `getattr` could we do an instance check to verify what the type of route_backend is and then based on the type do logic on whether cwd is present or absent

libs/deepagents/deepagents/middleware/filesystem.py
imnishitha: Added

libs/deepagents/tests/unit_tests/smoke_tests/test_system_prompt.py
```diff
@@ -95,6 +95,46 @@ def test_system_prompt_snapshot_with_execute(snapshots_dir: Path, *, update_snap
...
+        history[0]["tools"],
+        update_snapshots=update_snapshots,
+    )
+
+    messages = history[0]["messages"]
+    system_messages = [m for m in messages if isinstance(m, SystemMessage)]
+    assert len(system_messages) >= 1
+
+    # Redact the machine-specific host root so the snapshot is reproducible.
+    text = _system_message_as_text(system_messages[0]).replace(str(route.cwd), "<ROUTE_HOST_ROOT>")
```
eyurtsev: Instead of redaction here, let's use a realistic path; for example `/work/app`?

It'll make the system prompt easier to understand

eyurtsev:
What happens if CompositeBackend uses for virtual paths a StateBackend? (execute won't be able to access it?)

eyurtsev (dismissed):

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -523,6 +523,58 @@ class ExecuteSchema(BaseModel):
...
+    `LocalShellBackend`); other backends (e.g. store-backed) have none.
+
+    Returns an empty string if there are no routes to describe.
+    """
+    if not isinstance(backend, CompositeBackend):
+        return ""
+
+    host_mappings: list[tuple[str, str]] = []
+    no_host_routes: list[str] = []
+    for route_prefix, route_backend in backend.sorted_routes:
```
eyurtsev, line 586: Are all of these combinations handelled correctly w/ respect to the system prompt the agent sees?

* LocalShell vs. Sandbox
* FilesystemBackend no virtual mode, FilesystemBackend virtual mode, StateBackend

If not, do we need to handle all of them?

imnishitha: It's handled as below:

**Default = LocalShell (shell runs on the local disk)**

1. FS, virtual_mode=True 
- Prompt gives a mapping: /common/ -> <root_dir>.
- execute can reach it (model uses the host path). File tools work normally.

2. FS, virtual_mode=False
- Prompt gives a mapping: /common/ -> /.
- execute can reach it (model uses the path as-is). File tools work normally

3. StateBackend
- Prompt: "shell-inaccessible, use file tools."
- Data lives in memory, never on disk → execute genuinely can't reach it. File tools work.

**Default = Sandbox (shell runs inside the sandbox)**

4. FS, virtual_mode=True
- Prompt: "shell-inaccessible, use file tools."
- The files are on your local disk, but the shell runs in the sandbox → not reachable. File tools work (they read local disk).

5. FS, virtual_mode=False
- Same as 4
- File tools work (with the host-root quirk from 2).

6. StateBackend
- Prompt: "shell-inaccessible, use file tools."
- In memory, unreachable from the sandbox shell. File tools work.

libs/deepagents/tests/unit_tests/smoke_tests/snapshots/system_prompt_with_sandbox_default.md
```diff
@@ -0,0 +1,140 @@
...
+
+When a tool result is too large, it may be offloaded into the filesystem instead of being returned inline. In those cases, use `read_file` to inspect the saved result in chunks, or use `grep` within `/large_tool_results/` if you need to search across offloaded tool results and do not know the exact file path. Offloaded tool results are stored under `/large_tool_results/<tool_call_id>`.
+
+## Execute Tool `execute`
+
+You have access to an `execute` tool for running shell commands in a sandboxed environment.
+Use this tool to run commands, scripts, tests, builds, and other shell operations.
+
+- execute: run a shell command in the sandbox (returns output and exit code)
+
```
eyurtsev, line 91: This case is a bit degenerate case (no host mapping)


```
The `execute` tool can only access paths on the host filesystem. Some paths returned by file-system tools are virtual and cannot be accessed from the shell, including any path that begins with: `/common/`. 


Do not attempt to access these paths with `execute`.
```


web deployments aren't likely to use local file system

This sentence probably only relevant for Local file system + shell backend
> If a virtual mount has a host path mapping, replace its virtual prefix with the host prefix when running shell commands.

eyurtsev (approved):

PR #3843, opened 2026-06-10
fix(quickjs)!: remove `skills_backend`

hntrl:


sydney-runkle (dismissed):

ozyx:
Hi there! Early interpreter user here. No issue but curious about the rationale behind removing this?

hntrl:
@ozyx We removed it because we thought interpreter skills to be sort of orthogonal to how agents interact with files (e.g. overloading what a skill is by adding additional metadata defeats the "write once, use anywhere" purpose of skills). We're working to remediate this and come up with guidance on how to do this better, but in the mean time there's no hurt in just keeping your scripts inside of the skill directory, then in the skill instructions have it read the contents of the script file into the interpreter.

curious to hear any other feedback you have for code interpreter!

PR #3851, opened 2026-06-10
feat(sdk): delete dir support in delete tool

imnishitha:
The filesystem delete tool only removed single files and rejected directories, so clearing a directory (e.g. a subagent's working dir) meant one tool call per file and left the empty directories behind. This makes delete remove directories recursively, renames the tool from `delete_file` to `delete`, and makes the permission layer treat a recursive delete as a single all-or-nothing operation over the whole subtree.

What changed:
- `FilesystemBackend.delete` now removes directories recursively (`shutil.rmtree`). Symlinks are removed as links, never followed into their target.
- Renamed the tool `delete_file` to `delete` (the arg is still `file_path`).
- `delete` maps to the `write` operation. For a directory target, the tool expands the subtree and checks `write` on the target and every descendant; if any path is denied, the entire delete is refused and nothing is removed (all-or-nothing).
- `delete` is now registered for `interrupt`-mode permissions (bulk scope), so an interrupt rule on any path inside the target fires before the delete runs. It was previously absent, so interrupt rules never applied to deletes.

Why all-or-nothing: a recursive delete of a directory is one intent. If part of it is protected, refusing cleanly is safer and more predictable than leaving a half-deleted tree, and the agent can still delete the allowed paths explicitly.

### DIFF
--- a/libs/deepagents/deepagents/backends/protocol.py
+++ b/libs/deepagents/deepagents/backends/protocol.py
@@ -617,7 +617,7 @@ class BackendProtocol(abc.ABC):  # noqa: B024
         return await asyncio.to_thread(self.edit, file_path, old_string, new_string, replace_all)
 
     def delete(self, file_path: str) -> DeleteResult:
-        """Delete a single file from the filesystem.
+        """Delete a file (or directory) from the filesystem.
 
         This method is optional. Backends that do not implement it inherit this
         default, which raises `NotImplementedError`. Callers that need to support
@@ -625,16 +625,19 @@ class BackendProtocol(abc.ABC):  # noqa: B024
         [`supports_delete`][deepagents.backends.protocol.supports_delete] before
         calling, or catch `NotImplementedError`.
 
-        Directories are not supported. Directory-based backends return an error for
-        directory paths; key-value backends treat `file_path` as an exact key.
+        Directory handling is backend-dependent. Directory-based backends such as
+        [`FilesystemBackend`][deepagents.backends.filesystem.FilesystemBackend]
+        remove a directory recursively along with all of its contents; other
+        backends may return an error for directory paths. Key-value backends
+        treat `file_path` as an exact key.
 
         Args:
-            file_path: Absolute path to the file to delete. Must start with '/'
-                and refer to a file, not a directory.
+            file_path: Absolute path to the file or directory to delete. Must
+                start with '/'.
 
         Returns:
             `DeleteResult` with the deleted path on success, or an error if the
-                file does not exist or `file_path` refers to a directory.
+                path does not exist or removal fails.
 
         Raises:
             NotImplementedError: If the backend does not implement `delete`.
--- a/libs/deepagents/deepagents/backends/protocol.py
+++ b/libs/deepagents/deepagents/backends/protocol.py
@@ -617,7 +617,7 @@ class BackendProtocol(abc.ABC):  # noqa: B024
         return await asyncio.to_thread(self.edit, file_path, old_string, new_string, replace_all)
 
     def delete(self, file_path: str) -> DeleteResult:
-        """Delete a file (or directory) from the filesystem.
+        """Delete a path, recursively removing anything nested under it.
 
         This method is optional. Backends that do not implement it inherit this
         default, which raises `NotImplementedError`. Callers that need to support
@@ -625,19 +625,19 @@ class BackendProtocol(abc.ABC):  # noqa: B024
         [`supports_delete`][deepagents.backends.protocol.supports_delete] before
         calling, or catch `NotImplementedError`.
 
-        Directory handling is backend-dependent. Directory-based backends such as
-        [`FilesystemBackend`][deepagents.backends.filesystem.FilesystemBackend]
-        remove a directory recursively along with all of its contents; other
-        backends may return an error for directory paths. Key-value backends
-        treat `file_path` as an exact key.
+        Deletion is recursive: it removes `file_path` plus everything nested
+        under it. On hierarchical backends (e.g.
+        [`FilesystemBackend`][deepagents.backends.filesystem.FilesystemBackend])
+        that means a directory and its contents; on key-value backends it means
+        the exact key plus every key sharing the `file_path` + "/" prefix.
 
         Args:
-            file_path: Absolute path to the file or directory to delete. Must
-                start with '/'.
+            file_path: Absolute path to delete (a file, or a directory/prefix to
+                remove recursively). Must start with '/'.
 
         Returns:
-            `DeleteResult` with the deleted path on success, or an error if the
-                path does not exist or removal fails.
+            `DeleteResult` with the deleted path on success, or an error if
+                nothing exists at or under the path or removal fails.
 
         Raises:
             NotImplementedError: If the backend does not implement `delete`.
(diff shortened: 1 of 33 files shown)


libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -617,24 +617,27 @@ async def aedit(
         return await asyncio.to_thread(self.edit, file_path, old_string, new_string, replace_all)
 
     def delete(self, file_path: str) -> DeleteResult:
-        """Delete a single file from the filesystem.
+        """Delete a file (or directory) from the filesystem.
```
eyurtsev: nit: is there a more general name than directory (mostly thinking about s3 paths?)

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1266,93 +1270,119 @@ async def async_edit_file(
...
-            """Synchronous wrapper for delete_file tool."""
+            """Synchronous wrapper for delete tool."""
             resolved_backend = self._get_backend(runtime)
             try:
                 validated_path = validate_path(file_path)
             except ValueError as e:
                 return ToolMessage(
                     content=f"Error: {e}",
-                    name="delete_file",
+                    name="delete",
```
eyurtsev, line 1314: Could we confirm that name is used in other implementations? I can't remember off the top of my head, but thought that we end up using only tool call ids?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1266,93 +1270,119 @@ async def async_edit_file(
...
 
-    def _create_delete_file_tool(self) -> BaseTool:  # Tool wiring + permission/support handling
-        """Create the delete_file tool."""
-        tool_description = self._custom_tool_descriptions.get("delete_file") or DELETE_FILE_TOOL_DESCRIPTION
-
-        def sync_delete_file(
+    def _delete_denied_pattern(self, target: str) -> str | None:
+        """All-or-nothing write-deny check for a (possibly recursive) delete.
+
+        A recursive delete of ``target`` removes ``target`` plus every descendant,
```
eyurtsev: Could we follow google style doc-strings here?

```
Args:

Returns:
```

``deny`` -> `deny`

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1266,93 +1270,119 @@ async def async_edit_file(
             args_schema=EditFileSchema,
         )
 
-    def _create_delete_file_tool(self) -> BaseTool:  # Tool wiring + permission/support handling
-        """Create the delete_file tool."""
-        tool_description = self._custom_tool_descriptions.get("delete_file") or DELETE_FILE_TOOL_DESCRIPTION
-
-        def sync_delete_file(
+    def _delete_denied_pattern(self, target: str) -> str | None:
```
eyurtsev: Rename function to something else to make it clear that it's returning a list of violated permissions

eyurtsev (commented):
Strategy makes sense, let's pull the function to out of the class namespace and add exhaustive unit tests to check for the various edge cases

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -441,11 +441,13 @@ class ExecuteSchema(BaseModel):
...
 
-DELETE_FILE_TOOL_DESCRIPTION = """Deletes a file from the filesystem.
+DELETE_TOOL_DESCRIPTION = """Deletes a file or directory from the filesystem.
 
 Usage:
-- Permanently removes the file at the given absolute path.
-- This cannot be undone, so only delete files you are sure are no longer needed.
+- Permanently removes the file or directory at the given absolute path.
+- Deleting a directory removes it and everything inside it, recursively. Prefer
+  deleting a directory in one call over deleting each file individually.
```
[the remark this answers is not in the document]
imnishitha: it removes nested files too

libs/deepagents/deepagents/backends/store.py
```diff
@@ -685,36 +715,43 @@ async def aedit(
...
-        the file does not exist.
+        Returns:
+            `DeleteResult` with `file_path` on success, or an error if no key is
+                stored at or under it.
         """
         store = self._get_store()
         namespace = self._get_namespace()
 
-        if store.get(namespace, file_path) is None:
+        items = self._search_store_paginated(store, namespace)
```
eyurtsev, line 733: What is the performance of the implementation here? Anything we can do to save on data sent over the wire or number of network requests?

* Multiple network requests for data for a given namespace (pagination)?
* Are we listing only the keys or also the data at the path?
* Network request per delete action?

libs/deepagents/deepagents/backends/store.py
```diff
@@ -685,36 +715,41 @@ async def aedit(
...
+        to_delete = expand_delete_keys((str(item.key) for item in items), file_path)
```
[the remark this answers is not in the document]
eyurtsev: @imnishitha 

Optimizations that we can follow up with:

* Make sure that we have async implementations for grep, glob, ls (agrep, aglob, alas)
* Make sure that we don't eagerly load more information than we need. Pagination is OK but we don't need to load all pages into memory for things like grep, glob, ls
* When not needed we can work with keys only for glob and ls we only need the key not the value. (although full optimization would require changing how the store works in langgraph, and it's likely not that much of a high priority right now.)

* We want to be able to terminate glob/grep/ls early if there are too many results (this is not just for StoreBackend, but for all backends and would require modifying the schema of grep,glob and ls return type to accommodate some information indicating that that the result was truncated due to too many  matches. (And ask the agent in the tool message to narrow down its search). cc @sydney-runkle

libs/deepagents/tests/unit_tests/test_permissions.py
```diff
@@ -111,6 +113,211 @@ async def handler(_req):
...
+
+    Returns the `deny` write patterns whose glob overlaps the delete subtree
+    (target + descendants), capped at `limit`. Empty result == delete permitted.
+    """
+
+    @staticmethod
+    def _deny(*paths: str) -> FilesystemPermission:
+        return FilesystemPermission(operations=["write"], paths=list(paths), mode="deny")
+
+    def test_no_permissions_returns_empty(self):
```
eyurtsev: Here we can move to pytest.mark.parameterize to remove some boiler plate and make sure we cover all edge cases

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -135,6 +137,32 @@ def _check_fs_permission(
...
+    Returns:
+        Matching deny-write patterns, or an empty list if the delete is allowed.
+    """
+    denying: list[str] = []
+    seen: set[str] = set()
+    for rule in rules:
+        if rule.mode != "deny" or "write" not in rule.operations:
+            continue
+        for pattern in rule.paths:
+            if pattern not in seen and _paths_overlap(target, _glob_anchor(pattern)):
```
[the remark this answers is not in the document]
imnishitha: not the expected behaviour

libs/deepagents/deepagents/backends/store.py
```diff
@@ -685,36 +714,46 @@ async def aedit(
...
+        items = self._search_store_paginated(store, namespace)
+        # A recursive delete removes the exact key plus everything nested under it.
+        base = file_path.rstrip("/")
+        prefix = base + "/"
+        to_delete = [key for item in items if (key := str(item.key)) == base or key.startswith(prefix)]
+        if not to_delete:
             return DeleteResult(error=f"Error: File '{file_path}' not found")
 
-        store.delete(namespace, file_path)
+        store.batch([PutOp(namespace, key, None) for key in to_delete])
```
eyurtsev, line 741: None -> value=None

And i'd document that this does a deletion?

eyurtsev (approved):

PR #3863, opened 2026-06-11
fix(evals): preserve LangSmith sandbox workdir

mdrxy:
Restores the custom LangSmith Harbor environment adapter for eval runs so commands execute from the Docker image's `WORKDIR`, matching terminal-bench expectations. Harbor's built-in `langsmith` environment covers sandbox creation, but it does not preserve the image workdir when commands omit an explicit `cwd`, which can make verifier scripts run from the wrong directory.

## Changes
- Route only `sandbox_env=langsmith` Harbor workflow runs through `deepagents_harbor.langsmith_environment:LangSmithEnvironment` while leaving other sandbox providers on `--env`.
- Restore `LangSmithEnvironment` as a compatibility adapter between Harbor task configs and LangSmith sandboxes, including resource mapping, proxy config mapping, snapshot reuse, file transfer, and default cwd detection.
- Keep the `harbor[langsmith]` dependency so the adapter continues to use Harbor's LangSmith sandbox dependencies.

### DIFF
--- a/.github/workflows/harbor.yml
+++ b/.github/workflows/harbor.yml
@@ -361,7 +361,12 @@ jobs:
             n_tasks_flag="--n-tasks $HARBOR_N_TASKS"
           fi
 
-          env_flag="--env $HARBOR_SANDBOX_ENV"
+          # Keep LangSmith on our adapter so commands run from the image's Dockerfile WORKDIR.
+          if [ "$HARBOR_SANDBOX_ENV" = "langsmith" ]; then
+            env_flag="--environment-import-path deepagents_harbor.langsmith_environment:LangSmithEnvironment"
+          else
+            env_flag="--env $HARBOR_SANDBOX_ENV"
+          fi
 
           uv run harbor run \
             --agent-import-path deepagents_harbor:DeepAgentsWrapper \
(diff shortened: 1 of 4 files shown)


mdrxy:
Closing this in favor of fixing the behavior upstream in Harbor: https://github.com/harbor-framework/harbor/pull/1879

We’re willing to wait for a Harbor release with the LangSmith  fix rather than carry a DeepAgents-side adapter shim. Once that release is available, DeepAgents evals can depend on the fixed Harbor version and continue using the built-in  path.

mdrxy:
Correction: closing this in favor of fixing the behavior upstream in Harbor: https://github.com/harbor-framework/harbor/pull/1879

We’re willing to wait for a Harbor release with the LangSmith WORKDIR fix rather than carry a DeepAgents-side adapter shim. Once that release is available, DeepAgents evals can depend on the fixed Harbor version and continue using the built-in langsmith environment path.

mdrxy:
Correction/update: I verified the current released Harbor/LangSmith path against a fresh Dockerfile snapshot and it already inherits Dockerfile WORKDIR correctly.

Tested with WORKDIR /workspace, forced a fresh snapshot, and commands without explicit cwd returned /workspace and could read relative files from that directory. Explicit cwd=/ still overrode as expected.

So no DeepAgents adapter shim is needed, and no upstream Harbor patch is needed. If this failure appears again, the likely fix is to recreate the cached LangSmith snapshot.

PR #3875, opened 2026-06-11
feat(code): show model reason on tool approval prompts

### LINKED ISSUE #1374: Offer reason for interrupt/request to use tool
### Checked other resources

### Area (Required)

- [ ] deepagents (SDK)
- [x] code

### Feature description

When the CLI interrupts to request tool approval, display a model-generated **reason** explaining *why* it wants to run the tool. This mirrors how [Codex](https://github.com/openai/codex) surfaces intent before asking for permission:

```
Would you like to run the following command?

Reason: Do you want to allow running the targeted pytest file with uv so I can
validate this PR's behavior end-to-end?

$ uv run --group test pytest libs/cli/tests/unit_tests/test_chat_input.py -q
```

The CLI already has the plumbing — `ActionRequest.description` exists and the approval widget renders it ([`approval.py:291-298`](https://github.com/langchain-ai/deepagents/blob/main/libs/cli/deepagents_cli/widgets/approval.py#L291-L298)) — but nothing currently populates a model-generated reason, and the display treatment is generic dim text rather than a labeled, prominent field.

**Requirements:**

- Add a dedicated `reason: NotRequired[str]` field to `ActionRequest` (separate from `description`, which stays for static/factory use)
- Render as **`Reason: <italic text>`** in the approval widget, positioned between the title and tool-specific content — matching Codex's treatment
- Configurable per `InterruptOnConfig` which tools get reasons, defaulting to sensible high-risk tools (shell/exec at minimum)
- Always display the reason, including when auto-approved (thread-scope `a`)
- When the model doesn't provide a reason, show nothing (no fallback text)

**New `reason` field on `ActionRequest`:**

```python
class ActionRequest(TypedDict):
    name: str
    args: dict[str, Any]
    description: NotRequired[str]
    reason: NotRequired[str]  # NEW — model-generated justification
```

**Approval widget rendering:**

```python
reason = action_request.get("reason")
if reason:
    reason_widget = Static(
        f"Reason: [italic]{reason}[/italic]",
        classes="approval-reason",
    )
    await self._tool_info_container.mount(reason_widget, before=0)
```

**Open question — how the model populates `reason`:**

There are several approaches, each with trade-offs. This should be explored further before implementation:

| Approach | Pros | Cons |
|----------|------|------|
| **(a) Add `reason` as a parameter to every tool definition** | Cleanest; model fills it naturally as part of tool call schema | Touches tool definitions upstream in LangChain; token overhead on every call |
| **(b) Middleware injection** — intercept tool calls at the HITL layer, ask the model to justify before executing | No upstream changes; isolated to CLI | Extra LLM round-trip; latency cost |
| **(c) System prompt instruction** — tell the model to include reasoning in a structured way | Zero schema changes | Fragile; model may not comply consistently |

**Prior art:**

| Tool | Field | Display | Source |
|------|-------|---------|--------|
| Codex | `ExecApprovalRequestEvent.reason: Option<String>` | `Reason: <italic>` above command | [`approval_overlay.rs:363`](https://github.com/openai/codex/blob/main/codex-rs/tui/src/bottom_pane/approval_overlay.rs#L363) |
| OpenCode | `Permission.Info.message: String` | Event-driven via bus | [`permission/index.ts`](https://github.com/nicholasgriffintn/opencode/blob/main/packages/opencode/src/permission/index.ts) |
| Deep Agents (current) | `ActionRequest.description: NotRequired[str]` | `[dim]{description}[/dim]` | [`approval.py:291`](https://github.com/langchain-ai/deepagents/blob/main/libs/cli/deepagents_cli/widgets/approval.py#L291) |

mdrxy:
Closes #1374

Tool approval prompts now show a short model-generated reason explaining why the agent wants to run high-risk tools (shell, subagent dispatch).

---

When the app interrupts for approval of a high-risk tool, it now surfaces a short, model-generated justification — rendered as `Reason: <italic>` between the title and the tool-specific content. Previously the approval widget could render a static `description`, but nothing populated a model reason explaining *why* the tool was being run.

`ReasonInterruptMiddleware` subclasses the SDK's `HumanInTheLoopMiddleware` and attaches a `reason` to each configured `ActionRequest` before the interrupt fires, via one extra model call per interrupting batch. The call is tagged `lc_source="reason"` so it is filtered out of the streamed transcript (same mechanism the summarization middleware uses).

Which tools get a reason is configured per tool via a `reason: True` flag on the `InterruptOnConfig` (`ReasonInterruptOnConfig`), defaulting to the highest-risk gated tools (`execute`, `task`). The reason is also shown when a call is auto-approved (thread-scope `a`), and nothing is rendered when the model provides no reason.

To avoid a second `HumanInTheLoopMiddleware` on the main agent, `interrupt_on` is no longer forwarded to `create_deep_agent`; subagents instead carry `interrupt_on` on their own specs and keep their existing (reason-less) approval gating.

## Test Plan
- [ ] Trigger a shell-command approval and confirm a `Reason:` line appears above the command, and that it also shows when auto-approved.


mdrxy:
Closing in favor of #1374, the existing tracking issue for this feature.

PR #3876, opened 2026-06-11
fix(evals): make `DeepAgentsWrapper` temperature optional + fix langsmith make target

langchain-infra:
## Problem

`DeepAgentsWrapper` passes `temperature=0.0` to `init_chat_model` unconditionally:

```python
self._model = init_chat_model(model_name, temperature=temperature, **model_kwargs)
```

Anthropic **Opus 4.7+ and Fable** removed sampling parameters and reject `temperature` with:

```
anthropic.BadRequestError: 400 - {'type': 'invalid_request_error',
  'message': '`temperature` is deprecated for this model.'}
```

So those models fail at the **first model call**. This includes `anthropic:claude-opus-4-7`, which is in the `.github/workflows/harbor.yml` model matrix — meaning Opus runs in CI hit this too.

Surfaced while running terminal-bench on the LangSmith Harbor environment with `--model anthropic:claude-opus-4-8`.

## Fix

Default `temperature` to `None` and forward it to `init_chat_model` only when explicitly set:

```python
if temperature is not None:
    model_kwargs["temperature"] = temperature
self._model = init_chat_model(model_name, **model_kwargs)
```

- Models that accept `temperature` can still pass it (`--agent-kwarg temperature=0.0`).
- Models that reject it are no longer sent an invalid parameter.

Added unit tests covering omit-by-default and forward-when-set; updated the existing happy-path assertion (`_temperature` now defaults to `None`). `ruff`, `ty`, and the wrapper unit tests pass.

## 🙏 Looking for OSS input on the default

This changes the default from `temperature=0.0` → omitted. The upside is broad model support out of the box (Opus 4.7+/Fable just work). The tradeoff: models that *do* accept `temperature` (e.g. Sonnet) are no longer pinned to `0.0`, so they fall back to the provider default — a small **eval-reproducibility** shift.

Alternatives considered:
1. **This PR** — default `None` (omit), opt in per-run. Simple, model-agnostic.
2. Keep `0.0` default but skip `temperature` for known-incompatible model families. More brittle (hardcodes model lists).
3. Catch the 400 and retry without `temperature`. Hides the issue, costs a round-trip.

Preference / concerns welcome — happy to switch approaches.

### DIFF
--- a/libs/evals/Makefile
+++ b/libs/evals/Makefile
@@ -65,9 +65,9 @@ run-terminal-bench-runloop: ## Run terminal-bench on Runloop (10 concurrent)
 	@mkdir -p jobs/terminal-bench
 	harbor run --agent-import-path deepagents_harbor:DeepAgentsWrapper --dataset terminal-bench@2.0 -n 10 --jobs-dir jobs/terminal-bench --env runloop $(AGENT_KWARG)
 
-run-terminal-bench-langsmith: ## Run terminal-bench on LangSmith prod sandboxes (sequential)
+run-terminal-bench-langsmith: ## Run terminal-bench on LangSmith prod sandboxes (sequential; override with MODEL=<id>)
 	@mkdir -p jobs/terminal-bench
-	harbor run --agent-import-path deepagents_harbor:DeepAgentsWrapper --dataset terminal-bench@2.0 -n 1 --jobs-dir jobs/terminal-bench --env langsmith $(AGENT_KWARG)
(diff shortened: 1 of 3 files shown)


langchain-infra:
Verified end-to-end: with this fix, `anthropic:claude-opus-4-8` runs through the full DeepAgents loop on the LangSmith Harbor environment — 15m41s of real model calls, **no `temperature` 400**. Before the fix the same invocation died at the first model call in ~31s with `400 - temperature is deprecated for this model`.

(The single terminal-bench task then hit the default 900s agent timeout → reward 0.0, which is task difficulty, not the model-init path.)

langchain-infra:
Folded in the `run-terminal-bench-langsmith` Make-target fix (adds `--model anthropic:claude-opus-4-8`; the target was missing `--model` and failed with `model_name must be a non-empty string`). Supersedes #3878, which I'm closing in favor of this combined PR.

mdrxy (approved):

PR #3877, opened 2026-06-11
perf(sdk): count tokens once per model call in summarization middleware

nick-hollon-lc:
`SummarizationMiddleware.wrap_model_call` / `awrap_model_call` counted tokens over the full effective history twice on every model call — once inside `_truncate_args` and again for the should-summarize check — even on the common path where nothing is truncated or summarized. The count is expensive because counting with `tools=` converts every tool schema on every invocation; in a production deep-agent trace with 23 tools this middleware added roughly 280ms to every model call, and the duplicate count is half of that.

The count now runs once per call and is passed into `_truncate_args` via a new keyword-only `total_tokens` parameter that defaults to computing itself, so existing direct callers are unaffected. A recount happens only when truncation actually modified messages (which changes the total).

Related: langchain-ai/langchain#38073 addresses the underlying per-conversion tool-schema cost.

### DIFF
--- a/libs/deepagents/deepagents/middleware/summarization.py
+++ b/libs/deepagents/deepagents/middleware/summarization.py
@@ -711,11 +711,26 @@ A condensed summary follows:
             }
         return tool_call
 
+    def _count_tokens(
+        self,
+        messages: list[AnyMessage],
+        system_message: SystemMessage | None,
+        tools: list[BaseTool | dict[str, Any]] | None,
+    ) -> int:
+        """Count tokens for messages plus optional system message and tools."""
+        counted_messages = [system_message, *messages] if system_message is not None else messages
+        try:
+            return self.token_counter(counted_messages, tools=tools)  # ty: ignore[unknown-argument]
+        except TypeError:
+            return self.token_counter(counted_messages)
+
     def _truncate_args(
         self,
         messages: list[AnyMessage],
         system_message: SystemMessage | None,
         tools: list[BaseTool | dict[str, Any]] | None,
+        *,
+        total_tokens: int | None = None,
     ) -> tuple[list[AnyMessage], bool]:
         """Truncate large tool call arguments in old messages.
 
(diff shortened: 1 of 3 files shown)


libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -711,28 +711,44 @@ def _truncate_tool_call(self, tool_call: ToolCall) -> ToolCall:
...
+        except TypeError:
+            return self.token_counter(counted_messages)
+
     def _truncate_args(
         self,
         messages: list[AnyMessage],
         system_message: SystemMessage | None,
         tools: list[BaseTool | dict[str, Any]] | None,
+        *,
+        total_tokens: int | None = None,
```
eyurtsev: if all private then we don't worry about backwards compact (less code paths to maintain)

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -711,28 +711,44 @@ def _truncate_tool_call(self, tool_call: ToolCall) -> ToolCall:
...
+    def _count_tokens(
+        self,
+        messages: list[AnyMessage],
+        system_message: SystemMessage | None,
+        tools: list[BaseTool | dict[str, Any]] | None,
+    ) -> int:
+        """Count tokens for messages plus optional system message and tools."""
+        counted_messages = [system_message, *messages] if system_message is not None else messages
+        try:
+            return self.token_counter(counted_messages, tools=tools)  # ty: ignore[unknown-argument]
```
sydney-runkle, line 733: why is the signature having typing problemos here

sydney-runkle (commented):
idea seems right!

libs/deepagents/deepagents/middleware/summarization.py
nick-hollon-lc: carry over from before

eyurtsev (approved):

mdrxy (approved):

PR #3879, opened 2026-06-11
perf(sdk): speed up grep/read hot path, cache dynamic prompt

mdrxy:
Targeted performance improvements in `libs/deepagents` hot paths, with no public signature changes.

- The Python `grep` fallback (`FilesystemBackend._python_search`) now streams each file line-by-line instead of loading the whole file via `read_text().splitlines()`, does literal substring matching instead of compiling/running a regex over an `re.escape`d pattern, and checks the wall-clock deadline *within* long files so a single huge file can't blow the timeout or peak memory. Per-file read failures encountered mid-scan are now surfaced as a partial error instead of being silently dropped.
- Glob filters in the in-memory grep helpers (`_grep_search_files`, `grep_matches_from_files`) are compiled once via a cached `_compile_glob` instead of recompiling per file; the Context Hub `grep` hoists its `fnmatch` translation out of the loop.
- `slice_read_response` normalizes line endings only across the requested window rather than rewriting the entire (potentially huge) file before slicing.
- The filesystem middleware memoizes its dynamic system prompt (keyed on whether the execute tool is present and supported) instead of rebuilding it on every model call.

### Behavioral notes

The in-memory grep helpers, `slice_read_response`, and the middleware cache are byte-for-byte behavior-preserving. The Python `grep` fallback (exercised only when `ripgrep` is unavailable) has two intentional, non-breaking changes that bring it in line with both the sibling `read()` method and the `ripgrep` primary path:

- **Decoding is pinned to UTF-8** (`errors="strict"`) instead of the platform locale default. Every other read/write path in the backends already pins UTF-8; the fallback was the lone outlier since the original experimental backend. On a non-UTF-8 locale, fallback decoding now matches `read()` and `ripgrep` rather than the host locale.
- **Line splitting is narrower**: lines now break only on `\n` / `\r` / `\r\n` (matching `ripgrep`), not on the exotic separators that `str.splitlines()` also breaks on (form feed, vertical tab, file/group/record separators, NEL, and the Unicode line/paragraph separators). CRLF / CR match text is unchanged, since universal-newline translation still applies on read.

### Benchmarks

Focused local microbenchmarks compared this branch against `main` on Python 3.14.2:

- Normal `FilesystemBackend.grep` with `ripgrep` installed: no meaningful change (`main`: 0.022466s, PR: 0.022572s median; ~0.5% slower, effectively noise).
- Forced Python grep fallback: **1.06x faster** (`main`: 0.030186s, PR: 0.028388s median; ~6.0% improvement).
- State backend `_grep_search_files` with glob filtering: **3.50x faster** (`main`: 0.045913s, PR: 0.013116s median; ~71.4% improvement).
- State backend `grep_matches_from_files` with glob filtering: **3.71x faster** (`main`: 0.044786s, PR: 0.012071s median; ~73.0% improvement).
- `slice_read_response` on large CRLF content: **2.55x faster** for both first-window and late-window reads (~60.8% improvement), with **~29.2% lower peak memory**.

Takeaway: the common filesystem grep path remains dominated by `ripgrep`, while this PR substantially improves in-memory/state backend grep with glob filters and large-content read slicing.

### Tests

Added unit coverage for literal substring matching (including a negative case proving regex metacharacters stay inert), the mid-file per-line timeout (driven by a stubbed clock so it provably reaches the in-file branch rather than the outer guard), CRLF match-text stripping, non-UTF-8 file skipping, partial-read error surfacing, Context Hub `grep` glob filtering, glob filtering with brace expansion, a partial-window CRLF slice in `slice_read_response`, and the dynamic-prompt cache.

## Self-Hosted Release Note
Faster file `grep`/`read` operations and reduced per-request overhead in the filesystem middleware.

### DIFF
--- a/libs/deepagents/deepagents/backends/context_hub.py
+++ b/libs/deepagents/deepagents/backends/context_hub.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import fnmatch
 import logging
+import os
 import re
 from typing import TYPE_CHECKING
 
@@ -253,10 +254,12 @@ class ContextHubBackend(BackendProtocol):
 
         prefix = self._strip_prefix(path).rstrip("/") if path else ""
 
+        glob_re = re.compile(fnmatch.translate(os.path.normcase(glob))) if glob else None
+
         for file_path, content in cache.items():
             if prefix and not file_path.startswith(prefix):
                 continue
-            if glob and not fnmatch.fnmatch(file_path, glob):
+            if glob_re is not None and not glob_re.match(os.path.normcase(file_path)):
                 continue
             for i, line in enumerate(content.splitlines(), start=1):
                 if regex.search(line):
--- a/libs/deepagents/deepagents/backends/filesystem.py
+++ b/libs/deepagents/deepagents/backends/filesystem.py
@@ -6,7 +6,6 @@ import functools
 import json
 import logging
 import os
-import re
 import shutil
 import subprocess
 import time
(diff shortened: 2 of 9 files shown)


mdrxy:
Superseded by smaller focused PRs: #3886, #3887, #3888, and #3889.

PR #3882, opened 2026-06-11
feat(sdk): support configurable subagent response format

hntrl:
Adds SDK support for passing per-dispatch subagent response formats through `RunnableConfig.configurable` instead of runtime context. This gives bridge callers a stable way to request schema-backed task-tool dispatch for raw `SubAgent` specs while keeping compiled/runnable-backed subagents fixed at construction time.

The task tool now reads a private configurable response-format key, compiles temporary raw `SubAgent` variants with that response format, and rejects dynamic schemas for compiled subagents. The raw-spec compilation path stays centralized in `create_sub_agent`, and tests cover the dynamic response-format path, compiled-subagent rejection, name/config propagation, and custom state-schema compilation.

### DIFF
--- a/libs/deepagents/deepagents/middleware/subagents.py
+++ b/libs/deepagents/deepagents/middleware/subagents.py
@@ -4,7 +4,7 @@ import contextlib
 import dataclasses
 import json
 from collections.abc import Awaitable, Callable, Generator, Sequence
-from typing import Any, NotRequired, TypedDict, TypeGuard
+from typing import Any, NotRequired, TypedDict, cast
 
 from langchain.agents import create_agent
 from langchain.agents.middleware import HumanInTheLoopMiddleware, InterruptOnConfig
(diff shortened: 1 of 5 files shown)


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -232,6 +241,20 @@ def node(state):
...
+) -> TypeGuard[CompiledSubAgent]:
+    """Return whether `spec` is a compiled subagent spec."""
+    return "runnable" in spec
+
+
+def _is_raw_subagent(
+    spec: SubAgent | CompiledSubAgent,
+) -> TypeGuard[SubAgent]:
+    """Return whether `spec` is a raw subagent spec."""
+    return not _is_compiled_subagent(spec)
```
sydney-runkle: this is unnecessary? extra indirection

sydney-runkle (dismissed):
fast tracking repl support, this is largely internals

sydney-runkle (approved):

PR #3918, opened 2026-06-12
feat(sdk): add context_lines parameter to grep

### LINKED ISSUE #3109: Add `context_lines` parameter to `grep` for surrounding-line visibility
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] repl
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

### Feature description

The `grep` tool currently returns only the exact matching line for each result. When an agent is navigating an unfamiliar codebase, this is often not enough to understand what it found — it has to follow up with `read_file` just to make sense of the context around the match. This adds unnecessary round-trips and token spend.

**Example:** an agent searching for `"class Config"` gets back file path + line number. To understand whether this is the class it needs, it then calls `read_file` on the same file. With context lines, that information would already be in the grep result.

A `context_lines: int = 0` parameter (default `0`, preserving current behaviour exactly) would let agents — and users configuring them — trade a slightly larger tool result for fewer follow-up reads.

**User impact and usage**

A developer building an agent for codebase exploration or refactoring tasks can pass `context_lines=3` to get the three lines before and after each match in a single call — enough to see a function signature above a match or the closing logic below it, without an extra `read_file`. For agents running on large codebases where every tool call costs tokens and latency, this directly reduces the number of round-trips needed to understand search results. The default of `0` means no change for existing users.

Ripgrep already handles this natively. When called with `--json -C N`, it emits `"type": "context"` entries interleaved with `"type": "match"` entries in documented order:

```
{"type":"context", "data":{"path":...,"lines":{"text":"line two\n"},"line_number":2,...}}
{"type":"match",   "data":{"path":...,"lines":{"text":"target line\n"},"line_number":3,...}}
{"type":"context", "data":{"path":...,"lines":{"text":"line four\n"},"line_number":4,...}}
```

The current `_ripgrep_search` parser (`backends/filesystem.py`) already reads this JSON stream — it skips non-`"match"` entries with `if data.get("type") != "match": continue`. Collecting context entries in order (before vs. after a match) is straightforward with no structural change to the parser.

**Proposed API:**

```python
# GrepSchema — new optional field, backward-compatible
context_lines: int = Field(
    default=0,
    ge=0,
    description="Number of lines to include before and after each match (0 = matching line only).",
)

# GrepMatch — two new NotRequired fields, no breaking change
class GrepMatch(TypedDict):
    path: str
    line: int
    text: str
    context_before: NotRequired[list[tuple[int, str]]]  # (line_number, text) pairs preceding the match
    context_after: NotRequired[list[tuple[int, str]]]   # (line_number, text) pairs following the match
```

**Files that change** (all additive, no existing signatures broken):

| File | Change |
|---|---|
| `middleware/filesystem.py` | `context_lines` in `GrepSchema`; pass through `sync_grep` / `async_grep` to backend |
| `backends/protocol.py` | `context_before/after: NotRequired[list[tuple[int, str]]]` on `GrepMatch`; `context_lines: int = 0` on `grep()` / `agrep()` |
| `backends/filesystem.py` | `-C {context_lines}` to ripgrep call; update parser to collect context entries; ±N line slicing in Python fallback `_python_search` |
| `backends/sandbox.py` | `-C {context_lines}` to shell grep command when > 0; stateful parser for context output |
| `backends/utils.py` | Surrounding-line slice in `grep_matches_from_files`; context-aware rendering in `format_grep_matches` |
| `backends/state.py`, `backends/store.py`, `backends/composite.py` | `context_lines: int = 0` added to `grep()` / `agrep()` signatures; passed through to sub-backends |

`context_lines=0` (default) is identical to current behaviour at every layer.

- The system prompt explicitly discourages agents from using `execute(command="grep ...")` as a workaround, so the structured `grep` tool is the only sanctioned path for content search.
- This is a standard capability in most content search tools — the equivalent of `grep -C N` — and the absence of it in the structured tool is a consistent source of extra `read_file` round-trips.
- I am happy to implement this and open a PR if assigned.

i-anubhav-anand:
Fixes #3109

Adds an opt-in `context_lines: int = 0` parameter to the `grep` capability so each match can include up to N lines of surrounding context — saving the follow-up `read_file` round-trip the issue describes. Default `0` preserves current behavior exactly (existing matches/output are byte-identical; `GrepMatch` only gains `context_before`/`context_after` keys when context is requested).

**Design**

- Context is attached by re-reading each matched file once and slicing windows in Python, rather than parsing ripgrep's interleaved `"type":"context"` JSON events — so the ripgrep path, the Python fallback, and the in-memory backends (state, store, context-hub) behave identically. Values are clamped to `[0, 10]`.
- `CompositeBackend` forwards the parameter to routed backends only when set, so custom backends that still override the older 3-argument `grep` keep working as long as no context is requested (same guard in `BackendProtocol.agrep`).
- Sandbox backends accept the parameter but return matches without context (documented) — honoring it would cost an extra remote round-trip per matched file; can be a follow-up if desired.
- The `grep` tool schema exposes `context_lines` (0–10), and `content` output mode renders context grep `-C` style: context lines use `LINENO-`, matches keep the existing `LINENO:` shape, `--` separates match groups within a file.

**Testing**

- New `tests/unit_tests/backends/test_grep_context_lines.py` (9 tests): in-memory backend context before/after, file-boundary clamping, clamp bounds, local backend on both the ripgrep path and the forced Python fallback, formatter rendering with context, and no-context output unchanged.
- Tool-schema snapshots regenerated (`--update-snapshots`) since the grep tool gains a parameter.
- `make format`, `make lint` (ruff + ty) clean; full unit suite: 1807 passed, 0 failed.

### DIFF
--- a/libs/deepagents/deepagents/backends/composite.py
+++ b/libs/deepagents/deepagents/backends/composite.py
@@ -50,6 +50,16 @@ def _remap_grep_path(m: GrepMatch, route_prefix: str) -> GrepMatch:
     )
 
 
+def _ctx_kwargs(context_lines: int) -> dict[str, int]:
+    """Build optional grep kwargs for delegate calls.
+
+    Forwards `context_lines` only when set, so routed backends that still
+    use the older 3-argument `grep` signature keep working when no context
+    is requested.
+    """
+    return {"context_lines": context_lines} if context_lines else {}
+
+
 def _strip_route_from_pattern(pattern: str, route_prefix: str) -> str:
     """Strip a route prefix from a glob pattern when the pattern targets that route.
 
@@ -308,6 +318,7 @@ class CompositeBackend(BackendProtocol):
         pattern: str,
         path: str | None = None,
         glob: str | None = None,
+        context_lines: int = 0,
     ) -> GrepResult:
         """Search files for literal text pattern.
 
@@ -319,6 +330,8 @@ class CompositeBackend(BackendProtocol):
             path: Directory to search. None searches all backends.
             glob: Glob pattern to filter files (e.g., "*.py", "**/*.txt").
                 Filters by filename, not content.
+            context_lines: Lines of surrounding context to include with each
+                match, forwarded to the routed backend(s). Defaults to 0.
 
         Returns:
             GrepResult with matches or error.
@@ -337,7 +350,7 @@ class CompositeBackend(BackendProtocol):
                 path=path,
             )
             if route_prefix is not None:
-                grep_result = self._coerce_grep_result(backend.grep(pattern, backend_path, glob))
+                grep_result = self._coerce_grep_result(backend.grep(pattern, backend_path, glob, **_ctx_kwargs(context_lines)))
                 if grep_result.error:
                     return grep_result
                 return GrepResult(matches=[_remap_grep_path(m, route_prefix) for m in (grep_result.matches or [])])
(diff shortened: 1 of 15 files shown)


i-anubhav-anand:
Good catch @corridor-security — fixed in a6b14322: `_attach_context` now enforces the same `max_file_size_bytes` guard as the Python-fallback search path (files over the limit return their matches without context instead of being read). Added a regression test (`test_context_skips_files_over_size_limit`). Full suite re-run: green.

PR #3954, opened 2026-06-13
feat(code): `tool.use` and `tool.result` hook events

### LINKED ISSUE #3953: Add `tool.use` and `tool.result` hook events (`PreToolUse` / `PostToolUse`)
### Submission checklist

### Area (Required)

- [ ] deepagents (SDK)
- [ ] cli
- [x] code
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Feature description

The existing hook system covers session and context events (`session.start`, `user.prompt`, `tool.error`, etc.) but has no equivalent of Claude Code's `PreToolUse` / `PostToolUse`. Every tool call — `read_file`, `write_file`, `execute`, and MCP tools — runs without any observable hook for external scripts.

Two new events are proposed:

- **`tool.use`** — fires before a tool executes, with tool name and arguments
- **`tool.result`** — fires after completion, with name, arguments, output, and status

---

## Comparison with Claude Code

Claude Code's `PreToolUse` and `PostToolUse` are the most commonly used hooks in its ecosystem. They let users run external scripts that log tool calls, send notifications, enforce guardrails, and build observability dashboards — all without modifying agent code. deepagents users migrating from Claude Code currently have no equivalent.

The existing `tool.error` event is a partial PostToolUse that fires only on failure. `permission.request` fires only at the HITL dialog (not when auto-approval is on) and carries only tool names, not arguments.

---

## Use cases

- **Audit logging** — write every tool call and its output to a JSONL file for compliance or post-mortem debugging
- **Notifications** — send a Slack or webhook alert when the agent calls `execute` or writes a file, with the specific command or path in the message
- **Guardrails** — a hook script can inspect args before the tool runs and exit non-zero to signal a block (e.g. reject any `execute` call containing `rm -rf`)
- **Latency tracking** — `tool.use` sets a start time, `tool.result` provides the end; together they give per-tool P50/P95 across sessions

No SDK changes required — four insertion points in two files within `deepagents-code`.

**Event payloads:**

```jsonc
// tool.use — fires before the tool runs
{
  "event": "tool.use",
  "tool_name": "write_file",
  "tool_id": "toolu_abc123",
  "tool_args": { "path": "src/foo.py", "content": "..." }
}

// tool.result — fires after the tool completes
{
  "event": "tool.result",
  "tool_name": "write_file",
  "tool_id": "toolu_abc123",
  "tool_args": { "path": "src/foo.py", "content": "..." },
  "tool_status": "success",
  "tool_output": "Written 42 bytes to src/foo.py"
}
```

### `libs/code/deepagents_code/textual_adapter.py`

**`tool.use`** (~line 912) — after the tool call buffer is fully parsed, before the `ToolCallMessage` widget is mounted:

```python
await dispatch_hook("tool.use", {
    "tool_name": buffer_name, "tool_id": buffer_id, "tool_args": parsed_args,
})
```

**`tool.result`** (~line 708) — after `tool_msg.set_success()` / `tool_msg.set_error()`, alongside the existing `tool.error` dispatch:

```python
await dispatch_hook("tool.result", {
    "tool_name": tool_msg._tool_name, "tool_id": tool_id,
    "tool_args": tool_msg._args, "tool_status": tool_status, "tool_output": output_str,
})
```

### `libs/code/deepagents_code/non_interactive.py`

**`tool.use`** (~line 418) — when `chunk_name` is first confirmed in `_process_content_block`. Dispatch with `tool_args: {}` initially; enrich once the buffer is fully resolved later in the same function:

```python
dispatch_hook_fire_and_forget("tool.use", {
    "tool_name": chunk_name, "tool_id": chunk_id, "tool_args": {},
})
```

**`tool.result`** (~line 463) — in `_process_message_chunk` when a `ToolMessage` is handled:

```python
dispatch_hook_fire_and_forget("tool.result", {
    "tool_name": getattr(message_obj, "name", ""),
    "tool_id": getattr(message_obj, "tool_call_id", None),
    "tool_status": getattr(message_obj, "status", "success"),
    "tool_output": str(message_obj.content)[:2000],
})
```

---

## Implementation note

Both events must ship in the same PR — `tool.use` without `tool.result` breaks timing use cases, and `tool.result` alone breaks pre-call guardrails. The existing `tool.error` event becomes a subset of `tool.result`; keep firing both for backwards compatibility.

---

## Open questions

- **Block semantics**: should a non-zero exit from a `tool.use` hook block the tool call? For v1, observe-only is a safe default — document this clearly.
- **Payload size**: `tool_args` can be large (e.g. `write_file` with a full file). A cap (e.g. 4 KB) may be needed.
- **MCP tools**: `tool.result` should fire for MCP tools since they arrive as `ToolMessage` objects through the same path — worth a smoke-test to confirm.

_No response_

NinaadRao:
Closes #3953

---

Adds two hook events — `tool.use` (fires before a tool executes) and `tool.result` (fires after) — mirroring Claude Code's PreToolUse/PostToolUse hooks. They cover audit logging, notifications, guardrails, and latency tracking. Both events are wired into the two execution surfaces — the interactive Textual TUI and the headless (`-p`) runner — so hooks fire consistently regardless of how the agent is invoked. The existing `tool.error` event is untouched and continues to fire alongside `tool.result` on failure, so existing hooks are unaffected.

**Payloads**

`tool.use` carries `tool_name`, `tool_id`, and the parsed `tool_args`. `tool.result` adds `tool_status` (`"success"` or `"error"`) and `tool_output` (the tool's returned content, truncated to a shared `HOOK_TOOL_OUTPUT_LIMIT`). The full schema and a worked example live in the `hooks` module docstring.

**Resolving streamed arguments**

Tool-call arguments arrive as JSON string fragments while the model streams. Each in-progress call is tracked in a typed `_ToolCallBuffer`, and `_parse_tool_call_args` reassembles the fragments; `tool.use` dispatches only once the arguments parse into a complete value, so hooks receive the real arguments rather than a partial or empty payload. Arguments that look structurally complete but still fail to parse are logged rather than silently dropped, so a missed hook is never invisible.

**Dispatch and delivery**

Both events are dispatched fire-and-forget on both surfaces, so a slow hook command never blocks the agent or stalls the UI. Because they run as background tasks, each surface drains any in-flight hook tasks before its event loop tears down — the headless runner before the process exits, and the TUI on app shutdown — so the final `tool.result` is never cancelled mid-flight and silently dropped. Both surfaces also emit `tool.result` for every executed tool, including a tool call whose result arrives without a rendered widget, so audit consumers see the same event stream in either mode. Rejected HITL approvals that abort a turn now emit terminal `tool.error`/`tool.result` events, so hook consumers never observe an unterminated `tool.use` for a tool that did not run.

## Release note

Deep Agents Code now emits `tool.use` and `tool.result` hook events (in addition to the existing `tool.error`), fired before and after each tool call in both the interactive TUI and headless (`-p`) modes. Rejected HITL approvals also emit terminal hook events when the turn aborts, so every `tool.use` is closed by a matching result or error event. Configure them under `~/.deepagents/hooks.json`; see the `hooks` module docstring for the payload schema.

LinkedIn: https://www.linkedin.com/in/ninaadrrao/


mdrxy:
@open-swe please review this PR

NinaadRao:
@mdrxy , not sure if Open SWE worked but I just rebased my branch to latest main

NinaadRao:
@mdrxy , I just rebased this PR to the latest main. Let me know if there needs to be some edits to the implementation. Thanks!

PR #3988, opened 2026-06-15
fix(code): hide `[SYSTEM]` interrupt notices from `/threads` prompt

mdrxy:
When a task is interrupted, dcode persists a synthetic `[SYSTEM] Task interrupted by user.` `HumanMessage` into the thread's `messages` channel. Because `_initial_prompt_from_messages` returned the first human-typed message verbatim, this notice could surface as a thread's prompt in the `/threads` selector instead of the user's real request.

The parser now skips `[SYSTEM]`-prefixed human messages and falls through to the next genuine human message, mirroring the existing `[SYSTEM]` filtering already done when rendering a loaded thread's messages.


mdrxy:
Need to e2e test this from `/threads` modal

Issue #2873, opened 2026-04-21
Summarization middleware image loss


jacobkleiman-LC:
### Submission checklist

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] repl
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox
- [ ] Other / not sure / general

### Reproduction Steps / Example Code (Python)

```python
"""Reproduce the image-loss bug in deepagents' SummarizationMiddleware.

We bypass the LLM and directly invoke the two code paths that handle older
messages during a summarization event:

  1. `_offload_to_backend`  — writes the evicted messages to a markdown file
     that is supposed to be the archival record of the conversation.
  2. `_create_summary`      — feeds the evicted messages into the summarizer
     LLM to produce the replacement summary message.

Both paths route the messages through `get_buffer_string`, which extracts
only `message.text`. Any image content blocks (base64 or URL) are silently
dropped before the backend write AND before the summarizer LLM sees them.

Run with:
    uv run python reproduce.py
"""

from __future__ import annotations

import tempfile
from pathlib import Path

from langchain_core.language_models.fake_chat_models import FakeListChatModel
from langchain_core.messages import AIMessage, HumanMessage, get_buffer_string

from deepagents.backends.filesystem import FilesystemBackend
from deepagents.middleware.summarization import _DeepAgentsSummarizationMiddleware

IMAGE_B64 = "iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNgAAIAAAUAAeImBZsAAAAASUVORK5CYII="
IMAGE_URL = "https://example.com/secret-diagram.png"
SENTINEL_TEXT = "SENTINEL_IMAGE_PAYLOAD_MARKER"


def build_multimodal_messages() -> list:
    """A realistic older-context turn: user sends an image, assistant replies."""
    user_msg = HumanMessage(
        content=[
            {"type": "text", "text": "What is in this diagram?"},
            {
                "type": "image",
                "source_type": "base64",
                "mime_type": "image/png",
[...]

PR #3990, opened 2026-06-15
fix(sdk): preserve media references in summarization archives

imnishitha:
Fixes #[2873](https://github.com/langchain-ai/deepagents/issues/2873)

---

## User story

As a user whose conversation history includes messages with media, I expect that content to remain recoverable when `SummarizationMiddleware` offloads older messages during compaction.

---

When a thread grows past the summarization threshold, `SummarizationMiddleware` keeps the conversation small by moving older messages out of the active context. It first saves those older messages to `/conversation_history/{thread_id}.md`, then replaces them in the active conversation with a concise summary and a pointer to the saved file. See the [summarization docs](https://docs.langchain.com/oss/deepagents/context-engineering#summarization) for background on this flow. Crucially, when messages are offloaded to disk, they should preserve all content originally present.

Previously, the offload step relied on LangChain's default message rendering (via `get_buffer_string()`), which only preserved text content. Non-text content blocks were silently dropped from the saved history file. The issue surfaced with images, but the underlying problem was broader: URL-based media references could be omitted, and base64 media had no durable reference once the original messages were compacted away.

---

`SummarizationMiddleware` now preserves media references during offload. For media that already has a URL, the offload step now writes the message in `get_buffer_string()`'s XML format, which keeps the media reference in the saved history file.

For base64 image content, `SummarizationMiddleware` first writes the image bytes to the backend alongside the saved conversation history, under `/conversation_history/media/{content_hash}.{ext}`. It then rewrites the original base64 content block into a normal image reference pointing at that stored file. For example, a block containing inline base64 image data becomes a reference like:

```json
{
  "type": "image",
  "url": "/conversation_history/media/{content_hash}.png"
}
```

When the message is written to the saved history file, that reference is preserved in the XML-rendered output rather than embedding the raw base64 payload.

The helper handles all three base64 shapes recognized by langchain_core:

    - Explicit base64 field: {"type": "image", "base64": "...", "mime_type": "..."}
    - Top-level data: URL: {"image/png;base64,..."}
    - OpenAI-style image_url: {"type": "image_url", "image_url": {"url": "data:..."}}

If storing one of those files fails, the saved history file includes an explicit failed-offload marker for that block instead of making it look like the media was never there:

```xml
<image error="failed_to_offload" />
```

The difference is that this file now keeps durable references for media content instead of preserving only the surrounding text.

---

## Dependency note

This PR should land with a Deep Agents LangChain dependency bump to a version that includes [langchain-ai/langchain#38171](https://github.com/langchain-ai/langchain/pull/38171). That upstream change switches `SummarizationMiddleware` summary-input serialization to XML so URL-backed media survives the summarization prompt, matching the archive-preservation behavior here.


libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -376,6 +443,7 @@ def __init__(
         _root = artifacts_root.rstrip("/")
         self._history_path_prefix = f"{_root}/conversation_history"
         self._large_tool_results_prefix = f"{_root}/large_tool_results"
+        self._images_prefix = f"{_root}/conversation_history/images"
```
mdrxy: we should generalize to all multimodal
```suggestion
        self._images_prefix = f"{_root}/conversation_history/media"
```

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -850,6 +918,97 @@ def _truncate_args(
...
+                Messages without base64 content are returned unchanged.
+        """
+        import hashlib  # noqa: PLC0415
+
+        uploads: list[tuple[str, bytes]] = []
+        path_map: dict[str, str] = {}  # sha256_hex → backend path
+
+        # First pass: collect all unique images to upload.
+        for msg in messages:
+            if not isinstance(msg.content, list):
```
mdrxy: we should be using `.content_blocks`
```suggestion
            if not isinstance(msg.content_blocks, list):
```

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -260,6 +260,73 @@ def compute_summarization_defaults(model: BaseChatModel) -> SummarizationDefault
...
 
 
+_MIME_TO_EXT: dict[str, str] = {
+    "image/png": "png",
+    "image/jpeg": "jpg",
+    "image/jpg": "jpg",
+    "image/gif": "gif",
+    "image/webp": "webp",
+    "image/svg+xml": "svg",
+}
```
mdrxy: this feels like something we should not have to define/maintain ourselves, is there a different approach here?

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -260,6 +260,73 @@ def compute_summarization_defaults(model: BaseChatModel) -> SummarizationDefault
...
+
+
+def _decode_base64_block(block: Any) -> tuple[bytes, str] | None:
+    """Decode a base64-carrying content block to raw bytes and a file extension.
+
+    Mirrors the three shapes recognized by
+    `langchain_core.messages.utils._has_base64_data`:
```
mdrxy: is there an upstream function we should be reusing?

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -850,6 +918,97 @@ def _truncate_args(
 
         return truncated_messages, modified
 
+    def _externalize_base64_images(
```
ccurme: worth thinking through how it would look to support [file content blocks](https://github.com/langchain-ai/langchain/blob/e1fab4b9c61ebbe16ee70c0d12476ef8abe2828a/libs/core/langchain_core/messages/content.py#L721) (typically PDFs)
```python
{"type": "file", "base64": "...", "mime_type": "application.pdf"}
```
or
```python
{"type": "file", "url": "..."}
```

this is a private function so not super important, but ideally whatever we have here would be easy to extend to support PDFs / audio / other media.

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -260,6 +260,73 @@ def compute_summarization_defaults(model: BaseChatModel) -> SummarizationDefault
     }
 
 
+_MIME_TO_EXT: dict[str, str] = {
```
ccurme: should we use `mimetypes.guess_extension(mime)` here? (we already use `mimetypes` elsewhere in deepagents)

ccurme (commented):
General approach looks fine to me. I have mild concerns about overloading `url` with file paths. Should we update the [prompt](https://github.com/langchain-ai/deepagents/blob/309b054fdd040adb5086408a3bcb2550ba73439e/libs/deepagents/deepagents/middleware/summarization.py#L546-L555) to clarify that?

If we haven't already, would encourage testing a simple example with real model calls and looking at traces to ensure things make sense.

libs/deepagents/deepagents/middleware/summarization.py
imnishitha: Since the XML formatter only has handlers for `image`, `audio`, and `video`, does converting unsupported types like `file` into text annotations so the reference isn't dropped from the archive seem like the right approach?

libs/deepagents/deepagents/middleware/summarization.py
imnishitha: couldn't find any upstream equivalent.

libs/deepagents/deepagents/middleware/summarization.py
ccurme: we'd probably want to update `get_buffer_string` to handle

ccurme (approved):
LGTM pending:

1. updating associated prompts (if needed) to make it clear what these tags mean to the model
```
<image url="/conversation_history/media/{hash}.png" >
```
2. we are passing these tags to create_summary. this may be a separate PR, but ideally the summarization step actually sees the images. so instead of
```
{"type": "text", "text": (turn 1, turn 2, turn, 3, turn 4 with an image, turn 5, turn 6, ...}
```
we'd have
```
{"type": "text", "text": (turn 1, turn 2, turn 3)},
{"type": "image", ...},
{"type": "text": (turn 5, turn 6)}
```
(maybe there's another way of doing this, but this is my thought on allowing the summarization model to see the actual images when generating its summary)

libs/deepagents/deepagents/middleware/summarization.py
imnishitha: noted

mdrxy:
@ccurme for the second point, I think the intended flow here is that the summary preserves the media reference, and then the model consuming the summary can call `read_file` on that path if it determines the image is necessary for context. Passing the actual image block into the summarization model seems like a larger follow-up we can handle separately.

PR #3993, opened 2026-06-15
fix(code): guard against misconfigured LangSmith tracing

mdrxy:
Misconfigured LangSmith tracing (a tracing flag enabled without resolvable credentials) no longer floods the terminal with repeated 401 authentication errors, while keyless self-hosted/proxied tracing setups remain supported.

---

LangChain enables tracing whenever a tracing flag (`LANGCHAIN_TRACING_V2`, `LANGSMITH_TRACING`, etc.) is truthy, regardless of whether credentials exist. With the flag set but no key — a common leftover in a shell profile or stale `.env` — the background tracer keeps retrying ingestion and floods `langsmith.client` 401 errors into the TUI, most visibly at the atexit flush.

At bootstrap we now:

- **Route the `langsmith`/`langchain` SDK loggers off the terminal** — to the debug log when `DEEPAGENTS_CODE_DEBUG` is set, otherwise a `NullHandler` — so their errors cannot leak onto the alternate-screen TUI via Python's last-resort stderr handler.
- **Disable tracing before any traced run starts** when a tracing flag is on but no credentials resolve, avoiding wasted background ingestion work and a slow/hanging atexit flush against the default hosted LangSmith endpoint.
- **Resolve credentials from both env vars and LangSmith profiles.** Credentials include `LANGSMITH_API_KEY` / `LANGCHAIN_API_KEY`, plus the active LangSmith profile's stored `api_key` or OAuth access/refresh tokens, so users who ran `langsmith auth login` are not falsely caught.
- **Respect self-hosted / proxied setups from env vars or profiles.** Tracing is left untouched when a custom endpoint is configured through `LANGSMITH_ENDPOINT` / `LANGCHAIN_ENDPOINT` or through the active LangSmith profile's `api_url`, because those deployments may intentionally support keyless ingestion.
- **Surface a one-shot warning toast** explaining that tracing was disabled and how to re-enable it: set `LANGSMITH_API_KEY`, or run `langsmith auth login` when that CLI is on `PATH`. The toast is best-effort; the durable record is the `logger.warning` emitted at the disable site.

LangSmith profile resolution matches the SDK behavior:

- If `LANGSMITH_CONFIG_FILE` is set, that exact config file is loaded.
- Otherwise, LangSmith auto-detects `~/.langsmith/config.json`.
- If `LANGSMITH_PROFILE` is set, that profile is used.
- Otherwise, LangSmith uses `current_profile`, falling back to `default` when present.
- Endpoint env vars are not auto-detected; if unset, LangSmith checks the active profile's `api_url`, then falls back to `https://api.smith.langchain.com`.

Regression coverage now verifies the profile-only custom endpoint case: a config with only `api_url`, tracing enabled, no API key, and no endpoint env vars keeps tracing enabled and stages no orphaned-tracing notice.


omerfarukkural:
Çok harikasınız

PR #3996, opened 2026-06-15
fix(sdk): route `BaseSandbox async` helpers through aexecute

### LINKED ISSUE #665: `BaseSandbox` async helpers should flow through `aexecute()`, not `asyncio.to_thread(sync_method)`
### Problem                                                                                                                                                                                                      
                                                                                                                                                                                                                 
Currently, `BaseSandbox`'s async helper methods (`aread`, `awrite`, `aedit`, etc.) inherit the default `asyncio.to_thread()` wrapper from `BackendProtocol`:                                                     
                                                                                                                                                                                                                 
```python                                                                                                                                                                                                        
# BackendProtocol (protocol.py lines 221-228)                                                                                                                                                                    
async def aread(self, file_path: str, offset: int = 0, limit: int = 2000) -> str:                                                                                                                                
    return await asyncio.to_thread(self.read, file_path, offset, limit)                                                                                                                                          
```                                                                                                                                                                                                              
                                                                                                                                                                                                                 
And `BaseSandbox` implements the sync methods to call `execute()`:                                                                                                                                               
                                                                                                                                                                                                                 
```python                                                                                                                                                                                                        
# BaseSandbox (sandbox.py lines 199-216)                                                                                                                                                                         
def read(self, file_path: str, offset: int = 0, limit: int = 2000) -> str:                                                                                                                                       
    cmd = _READ_COMMAND_TEMPLATE.format(...)                                                                                                                                                                     
    result = self.execute(cmd)                                                                                                                                                                                   
    # ...                                                                                                                                                                                                        
```                                                                                                                                                                                                              
                                                                                                                                                                                                                 
This creates the following flow:                                                                                                                                                                                 
                                                                                                                                                                                                                 
```                                                                                                                                                                                                              
aread() → asyncio.to_thread(self.read) → read() → self.execute()                                                                                                                                                 
```

The problem: **overriding `aexecute()` with a native async implementation has no effect on the helper methods**—they still go through the sync `execute()` path via `asyncio.to_thread`.

### Proposed Fix

`BaseSandbox` should override the async helpers to flow through `aexecute()` directly:

```python
# In BaseSandbox
async def aread(self, file_path: str, offset: int = 0, limit: int = 2000) -> str:
    cmd = _READ_COMMAND_TEMPLATE.format(file_path=file_path, offset=offset, limit=limit)
    result = await self.aexecute(cmd)  # ← Use aexecute, not execute

    output = result.output.rstrip()
    if result.exit_code != 0 or "Error: File not found" in output:
        return f"Error: File '{file_path}' not found"
    return output
```

This creates the correct flow:

```
aread() → await self.aexecute(cmd)
  → [default] asyncio.to_thread(self.execute)  # Works exactly as before
  → [overridden] native async implementation    # Now works correctly!
```

### Why This Matters

For backends wrapping SDKs with native async support (e.g., Daytona's `AsyncSandbox`, or any HTTP-based sandbox with `aiohttp`), the current design requires overriding **9 methods** to get native async behavio
r:

- `aexecute`, `aupload_files`, `adownload_files` (core)
- `aread`, `awrite`, `aedit`, `agrep_raw`, `aglob_info`, `als_info` (helpers)

With this fix, implementers would only need to override **3 methods** (`aexecute`, `aupload_files`, `adownload_files`), and all helpers automatically use the native async path.

### Backward Compatibility

This change is fully backward compatible:

| Scenario | Current Behavior | New Behavior |
|----------|-----------------|--------------|
| No overrides | `aread` → `to_thread(read)` → `execute` | `aread` → `aexecute` → `to_thread(execute)` |
| Override `execute()` only | Works | Works (unchanged) |
| Override `aexecute()` | Helpers ignore it | Helpers use it correctly |

The only difference is an extra function call in the default path (`aexecute` → `to_thread(execute)` instead of `to_thread(read)` → `execute`), which is negligible.

### Implementation

The fix requires adding ~6 async method overrides in `BaseSandbox`, each mirroring the sync implementation but calling `await self.aexecute(cmd)` instead of `self.execute(cmd)`. I'm happy to submit a PR if this direction makes sense.

imnishitha:
Fixes #665

## Problem

`BaseSandbox`'s async helpers (`als`, `aread`, `agrep`, `aglob`, `awrite`, `aedit`) previously wrapped the corresponding sync methods in `asyncio.to_thread`, meaning overriding `aexecute` with a native async implementation (e.g. wrapping an async SDK like Daytona's `AsyncSandbox`) had no effect on those helpers, they still ran the sync path in a thread pool. Subclasses were forced to override all 9 file-operation methods to get native async behavior.

## Fix

This PR overrides all six async helpers in `BaseSandbox` to call `aexecute`/`aupload_files` directly. Subclasses now only need to override those three core methods. To avoid duplicating command-building and output-parsing logic between sync and async paths, each method's pure logic was factored into private helpers shared by both. Sync methods and their behavior are completely unchanged.

## Tests

Added a `NativeAsyncSandbox` test fixture where `exes `RuntimeError`, if any async helper falls backto the sync path, the test fails immediately. Seven new `@pytest.mark.asyncio` tests cover each overridden method (`als`, `aread`, `agrep`, `aglob`, `awrite`, inline `aedit`, and large-payload `aedit` via upload), asserting that `aexecute`/`aupload_files` are
called the expected number of times and results are tests continue to pass.


libs/deepagents/tests/unit_tests/backends/test_sandbox_backend.py
```diff
@@ -1326,3 +1326,140 @@ def test_sandbox_edit_inline_permission_denied() -> None:
...
+
+    async def aupload_files(self, files: list[tuple[str, bytes]]) -> list[FileUploadResponse]:
+        self._aupload_calls.append(files)
+        return [FileUploadResponse(path=f[0], error=None) for f in files]
+
+    def download_files(self, paths: list[str]) -> list[FileDownloadResponse]:
+        raise RuntimeError("sync download_files() must not be called from async helpers")
+
+
+@pytest.mark.asyncio
```
sydney-runkle: don't think we need these, should be auto?

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -756,97 +865,80 @@ def _map_edit_error(error: str, file_path: str, old_string: str) -> EditResult:
...
             return GrepResult(matches=[])
-
-        # Parse grep output into GrepMatch objects
         matches: list[GrepMatch] = []
         parse_error: str | None = None
         for line in output.split("\n"):
-            # Format is: path\0line_number:text
             parts = line.split("\0", 1)
-            if len(parts) != 2:  # noqa: PLR2004  # Grep output field count
+            if len(parts) != 2:  # noqa: PLR2004
```
sydney-runkle: why are we removing these informative comments?

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -639,43 +680,66 @@ def edit(
...
+        file_path: str,
+        old_string: str,
+        new_string: str,
+        replace_all: bool,  # noqa: FBT001
+    ) -> EditResult:
+        """Async version of `_edit_inline`, delegating to `aexecute`."""
+        result = await self.aexecute(self._build_edit_inline_cmd(file_path, old_string, new_string, replace_all))
+        return self._parse_edit_output(result.output, file_path, old_string)
+
+    def _build_edit_tmpfile_cmd(self, file_path: str, old_tmp: str, new_tmp: str, replace_all: bool) -> str:  # noqa: FBT001
```
sydney-runkle: what's this ignore for?

sydney-runkle (commented):
just an initial review, seems like there's some unrelated changes?
are we losing agrep timeout functionality?

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -829,24 +909,47 @@ def grep(
...
+        pattern: str,
+        path: str | None = None,
+        glob: str | None = None,
+    ) -> GrepResult:
+        """Search file contents for a literal string using `grep -F`.
+
+        Args:
+            pattern: Literal string to search for (not a regex).
+            path: Directory or file to search in.
+
```
eyurtsev: nit: i think we don't need a new line here

```suggestion
```

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -639,43 +680,66 @@ def edit(
...
-                "replace_all": replace_all,
-            }
-        )
+        """Async version of `edit`, delegating to `aexecute` and `aupload_files`."""
+        payload_size = len(old_string.encode("utf-8")) + len(new_string.encode("utf-8"))
+        if payload_size <= _EDIT_INLINE_MAX_BYTES:
+            return await self._aedit_inline(file_path, old_string, new_string, replace_all)
+        return await self._aedit_via_upload(file_path, old_string, new_string, replace_all)
+
+    def _build_edit_inline_cmd(self, file_path: str, old_string: str, new_string: str, replace_all: bool) -> str:  # noqa: FBT001
```
eyurtsev: Let's remove the ignores in new functions, this is for boolean traps so can create a named variable

libs/deepagents/deepagents/backends/protocol.py
```diff
@@ -567,7 +567,8 @@ def edit(
         file_path: str,
         old_string: str,
         new_string: str,
-        replace_all: bool = False,  # noqa: FBT001, FBT002
+        *,
+        replace_all: bool = False,
```
sydney-runkle: this is breaking?

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -635,55 +790,55 @@ def edit(
...
 
-    def _edit_inline(
+    async def aedit(
         self,
         file_path: str,
         old_string: str,
         new_string: str,
-        replace_all: bool,  # noqa: FBT001
+        *,
+        replace_all: bool = False,
```
sydney-runkle: also breaking
we need to be really careful about changes in the codebase that could affect downstream users

eyurtsev (approved):

PR #4004, opened 2026-06-16
feat(code): add provider-specific auth guidance

mdrxy:
Auth setup now leads users through getting the right provider API key before asking them to paste it, while keeping endpoint and environment-variable details behind an Advanced disclosure. Provider configuration can also supply human-readable labels and key-management URLs so custom providers read cleanly in `/auth`.

## Changes
- Added provider-specific display names and API-key links for the auth flow, including custom `display_name` and `api_key_url` metadata through `ProviderConfig`.
- Reworked `AuthPromptScreen` to show key acquisition guidance first, call out active environment-sourced credentials, and distinguish scoped `DEEPAGENTS_CODE_` variables from shared provider variables.
- Moved base URL overrides and env-var precedence details into an Advanced section toggled by F2 or click, with Unicode/ASCII disclosure glyphs.
- Updated auth manager provider labels to use `_provider_display_name` instead of raw provider config keys.
- Added pointer cursor behavior for clickable links and disclosure rows across auth, Codex login, MCP login, and thread selector modals.
- Renamed environment credential badges from `[env: ...]` to `[env set: ...]` for clearer status labels.


mdrxy:
we should make model switcher headers not display api key name

PR #4072, opened 2026-06-18
refactor(sdk): use `RemoveMessage` sentinel in FS and patch tool calls middleware

sydney-runkle:
`FilesystemMiddleware` and `PatchToolCallsMiddleware` previously emitted `Overwrite(...)` to atomically replace the `messages` channel. With the local `_messages_delta_reducer` and `DeltaChannel`, the same semantics are expressible as a `RemoveMessage(REMOVE_ALL_MESSAGES)` sentinel followed by the desired message list — which is the standard `add_messages`-compatible idiom and avoids depending on `Overwrite`.

Three call sites change:

- `PatchToolCallsMiddleware.before_agent` now returns `[RemoveMessage(id=REMOVE_ALL_MESSAGES), *patched_messages]`.
- `FilesystemMiddleware._apply_eviction_and_truncate` emits the same sentinel-prefixed list when tagging a newly evicted `HumanMessage`, preserving the "atomic replace so DeltaChannel replay doesn't duplicate the `id=None` write" property.
- `FilesystemMiddleware._intercept_large_tool_result` (and its async sibling) now detect a leading `REMOVE_ALL_MESSAGES` sentinel on incoming tool `Command` updates instead of an `Overwrite` wrapper, and re-emit the sentinel after eviction processing.

`Overwrite` is no longer imported by either middleware. Unit tests covering both middlewares are updated to assert the new sentinel-prefixed list shape.

### DIFF
--- a/libs/deepagents/tests/unit_tests/test_eviction_replay.py
+++ b/libs/deepagents/tests/unit_tests/test_eviction_replay.py
@@ -20,16 +20,16 @@ class _FakeModel:
 
     name = "fake"
 
-    def bind_tools(self, *args, **kwargs):
+    def bind_tools(self, *args, **kwargs):  # noqa: ANN002, ANN003
         return self
 
-    def with_structured_output(self, *args, **kwargs):
+    def with_structured_output(self, *args, **kwargs):  # noqa: ANN002, ANN003
         return self
 
-    def invoke(self, messages, config=None, **kw):
+    def invoke(self, messages, config=None, **kw):  # noqa: ANN003
         return AIMessage(content="ok")
 
-    async def ainvoke(self, messages, config=None, **kw):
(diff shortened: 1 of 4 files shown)


ccurme (approved):

libs/deepagents/tests/unit_tests/test_eviction_replay.py
```diff
@@ -0,0 +1,83 @@
...
+from langchain.agents import create_agent
+from langchain_core.messages import AIMessage, HumanMessage
+from langgraph.checkpoint.memory import InMemorySaver
+from langgraph.store.memory import InMemoryStore
+
+from deepagents.backends import StoreBackend
+from deepagents.middleware.filesystem import FilesystemMiddleware
+
+
+class _FakeModel:
```
ccurme, line 18: why not update test_end_to_end.py and use the same fake chat models there

PR #4093, opened 2026-06-19
feat(sdk): extract video frames via PyAV on read_file

jkennedyvz:
Revives and modernizes the lost `origin/jk/video-frame-extraction-read-file` work (baseline `d3cecf9b`).

`read_file` on `.mp4` / `.mov` / `.webm` / `.mkv` now routes through a PyAV-based extractor that reinterprets the existing `offset`/`limit` args as seconds: `offset` = seconds to skip, `limit` = seconds of source to sample. Frames are sampled at a constructor-defined rate (`FilesystemMiddleware(video_sampling_rate=...)`, default 0.5 fps). The tool result is interleaved text+image content blocks with a leading row describing the window, and raw video bytes from the backend no longer leak to the model.

PyAV (av>=17) is an **optional `[video]` extra** pulled in via `uv add "deepagents[video]"`; core users without it see a clear install hint. The PyAV import is lazy so the core package stays lightweight when `[video]` isn't installed. Per the AGENTS.md scope rules and Conventional Commits, the work is split across three commits:

- `feat(deepagents): extract sampled frames from video reads` — `_video.py`, `filesystem.py` dispatch + `_handle_video_read` / `_emit_binary_block` helpers.
- `test(deepagents): cover video read_file dispatch and snapshots` — constructor + dispatch tests + regenerated smoke snapshots.
- `build(deepagents): add [video] extra and document it` — `PROJECT.OPTIONAL_DEPENDENCIES video = ["av>=17.0.0"]`, dev toolchain adds av + pillow for `ty`, README documents the new extra.

Why PyAV and not imageio-ffmpeg: PyAV is the more modern binding — the FFmpeg-python union is mostly being phased out, and PyAV handles seek/scan operations straightforwardly.

Why a deploy knob, not an agent-facing kwarg: `video_sampling_rate` is constructor-scoped (`FilesystemMiddleware(video_sampling_rate=...)`) so the application decides the right sample density for its domain (default 0.5 = one frame every two seconds) rather than letting the agent overspend on context.

Closes the revival ask — branch `jk/video-frame-extraction-read-file` had two commits (`899f1ade` + `0dec6b99`) that were never merged, and the original branch's first commit (`origin/jk/video-frame-extraction`) added a separate middleware that took agent-facing kwargs. This modernized form factors PyAV behind the same `offset`/`limit` paged signature the rest of `read_file` already uses, folds the FPS knob into `FilesystemMiddleware.__init__`, and ships behind a runtime-optional extra so we don't push a 17 MB binary into the core install.


jkennedyvz:
Superseded by #4094, which uses the same branch name from `langchain-ai/deepagents` as the PR head and targets `main`.

PR #4097, opened 2026-06-20
feat(talon): add Telegram channel adapter, CLI wiring, and offset persistence

jkennedyvz:
Implements a Telegram Bot API channel for Talon and updates the shared channel policy so Telegram and WhatsApp use the same exposure rules.

---

## Telegram channel

Adds a pure-Python Telegram channel adapter that uses the Bot API directly rather than a local bridge. It supports long polling, persisted update offsets, inbound media downloads, outbound text and media, typing indicators, message edits, and plain-text message delivery.

Telegram configuration is wired through `DEEPAGENTS_TALON_TELEGRAM_*`, with `TELEGRAM_BOT_TOKEN` accepted as a convenience alias. The host can run Telegram, WhatsApp, or both channels against the same Talon runtime.

## Shared Channel Exposure

Moves provider-specific exposure parsing into the shared channel abstraction via `ChannelExposureEnv` and `channel_exposure_from_env`. Telegram and WhatsApp now share the same handling for `self`, `allowlist`, `open`, mention patterns, open-exposure acknowledgement, and operator identity.

`ChannelExposure` now supports multiple operator IDs while preserving the existing single `operator_id` API. The existing `DEEPAGENTS_TALON_TELEGRAM_OPERATOR_ID` and `DEEPAGENTS_TALON_WHATSAPP_OPERATOR_ID` env vars accept comma-separated IDs, so `self` exposure can allow more than one trusted operator without duplicating logic in each provider.

## Media And Runtime Handling

Adds shared media helpers for inbound media metadata, outbound media validation, and configured media-size caps. Telegram and WhatsApp both use the common cap handling, while WhatsApp keeps a lower provider-specific cap because its bridge materializes downloads in memory.

The host and runtime wiring now preserve channel metadata, support channel-specific approval handling, and allow Telegram channel posts through the same serialized agent execution path as other Talon channels.

## Example App

Refreshes the example app from a WhatsApp-only example into a general Talon Docker example. The example now documents running WhatsApp, Telegram, or both; adds placeholder-only env settings for Telegram allowlists and channel toggles; keeps local secrets in an ignored `.env`; and updates the Docker setup so the packaged WhatsApp bridge and Talon runtime work together from the same container.

## Review Notes

The exposure changes are intentionally centralized in the shared channel layer. Provider modules should only declare their env var names and provider-specific requirements, such as Telegram requiring an operator ID for `self` exposure.

The example app changes are mostly naming and configuration surface updates so it reflects the new multi-channel Talon host rather than only WhatsApp.


libs/talon/deepagents_talon/channels/base.py
```diff
@@ -37,7 +47,26 @@ class ExposureMode(StrEnum):
...
+    """Raised when channel media cannot be handled safely."""
+
+
+@dataclass(frozen=True, slots=True)
+class ChannelExposureEnv:
+    """Environment variable prefix and options for channel exposure policy.
+
+    Args:
+        provider: Human-readable provider name used in error messages.
+        env_prefix: Prefix for exposure env vars (e.g. ``DEEPAGENTS_TALON_TELEGRAM``).
```
mdrxy: nit

```suggestion
        env_prefix: Prefix for exposure env vars (e.g. `DEEPAGENTS_TALON_TELEGRAM`).
```

libs/talon/deepagents_talon/channels/base.py
```diff
@@ -37,7 +47,26 @@ class ExposureMode(StrEnum):
...
 class ChannelMediaError(ValueError):
-    """Raised when outbound media cannot be sent safely."""
+    """Raised when channel media cannot be handled safely."""
+
+
+@dataclass(frozen=True, slots=True)
+class ChannelExposureEnv:
+    """Environment variable prefix and options for channel exposure policy.
+
+    Args:
```
mdrxy: i'd move these to docstrings under the attribute definitions themselves (so as to not break ref doc parsers)

mdrxy:
what's the user story here?

mdrxy (approved):

PR #4108, opened 2026-06-22
feat(sdk): add Bedrock prompt caching middleware

aolsenjazz:
Bedrock models now receive `BedrockPromptCachingMiddleware` automatically when `langchain-aws` is installed.

Adds Python SDK parity with [deepagentsjs](https://github.com/langchain-ai/deepagentsjs/pull/611) by wiring Bedrock prompt caching into main and subagent middleware stacks for AWS Bedrock models.

Docs PR: https://github.com/langchain-ai/docs/pull/4587

Made by [Open SWE](https://openswe.vercel.app/agents/696a9011-2123-6396-5992-265911c8d2af)

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -33,7 +33,7 @@ from deepagents._excluded_middleware import (
     _verify_excluded_middleware_coverage,
 )
 from deepagents._messages_reducer import _messages_delta_reducer
-from deepagents._models import is_bedrock_model, resolve_model
+from deepagents._models import resolve_model
 from deepagents._tools import _apply_tool_description_overrides
 from deepagents._version import __version__
 from deepagents.backends import StateBackend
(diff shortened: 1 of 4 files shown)


mdrxy (approved):

libs/deepagents/deepagents/graph.py
```diff
@@ -349,6 +376,8 @@ def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly
             - `_ToolExclusionMiddleware` (if profile has `excluded_tools`)
             - [`AnthropicPromptCachingMiddleware`][langchain_anthropic.middleware.AnthropicPromptCachingMiddleware] (unconditional; no-ops for
                 non-Anthropic models)
+            - `BedrockPromptCachingMiddleware` for AWS Bedrock models when
+                `langchain-aws` is installed
```
sydney-runkle: api link

libs/deepagents/deepagents/graph.py
```diff
@@ -185,6 +186,32 @@ def get_default_model() -> ChatAnthropic:
...
+def _append_prompt_caching_middleware(
+    middleware: list[AgentMiddleware[Any, Any, Any]],
+    model: BaseChatModel,
+) -> None:
+    """Append provider-specific prompt caching middleware."""
+    middleware.append(AnthropicPromptCachingMiddleware(unsupported_model_behavior="ignore"))
+    if is_bedrock_model(model):
+        bedrock_middleware = _create_bedrock_prompt_caching_middleware()
+        if bedrock_middleware is not None:
+            middleware.append(bedrock_middleware)
```
sydney-runkle: this doesn't make sense right, model can change dynamically at runtime? so we'd want to support that?

sydney-runkle (commented):
some questions!

libs/deepagents/deepagents/graph.py
aolsenjazz: FWIW, I implemented in JS with the check to match what already existed w/Anthropic middleware - a model check before adding

If model can change at runtime I'd see why we would want to default-include this. I remember folks were saying that traces were noisy due to the default middleware stack - is that a concern?

sydney-runkle (approved):

PR #4109, opened 2026-06-22
feat(sdk): allow ```write_file``` to overwrite existing files

### LINKED ISSUE #3731: Agents cannot efficiently rewrite existing files using `write_file`
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
The agent typically does:

read_file README.md
edit_file(
    old_string=<entire file>,
    new_string=<new content>
)

instead of:

write_file(
    file_path="README.md",
    content=<new content>
)
```

```shell

```

### Description

When asked to completely rewrite an existing file, agents currently tend to use edit_file and pass the entire file contents as old_string, replacing the file through a large edit operation.

Using write_file would be a more natural fit for full-file rewrites, but write_file currently fails when the target file already exists because it only supports creating new files.

As a result:
Full-file rewrites require the model to read and resend the entire existing file.
Tool calls become unnecessarily large and token-intensive.

Expected behavior:
Agents should have an efficient way to replace the contents of an existing file without needing to include the entire previous file contents in the tool call.

### Environment / System Info

_No response_

imnishitha:
Fixes #3731 

`write_file` now creates or overwrites a file rather than failing when the target already exists. This removes the need for agents to use `edit_file` with the entire file as `old_string` for full rewrites, reducing unnecessary token usage. Tool description updated to guide the agent: use `write_file` for full rewrites, `edit_file` for targeted changes.

### DIFF
--- a/libs/deepagents/deepagents/backends/filesystem.py
+++ b/libs/deepagents/deepagents/backends/filesystem.py
@@ -569,9 +569,6 @@ class FilesystemBackend(BackendProtocol):
         try:
             if not resolved_path.exists() and not resolved_path.is_symlink():
                 return DeleteResult(error=f"Error: '{file_path}' not found")
-            # Check is_symlink() before is_dir(): is_dir() follows symlinks, so a
-            # symlink pointing at a directory would otherwise be recursed into and
-            # delete the target's contents. Removing the link itself is correct.
             if resolved_path.is_symlink():
                 resolved_path.unlink()
             elif resolved_path.is_dir():
--- a/libs/deepagents/deepagents/backends/filesystem.py
+++ b/libs/deepagents/deepagents/backends/filesystem.py
@@ -546,29 +546,41 @@ class FilesystemBackend(BackendProtocol):
             return EditResult(error=f"Error editing file '{file_path}': {e}")
 
     def delete(self, file_path: str) -> DeleteResult:
-        """Delete a file from the filesystem.
+        """Delete a file or directory from the filesystem.
+
+        Files are unlinked. Directories are removed recursively along with all
+        of their contents. Symlinks are removed as links and never followed into
+        their target (so deleting a symlink to a directory removes only the link).
 
         Args:
-            file_path: Path to the file to delete.
+            file_path: Path to the file or directory to delete.
 
         Returns:
             `DeleteResult` with the deleted path on success, or an error if the
-                file does not exist, is a directory, or removal fails.
+                path does not exist or removal fails. A recursive directory
+                removal may delete some entries before failing partway (for
+                example when a nested entry is not writable).
         """
         try:
             resolved_path = self._resolve_path(file_path)
         except (OSError, RuntimeError) as e:
-            return DeleteResult(error=f"Error deleting file '{file_path}': {e}")
+            return DeleteResult(error=f"Error deleting '{file_path}': {e}")
 
         try:
-            if resolved_path.is_dir():
-                return DeleteResult(error=f"Error: '{file_path}' is a directory, not a file")
             if not resolved_path.exists() and not resolved_path.is_symlink():
-                return DeleteResult(error=f"Error: File '{file_path}' not found")
-            resolved_path.unlink()
+                return DeleteResult(error=f"Error: '{file_path}' not found")
+            # Check is_symlink() before is_dir(): is_dir() follows symlinks, so a
+            # symlink pointing at a directory would otherwise be recursed into and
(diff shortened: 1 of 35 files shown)


libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -446,26 +446,21 @@ def write(
         file_path: str,
         content: str,
```
eyurtsev, line 447: write has two modes:

1. overwrite
2. append

This implements (1) which is probably correct

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -462,11 +463,11 @@ class ExecuteSchema(BaseModel):
 - Only use emojis if the user explicitly requests it."""
 
 
-WRITE_FILE_TOOL_DESCRIPTION = """Writes to a new file in the filesystem.
+WRITE_FILE_TOOL_DESCRIPTION = """Writes content to a file. Creates the file if it does not exist; replaces it entirely if it does.
 
 Usage:
-- The write_file tool will create the a new file.
-- Prefer to edit existing files (with the edit_file tool) over creating new ones when possible.
```
ccurme, line 469: why this change to the guidance on `edit_file`?

changing prompts has a cost (e.g., can result in subtle behavior changes out from under users). not clear this one is needed.

libs/deepagents/tests/integration_tests/test_filesystem_middleware.py
```diff
@@ -346,7 +346,7 @@ def test_write_file_longterm(self):
         assert isinstance(content, str), f"Expected str content, got {type(content)}"
         assert "fiery" in content or "Fiery" in content
 
-    def test_write_file_fail_already_exists_in_local(self):
+    def test_write_file_overwrites_existing_in_local(self):
```
ccurme, line 349: I'm not sure these integration tests ever run fwiw

libs/deepagents/tests/unit_tests/backends/test_sandbox_backend.py
```diff
@@ -537,17 +537,17 @@ def test_sandbox_write_with_special_content() -> None:
     assert sandbox._uploaded[0][1] == content.encode("utf-8")
 
 
-def test_sandbox_write_returns_error_on_existing_file() -> None:
-    """Test that write() returns an error when the check command fails."""
+def test_sandbox_write_returns_error_on_preflight_failure() -> None:
```
ccurme, line 554: Why aren't we testing that we overwrite successfully, like we do for StoreBackend in test_store_backend.py?

libs/evals/tests/evals/test_file_operations.py
```diff
@@ -49,6 +50,89 @@ def test_read_file_seeded_state_backend_file(model: BaseChatModel) -> None:
     )
 
 
+@pytest.mark.eval_tier("baseline")
+@pytest.mark.eval_category("file_operations")
+@pytest.mark.langsmith
+def test_write_file_overwrites_existing(model: BaseChatModel) -> None:
```
ccurme, line 56: these pass for me

ccurme (dismissed):

libs/deepagents/deepagents/backends/store.py
```diff
@@ -596,20 +596,19 @@ def write(
...
         namespace = self._get_namespace()
 
-        # Check if file exists
         existing = store.get(namespace, file_path)
         if existing is not None:
-            return WriteResult(error=f"Cannot write to {file_path} because it already exists. Read and then make an edit, or write to a new path.")
-
-        # Create new file
-        file_data = create_file_data(content)
+            existing_file_data = self._convert_store_item_to_file_data(existing)
```
eyurtsev, line 608: Old behavior was inefficient in terms of network requests. Current behavior is consistent with old behavior. We can revisit whether we want to preserve creation timestamp. (The cost of this can be 10-100 ms of latency.)

libs/deepagents/deepagents/backends/sandbox.py
eyurtsev, line 584: we should see if we can eliminate _write_preflight path entirely in a follow up PR.

Now that there's no file check existence, the responsibility for ensuring the parent directory exists should lie entirely with the `upload_files` API.

eyurtsev (approved):

PR #4193, opened 2026-06-24
feat(code): add LangSmith tracing config to `/auth`

mdrxy:
LangSmith tracing can now be configured directly from `/auth` (and `dcode auth set langsmith`), including enabling tracing, a temporary opt-out via `DEEPAGENTS_CODE_LANGSMITH_TRACING=false`, and a custom project name.

---

Wires LangSmith tracing into the `/auth` surface as a configurable service, mirroring how Tavily is already handled. Storing a LangSmith key via `/auth` persists it to the credential store, bridges it onto `LANGSMITH_API_KEY`, and opts the user into tracing — so configuring tracing no longer requires exporting environment variables before launch.

Key behaviors:
- A stored key turns tracing on by default. An explicit falsy tracing flag (most simply `DEEPAGENTS_CODE_LANGSMITH_TRACING=false`) is honored as a non-destructive, session-scoped opt-out — pause tracing without deleting the key. The TUI prompt surfaces this path.
- An optional project name (advanced field in the TUI prompt, `--project` on `dcode auth set`) routes traces to a custom `LANGSMITH_PROJECT`; the default stays `deepagents-code`.
- Services are now surfaced in `dcode auth list`/`status` (this also fixes a pre-existing gap where Tavily was invisible to the CLI).
- Bootstrap reordered so a `/auth`-stored key is applied before orphaned-tracing detection, keeping tracing alive instead of being disabled.

Made by [Open SWE](https://openswe.vercel.app/agents/2ba1485c-7c70-59b6-82bf-18302c214c6c)

### DIFF
--- a/libs/code/COMMANDS.md
+++ b/libs/code/COMMANDS.md
@@ -12,7 +12,7 @@ aliases, descriptions, visibility, or hidden-command metadata.
 | Command | Aliases | Description |
 | --- | --- | --- |
 | `/agents` |  | Browse and switch between available agents |
-| `/auth` | `/connect` | Connect and manage model provider credentials |
+| `/auth` | `/connect` | Connect and manage provider and service credentials |
 | `/auto-update` |  | Turn automatic updates on or off |
 | `/changelog` |  | Open the changelog in a browser |
 | `/clear` |  | Clear the chat and start a new thread |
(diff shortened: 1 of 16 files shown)


mdrxy:
* need EU option checkbox
* need to make part of onboarding

PR #4230, opened 2026-06-24
feat(sdk): reduce round trips when offloading large tool results with `sandbox.execute`

ccurme:
When the `execute` tool receives a large tool result and you're using a sandbox:
1. The result is returned to the agent server
2. The agent server offloads it back to the sandbox (assuming no re-routing of large tool results using a `CompositeBackend`)

Here we remove the round trip by wrapping **all** `execute` commands with additional shell commands inside the `execute` tool. The additional shell commands:
- Redirect the command's combined stdout/stderr to a file in the sandbox (`> "$__da_f" 2>&1`)
- Measures the captured size (`wc -c`)
- If under eviction threshold, prints the full output inline and deletes the file (no offloading)
- If above eviction threshold, leaves the file at `artifacts_root/large_tool_results/tool_call_id` and prints the content preview
- Emits a one-line machine-readable header that the agent server parses to rebuilt the ExecuteResponse.

This only engages when the backend is a BaseSandbox and the offload path routes to the same backend (so no changes if we're routing large tool results away with CompositeBackend).

We also add a flag `enable_capture_offload` to BaseSandbox as an escape hatch.

Also have a standard integration test here. Passes with LangsmithSandbox although I haven't got creds yet for any others: https://github.com/langchain-ai/langchain/compare/cc/standard-tests-sandboxes


libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -583,6 +583,128 @@ def _build_edit_tmpfile_cmd(file_path: str, old_tmp: str, new_tmp: str, *, repla
...
+"""
+
+# The captured stream is piped through `head -c` so the on-disk file can never
+# exceed the cap regardless of how the command behaves. Because that puts the
+# command in a pipeline, its real exit code is recovered from a sidecar file
+# rather than `$?` (which would be `head`'s). The command runs in a subshell so a
+# command `exit` cannot abort the wrapper, and `eval` preserves the backend's own
+# shell/env. The command is embedded via a quoted heredoc with a random delimiter
+# to avoid shell-quoting issues; the (internal, sanitized) path is shell-quoted.
+_EXECUTE_CAPTURE_CMD_TEMPLATE = """# ===== deepagents capture-at-source offload (auto-generated wrapper) =====
```
ccurme, line 612: Breakdown of what's happening here:

> - Redirect the command's combined stdout/stderr to a file in the sandbox (> "$__da_f" 2>&1) instead of streaming it back, so the full payload never leaves the sandbox.
> - Embed the requested command verbatim via a quoted heredoc (cat <<'DELIM' … DELIM) and run it in a subshell via eval (( eval "$__da_cmd" )). The heredoc avoids any shell-quoting pitfalls; the subshell+eval runs the command in the sandbox's own shell/env and ensures a stray exit in the user's command can't abort the wrapper.
> - Pipe the captured stream through head -c <max> (default 10 MiB) so the on-disk file can never exceed a hard cap — this bounds sandbox disk usage and restores backpressure (the writer gets SIGPIPE) for runaway output like yes.
> - Recover the command's real exit code from a sidecar file (echo "$?" > "$__da_f.ec"). The capture pipe would otherwise leave $? as head's exit code, not the command's.
> - Measure the captured size (wc -c) and branch:
>   - Under the eviction threshold → print the full output inline and delete the file. Indistinguishable from today's small-result behavior.
>   - Over the threshold → leave the file at <artifacts_root>/large_tool_results/<tool_call_id> and print only a head/tail preview (whole lines, byte-bounded) with a [N lines truncated] marker.
> - Emit a one-line machine-readable header — <sentinel> <exit_code> <offloaded> <capped> — that the agent server parses to rebuild the ExecuteResponse and decide between returning the output inline vs. a preview + a read_file pointer.

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -583,6 +583,128 @@ def _build_edit_tmpfile_cmd(file_path: str, old_tmp: str, new_tmp: str, *, repla
...
+  if [ "$__da_omitted" -gt 0 ]; then
+    head -c __HEAD__ "$__da_f" | head -n __HEADLINES__
+    printf '... [%s lines truncated] ...\\n' "$__da_omitted"
+    tail -c __TAIL__ "$__da_f" | tail -n __TAILLINES__
+  else
+    head -c $((__HEAD__ + __TAIL__)) "$__da_f"
+  fi
+fi
+"""
+"""Pure POSIX sh wrapper for capture-at-source `execute`. See the comment above."""
```
eyurtsev: nit: maybe `#` instead of `"""` to delineate from the code?

hntrl (approved):

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -580,6 +580,130 @@ def _build_edit_tmpfile_cmd(file_path: str, old_tmp: str, new_tmp: str, *, repla
...
+
+def _build_capture_execute_cmd(command: str, capture_path: str, *, inline_budget: int) -> str:
+    """Build the capture-at-source wrapper command for `execute`.
+
+    `inline_budget` is the byte threshold at or below which output is returned
+    inline; above it the output is left at `capture_path` and only a head/tail
+    preview is returned. Captured output is hard-capped at
+    `_EXECUTE_CAPTURE_MAX_BYTES`; beyond that it is truncated and flagged.
+    """
+    delim = "__DEEPAGENTS_CMD_" + base64.b32encode(os.urandom(10)).decode("ascii").rstrip("=") + "__"
```
eyurtsev: Let's check that `command` doesn't contain `delim`

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -580,6 +580,130 @@ def _build_edit_tmpfile_cmd(file_path: str, old_tmp: str, new_tmp: str, *, repla
...
+
+    `offloaded` is kept separate from `ExecuteResponse` because it describes the
+    capture mechanism, not the command result (an ordinary `execute` never sets
+    it). Falls back to `(False, raw output)` when the meta line is absent — e.g.
+    if the backend truncated transport; the caller must not re-run the command in
+    that case. `ExecuteResponse.truncated` is set when the captured output hit the
+    size cap (the saved file is incomplete) or `backend_truncated` is passed
+    through from the underlying `execute`.
+    """
+    first, _, body = output.partition("\n")
```
eyurtsev, line 711: Could you document what we're parsing here more explicitly? e.g., it's hard to tell why parts should have len(4)

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1818,6 +1825,65 @@ async def async_grep(
...
 
+    def _capture_path_if_local(self, resolved_backend: BackendProtocol, tool_call_id: str | None) -> str | None:
+        """Return the sandbox-local offload path for capture-at-source, or `None` to skip.
+
+        Capture-at-source writes output to a literal path via the sandbox shell
+        and later reads it back through the backend, so it is only sound when
+        `execute()` and `read_file` share one filesystem at that path. Only
+        `BaseSandbox` guarantees this, so the optimization is gated on it (a
+        stub or host-shell backend falls back to inline execute plus generic
+        eviction). Also returns `None` when the sandbox opts out via
```
eyurtsev: Could we follow Returns: style rather than prose if we're documenting a return?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1818,6 +1825,65 @@ async def async_grep(
             args_schema=GrepSchema,
         )
 
+    def _capture_path_if_local(self, resolved_backend: BackendProtocol, tool_call_id: str | None) -> str | None:
+        """Return the sandbox-local offload path for capture-at-source, or `None` to skip.
+
+        Capture-at-source writes output to a literal path via the sandbox shell
+        and later reads it back through the backend, so it is only sound when
```
eyurtsev: nit: "it is only sound" -> paraphrase to technical language?

eyurtsev (approved):
lgtm, but maybe we could push more code into sandbox and out of file system (perhaps a method on base sandbox that accepts max output size?)

libs/deepagents/deepagents/backends/sandbox.py
```diff
@@ -621,6 +782,56 @@ def execute(
...
+        """Run `command`, offloading large output to a file in the sandbox.
+
+        Captures the command's combined output: returned inline when it is at or
+        below `max_inline_bytes`, otherwise left at `capture_path` (so the caller
+        can surface a `read_file` pointer) with only a head/tail preview returned.
+        Captured output is hard-capped at `max_capture_bytes` (default
+        `_EXECUTE_CAPTURE_MAX_BYTES`) without killing the command, so the exit
+        code is preserved.
+
+        Returns `(offloaded, ExecuteResponse)`, where `offloaded` is `True` when
```
eyurtsev: This should be `Returns:` as we're not using prose based doc-strings

mdrxy (dismissed):
doctring

PR #4251, opened 2026-06-25
feat(sdk): allow users to override default middleware by name

### LINKED ISSUE #3783: Customizable default harness stack: middleware, profiles, and system prompt overrides
## Problem

The default Deep Agents harness currently lacks knobs for customization in a coherent way. Several open issues are different symptoms of the same underlying limitation: `create_deep_agent(...)` wires together profiles, middleware, system prompts, and built-in subagent behavior through mostly fixed internal defaults.

This makes common customization use cases awkward or impossible without relying on provider/model-specific profile registration, post-construction mutation, or duplicating internal stack-building logic.

Related issues:

- #2126 — allow injection to the default middleware stack factory
- #3139 — allow overriding `FilesystemMiddleware`
- #3137 — allow customizing `SkillsMiddleware` system prompt template
- #3425 — allow passing profiles directly into `create_deep_agent`
- #3580 — way to override `BASE_AGENT_PROMPT` globally
- #2622 — allow customizing `SummarizationMiddleware` model and parameters in `create_deep_agent`

## Motivation

These issues are tightly related because they all point at the same design gap: users need supported extension points for customizing the default harness without abandoning the convenience of `create_deep_agent(...)`.

Examples:

- A user wants to replace or configure `FilesystemMiddleware`, but it is hard-pinned in the required middleware stack.
- A user wants to customize the `SkillsMiddleware` prompt template, but there is no public constructor/profile path for doing so.
- A user wants to customize `SummarizationMiddleware` model/threshold/keep/trim behavior from `create_deep_agent(...)`, but those options are not currently forwarded through the harness.
- A user wants an autonomous/headless base prompt, but prompt override is tied to provider/model profile keys.
- A user wants to pass a harness profile directly to a specific agent instead of registering it globally by model ID.

Taken individually, each issue can be patched. Taken together, they suggest we need a small, coherent customization model for the default harness.

## Desired outcome

Design and implement supported customization hooks for the default Deep Agents harness, covering:

- [x] Global/default profile or base prompt override
- [x] Middleware stack customization/injection
- [x] Replacing or configuring required middleware such as `FilesystemMiddleware`
- [x] Customizing built-in middleware parameters such as `SummarizationMiddleware` model, thresholds, and trim behavior
- [x] Customizing `SkillsMiddleware` prompt/template behavior
- [x] Clear behavior for how customizations apply to default/general-purpose subagents

## Design questions

1. Should `create_deep_agent(...)` accept a `HarnessProfile` directly?
2. Should there be a wildcard/global profile key such as `"*"`?
3. Should middleware customization be expressed as:
   - direct middleware lists,
   - middleware factory hooks,
   - profile fields,
   - or some combination?
4. How should required middleware be handled?
   - truly required and non-replaceable,
   - replaceable by compatible subclasses,
   - configurable through profile/options,
   - or injectable through a stack factory?
5. Should the default general-purpose subagent inherit the parent’s custom middleware/profile configuration by default?
6. How do we preserve backward compatibility and avoid making `create_deep_agent(...)` overly complex?

## Non-goals

- Forcing all users to manually assemble middleware stacks
- Removing the default convenience behavior of `create_deep_agent(...)`
- Making provider/model profiles responsible for unrelated per-agent harness configuration

## Proposed direction

A possible shape:

```python
agent = create_deep_agent(
    model=model,
    tools=tools,
    profile=HarnessProfile(...),
    middleware_factory=...,
)
```

or:

```python
register_harness_profile("*", HarnessProfile(...))
```

or some combination where:

- global/default profiles provide broad defaults,
- direct `create_deep_agent(..., profile=...)` overrides per-agent behavior,
- middleware factory hooks provide advanced customization without copying internal stack assembly.

The key goal is to make customization explicit, supported, and composable while keeping the default harness easy to use.

imnishitha:
### Summary

Passing a custom middleware instance whose `.name` matches a default (e.g. `SummarizationMiddleware`) now replaces the default in-place.

New middleware with no matching default is inserted after the last core middleware entry and before the profile/prompt-caching/memory tail, so `AnthropicPromptCachingMiddleware` and `MemoryMiddleware` still sit at the end of the stack where they need to be. `_ToolExclusionMiddleware` is always appended last so it strips tools after all tool-injecting middleware (filesystem, subagent, etc.) has run.

The general-purpose subagent inherits overrides for its default middleware from the main agent without carrying over middleware that's specific to the main agent.
Declarative `subagents=` continue build their stacks independently, with no overrides taking place.

### Motivation

Before this change, passing a custom `SummarizationMiddleware` (e.g. with a different `model` or `summary_prompt`) to `middleware=` caused `AssertionError: Please remove duplicate middleware instances`

The only workaround was to also add it to `HarnessProfile.excluded_middleware`. Now you can pass the replacement directly:

```Python
custom_summarization = SummarizationMiddleware(
    model="openai:gpt-5.5",
    backend=backend,
    summary_prompt="Your custom summary prompt here.",
)

agent = create_deep_agent(
    model="openai:gpt-5.5",
    middleware=[custom_summarization],  # replaces the default SummarizationMiddleware
)
```


libs/deepagents/deepagents/graph.py
```diff
@@ -204,6 +204,54 @@ def _merge_fs_interrupt_on(
...
+def _apply_custom_middleware(
+    base: list[AgentMiddleware[Any, Any, Any]],
+    custom: Sequence[AgentMiddleware[Any, Any, Any]],
+    original_name_to_index: dict[str, int] | None = None,
+) -> list[AgentMiddleware[Any, Any, Any]]:
+    """Merge custom middleware into the base stack by name.
+
+    For each custom entry:
+
+    - If its ``.name`` matches a name still present in ``base``: replace in-place,
```
ccurme:
```suggestion
    - If its `.name` matches a name still present in `base`: replace in-place,
```

libs/deepagents/deepagents/graph.py
```diff
@@ -204,6 +204,54 @@ def _merge_fs_interrupt_on(
     return merged
 
 
+def _apply_custom_middleware(
+    base: list[AgentMiddleware[Any, Any, Any]],
+    custom: Sequence[AgentMiddleware[Any, Any, Any]],
+    original_name_to_index: dict[str, int] | None = None,
+) -> list[AgentMiddleware[Any, Any, Any]]:
+    """Merge custom middleware into the base stack by name.
```
sydney-runkle, line 236: why by name and not by type?

libs/deepagents/deepagents/graph.py
```diff
@@ -204,6 +204,54 @@ def _merge_fs_interrupt_on(
     return merged
 
 
+def _apply_custom_middleware(
+    base: list[AgentMiddleware[Any, Any, Any]],
```
sydney-runkle, line 231: i think we want to thread through proper types here? good to at least understand what they all are

libs/deepagents/tests/unit_tests/test_graph.py
```diff
@@ -2081,3 +2083,232 @@ def test_get_default_model_emits_langchain_deprecation_warning(self) -> None:
...
         assert "https://docs.langchain.com/oss/python/deepagents/models" in msg
+
+
+def _named_mw(name: str) -> AgentMiddleware[Any, Any, Any]:
+    """Return a minimal AgentMiddleware whose .name returns `name`."""
+
+    class _MW(AgentMiddleware[Any, Any, Any]):
+        @property
+        def name(self) -> str:
+            return name
```
sydney-runkle, line 2240: what happens if you don't explicitly set name?

libs/deepagents/deepagents/graph.py
```diff
@@ -204,6 +204,54 @@ def _merge_fs_interrupt_on(
...
+    original_name_to_index: dict[str, int] | None = None,
+) -> list[AgentMiddleware[Any, Any, Any]]:
+    """Merge custom middleware into the base stack by name.
+
+    For each custom entry:
+
+    - If its ``.name`` matches a name still present in ``base``: replace in-place,
+      preserving stack order.
+    - If it matches a middleware that was excluded, insert it where that middleware
+    originally appeared.
```
sydney-runkle: do we have any harness profiles that actually exclude middleware? i'd be not inclined to add this complexity

libs/deepagents/tests/unit_tests/test_graph.py
```diff
@@ -2081,3 +2083,232 @@ def test_get_default_model_emits_langchain_deprecation_warning(self) -> None:
...
+
+        summ_entries = [m for m in stack if isinstance(m, _DeepAgentsSummarizationMiddleware)]
+        assert len(summ_entries) == 1
+        assert summ_entries[0] is custom
+
+    def test_diff_name_subclass_is_appended(self) -> None:
+        """A subclass with a different .name doesn't replace the default. Both are present."""
+
+        class CompactSummarizationMiddleware(SummarizationMiddleware):
+            pass
```
sydney-runkle, line 2361: hmm i would think we want this to replace?

sydney-runkle (commented):
ah also can we format the code snippet in the PR description w/ `py`, thanks

looking like a good start, this is a highly requested feature so i'm excited to support!

libs/deepagents/deepagents/graph.py
imnishitha: Avoids ambiguous inheritance-based behavior. With isinstance() it's unclear whether subclasses should replace the default middleware or be treated as additional middleware. For example, if a user provides two different subclasses of SummarizationMiddleware, both satisfy isinstance(..., SummarizationMiddleware). The system then has to guess which one replaces the default, whether both should replace it, or whether one should be appended. Name-based replacement avoids this ambiguity by making the user's intent explicit: same name means replace, different name means add.

libs/deepagents/tests/unit_tests/test_graph.py
imnishitha: it's gonna be the name of the class (base class `.name` is `self.__class__.__name__`)

libs/deepagents/deepagents/graph.py
ccurme: no pre-baked harness profiles do. this would be to protect enterprising users who are excluding default middleware and then passing in their own with a custom harness profile. it's ~10 lines of code, but to your point, it's a corner case and I wouldn't be surprised if no one is doing this.

ccurme: I think there are valid (likely uncommon) usage patterns now for which we'd start raising a ValueError if we don't handle.

imnishitha: This is the scenario that needs to be taken care of (for someone who's probably using this work around for using a custom middleware by excluding the default middleware from a harness profile and supplying their own replacement via middleware=[...]):

```python
register_harness_profile(
    "my-profile",
    HarnessProfile(
        excluded_middleware=frozenset(
            {SummarizationMiddleware, TodoListMiddleware}
        ),
    ),
)

agent = create_deep_agent(
    model="my-profile:some-model",
    middleware=[
        MySummarizationMiddleware(...)  # replaces the default SummarizationMiddleware
    ],
)
```

Without the position preservation logic, the replacement is appended to the end of the stack, changing middleware execution order.:


[Filesystem, SubAgent, PatchToolCalls, Caching, Memory, Interrupt, MySummarization]

libs/deepagents/deepagents/graph.py
imnishitha: `Any, Any, Any` is intentional because this helper only reorders middleware objects by name and never uses their state, context, or response types.

libs/deepagents/tests/unit_tests/test_graph.py
```diff
@@ -2164,3 +2229,391 @@ def test_get_default_model_emits_langchain_deprecation_warning(self) -> None:
...
+        custom = _named_mw("CustomMW")
+        stack = self._run([custom])
+
+        custom_idx = next(i for i, m in enumerate(stack) if m is custom)
+        pc_idx = next(
+            (i for i, m in enumerate(stack) if m.name == "AnthropicPromptCachingMiddleware"),
+            None,
+        )
+        assert pc_idx is not None, f"expected prompt caching middleware in stack; got {[m.name for m in stack]}"
+        assert custom_idx < pc_idx, f"user middleware must precede prompt caching; got order {[m.name for m in stack]}"
```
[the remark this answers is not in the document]
ccurme: This was deliberately added to demonstrate the issue.

ccurme (approved):

PR #4259, opened 2026-06-25
ci(evals): clarify evals dispatch defaults, order advanced inputs last

nick-hollon-lc:
## What

Two display-only tweaks to the `📊 Evals` dispatch form to lower launch-time cognitive load (Tier 1.3). No behavior change.

- **`models` (first field):** state that leaving every field blank and clicking Run performs the standard full eval (all models / categories / tiers), and that the fields below only *narrow* the run.
- **Reorder the advanced cluster** so run-shaping knobs come before post-run analysis:
  `… eval_tiers_override → openai_reasoning_effort → repl → openrouter_provider → openrouter_allow_fallbacks → analyze_failures → analysis_model → notes`

## Why

The form has 14 inputs, and GitHub `workflow_dispatch` offers **no form-level help text and no field grouping/collapsing**. The only levers are field order and per-field copy:

- The first field's description is the only prominent spot for "you can just hit Run" guidance — most launches want the full eval and shouldn't have to reason about 14 fields.
- The core selection (models / categories / tiers) was already first; this only regroups the advanced tail so the knobs that shape *how the eval runs* precede the *post-run* failure-analysis toggles.

## Scope / safety

- **Display-only.** Pure key reorder + one description reword. Verified the input set is unchanged (still 14; only `models`'s definition differs) and every other input definition is byte-identical.
- Inputs are referenced by name downstream, so order is cosmetic — no job/logic/secret changes.
- `python -m pytest .github/scripts/test_*.py` → **180 passed** (incl. the evals.yml dropdown-sync guard); workflow parses.

## Tier 1 status

This is the last of the Tier 1 cognitive-load items (#4250 split clbench, #4255 unify names — both merged). Deliberately *not* here: collapsing the dropdown+override input pairs (models/categories/tiers), which needs a generator + CI sync check — that's the heavier Tier 3 follow-up.


.github/workflows/evals.yml
```diff
@@ -36,9 +36,11 @@ on:
   workflow_dispatch:
     inputs:
       models:
-        description: "Model set to evaluate. Set definitions:
-          libs/evals/MODEL_GROUPS.md. Leave empty to use models_override
-          instead. Defaults to all models if both are empty."
+        description: "Model set to evaluate. Standard run: leave every field on
+          this form blank and click Run — that evaluates all models, all
```
mdrxy, line 40: i don't think evaluating ALL MODELS is standard

nick-hollon-lc: yeah good call will fix. how often do we think this is actually useful?

mdrxy: rarely, i'd rather default to a preset curated list that's no more than 5 models

PR #4267, opened 2026-06-25
fix(code): show months instead of "0y ago" for 360-364 day old timestamps

### LINKED ISSUE #4265: `format_relative_timestamp` shows `"0y ago"` for timestamps 360–364 days old
### Submission checklist

### Area (Required)

- [ ] deepagents (SDK)
- [ ] cli
- [x] code
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

None found

```python
from datetime import UTC, datetime, timedelta

from deepagents_code.sessions import format_relative_timestamp

for days in (359, 360, 362, 364, 365):
    ts = (datetime.now(tz=UTC) - timedelta(days=days, hours=1)).isoformat()
    print(f"days={days:>4} -> {format_relative_timestamp(ts)!r}")
```

```shell
days= 359 -> '11mo ago'
days= 360 -> '0y ago'
days= 362 -> '0y ago'
days= 364 -> '0y ago'
days= 365 -> '1y ago'
```

### Description

A thread (or a release, via `update_check`) that was last touched between 360 and 364 days ago shows up as **"0y ago"** in the `/threads` selector. I expected it to say something like "12mo ago" or "1y ago" - "0y ago" reads as a bug.

The cause is in `format_relative_timestamp`. The months bucket uses a 30-day divisor and the years bucket uses a 365-day divisor, and the two don't line up:

```python
months = days // 30
if months < 12:                 # only true while days < 360
    return f"{months}mo ago"
years = days // 365             # 0 until days >= 365
return f"{years}y ago"
```

So for `days` in 360–364, the months branch is skipped (`days // 30` is already 12, not `< 12`), but `days // 365` is still 0, and you get `"0y ago"`. It's a small 5-day gap, but it's always wrong when you land in it. The existing tests use 90 days and 400 days, so they step right over this range.

---

## Suggested fix
The cleanest fix is to gate the months branch on the **day** threshold that matches the years divisor, instead of on the derived `months` value. Switch the boundary to 365 days so months covers everything below a year:

```python
days = hours // 24
if days < 30:
    return f"{days}d ago"
if days < 365:                  # was: months = days // 30; if months < 12
    months = days // 30
    return f"{months}mo ago"
years = days // 365
return f"{years}y ago"
```

With that change, 360–364 days falls into the months branch and renders `"12mo ago"`, and 365+ days renders `"1y ago"` as before. Nothing else in the cascade changes.

(If you'd rather keep the boundary at exactly 12 months, the same fix can be written as `months = days // 30; if months < 12:` → no change there, plus computing `years` from `months // 12` so the two buckets share one divisor. But the day-threshold version above is the smaller, clearer edit.)

### Environment / System Info

```
deepagents-code: 0.1.23
Python: 3.12
OS: Ubuntu (WSL2) / Windows 11
```

parveshsaini:
Fixes #4265

---

## Summary
`format_relative_timestamp` showed `"0y ago"` for timestamps 360–364 days old, since the months branch stopped at `days < 360` but the years branch only kicks in at `days >= 365`. This closes that gap so those dates show `"12mo ago"`.

User story: with relative timestamps enabled, opening the `/threads` switcher to resume an old session shows each thread's age. A thread last touched ~360–364 days ago reads `"0y ago"`, which looks broken and is hard to tell apart from a brand-new thread. The same helper formats the `released … ago` line on the update notice, so that hits the gap too. After this fix both read `"12mo ago"`.

I came across it while reading the timestamp bucketing, the month bucket divides by 30 and the year bucket by 365, so I checked the boundary between them and confirmed 360–364 days falls through to `"0y ago"`.

No breaking changes; only the 360–364 day range is affected.

**How I verified**: added a boundary test that fails on `main` and passes here, and ran `make format`, `make lint`, and `make test` for the `code` package.

## Social handles (optional)

Twitter: @PrvshHasNoLife
LinkedIn: https://www.linkedin.com/in/saini-parvesh/

### DIFF
--- a/libs/code/deepagents_code/sessions.py
+++ b/libs/code/deepagents_code/sessions.py
@@ -230,8 +230,8 @@ def format_relative_timestamp(iso_timestamp: str | None) -> str:
     days = hours // 24
     if days < 30:  # noqa: PLR2004
         return f"{days}d ago"
-    months = days // 30
-    if months < 12:  # noqa: PLR2004
+    if days < 365:  # noqa: PLR2004
+        months = days // 30
         return f"{months}mo ago"
     years = days // 365
     return f"{years}y ago"
--- a/libs/code/tests/unit_tests/test_sessions.py
+++ b/libs/code/tests/unit_tests/test_sessions.py
@@ -449,6 +449,13 @@ class TestFormatRelativeTimestamp:
         result = sessions.format_relative_timestamp(ts)
         assert result.endswith("s ago")
 
+    def test_boundary_360_to_364_days_shows_months(self) -> None:
+        """360-364 days old should show months, never the bogus '0y ago'."""
+        for days in (360, 362, 364):
+            ts = (datetime.now(tz=UTC) - timedelta(days=days, hours=1)).isoformat()
+            result = sessions.format_relative_timestamp(ts)
+            assert result == "12mo ago"
+
 
 class TestFormatPath:
     """Tests for format_path helper."""
--- a/libs/code/tests/unit_tests/test_sessions.py
+++ b/libs/code/tests/unit_tests/test_sessions.py
@@ -456,6 +456,12 @@ class TestFormatRelativeTimestamp:
             result = sessions.format_relative_timestamp(ts)
             assert result == "12mo ago"
 
+    def test_boundary_365_days_shows_years(self) -> None:
+        """At exactly 365 days, the year bucket takes over with '1y ago'."""
+        ts = (datetime.now(tz=UTC) - timedelta(days=365, hours=1)).isoformat()
+        result = sessions.format_relative_timestamp(ts)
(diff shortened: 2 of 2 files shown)


mdrxy:
thanks - how did you encounter this OOC? can you update PR description with a user story of where this resolves?

parveshsaini:
From a user's perspective, this can show up in places like the /threads switcher (which displays thread age) and the release update notice ("released ... ago"). So if a thread or release is almost a year old, it can end up showing "0y ago", which doesn't look quite right.

I've updated the PR description with that user-facing example as well.

@mdrxy

PR #4324, opened 2026-06-26
fix(code): offload `create_model` in server graph factory to unblock Codex

mdrxy:
Fix `Blocking call to os.mkdir` server startup failure when using `openai_codex` (ChatGPT) models.

---

The async langgraph dev server graph factory (`server_graph._make_graph`) called `create_model` synchronously on the event loop. For the `openai_codex` provider, `create_model` → `build_chat_model` → `provider.get_token()` acquires a cross-process file lock that *unconditionally* calls `os.mkdir` (on every call, regardless of token freshness). The dev server runs `blockbuster`, which rejects blocking IO on the loop, so graph readiness fails with a 500: `Blocking call to os.mkdir`. This made the server unstartable for ChatGPT/Codex users, and re-running `/auth` did not help because the `mkdir` is unconditional. Fix: offload `create_model` via `asyncio.to_thread`, matching the existing pattern in `configurable_model._apply_overrides_async`.

Made by [Open SWE](https://openswe.vercel.app/agents/2b7c33d3-d7a8-d82b-c254-affd26a14334)

## References
- Plan: https://openswe.vercel.app/agents/2b7c33d3-d7a8-d82b-c254-affd26a14334/plan

### DIFF
--- a/libs/code/deepagents_code/server_graph.py
+++ b/libs/code/deepagents_code/server_graph.py
@@ -164,8 +164,9 @@ async def _make_graph() -> Any:  # noqa: ANN401
         settings.reload_from_environment(start_path=project_context.user_cwd)
 
     # Offload to a worker thread: `create_model` does blocking disk IO for some
-    # providers (e.g. the `openai_codex` token store acquires a file lock that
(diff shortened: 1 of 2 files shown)


tom21100227:
bump i want this

PR #4325, opened 2026-06-26
feat(sdk):  add `tools` allowlist to `FilesystemMiddleware`

imnishitha:
### Summary

- Adds a `tools` allowlist to `FilesystemMiddleware`. The default keeps all tools enabled; passing an explicit set restricts the model to only those tools.
- The system prompt dynamically reflects which tools are actually visible — a tool not in the allowlist is omitted from the prompt entirely, so the model is never told about tools it can't use.
- The allowlist can only *narrow* the tool set, never widen it. Some tools depend on backend capabilities that the allowlist can't override: `execute` is only available on sandbox backends, and `delete` is only available on backends that support deletion. Listing either against a backend that can't serve it is a no-op — the tool is silently filtered out, same as before this PR.
- `tools` is threaded through `create_deep_agent`, so the main agent and the general-purpose subagent respect it automatically when you pass a `FilesystemMiddleware(tools=...)` instance via `middleware=`. Declarative subagents use their own `FilesystemMiddleware` spec independently — include one in that subagent's `middleware` to restrict it.
- `read_file` cannot be excluded — passing a `tools` list without it raises `ValueError` at construction time.
- User-provided (non-filesystem) tools are never affected by the `tools` list.

<details>
<summary><b>How the allowlist interacts with <code>create_deep_agent</code></b></summary>

`create_deep_agent` builds separate `FilesystemMiddleware` instances internally: one for the main agent, one for the general-purpose subagent, and one per declarative subagent. The SDK uses name-based middleware merging — when you pass a `FilesystemMiddleware(tools=...)` instance via `middleware=`, it replaces the default FS middleware in-place on the main agent's stack. The general-purpose subagent inherits it too, because its base stack has a `FilesystemMiddleware` slot with the same `.name`.

Declarative subagents are different: each has its own independent base stack and only picks up the restriction if you include a `FilesystemMiddleware(tools=...)` in that subagent's own `middleware` spec. By design — declarative subagents often need different (broader or narrower) tool sets than the parent.

A single `middleware=[FilesystemMiddleware(backend=backend, tools=...)]` argument covers the main agent and GP subagent in one shot; declarative subagents require per-subagent opt-in.

</details>

<details>
<summary><b>Replacing the default FS middleware</b></summary>

Passing your own `FilesystemMiddleware` instance via `middleware=` replaces the SDK's default **entirely** — it is a whole-object swap, not a field-level merge. The default instance was pre-configured with `backend`, `custom_tool_descriptions`, and `_permissions` from the top-level `create_deep_agent` arguments; any parameter you omit on your own instance falls back to the constructor default, not to what `create_deep_agent` would have used.

For example, if you set `permissions=` on `create_deep_agent`, you must repeat `permissions=` on your `FilesystemMiddleware` instance or the rules won't apply.

</details>

<details>
<summary><b>User-provided tools</b></summary>

The allowlist only governs the eight built-in filesystem tool names (`ls`, `read_file`, `write_file`, `edit_file`, `delete`, `glob`, `grep`, `execute`). Any tool whose name isn't in that set passes through untouched — so you can safely layer filesystem restrictions on top of an agent that also has custom domain-specific tools without those tools disappearing.

</details>

### Motivation

Gives callers explicit control over which filesystem tools are exposed. New tools added to `FilesystemMiddleware` in the future are automatically hidden from restricted agents without requiring config updates.

### Example

```Python
# Read-only agent: write, edit, delete, and execute are never shown
agent = create_deep_agent(
    model="openai:gpt-5.5",
    middleware=[
        FilesystemMiddleware(
            backend=backend,
            tools={"read_file", "ls", "glob", "grep"},
        )
    ],
)
```


libs/deepagents/deepagents/graph.py
```diff
@@ -651,6 +651,17 @@ class MyState(DeepAgentState):
...
     backend = backend if backend is not None else StateBackend()
 
+    # Extract a caller-supplied FilesystemMiddleware from `middleware`, if any.
+    # Its enabled_tools is forwarded to the main agent and GP subagent so both
+    # share the same tool restrictions without requiring a separate parameter.
+    _caller_fs_mw: FilesystemMiddleware | None = None
+    _caller_middleware: list[AgentMiddleware[Any, Any, Any]] = list(middleware)
+    for _i, _mw in enumerate(_caller_middleware):
+        if isinstance(_mw, FilesystemMiddleware):
+            _caller_fs_mw = cast("FilesystemMiddleware", _caller_middleware.pop(_i))
```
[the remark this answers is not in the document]
imnishitha: this is expected - the user needs to pass the entire instance of the mw while passing a custom mw.

libs/deepagents/deepagents/graph.py
```diff
@@ -891,6 +915,7 @@ class MyState(DeepAgentState):
     )
     if main_interrupt_on is not None:
         deepagent_middleware.append(HumanInTheLoopMiddleware(interrupt_on=main_interrupt_on))
+    _main_original_name_to_index = {m.name: i for i, m in enumerate(deepagent_middleware)}
```
ccurme: what's this and `_subagent_original_name_to_index` above for?

libs/deepagents/deepagents/graph.py
```diff
@@ -651,6 +651,17 @@ class MyState(DeepAgentState):
 
     backend = backend if backend is not None else StateBackend()
 
+    # Extract a caller-supplied FilesystemMiddleware from `middleware`, if any.
```
ccurme: Do we need the changes to graph.py? If user passes FilesystemMiddleware into `create_deep_agent(middleware=[...])`, should we just use that instance? otherwise, what happens to the instance-- do we discard it save for the `enabled_tools` setting?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -889,6 +907,7 @@ def __init__(
         tool_token_limit_before_evict: int | None = 20000,
         human_message_token_limit_before_evict: int | None = 50000,
         max_execute_timeout: int = 3600,
+        enabled_tools: frozenset[str] | set[str] | None = None,
```
ccurme: shall we type this as `list[Literal["read_file", "write_file", ...]]` so people get autocomplete, type checkers can flag errors, etc?

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -906,13 +925,24 @@ def __init__(
 
                 Defaults to 3600 seconds (1 hour). Any per-command timeout
                 exceeding this value will be rejected with an error message.
+            enabled_tools: Allowlist of tool names to expose to the model.
+                ``None`` (the default) enables all tools. When a set is
```
ccurme:
```suggestion
                `None` (the default) enables all tools. When a set is
```
I see double backticks throughout (agent assuming sphinx docs?), we use single.

libs/deepagents/deepagents/middleware/filesystem.py
ccurme, line 1202: do we need to filter here

imnishitha: Filtering happens in `_filter_unsupported_tools_and_apply_prompt `in `wrap_model_call`. It's not filtered out when instantiating the FS middleware

libs/deepagents/deepagents/middleware/filesystem.py
imnishitha: sounds good!

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -889,6 +910,7 @@ def __init__(
         tool_token_limit_before_evict: int | None = 20000,
         human_message_token_limit_before_evict: int | None = 50000,
         max_execute_timeout: int = 3600,
+        enabled_tools: Collection[FsToolName] | None = None,
```
ccurme: I'd propose we rename to `tools` and update the type:
```python
tools: list[FsToolName] | Literal["all"] | None = None
```
and have None default to `"all"` under the hood.

reasons:
- Collection works technically but is arguably too broad (a `dict` is a collection). in practice python libraries often expose `list` even if a more abstract type would be cleaner for ergonomics.
- nice to support explicit `tools="all"` instead of requiring `tools=None` to specify all (None is often a sentinel value in python but can also be meaningful; `tools=None` reads like "no tools")
- having a sentinel None lets us distinguish users who are explicitly passing in `tools="all"` from users who are accepting the default, which lets us manage future changes better.

ccurme (commented):
can we add at least one minimal end-to-end test in tests/test_end_to_end.py demonstrating how the system prompt is changed, and spying on the request to demonstrate that disabled tools are not there?

libs/deepagents/deepagents/middleware/filesystem.py
imnishitha: renamed it to tools and changed the type to list, with literal "all"

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1097,13 +1119,24 @@ def __init__(
 
                 Defaults to 3600 seconds (1 hour). Any per-command timeout
                 exceeding this value will be rejected with an error message.
+            tools: Allowlist of `FsToolName` values to expose to the
+                model. `None` (the default) and `"all"` both enable every
+                tool; `None` signals "user did not specify" while `"all"`
+                is an explicit opt-in. Pass a list to restrict the model to
```
ccurme: (nit) the fact that `None` signals "user did not specify" is more relevant for maintainers. users shouldn't have to think about this. I'd rephrase to something like `Allowlist of tool names to expose to the model. "all" indicates all tools. If unset, defaults to `"all"`.

Might be good for the logic to actually read like this as well. atm we ignore the string literal `"all"` completely, which is technically correct, but if we decide to change the default in the future, we'll have to code in logic for the `"all"` case.

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1097,13 +1119,24 @@ def __init__(
 
                 Defaults to 3600 seconds (1 hour). Any per-command timeout
                 exceeding this value will be rejected with an error message.
+            tools: Allowlist of `FsToolName` values to expose to the
```
ccurme: I would document the supported tool names here instead of `FsToolName`.

ccurme (approved):

PR #4350, opened 2026-06-28
feat(code): add OpenAI hosted web search

semistrict:
Adds provider-hosted web search support for OpenAI-backed models, including rendering hosted search events in the terminal UI and live integration coverage for the provider call path.

Screenshot 2026-06-28 at 10 07 50 AM

### DIFF
--- a/libs/code/deepagents_code/app.py
+++ b/libs/code/deepagents_code/app.py
@@ -6255,7 +6255,7 @@ class DeepAgentsApp(App):
                 return
 
             model_spec, provider = result
-            await self._prompt_launch_tavily()
+            await self._prompt_launch_tavily(provider=provider)
             if self._connecting:
                 # Bound the wait so a stuck server never traps onboarding.
                 # Server startup typically completes in seconds; a minute is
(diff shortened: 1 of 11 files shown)


libs/code/deepagents_code/agent.py
```diff
@@ -216,6 +216,42 @@ async def awrap_tool_call(
         return await handler(request)
 
 
+class OpenAIHostedWebSearchMiddleware(AgentMiddleware):
```
ccurme: do we need a separate middleware for this? can we just pass in the tool dict to `tools`?

libs/code/deepagents_code/app.py
```diff
@@ -6340,20 +6340,22 @@ async def _switch_or_install_launch_model(
...
 
-        Skipped when a Tavily key is already configured (env or stored). A
-        blank submission or Escape stores nothing; a non-empty key is persisted
-        via the same `auth_store` path `/auth` uses. The key is also exported to
-        the process environment (`apply_stored_service_credentials`) so a server
-        respawn this session picks it up; the already-running server keeps its
-        spawn-time tools, so web search takes full effect on the next launch (or
-        after a restart).
+        Skipped when a Tavily key is already configured (env or stored), or when
+        the selected model provider has hosted web search. A blank submission or
```
ccurme, line 6347: @mdrxy do we want to expose any configurability here?

libs/code/deepagents_code/textual_adapter.py
```diff
@@ -1092,6 +1092,72 @@ async def execute_task_textual(
 
                             tool_call_buffers.pop(buffer_key, None)
 
+                        elif block_type == "server_tool_call":
```
ccurme, line 1095: this is good to have as it should extend to other server-side tools.

libs/deepagents/deepagents/openai.py
```diff
@@ -0,0 +1,50 @@
...
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Literal
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+
+def openai_web_search_tool(
```
ccurme: can we do without this function? atm it looks like we just call `openai_web_search_tool()`. we'll also need to update this if openai evolves their spec. if we want validation, why not just do
```python
from openai.types.responses import WebSearchTool

WebSearchTool(type="web_search").model_dump()
```

mdrxy:
Closing in favor of #3979, the existing tracking issue for pluggable/arbitrary web search providers. The OpenAI hosted search work here can be folded under that issue.

PR #4369, opened 2026-06-29
style(code): order `/model` selector by provider availability

mdrxy:
The `/model` selector now lists providers you can use (have credentials for) at the top.

---

In the `/model` selector, the provider groups were rendered in a fixed discovery order, so a user whose only usable provider (e.g. `openai_codex`) sat low in the list had to scroll or search to reach it. This sorts the unfiltered default view so providers the user can actually use float to the top, with missing-credential providers above not-installed ones. A search still ranks by fuzzy score, so that path is unchanged. The pinned "Recent" section stays first.

Requested by @tom21100227.

Made by [Open SWE](https://openswe.vercel.app/agents/ad6b5403-7427-49c7-620e-0d070f2f9a40)


tom21100227:
@mdrxy hi when would this be out

PR #4391, opened 2026-06-30
fix(evals): make run-hello-world actually runnable

nick-hollon-lc:
## Summary
- Follow-up to #4383, addressing the OpenSWE review comment: `--task hello-world` expects an `org/name` registry id, not a dataset task-name filter, so `make run-hello-world` errored immediately. Switched to `--include-task-name hello-world`.
- That fix alone wasn't enough to make the target actually run end-to-end, so this PR also fixes the rest of the chain:
  - `run-hello-world` never passed `--model`, so the LangGraph agent had no model to invoke. Added `--model` (default `fireworks:accounts/fireworks/models/glm-5p2`), overridable via `MODEL=`.
  - The pinned `langchain-fireworks>=1.4.2,<1.5.0` in `langgraph.json` only resolves against a `fireworks-ai` prerelease (`>=1.2.0a71`); forwarded `UV_PRERELEASE=allow` via `--agent-env` so harbor's install step (which runs inside the scoped exec-env overlay) lets `uv` pick it up, rather than pinning the dependency back to an older line.


mdrxy (approved):

nick-hollon-lc:
Superseded by #4402 — reopened from a branch on this repo instead of a fork.

PR #4395, opened 2026-06-30
hotfix(talon): allow prerelease release dependencies

jkennedyvz:
The Talon release pre-check installs the built wheel from a clean venv. Talon depends on `deepagents-code`, and the currently published `deepagents-code` line pins the alpha SDK `deepagents==0.7.0a2`, so uv must allow prerelease dependency resolution for Talon's release install checks.

This updates Talon to require the latest `deepagents-code` package and scopes `--prerelease allow` to `deepagents-talon` in the release workflow. Other packages continue to install built wheels with the default stable dependency resolution.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -909,7 +909,7 @@ jobs:
         # Install directly from the locally-built wheel (no index resolution needed)
         run: |
           uv venv
-          VIRTUAL_ENV=.venv uv pip install dist/*.whl
+          VIRTUAL_ENV=.venv uv pip install --prerelease allow dist/*.whl
 
           # Replace all dashes in the package name with underscores,
           # since that's how Python imports packages with dashes in the name.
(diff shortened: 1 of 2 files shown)


mdrxy (approved):

.github/workflows/release.yml
```diff
@@ -909,7 +909,7 @@ jobs:
         # Install directly from the locally-built wheel (no index resolution needed)
         run: |
           uv venv
-          VIRTUAL_ENV=.venv uv pip install dist/*.whl
+          VIRTUAL_ENV=.venv uv pip install --prerelease allow dist/*.whl
```
mdrxy: why is this needed? is it possible to scope to talon?

mdrxy: Scoped this to `deepagents-talon` only and added a short comment at both install sites explaining why Talon needs prerelease resolution: it transitively requires the prerelease `deepagents` pin via `deepagents-code`.

PR #4412, opened 2026-07-01
feat(evals): goal-tool over-eager-calling evals + gated prompt copy

mdrxy:
Models over-eagerly call `dcode`'s `get_rubric` / `get_goal` even when no goal or rubric was ever set, because `GOAL_TOOLS_SYSTEM_PROMPT` leads with an unconditional "inspect before finishing" instruction. This adds evals to pin the behavior down and rewrites the copy to gate on the precondition. Specifically:

- New `tool_not_called` hard-fail success assertion in `libs/evals/tests/evals/utils.py` (the framework only had text-based success assertions + efficiency presence checks).
- New `libs/evals/tests/evals/test_goal_tools.py` exercising the real `GoalToolsMiddleware` via `create_agent` (mirrors `test_langchain_middleware_todo.py`): two baseline regression gates that `get_rubric` / `get_goal` stay untouched when no goal/rubric is set (trivial task + real multi-step tool use), and a hillclimb case that a consulted rubric is still allowed (seeded via the public `rubric` input, since goal channels aren't part of the public graph input in an isolated harness).
- Rewrote `GOAL_TOOLS_SYSTEM_PROMPT` to lead with "a goal/rubric is only present if one was set earlier; if none was set, do not call these tools."
- Regenerated `EVAL_CATALOG.md`.

Made by [Open SWE](https://openswe.vercel.app/agents/de848f56-d39d-53f9-1959-7cd5cca9eee6)

## References
- Plan: https://openswe.vercel.app/agents/de848f56-d39d-53f9-1959-7cd5cca9eee6/plan


mdrxy:
maybe instead of system prompt artificla msg when goal is added?

mdrxy:
Superseded by #4980.

The replacement carries forward the goal-state notices, goal-tool gating, internal-message filtering, and relevant eval coverage, reconciled with #4944’s accepted-goal continuation. This avoids landing two overlapping hidden-message protocols for the same goal-activation transition.

Closing this PR in favor of #4980.

PR #4442, opened 2026-07-02
feat(talon): add Fleet run manifest schema

jkennedyvz:
## Why this is useful

Durable local manifest for Fleet-backed Talon runs lets follow-up setup agents read the selected channel, Fleet source, resolved model, and minimal MCP tool requirements from disk instead of re-deriving them each startup. Operators can inspect `<assistant-home>/fleet-run-manifest.json` to confirm the run configuration and see which Fleet MCP tool names are requested or currently loaded without persisting Fleet server URLs or auth metadata.

---

Adds an assistant-scoped Fleet run manifest for Talon Fleet runs. The manifest is refreshed on Fleet runtime startup and credential reloads, preserves its original creation timestamp, points local MCP setup to `<assistant-home>/.mcp.json`, and records only each requested MCP tool's name, scope, and loaded status.

## Release note
Talon now writes a Fleet run manifest at `<assistant-home>/fleet-run-manifest.json` for Fleet-backed local runs.

Made by [Open SWE](https://openswe.vercel.app/agents/0df7a2a8-81b9-d4cc-1b8d-4c5c10649e12)


jkennedyvz:
This still preserves too much Fleet MCP metadata from `tools.json`. We should not persist Fleet MCP server URLs or auth-path details, even sanitized. The local manifest should only tell the operator/follow-up agent what tool names are missing or needed and that they should create `<assistant-home>/.mcp.json`.

Please remove the per-tool `mcp_server_url`, `auth_path`, and setup-task IDs derived from Fleet server metadata. A minimal shape like `{tool_name, scope, loaded}` plus `local_mcp_config_path` is enough. The summary can say which Fleet tool names were requested and which are currently loaded; the actual MCP server config should be authored fresh by the user/agent in `.mcp.json`.

The core issue is that this code is trying to safely persist Fleet MCP data we do not need to persist at all.

PR #4471, opened 2026-07-03
feat(code): running session cost in the status bar

sydney-runkle:
Adds a running, cumulative estimate of session spend to the `deepagents-code` status bar, next to the existing token count.

Cost is estimated per request with [`pydantic/genai-prices`](https://github.com/pydantic/genai-prices) (bundled, offline price data — no network on the startup hot path). The estimate is folded into a monotonic session total that is persisted in a new private `_session_cost` checkpoint channel, written from inside the graph by `ResumeStateMiddleware.after_model` so it rides the same checkpoint as `_context_tokens` and is restored on `-r` resume and `/threads` switches.

Unlike the token count — which reflects *current context size* and resets on `/compact` — cost is a cumulative session total, so the two stay distinct. Unknown/self-hosted models that are not in the price data degrade quietly (the running total is left unchanged and nothing is shown) rather than crashing streaming. The cost also appears in `/tokens` and the headless (`-x`) usage table.

`genai-prices` is added as a required dependency of `deepagents-code`; no `deepagents` SDK pin change is needed since this builds on existing middleware APIs.

## Release note

The `deepagents-code` status bar now shows a running estimate of session cost (USD) alongside the token count, persisted across restarts and thread resume. Also surfaced in `/tokens` and the headless usage table.

### DIFF
--- a/libs/code/deepagents_code/_cost.py
+++ b/libs/code/deepagents_code/_cost.py
@@ -13,8 +13,31 @@ returns `None` and callers leave the running total unchanged rather than crash.
 from __future__ import annotations
 
 import logging
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
 
 logger = logging.getLogger(__name__)
+
+
+def cache_tokens_from_usage(usage: Mapping | None) -> tuple[int, int]:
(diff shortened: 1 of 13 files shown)


mdrxy:
Closing this PR in favor of tracking the feature request in #4520: https://github.com/langchain-ai/deepagents/issues/4520

PR #4526, opened 2026-07-06
docs(code): clarify /offload command summarizes messages

mdrxy:
New users were confused that `/offload` (alias `/compact`) only "offloads" context, when it actually runs an LLM summarization of older messages before writing the originals to a recovery file. This updates the slash-command description to mention summarization, staying under 60 characters ("Summarize and offload older messages to free context", 52 chars), and regenerates the `COMMANDS.md` catalog to match.

Made by [Open SWE](https://openswe.vercel.app/agents/d78e6bda-ad23-13f1-e566-a01f25a05364)


aolsenjazz:
Yeah, I mean this is one step better - I think the term `offload` can still be confounding, but this is good enough that i don't really care

PR #4531, opened 2026-07-06
fix(sdk): isolate composite glob paths

### LINKED ISSUE #4510: `CompositeBackend.glob` leaks results from unrelated routes on a path-scoped search
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

#4508 

```python
from deepagents.backends.composite import CompositeBackend

mem = InMemoryStore()
comp = CompositeBackend(
    default=StoreBackend(store=mem, namespace=lambda _rt: ("default",)),
    routes={"/memories/": StoreBackend(store=mem, namespace=lambda _rt: ("filesystem",))},
)
comp.write("/tools/hammer.md", "content")
comp.write("/memories/secret.md", "content")

# grep scoped to /tools correctly stays within the default backend:
print(comp.grep("content", path="/tools").matches)
# -> [{'path': '/tools/hammer.md', ...}]

# glob scoped to the SAME /tools path also returns the /memories route (bug):
print([fi["path"] for fi in comp.glob("*.md", path="/tools").matches])
# -> ['/memories/secret.md', '/tools/hammer.md']
```

```shell
No exception is raised. The failure is silent: a path-scoped glob returns extra
results from an unrelated route instead of restricting to the scoped path.
```

### Description

Using CompositeBackend with a default backend plus a routed backend at /memories/,
I call a path-scoped search on a path that belongs to the default backend (/tools).

Expected: like grep, a path-scoped glob searches only the backend that owns that
path. grep already enforces this and it is guarded by
test_composite_backend_grep_path_isolation.

Actual: glob (and its async twin aglob) fall through to the merge branch that also
searches every routed backend, so files under /memories/ appear in a /tools-scoped
glob. This breaks path isolation; with sensitive data behind a route it discloses
paths the caller intended to exclude.

Root cause: glob/aglob lack the "explicit non-route path -> default backend only"
guard that grep/agrep have in deepagents/backends/composite.py. A fix mirroring
grep, with sync and async regression tests (failing on main, passing with the fix),
is ready in PR #4508.

### Environment / System Info

OS: macOS 26.5.1
Python: 3.12.12
deepagents: 0.6.12
langchain: 1.3.x (constraint >=1.3.11,<2.0.0)

mdrxy:
Fixes #4510

Corrected `CompositeBackend.glob/aglob` so explicit default-backend paths no longer include results from routed backends.

---

`CompositeBackend` routes file operations by path prefix. `grep`/`agrep` already handle the three routing cases correctly:

- Route-scoped paths like `/memories/` search only the routed backend.
- Root and unscoped searches (`path="/"` or `path=None`) aggregate the default backend plus all routed backends.
- Explicit paths that do not match a route, like `/tools`, search only the default backend.

`glob`/`aglob` handled route-scoped searches correctly, but unmatched explicit paths fell through to the aggregate branch. That meant a search scoped to `/tools` could incorrectly return matching files from `/memories`.

## What changed

- Updated `CompositeBackend.glob` and `CompositeBackend.aglob` to mirror the existing `grep` / `agrep` routing rules.
- Preserved root and unscoped aggregation across the default backend and routed backends.
- Preserved route-scoped glob behavior, including remapping result paths back under the route prefix.
- Added sync and async regression coverage for the issue repro: `glob("*.md", path="/tools")` returns `/tools/hammer.md` and does not return `/memories/secret.md`.

## User story

As a Deep Agents user with `CompositeBackend` routing `/memories/` to a separate backend, I expect a scoped glob like `glob("*.md", path="/tools")` to search only `/tools` in the default backend. Files from routed paths, such as `/memories/secret.md`, should only appear when I explicitly search that route, search `/`, or run an unscoped glob.


libs/deepagents/deepagents/backends/composite.py
```diff
@@ -419,9 +419,12 @@ async def agrep(
         return self._coerce_grep_result(await self.default.agrep(pattern, path, glob))
 
     def glob(self, pattern: str, path: str | None = None) -> GlobResult:
-        """Find files matching a glob pattern, routing by path prefix."""
-        results: list[FileInfo] = []
+        """Find files matching a glob pattern, routing by path prefix.
 
+        Routes to backends based on path: a specific route searches one backend,
+        `"/"` or `None` searches all backends, otherwise searches the
```
eyurtsev: Is the "otherwise" part of the documentation already handeled by specific route? (If not, I don't quite understand what otherwise means)

libs/deepagents/deepagents/backends/composite.py
```diff
@@ -438,34 +441,40 @@ def glob(self, pattern: str, path: str | None = None) -> GlobResult:
...
+            # grep merge path); surface it instead of returning default-only matches.
+            if isinstance(default_result, GlobResult) and default_result.error:
+                return default_result
+            default_matches = default_result.matches if isinstance(default_result, GlobResult) else default_result
+            results.extend(default_matches or [])
+            truncated = truncated or _glob_truncated(default_result)
+
+            for route_prefix, backend in self.routes.items():
+                route_pattern = _strip_route_from_pattern(pattern, route_prefix)
+                sub_result = backend.glob(route_pattern, "/")
```
eyurtsev: We should probably optimize this and run the globs in parallel (can be done separately)

eyurtsev (approved):

PR #4536, opened 2026-07-06
fix(code): add `/exit` alias for quit

johannes117:
Adds `/exit` as an alias for `/quit` so users can exit the TUI with either command.

Also updates the generated command catalog and covers the alias in slash-command bypass tests.


mdrxy:
is this not already a hidden alias?

johannes117:
> is this not already a hidden alias?

yeah there is no functional difference after testing

PR #4541, opened 2026-07-07
refactor(sdk)!: remove deprecated backend compatibility shims

mdrxy:
Deprecated backend APIs and compatibility symbols are now removed. Users should pass backend instances instead of callable factories, configure `StoreBackend` with an explicit namespace, and use the current `ls` / `glob` / `grep` / `ReadResult` APIs. New files use string `FileData.content`; files persisted by older releases with `list[str]` content remain readable and are converted to the current format when overwritten or edited. The deprecated `runtime` parameters on `StateBackend` and `StoreBackend` are removed, along with the compatibility symbols `BackendFactory`, `BACKEND_TYPES`, `FileFormat`, and `Unset`.

---

Removes expired SDK compatibility paths and related legacy storage controls, so backend APIs now use the current storage and middleware contracts directly without making persisted legacy files unreadable. This tightens the public surface around `BackendProtocol` instances, explicit `StoreBackend` namespaces, modern file data storage, and virtual filesystem defaults.

## Changes
- **Backend construction:** removes callable `backend` factories across `create_deep_agent` and SDK middleware; callers now pass concrete `BackendProtocol` instances. The `BackendFactory` and `BACKEND_TYPES` aliases are also removed.
- **Backend constructors:** removes the deprecated `runtime` parameters from `StateBackend` and `StoreBackend`.
- **Backend protocol cleanup:** drops deprecated `files_update` fields and their `Unset` sentinel, legacy `ls_info` / `glob_info` / `grep_raw` shims, and plain-string `read()` compatibility.
- **File storage format:** removes the `FileFormat` alias and `file_format` controls, so new writes always store string content. Existing `list[str]` content remains readable without modifying stored data and is converted to the string format when overwritten or edited.
- **Persistent store behavior:** requires explicit `StoreBackend(namespace=...)`, removes `BackendContext`, and removes legacy assistant-id namespace fallback.
- **Filesystem defaults:** makes `FilesystemBackend` and `LocalShellBackend` default to `virtual_mode=True` without the old migration warning.
- **Summarization history:** removes deprecated `history_path_prefix` and uses the configured backend instance directly for conversation-history offloading.


libs/deepagents/deepagents/backends/store.py
```diff
@@ -176,59 +92,31 @@ class StoreBackend(BackendProtocol):
...
-    The namespace can include an optional assistant_id for multi-agent isolation.
+    Files are scoped by the caller-supplied `namespace` factory (e.g. per-user
+    or per-assistant isolation).
     """
 
     def __init__(
         self,
-        runtime: object = None,
         *,
+        namespace: NamespaceFactory,
```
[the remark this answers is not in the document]
mdrxy: ignore for now

libs/deepagents/deepagents/backends/store.py
```diff
@@ -320,24 +181,10 @@ def _convert_store_item_to_file_data(self, store_item: Item) -> FileData:
...
-                package="deepagents",
-            )
-            content = "\n".join(raw_content)
-        elif isinstance(raw_content, str):
-            content = raw_content
-        else:
-            msg = f"Store item does not contain valid content field. Got: {store_item.value.keys()}"
+        if not isinstance(raw_content, str):
+            msg = f"Store item `content` must be a `str`, got {type(raw_content).__name__}. Legacy `list[str]` content is no longer supported."
             raise TypeError(msg)
```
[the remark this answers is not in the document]
mdrxy: this is fine probably since it's a breaking change?

mdrxy:
TODO (follow-up PR): update `libs/evals/tests/evals/test_memory.py` for the removed `StoreBackend` compatibility paths.

The eval still constructs `StoreBackend(store=store)` without the now-required `namespace`, and seeds the store with legacy `list[str]` content. To keep eval-package changes outside this SDK cleanup PR, a follow-up should:

- pass an explicit namespace factory to `StoreBackend`
- seed current-format `FileData` with string `content` and an `encoding`

mdrxy:
Open question: how should we handle `FileFormat` and the public `file_format=` parameters here?

The v1 `list[str]` storage behavior was deprecated, so removing v1 itself was signaled. However, the `FileFormat` alias and the `file_format=` parameters on `StateBackend` and `StoreBackend` were not directly deprecated and emitted no warning. Users could reasonably expect v1 to disappear without expecting the alias and parameters to be removed entirely.

Should we:

- keep the current removal for 0.7;
- restore the alias and parameters for one deprecation cycle, potentially accepting only `"v2"`; or
- take another compatibility approach?

`BackendFactory`, `BACKEND_TYPES`, and `Unset` seem more directly inferable as collateral removals of their explicitly deprecated factory and `files_update` paths, so `FileFormat` is the uncertain case.

libs/deepagents/deepagents/backends/local_shell.py
```diff
@@ -109,7 +108,7 @@ def __init__(
         self,
         root_dir: str | Path | None = None,
         *,
-        virtual_mode: bool | None = None,
+        virtual_mode: bool = True,
```
sydney-runkle, line 111: can we have eugene review this

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -1564,28 +1562,8 @@ def _build_dynamic_system_prompt(self, *, include_execution: bool) -> str:
...
-                    "and will be removed in deepagents==0.7.0. Pass a "
-                    "`BackendProtocol` instance directly instead "
-                    "(e.g. `StateBackend()`)."
-                ),
-                package="deepagents",
-            )
-            return _resolve_backend(self.backend, runtime)
+    def _get_backend(self, _runtime: ToolRuntime[Any, Any]) -> BackendProtocol:
+        """Return the backend instance."""
         return self.backend
```
sydney-runkle: we should not need these anymore

sydney-runkle (commented):
initial review
did we plan to remove the file format support?

libs/deepagents/deepagents/backends/filesystem.py
```diff
@@ -132,14 +131,14 @@ class FilesystemBackend(BackendProtocol):
...
-            provides no security even with `root_dir` set.
+            provide sandboxing or process isolation. Set `virtual_mode=False` only for
+            trusted local development workflows that require unrestricted host paths.
     """
 
     def __init__(
         self,
         root_dir: str | Path | None = None,
-        virtual_mode: bool | None = None,  # noqa: FBT001
+        virtual_mode: bool = True,  # noqa: FBT001, FBT002
```
ccurme, line 141: why drop the sentinel value?

eyurtsev (approved):

libs/deepagents/deepagents/backends/filesystem.py
mdrxy: The `None` sentinel existed only to distinguish an omitted argument from explicit `False` during the default-change deprecation: omission warned and temporarily resolved to `False`. Now that the deprecation period is over and the default has changed to `True`, omission and `True` intentionally mean the same thing. `None` was never a meaningful mode, so retaining it would preserve obsolete migration behavior and an unnecessarily broad public type.

ccurme (approved):

libs/deepagents/deepagents/backends/filesystem.py
ccurme: `None` is nice generally to manage future changes since we can see what users are passing in explicitly, vs. accepting the default. But don't feel strongly / subjective preference.

mdrxy:
### Follow-up needed in `deepagents-code` (not addressed in this PR)

This PR removes `SkillsMiddleware._get_backend`, which the sibling `deepagents-code` package still depends on. That fix is intentionally **not** included here and needs a coordinated follow-up before/alongside a `deepagents` release bump.

**Where it breaks:** `libs/code/deepagents_code/plugins/adapters/skills_middleware.py`

```
297:        backend = self._get_backend(state, runtime, config)
343:        backend = self._get_backend(state, runtime, config)
```

`PluginSkillsMiddleware(SkillsMiddleware)` calls the removed method in both `before_agent` and `abefore_agent` and defines no local override, so these raise `AttributeError` once the parent method is gone.

**Not shielded by the version pin.** `libs/code` declares `deepagents==0.7.0a8` (`pyproject.toml:29`) but overrides resolution to local editable source:

```
# libs/code/pyproject.toml:197
[tool.uv.sources]
deepagents = { path = "../deepagents", editable = true }
```

So `deepagents-code` runs against this branch's source, and its test suite will fail against the removal — this is a break today, not just on a future pin bump.

**Suggested fix (follow-up PR):** inline the backend at both call sites and drop the now-unused args:

```python
backend = self._backend
```

Safe now that backend factories are rejected at `__init__` (`filesystem.py:1530`), so `self._backend` is always a concrete `BackendProtocol` — there is no factory-resolution behavior left to preserve. After the change, run the `deepagents-code` unit tests to confirm.

PR #4570, opened 2026-07-08
feat(sdk): add total match cap and streaming to grep

### LINKED ISSUE #2834: Add `max_count` to `grep_raw`
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] repl
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox

### Feature description

grep_raw is currently uncapped. While running some work on the linux kernel, my process was crashing due to OOM. Upon profiling/inspection, ripgrep was buffering 7gb worth of results to stdout. 

Implement a max_count with some sane default (perhaps overiddable) and return a message when results are capped.

_No response_

mdrxy:
Closes #2834

`FilesystemMiddleware`'s `grep` tool now returns at most **1000 matches by default** (previously unbounded). When the cap is hit, results are truncated and a note tells the model to narrow the search or raise `max_count`. This bounds memory and context size on very large repositories.

- The model can raise or lower the cap per call via the new `max_count` argument on the `grep` tool.
- To restore the previous unbounded behavior, construct `FilesystemMiddleware(..., grep_max_count=None)`.
- Direct backend callers (`backend.grep(...)`) are unaffected: `max_count` defaults to `None` (no cap) unless explicitly passed.

---

Deep Agents' filesystem `grep` was already timeout-bounded, but it could still buffer an unbounded number of matches: the local ripgrep path ran `subprocess.run(capture_output=True)`, so a broad pattern on a huge repo (e.g. the Linux kernel) could buffer gigabytes of stdout before the tool layer got a chance to truncate, risking OOM.

This adds a total-match cap and makes the local ripgrep path stream:

- New keyword-only `max_count: int | None = None` on `BackendProtocol.grep`/`agrep` and every backend implementation (`filesystem`, `state`, `store`, `sandbox`, `composite`, `context_hub`). It is a total cap across all files, not per-file. Defaulting to `None` keeps existing direct-backend behavior unchanged.
- `LocalFilesystemBackend` now streams ripgrep's `--json` output via `subprocess.Popen` line-by-line, stops and `terminate()`s the process once the cap is reached, and passes `-m` as a cheap secondary per-file guard. A `threading.Timer` watchdog preserves the existing time-limit behavior. The Python fallback early-returns at the cap.
- `CompositeBackend` enforces the cap globally across routed backends, short-circuiting remaining routes once it is met.
- `GrepResult.truncated` (already present) is set when the cap is hit, and `FilesystemMiddleware` gains a `grep_max_count` default of `1000` (configurable, and overridable per call via the tool's `max_count` argument). The truncation note shown to the model now covers both the time-limit and match-cap cases.

Reviewers may want to focus on the rewritten `_ripgrep_search` streaming/termination/watchdog logic and the composite cross-route budget accounting.

Made by [Open SWE](https://openswe.vercel.app/agents/389434e7-284f-e1ff-11ba-8c65c21683af)


mdrxy:
TODO after #4541 merges: rebase this PR and remove the `grep_raw` compatibility path from `BackendProtocol.grep`. #4541 intentionally removes that deprecated shim, so conflict resolution here should not restore it. Keep the compatibility handling for current custom backends whose `grep` / `agrep` methods use the existing signature without `max_count`.

PR #4597, opened 2026-07-09
style(code): inline confirmation after MCP `/mcp login`

mdrxy:
`/mcp login` now leaves an inline "Logged in to '<server>'." message in the transcript on success.

---

After a successful `/mcp login <server>` OAuth flow, the only confirmation was the login modal and reconnect toasts — both transient. This mounts a persistent `AppMessage` in the transcript (`Logged in to '<server>'.`) so there's a lasting record once those dismiss.

Made by [Open SWE](https://openswe.vercel.app/agents/b6757761-527b-a793-94a3-41fa5727eb66)


mdrxy:
we may want to just remove the toast?

PR #4600, opened 2026-07-09
feat(code): add `GPT-5.6` model family

johannes117:
Deep Agents Code now supports GPT-5.6 Sol, Terra, and Luna through OpenAI API keys and ChatGPT sign-in, including the new `max` reasoning effort.

---

OpenAI's GPT-5.6 profiles have not reached `langchain-openai` yet, which would otherwise keep the new family out of the model selector. This adds Sol, Terra, and Luna as curated recommendations while preserving upstream profiles as the authoritative metadata source once they are released.

- Surface all three models through the OpenAI API and ChatGPT OAuth/Codex paths.
- Support the documented `medium` default and new `max` reasoning effort.
- Cover model discovery, display names, Codex allowlisting, and reasoning behavior.


libs/code/deepagents_code/model_config.py
```diff
@@ -594,6 +594,9 @@ class ProviderConfig(TypedDict, total=False):
 
 CODEX_MODELS: frozenset[str] = frozenset(
     {
+        "gpt-5.6-luna",
+        "gpt-5.6-sol",
+        "gpt-5.6-terra",
```
mdrxy, line 599: do we know this for sure?

mdrxy (approved):

libs/code/deepagents_code/model_config.py
johannes117: Yes. OpenAI’s [Codex model docs](https://developers.openai.com/codex/models) explicitly say Codex offers all three and identify them as `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna`. The current [`openai/codex` model catalog](https://github.com/openai/codex/blob/d72d669ca727a26765315ffe03db3dd2594d9b91/codex-rs/models-manager/models.json) also contains those exact three slugs, so these are confirmed ChatGPT-sign-in/Codex identifiers rather than inferred from API availability.

PR #4613, opened 2026-07-09
feat(sdk): support prefixed skill sources

johannes117:
Skill sources can now provide an optional name prefix and can point directly at a skill directory containing `SKILL.md`.

---

Consumers that aggregate skills from multiple independent sources need to qualify names without modifying source files or maintaining a projected cache.

`SkillsMiddleware` now accepts `(path, label, name_prefix)` sources and applies the prefix after validating the underlying skill metadata. Sources can also point directly to a skill directory containing `SKILL.md`. Existing string and two-item source forms retain their behavior.

### DIFF
--- a/libs/deepagents/deepagents/middleware/skills.py
+++ b/libs/deepagents/deepagents/middleware/skills.py
@@ -53,14 +53,15 @@ Parsed from YAML frontmatter per Agent Skills specification:
 
 ## Sources
 
-Sources point to skill directories in the backend. Each source is either a bare
-path or a `(path, label)` tuple. With a bare path the label is derived from the
-last path component capitalized (e.g., `/skills/user/` -> `User`), with two
-special cases: `built_in_skills` collapses to `Built-in`, and a literal `skills`
-leaf climbs one level so `~/.claude/skills` renders as `Claude` rather than the
-duplicative `Skills Skills`. Pass an explicit tuple to disambiguate sources
-whose leaf directories would collide (e.g. user- vs project-scoped
-`.claude/skills`).
+Sources point to skill directories in the backend. Each source is a bare path,
+a `(path, label)` tuple, or a `(path, label, name_prefix)` tuple. With a bare
+path the label is derived from the last path component capitalized (e.g.,
+`/skills/user/` -> `User`), with two special cases: `built_in_skills` collapses
+to `Built-in`, and a literal `skills` leaf climbs one level so
+`~/.claude/skills` renders as `Claude` rather than the duplicative
+`Skills Skills`. Pass an explicit tuple to disambiguate labels or qualify names
+from a source. Prefixes are applied after source skill validation, so qualified
+names may use namespace separators that are not valid in source frontmatter.
 
 Example sources:
 ```python
@@ -69,6 +70,7 @@ Example sources:
     "/skills/project/",
     ("/home/me/.claude/skills", "User Claude"),
     ("/repo/.claude/skills", "Project Claude"),
+    ("/plugins/review/skills", "Review Plugin", "review:"),
 ]
 ```
 
@@ -140,6 +142,7 @@ logger = logging.getLogger(__name__)
 MAX_SKILL_FILE_SIZE = 10 * 1024 * 1024
 MAX_SKILLS_LOAD_WARNINGS = 20
 MAX_SKILL_LOAD_WARNING_LENGTH = 1000
+_SKILL_SOURCE_WITH_PREFIX_LEN = 3
 _SKILL_LOAD_WARNING_TRUNCATION_SUFFIX = "... [truncated]"
 
 # Agent Skills specification constraints (https://agentskills.io/specification)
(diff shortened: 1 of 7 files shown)


libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -140,21 +142,26 @@
...
 _SKILL_LOAD_WARNING_TRUNCATION_SUFFIX = "... [truncated]"
 
 # Agent Skills specification constraints (https://agentskills.io/specification)
 MAX_SKILL_NAME_LENGTH = 64
 MAX_SKILL_DESCRIPTION_LENGTH = 1024
 MAX_SKILL_COMPATIBILITY_LENGTH = 500
 
-SkillSource = str | tuple[str, str]
-"""A skill source: either a bare path or a `(path, label)` pair.
+SkillSource = str | tuple[str, str] | tuple[str, str, str]
```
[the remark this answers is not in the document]
johannes117: Addressed in 9205999eb. `create_deep_agent.skills` and `SubAgent.skills` now use `list[SkillSource]`, their public documentation covers all three source forms, and the existing end-to-end tests exercise prefixed sources through both entry points. Focused tests and `make lint` pass.

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -166,11 +173,12 @@ def _validate_tuple_source(source: tuple[object, ...]) -> None:
     middleware or a silently-coerced non-string path downstream.
     """
     if (
-        len(source) != 2  # noqa: PLR2004  # SkillSource tuple is exactly (path, label)
+        len(source) not in {2, _SKILL_SOURCE_WITH_PREFIX_LEN}
```
aolsenjazz, line 176: nit: If we're going to use a const for skill source with prefix len, we should probably just use a const for non-prefixed tuple length as well

libs/deepagents/deepagents/middleware/skills.py
aolsenjazz, line 169:
```suggestion
    """Raise `TypeError` if a tuple source is not str | (str, str) | (str, str, str).
```

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -190,6 +198,14 @@ def _truncate_skill_load_warning(error: str) -> str:
     return f"{error[:length]}{_SKILL_LOAD_WARNING_TRUNCATION_SUFFIX}"
 
 
+def _source_name_prefix(source: SkillSource) -> str:
+    """Return an optional prefix applied to skill names from a source."""
+    if isinstance(source, str):
+        return ""
+    _validate_tuple_source(source)
+    return source[2] if len(source) == _SKILL_SOURCE_WITH_PREFIX_LEN else ""
```
aolsenjazz, line 206: nit: use const value vs magic number

libs/deepagents/deepagents/middleware/skills.py
aolsenjazz, line 1: Need to preface this by saying I'm a typescript guy who has written a lot of python, but I haven't written python daily in years

Main question (and I'm not suggesting this is incorrect) is - how did you make the decision to concat skill prefix at load time (or whenever) vs adding an optional prefix field to the `SkillMetadata` TypeDict? My gut says that a lossy concat may make things ugly in the future since we can reliably parse back to prefix/name

aolsenjazz, line 696: Lets update this docstring to reflect the direct_skill_md_path potential

aolsenjazz: We're also violating the `SkillMetadata.name` invariant documented here:

```
    name: str
    """Skill identifier.
    
      Constraints per Agent Skills specification:
    
      - 1-64 characters
      - Unicode lowercase alphanumeric and hyphens only (`a-z` and `-`).
      - Must not start or end with `-`
      - Must not contain consecutive `--`
      - Must match the parent directory name containing the `SKILL.md` file
      """
```

unsure if any consequences with this

aolsenjazz: @mdrxy ?

aolsenjazz, line 328: This function should also in theory fail for prefixed skill names

libs/deepagents/tests/unit_tests/middleware/test_skills_middleware.py
```diff
@@ -715,13 +741,14 @@ def test_list_skills_logs_ls_error(caplog: pytest.LogCaptureFixture) -> None:
...
     backend.ls.return_value = LsResult(error="Cannot list '/bad': denied", entries=[])
+    backend.download_files.return_value = [FileDownloadResponse(path="/bad/SKILL.md", content=None, error="file_not_found")]
 
     with caplog.at_level(logging.WARNING):
         skills = _list_skills(backend, "/bad")
 
     assert skills == []
     assert "Cannot load skills from '/bad'" in caplog.text
-    backend.download_files.assert_not_called()
+    backend.download_files.assert_called_once_with(["/bad/SKILL.md"])
```
aolsenjazz, line 751: Looks like this is fixing a preexisting test yes?

johannes117:
closing as we are going to go in a different direction

PR #4623, opened 2026-07-10
perf(infra): install only the model integrations a Harbor job needs

nick-hollon-lc:
Only include the model integrations that the model list needs to improve the performance of Harbor jobs.

Also fixes an issue where non-Fireworks jobs failed to install because of the Fireworks prerelease dependency: `langchain-fireworks` was being installed every time but without `prerelease=allow`.

### DIFF
--- a/.github/scripts/prune_agent_deps.py
+++ b/.github/scripts/prune_agent_deps.py
@@ -128,14 +128,19 @@ def main() -> None:
     deps = config.get("dependencies", [])
 
     if provider not in PROVIDER_TO_PACKAGE:
-        # Ad-hoc / newly-added provider (e.g. a one-off `models_override`): we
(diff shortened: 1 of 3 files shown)


.github/scripts/prune_agent_deps.py
```diff
@@ -0,0 +1,181 @@
...
+        python3 prune_agent_deps.py path/to/langgraph.json
+"""
+
+from __future__ import annotations
+
+import json
+import os
+import sys
+
+PROVIDER_TO_PACKAGE: dict[str, str] = {
```
mdrxy, line 29: (nit/nb) i feel like this is something that probably exists somewhere else in our codebase we could re-use

mdrxy (approved):

PR #4626, opened 2026-07-10
fix(code): preserve `Ctrl+D` deletion in non-empty input

wbbradley:
Ctrl+D now deletes the character to the right in a non-empty prompt and only exits from an empty prompt.

---

A priority app binding intercepted Ctrl+D before the focused chat text area could apply its standard delete-right behavior. Delegate Ctrl+D to the focused text area while a draft exists, and retain quit behavior for an empty prompt and existing modal flows.

Made by [Open SWE](https://openswe.vercel.app/agents/9c524bf4-b1c3-5e11-d131-8d5d44be990b)


libs/code/deepagents_code/app.py
```diff
@@ -12587,6 +12587,12 @@ def action_quit_app(self) -> None:
                 return
             self._arm_quit_pending("Ctrl+D")
             return
+        chat_input = self._chat_input
```
aolsenjazz, line 12693:
```suggestion

        chat_input = self._chat_input
```

libs/code/deepagents_code/app.py
```diff
@@ -12587,6 +12587,12 @@ def action_quit_app(self) -> None:
...
+        if chat_input is not None:
+            text_area = chat_input.input_widget
+            if text_area is not None and self.focused is text_area and chat_input.value:
+                text_area.action_delete_right()
+                return
```
aolsenjazz, line 12698:
```suggestion
                return

```

aolsenjazz (approved):
Just a couple new lines for readability, nbd either way. Thanks for submitting this!

PR #4629, opened 2026-07-10
fix(code): reap langgraph dev server when startup is cancelled

wbbradley:
## Problem

Pressing `Ctrl+D` shortly after launching `dcode` returns you to the terminal but leaves an orphaned `python -m langgraph_cli dev …` process running:

```
$ ps aux | grep deepagent
wbbradley  69643  ...  python3 -m langgraph_cli dev --host 127.0.0.1 --port 59934 --no-browser --no-reload --config .../langgraph.json
```

This is not intentional — it's a process leak.

## Root cause

To make the splash screen appear instantly, the LangGraph server is started in a background Textual worker (`_start_server_background`) rather than blocking launch. That worker calls `start_server_and_get_agent`, which spawns the `langgraph dev` subprocess early (inside `ServerProcess.start()`) and then spends a few seconds polling `/ok` for health + graph readiness.

If the user quits during that window:

1. `super().exit()` tears down Textual, which **cancels all workers**, injecting `asyncio.CancelledError` into the in-flight `server.start()`.
2. `start_server_and_get_agent` guarded cleanup with `except Exception:` — but `CancelledError` is a `BaseException`, **not** an `Exception`, so `server.stop()` was skipped and the exception re-raised.
3. Because the function never returned, `DeepAgentsApp._server_proc` was never assigned, so the belt-and-suspenders `finally` in `run_textual_app` also saw `None` and skipped cleanup.

Net result: the `langgraph dev` subprocess is orphaned. When the server has fully finished starting before the quit, `stop()` runs correctly and there is no leak — the leak is specific to quitting *during* startup.

## Fix

Catch `BaseException` (which includes `asyncio.CancelledError`) around `server.start()` / `wait_for_graph_ready()` so a cancelled startup still stops the half-started subprocess, then re-raise to preserve cancellation/shutdown semantics. This matches the existing cleanup-on-cancel-and-reraise pattern already used in `skills/trust.py`.

### DIFF
--- a/libs/code/deepagents_code/client/launch/server.py
+++ b/libs/code/deepagents_code/client/launch/server.py
@@ -528,6 +528,7 @@ class ServerProcess:
             stderr=subprocess.STDOUT,
         )
 
+        started = False
         try:
             await wait_for_server_healthy(
                 self.url,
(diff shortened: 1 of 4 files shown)


aolsenjazz:
LGTM, @mdrxy would you confirm that widening this exception type is the most-correct solution?

mdrxy (approved):

PR #4637, opened 2026-07-10
feat(evals): add context-retrieval-evals Harbor dataset (30 calibrated Context-Bench tasks)

nick-hollon-lc:
Adds `context-retrieval-evals`: a 30-task Harbor dataset measuring multi-file retrieval + reasoning, derived from Context-Bench (cloud suite). Each task ships the whole corpus so the agent must retrieve/join/aggregate rather than guess which files matter.

**Tasks — frozen from calibration** (Opus 4.8 / bare harness / 3 rollouts): 5 easy (solved cold) · 9 medium (intermittent) · 16 hard (failed cold). Context-Bench is bimodal on Opus 4.8, so tiers are assigned by measured pass-rate; each task's pass-rate + what it tests is documented in the dataset README as tier justification.

**Adapter** (`harbor_adapters/contextbench`): generates task dirs from the vendored `filesystem_cloud.jsonl`.

**Corpus single-sourced**: the 64.7K-line corpus lives once under `vendor/files/`; per-task `environment/files/` is git-ignored and regenerated via `--populate` before runs (CI + local), keeping the diff small.

**`harbor.yml`**: local-dataset sharding, `n_retries` (→ Harbor `--max-retries`), and a corpus-populate step.

Follow-up (non-blocking): `pass@bare` is the weakest-harness floor — recalibrate on the `dcode` harness before reporting scorecard numbers (it will solve some of the hard tier).


mdrxy:
There is a verifier correctness issue here for currency answers.

The current normalization strips both commas and decimal points before comparison, so different amounts can collapse to the same string. For example, `$131,620.25` and `$1,316,202.5` both normalize to `13162025`, which could incorrectly award credit for a wrong answer.

Can we preserve decimal separators or parse these as numeric currency values while still tolerating `$` and thousands separators?

mdrxy:
Following on from the currency issue above — the same `tr -cd '[:alnum:][:space:]'` normalization also produces the *opposite* failure (correct answers scored `0.0`), so it's worth fixing both directions in one pass.

**False negatives from whitespace.** The normalization keeps `[:space:]` but never trims or collapses it, and `$(...)` strips only a trailing *newline*. So these correct answers all score `0.0` (verified in `dash`, the `python:3.12-slim` `/bin/sh`):

| agent writes | vs expected `Linda Robbins` | reward |
|---|---|:--:|
| `Linda Robbins` | ✓ | `1.0` |
| ` Linda Robbins` (leading space) | ✓ | **`0.0`** |
| `Linda Robbins ` (trailing space) | ✓ | **`0.0`** |
| `Linda  Robbins` (double space) | ✓ | **`0.0`** |
| `Linda<tab>Robbins` (tab) | ✓ | **`0.0`** |

27 of the 30 ground-truth answers are multi-word names, and trailing/irregular whitespace is a common LLM artifact even with the "answer and nothing else" instruction — so this likely deflates measured `pass@bare`, which matters because the tier calibration in `calibration.json` was produced by this same grader.

**Answer-format ambiguity** compounds it: `cb-cloud-16` expects bare `17`, but `cb-cloud-38` expects `18 records`. An agent can't infer which — `"18"` fails cb-cloud-38 and `"17 records"` fails cb-cloud-16.

**Suggested direction** — since `question_type` is already in `task.toml`, split the emitted comparison:

- **String/name answers:** lowercase, trim, and collapse internal whitespace (e.g. pipe both sides through `tr -s '[:space:]' ' '` + strip leading/trailing), then compare. Drop the `tr -cd` character deletion.
- **Numeric/currency answers** (`aggregation`, counting): strip `$` and thousands separators but **keep** `.` and `-`, then compare numerically. This resolves the `$131,620.25` ≡ `$1,316,202.5` collision and the whitespace false-negatives together, and lets count answers be format-tolerant (`17` ≡ `17 records`).

Worth adding a small test that actually *executes* the generated `test.sh` against sample answers (match / mismatch / whitespace variant / currency variant) — right now `test_generate_task_*` asserts the script's text but never runs it, so neither failure mode is caught.

nick-hollon-lc:
@mdrxy deferring the above two issues unless we see them in traces at which point they can be fixed. the verifier scripts come directly from context bench so modifying them changes our ability to compare to the underlying benchmark

mdrxy (approved):

PR #4650, opened 2026-07-10
feat(code): add Meta model provider

ccurme:
Adds the Meta Model API to the model switcher, credential manager, model discovery, and optional-provider installer.

The new `meta` extra installs the MIT-licensed `langchain-meta>=0.5.0` integration (published by the `langchain-ai` org), and the base `langchain` dependency is bumped to `>=1.3.13` so `init_chat_model` resolves the `meta` provider through the standard LangChain provider path. Verified against `langchain-meta` 0.5.0: `ChatMetaModel` reads `MODEL_API_KEY` for the key and `MODEL_API_BASE` for the base URL.

Made by [Open SWE](https://openswe.vercel.app/agents/4010ad30-bfc5-e865-9ab2-c08a10518889)

### DIFF
--- a/libs/code/deepagents_code/_constants.py
+++ b/libs/code/deepagents_code/_constants.py
@@ -13,6 +13,12 @@ from typing import Final
 DEFAULT_AGENT_NAME: Final[str] = "agent"
 """Default agent / assistant identifier when no `-a` flag is given."""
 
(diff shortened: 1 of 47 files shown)


aolsenjazz:
I think we would also want to add this to `_RECOMMENDED_MODELS` in `model_selector.py` so that it shows up in the model selector list without having to install the package first e.g. `"meta:muse-spark-1.1": "Muse Spark 1.1",`

PR #4651, opened 2026-07-10
ci(evals): add unified multi-model Harbor evals workflow

nick-hollon-lc:
Harbor evaluations currently require a separate workflow dispatch for each model and benchmark. This makes cross-model comparisons cumbersome and leaves Harbor without the multi-model orchestration already available in `evals.yml`.

This PR adds a unified Harbor entry point that accepts a comma-separated list of `provider:model` IDs and any combination of the autonomous, conversation, and context categories.

| Before | After |
|---|---|
| Each dispatch evaluated exactly one model. | One dispatch evaluates a list of models. |
| The dataset and agent implementation were selected manually for each run. | Each category selects the appropriate benchmark and agent configuration automatically. |
| Evaluating three categories across several models required a separate run for every pair. | The workflow coordinates the complete model × category matrix. |
| Each run produced an independent result. | Results are combined into a leaderboard, JSON summary, and radar chart. |
| Harbor execution was embedded in the single-model workflow. | Single-model and unified dispatches share the same reusable Harbor execution workflow. |

The categories map to the existing Harbor-backed evaluations:

- **Autonomous:** Harbor Index with the dcode harness
- **Conversation:** the tau3 subset with the conversational agent
- **Context:** the context-retrieval evaluation with the dcode harness

Models are grouped by provider so calls to a single provider are serialized while different providers can run in parallel. Per-model and global concurrency limits are enforced, and shard counts remain configurable by category. Each `(model × category)` pair is tracked in its own LangSmith dataset.

The final report includes pass@K and avg@K for each pair, a macro overall that weights categories equally, and a micro overall weighted by task count. The expected model × category grid is retained so missing or incomplete runs are reported rather than silently omitted.

The existing single-model Harbor dispatch keeps its current interface. Extracting its execution into a reusable workflow is an enabling refactor; multi-model, cross-category orchestration and combined reporting are the net-new functionality.


mdrxy:
I think this misses a valid result shape from `gh run download`. When the workflow only produces one `harbor-combined-*` artifact, GitHub extracts that artifact directly into the target directory, so the downloaded `summary.json` can live at `_leaves/summary.json` instead of `_leaves/<artifact-name>/summary.json`.

`aggregate_unified.py` currently discovers summaries by iterating child directories under `_leaves`, so the single-artifact case is skipped and later reported as a missing expected leaf even though the summary was downloaded successfully.

A concrete repro shape is one selected model plus one selected category. The aggregator should treat `_leaves/summary.json` as a valid leaf summary in addition to scanning per-artifact subdirectories.

mdrxy:
I think the default unified workflow can currently select a path that users cannot make runnable from the dispatch form. `unified_prep.py` maps the default `conversation` category to `agent_impl: tau3`. The tau3 path currently requires Harbor to forward MCP servers, and the existing error text says the pinned Harbor release does not do that yet unless a Harbor package override is provided.

The problem is that this unified workflow does not expose or pass `harbor_package_override`, so a user who runs the default category set can hit a tau3/Harbor failure without any workflow input they can use to supply the required override.

I think this needs one of these fixes before `conversation` remains in the default unified set: expose and pass through the Harbor package override for the leaf jobs, or change the default conversation config to an implementation that works with the pinned Harbor package.

mdrxy:
### Fix before the first live smoke run
- **`context` can't trigger a first-time snapshot build.** No `force_build` input is exposed/threaded, so it defaults `false`. On a fresh sandbox the `context` leaves will fail to build. — `unified_evals.yml`
- **Unvalidated `int()` on dispatch inputs** → bare traceback instead of a clean `SystemExit`; also no range check on `shard_parallel`/`n_shards` (negatives are silently floored to 1).

### Fix before trusting the numbers
- **`combine` download swallows all failures as "no artifacts."** Auth/network/rate-limit errors become success-with-warning → false ⚠️ "partial data" rows, job stays green. The sibling leaf already classifies + retries + `exit 1`; mirror it.
- **One malformed leaf aborts the entire combine.** `read_leaf`'s `json.loads` / `summary["rollouts_per_task"]` are unguarded, so a truncated artifact (exactly the failure the design claims to tolerate) kills ranking for *all* models. Wrap per-leaf, `::warning::` + skip.
- **`combine` exits 0 even when every model is incomplete** → fully-degraded run shows green. Emit `::error::` + non-zero when all-incomplete.

### Nice-to-have / confirm
- Dangling doc reference: `docs/superpowers/specs/2026-07-10-unified-evals-ci-design.md` doesn't exist in-tree or in history.
- Test comment describes micro math the code doesn't do (code weights *both* pass@k and avg@k by tasks; assertion itself is correct).
- Confirm per-model LangSmith datasets are intentional (defeats native cross-model view; likely deliberate) + add a one-line note.
- `other`-bucket models run with no provider key → late runtime failure; fail fast in the credential step.
- Micro/macro `None` handling is asymmetric but currently *unreachable* (`None` iff `tasks==0`); add a regression-pinning test + comment.
- Housekeeping: dead `providers` output (`unified_prep.py:173`); `eval-assets` branch grows unbounded; `_load_list_env` unguarded `json.loads`.

mdrxy (approved):

PR #4657, opened 2026-07-11
feat(code): notify when a background async task finishes

### LINKED ISSUE #4656: Notify when a background async task finishes
### Submission checklist

### Area (Required)

- [ ] deepagents (SDK)
- [ ] cli
- [x] code
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Feature description

Async subagent tasks (`start_async_task` / `check_async_task` / `list_async_tasks` in `libs/deepagents/deepagents/middleware/async_subagents.py`) are tracked in `AsyncSubAgentState.async_tasks` (`async_subagents.py:125-130`), a `dict[str, AsyncTask]` keyed by `task_id`, where each `AsyncTask` carries `status`, `agent_name`, `thread_id`, `created_at`, `last_checked_at`, `last_updated_at` (`async_subagents.py:73-108`).

The tool description is explicit that the *model* should not poll this state in a loop:

> "If the status is 'running', report that and stop - do not poll in a loop." (`async_subagents.py:195`)
> "If a check returns 'running', tell the user and wait for them to ask again." (`async_subagents.py:206`)
> "Task statuses in conversation history are ALWAYS stale - a task that was 'running' may now be done." (`async_subagents.py:207`)

That's the right call for the model (polling in-loop burns turns and tokens for no reason), but it leaves a real gap on the human side: **nothing tells the user when a background task finishes either.** `async_tasks` is pure SDK/graph state — `libs/code/deepagents_code` (the TUI) has zero references to it anywhere in the package today (confirmed via full-package search). So a user who starts a long-running async task and switches away has no way to find out it's done except manually returning and asking the agent to check again.

Add a lightweight, TUI-side background poller — entirely out-of-band from the model loop, so it costs no tokens and doesn't touch agent reasoning at all — that watches tracked async task IDs and fires a toast (and a new hook event) when one transitions to a terminal state.

**Polling loop** — `libs/code/deepagents_code/app.py`, modeled directly on the existing periodic-worker pattern used for update checks (`app.py:3348-3354`):

```python
_ASYNC_TASK_POLL_INTERVAL_SECONDS = 15

self.set_interval(
    _ASYNC_TASK_POLL_INTERVAL_SECONDS,
    lambda: self.run_worker(
        self._poll_async_tasks(),
        exclusive=True,
        group="async-task-poll",
    ),
)
```

**State read** — reuse the existing `aget_state` mechanism already used elsewhere in `app.py` (e.g. `app.py:10386`, `app.py:11199`) to pull `async_tasks` off the current thread's checkpoint:

```python
async def _poll_async_tasks(self) -> None:
    if not self._agent or not self._async_task_ids:
        return
    state = await self._agent.aget_state(self._current_config())
    tasks = state.values.get("async_tasks", {})
    for task_id in list(self._async_task_ids):
        task = tasks.get(task_id)
        if task is None:
            continue
        if task["status"] in {"success", "error", "cancelled"} and task_id not in self._notified_task_ids:
            self._notify_async_task_done(task)
            self._notified_task_ids.add(task_id)
```

**Notification** — reuse the existing toast mechanism (`self.notify(...)`, e.g. `app.py:15415`):

```python
def _notify_async_task_done(self, task: AsyncTask) -> None:
    self.notify(
        f"Background task '{task['agent_name']}' finished ({task['status']}).",
        timeout=8,
        markup=False,
    )
    from deepagents_code.hooks import dispatch_hook_fire_and_forget
    dispatch_hook_fire_and_forget("async_task.complete", {
        "task_id": task["task_id"],
        "agent_name": task["agent_name"],
        "status": task["status"],
    })
```

A new hook event name, `async_task.complete`, is deliberately used instead of the existing `task.complete` event. `task.complete` is already dispatched exclusively for the headless run finishing (`client/non_interactive.py:1195`) — reusing it here would conflate "the current dcode invocation is done" with "a background subagent finished," which is a different signal a hook script would want to distinguish. This gives users with existing hook scripts (`hooks.py:59` documents the event dispatch system) an extension point to route to a desktop notification (`terminal-notifier`, `osascript`, etc.) without dcode needing to own OS-notification code itself.

`self._async_task_ids` would need to be populated whenever `start_async_task`/`astart_async_task` returns a new task — the simplest hook point is the same place the TUI already surfaces tool results in the message stream, since the task ID is already present in that tool's return payload.

Scoped deliberately to session-local tracking: only tasks started in the current TUI session are watched. Tasks resumed from a prior session's `async_tasks` state (e.g. after `-r`) would need `_async_task_ids` seeded from that state on thread load — noting as a possible follow-up, not blocking v1.

- `AsyncTask` fields: `async_subagents.py:73-108`.
- `AsyncSubAgentState.async_tasks`: `async_subagents.py:125-130`.
- No-polling guidance for the model: `async_subagents.py:195`, `206`, `207`.
- Existing periodic-worker precedent (`set_interval` + `run_worker(exclusive=True, group=...)`): `app.py:3348-3354`.
- Existing `agent_state` usage from the TUI layer: `app.py:10386`, `app.py:11199`.
- Existing toast precedent: `app.py:15415` (`self.notify(...)`).
- Existing hook dispatch precedent and event list: `hooks.py:59`, `client/non_interactive.py:1195` (`task.complete`, headless-run-only — not reused here to avoid ambiguity).
- Related but distinct: #1736 ("Tail background task logs") is about streaming *shell* background-task stdout live, not async-subagent terminal-state notification. #4343 ("parent-side lifecycle hooks for synchronous subagent task completion") is SDK-level and about the *sync* `task` tool's in-graph post-processing hooks, not user-facing notification for the *async* task tool. Neither overlaps with this proposal.

Happy to open the PR for this

NinaadRao:
Fixes #4656

---

Async subagent tasks started via `start_async_task` are tracked in `AsyncSubAgentState.async_tasks`, but the tool description explicitly tells the model not to poll `check_async_task` in a loop — which is the right call for the model, but left nothing telling the *user* when a background task finishes. Someone who kicks off a long-running async task and looks away had no way to find out short of manually asking the agent to check again.

This adds a lightweight, TUI-side background poll (`_poll_async_tasks`, driven by a `set_interval` tick registered in `_post_paint_init`) that checks the current thread's `async_tasks` state every 15 seconds. The first time a tracked task reaches `success`, `error`, or `cancelled`, it shows a toast via `self.notify(...)` and dispatches a new `async_task.complete` hook event. It's entirely out-of-band from the agent loop, so it costs no tokens and never touches model reasoning, and it's a no-op whenever the current thread has no tracked async tasks (the common case).

A new hook event name (`async_task.complete`) is used rather than reusing the existing `task.complete`, which already reports something different — the current headless run finishing — so overloading it would make a hook script unable to tell the two apart.

**Automated:**
- Added `TestAsyncTaskCompletionNotifications` (6 tests) in `tests/unit_tests/test_app.py`: notifies on terminal status, no duplicate notification on a repeated poll, ignores `running` tasks, dispatches the hook with the expected payload, no-op when there are no tracked async tasks, no-op when there's no agent yet.
- `uv run --group test ruff check .`, `ruff format --check .`, and `ty check .` — all clean.
- `uv run --group test pytest tests/unit_tests/ -q` — full suite: 8470 passed, 3 skipped, plus the 6 new tests. 4 unrelated failures (`test_auth_display.py`, `test_status.py::TestBranchDisplay` — glyph/branch-icon rendering, nothing touched by this change) reproduce identically on `main` and pass when run in isolation, confirming pre-existing test-order flakiness, not a regression from this change.

**Manual, against the real running app:**
Standing up a real remote async-subagent server just to get a task to `success` is heavyweight for a one-off check, so I stubbed only the two things that server would normally produce (`agent=MagicMock()` — the same stand-in the app's own test suite already uses — and a patched `_get_thread_state_values` returning an `async_tasks` dict with one task at `status: "success"`), then ran the real, unmodified `DeepAgentsApp` in a real terminal (via `tmux`) with the poll interval shortened to 1s. The toast "Background task 'researcher' finished (success)." appeared exactly as expected. 

async_task_notification

### DIFF
--- a/libs/code/deepagents_code/app.py
+++ b/libs/code/deepagents_code/app.py
@@ -392,6 +392,15 @@ state as the single app-wide progress indicator.
 _UPDATE_RECHECK_INTERVAL_SECONDS = 60 * 60
 """How often long-running TUI sessions quietly re-check for app updates."""
 
+_ASYNC_TASK_POLL_INTERVAL_SECONDS = 15
+"""How often to check tracked async subagent tasks for a terminal status.
+
+Purely a TUI-side poll against checkpointed state (`_get_thread_state_values`)
+so a user who starts a background task and looks away finds out when it's
+done — the agent itself is explicitly told not to poll `check_async_task` in
+a loop, so nothing else surfaces this.
+"""
+
 _MODAL_WATCHDOG_TIMEOUT_SECONDS = 600.0
 """Upper bound on awaiting a confirmation modal's dismissal.
 
@@ -2270,6 +2279,13 @@ class DeepAgentsApp(App):
         self._pending_mcp_disable_reconnect_servers: set[str] = set()
         """MCP servers with disable-state changes waiting for reconnect."""
 
+        self._async_task_notified: set[str] = set()
+        """Async subagent `task_id`s already surfaced by `_notify_async_task_done`.
+
+        Prevents `_poll_async_tasks` from re-toasting the same completed task
+        on every subsequent poll tick.
+        """
+
         self._mcp_tool_count = sum(len(s.tools) for s in (mcp_server_info or []))
         """Total tool count across MCP servers, displayed in the status bar."""
 
@@ -3358,6 +3374,19 @@ class DeepAgentsApp(App):
                 exclusive=True,
                 group="startup-whats-new",
             )
+
+        # Watch for background async subagent tasks completing while the
+        # user's attention is elsewhere. Unconditional (not gated behind a
+        # config flag): `_poll_async_tasks` is a no-op whenever the current
+        # thread has no tracked async tasks, which is the common case.
+        self.set_interval(
+            _ASYNC_TASK_POLL_INTERVAL_SECONDS,
+            lambda: self.run_worker(
+                self._poll_async_tasks(),
+                exclusive=True,
+                group="async-task-poll",
+            ),
+        )
 
         # Prewarm model discovery and profile caches unconditionally so
         # /model opens instantly even before the agent/server is ready.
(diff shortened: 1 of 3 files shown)


mdrxy:
Thanks for taking the initiative here—we appreciate the thought and effort you put into this proposal. This change has broader API, security, and architecture implications and requires more design review before we pursue an implementation, so we’re closing this PR for now.

Going forward, please don’t open implementation PRs unless a maintainer has explicitly assigned the work to you. You’re welcome to share ideas on an issue or discussion first and wait for design alignment and assignment before starting implementation.

PR #4660, opened 2026-07-11
feat(sdk): add ToolSelectionMiddleware to filter tools sent to the model per turn

### LINKED ISSUE #4658: Add `ToolSelectionMiddleware` for per-turn tool filtering
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Feature description

The middleware stack in `graph.py` assembles the tool list once at agent construction (`graph.py:745-974` builds `gp_middleware`/`deepagent_middleware` per profile) and it stays fixed for the life of the session. `_ToolExclusionMiddleware` (`libs/deepagents/deepagents/middleware/_tool_exclusion.py`) can drop tools, but only statically — a name is either excluded for the whole session via a harness profile or it isn't; there's no way to change what's visible turn-to-turn based on what the agent is actually doing.

With just the built-in tools (filesystem tools, `execute`, `task`, async task management, `write_todos`) you're already at roughly a dozen schemas. Add one MCP server with 10-15 tools and a skill or two, and a session is well past 25-30 tool schemas sent on *every single model call*, regardless of whether that turn touches any of them. Concretely: `app.py:2273` already tracks `_mcp_tool_count = sum(len(s.tools) for s in mcp_server_info)` purely for UI display — that count has no effect on what's actually sent to the model.

This was raised before as #3672, with a fairly complete design; the idea was never actually built or explicitly rejected by a maintainer. Re-raising with a narrower, dependency-free default.

An opt-in `AgentMiddleware` that filters `request.tools` in `wrap_model_call`/`awrap_model_call`, following the exact pattern already used by `_ToolExclusionMiddleware` (`_tool_exclusion.py:45-53`, `request.override(tools=filtered)`), but scoring dynamically instead of filtering a static excluded-set.

```python
class ToolSelectionMiddleware(AgentMiddleware[Any, Any, Any]):
    """Filters the tool list sent to the model down to the top-K most
    relevant tools for the current turn, plus an always-include set.
    """

    def __init__(
        self,
        *,
        top_k: int = 15,
        always_include: frozenset[str] = frozenset({"read_file", "write_file", "edit_file", "ls", "task"}),
        embeddings: Embeddings | None = None,
    ) -> None:
        self._top_k = top_k
        self._always_include = always_include
        self._embeddings = embeddings  # optional; falls back to lexical scoring
```

Two deliberate design choices to keep this mergeable without a new hard dependency:

1. **No new required dependency.** `libs/deepagents/pyproject.toml` currently has a lean dependency set (`langchain`, `langchain-core`, `langchain-anthropic`, `langchain-google-genai`, `langsmith`, `wcmatch`) with `aws`/`quickjs`/`video` as optional extras (`pyproject.toml:1-13`) — no embedding library. Default scoring should be a zero-dependency lexical overlap (tokenize the tool's name + description, tokenize the latest human message, score by overlap) — cheap, deterministic, and good enough to separate "clearly irrelevant" from "plausibly relevant" at the top-K cutoff this is meant for.
2. **Pluggable embeddings, not bundled ones.** `langchain_core.embeddings.Embeddings` is already part of the existing `langchain-core` dependency, so accepting an optional `embeddings: Embeddings | None` for higher-quality scoring costs nothing new — the caller brings their own configured embedding model (or none, and gets the lexical default).

**Escape hatch.** A `discover_tools(query: str)` tool is added alongside the middleware (or embedded in it) so the model is never permanently blind to a tool that didn't make the cut — it searches the *full* registry and the match is added to that turn's `always_include` set for the next call. This is the piece that makes filtering safe: worst case the model spends one extra tool call finding what it needs, instead of silently never knowing a tool exists.

```python
agent = create_deep_agent(
    model="anthropic:claude-sonnet-5",
    tools=[...],
    middleware=[ToolSelectionMiddleware(top_k=15)],
)
```

Fully opt-in via `create_deep_agent(middleware=[...])` (`graph.py:364`) — no change to any existing default behavior.

_No response_

NinaadRao:
[tool_selection_middleware_demo.py](https://github.com/user-attachments/files/29931715/tool_selection_middleware_demo.py)
[tool_selection_middleware_demo_before_main.py](https://github.com/user-attachments/files/29931716/tool_selection_middleware_demo_before_main.py)
Fixes #4658

---

The middleware stack assembles the tool list once at agent construction and it stays fixed for the life of the session. `_ToolExclusionMiddleware` can drop tools, but only statically — a name is either excluded for the whole session or it isn't; there's no way to change what's visible turn-to-turn based on what the agent is actually doing. With just the built-in tools you're already at roughly a dozen schemas; add an MCP server or two and a session is well past 25-30 tool schemas sent on *every* model call, regardless of whether that turn touches any of them.

This adds an opt-in `ToolSelectionMiddleware` that scores `request.tools` against the latest human message in `wrap_model_call`/`awrap_model_call` and filters down to the top-K most relevant tools plus a configurable `always_include` set, following the same `request.override(tools=...)` pattern `_ToolExclusionMiddleware` already uses. Scoring defaults to zero-dependency lexical token overlap (no new dependency needed); passing an `embeddings: Embeddings` instance switches to cosine-similarity scoring instead — `langchain_core.embeddings.Embeddings` is already part of the existing `langchain-core` dependency, so this costs nothing new.

The important part is the escape hatch: alongside the middleware, a `discover_tools` tool is injected that searches the *full* tool registry (via `ToolRuntime.tools`, which always holds the complete set the `ToolNode` was constructed with, unaffected by what a given turn filtered down to) and pins a match into the always-include set for the rest of the conversation. This is what makes filtering safe — worst case the model spends one extra tool call finding what it needs, instead of silently never knowing a tool exists. The pin is tracked as private per-thread graph state (`tool_selection_pinned`), not middleware-instance state, so it can't leak between concurrent sessions sharing the same middleware instance.

One implementation detail worth flagging for review: `discover_tools` takes a `runtime: ToolRuntime` parameter that the `ToolNode` injects automatically at call time, and this only works because the module does *not* use `from __future__ import annotations` — the injection detection inspects the tool's live function signature, and a postponed (stringified) annotation breaks it silently until the tool is actually invoked through a real agent. `async_subagents.py` and `subagents.py` already avoid the future import for the same reason; this follows that precedent.

Fully opt-in via `create_deep_agent(middleware=[ToolSelectionMiddleware(top_k=15)])` — no change to any existing default behavior.

Added `test_tool_selection.py` (11 tests) covering top-k limiting, `always_include` handling, lexical scoring, `discover_tools` pinning and no-match reporting, embeddings-based scoring, both sync and async `wrap_model_call`, and — most importantly — a regression test that drives `discover_tools` through a real `create_agent` + `ToolNode` invocation rather than calling the tool's underlying function directly. That last test exists because the direct-call style of the other tests didn't catch a real bug during manual verification: running the middleware end-to-end through an actual agent turned up the `ToolRuntime` injection issue described above, which unit tests calling `.func()` directly never exercised. `uv run --group test pytest`, `ruff check`, `ruff format --check`, and `ty check` are all clean, and the full existing middleware + graph test suites (561 tests) pass with no regressions.

I also ran the same 3-turn scenario (4 domain tools, a weather question, a math question, a translation question) against a real checkpointed agent both with and without the middleware, to see the actual effect on what gets sent to the model:

- Without the middleware (current `main` behavior): all 4 tool schemas are sent on every single turn, regardless of relevance.
- With `ToolSelectionMiddleware(top_k=1)`: the weather turn sends only `weather_lookup` + `discover_tools`; the math turn (worded so `calculator` doesn't score highly, forcing a real `discover_tools` call) correctly finds and pins `calculator`; the translation turn sends `translate_text` + the now-pinned `calculator` + `discover_tools` — `calculator` staying visible in a turn that has nothing to do with math is the pin surviving across the checkpointed thread, not a fluke.

Before the changes from this PR:
image

After the changes from this PR: 
image

### DIFF
--- a/libs/deepagents/deepagents/middleware/__init__.py
+++ b/libs/deepagents/deepagents/middleware/__init__.py
@@ -75,6 +75,7 @@ from deepagents.middleware.summarization import (
     SummarizationToolMiddleware,
     create_summarization_tool_middleware,
 )
+from deepagents.middleware.tool_selection import ToolSelectionMiddleware
 
 __all__ = [
     "DEEPAGENTS_DEFAULT_SUMMARY_PROMPT",
@@ -100,5 +101,6 @@ __all__ = [
     "SubAgentMiddleware",
     "SummarizationMiddleware",
     "SummarizationToolMiddleware",
+    "ToolSelectionMiddleware",
     "create_summarization_tool_middleware",
 ]
--- a/libs/deepagents/deepagents/middleware/tool_selection.py
+++ b/libs/deepagents/deepagents/middleware/tool_selection.py
@@ -122,10 +122,7 @@ def _lexical_score(tool: BaseTool | dict[str, Any], query_tokens: set[str]) -> f
     tool_tokens = _tokenize(f"{_tool_name(tool) or ''} {_tool_description(tool)}")
     if not tool_tokens:
         return 0.0
-    union = tool_tokens | query_tokens
-    if not union:
-        return 0.0
-    return len(tool_tokens & query_tokens) / len(union)
+    return len(tool_tokens & query_tokens) / len(tool_tokens | query_tokens)
 
 
 def _cosine_similarity(a: list[float], b: list[float]) -> float:
--- a/libs/deepagents/deepagents/middleware/tool_selection.py
+++ b/libs/deepagents/deepagents/middleware/tool_selection.py
@@ -34,37 +34,34 @@ Scoring defaults to zero-dependency lexical token overlap between the latest
 Embeddings` instance for cosine-similarity scoring instead.
 """
 
-from __future__ import annotations
-
 import re
-from typing import TYPE_CHECKING, Annotated, Any, NotRequired
+from collections.abc import Awaitable, Callable
+from typing import Annotated, Any, NotRequired
 
 from langchain.agents.middleware.types import (
     AgentMiddleware,
     AgentState,
     ContextT,
+    ExtendedModelResponse,
+    ModelRequest,
+    ModelResponse,
     PrivateStateAttr,
     ResponseT,
 )
-from langchain.tools import (
-    ToolRuntime,  # noqa: TC002  # must stay a real import: the tool node resolves this annotation at runtime to inject `ToolRuntime`
-)
-from langchain_core.messages import HumanMessage, ToolMessage
-from langchain_core.tools import StructuredTool
+from langchain.tools import ToolRuntime
+from langchain_core.embeddings import Embeddings
+from langchain_core.messages import AIMessage, AnyMessage, HumanMessage, ToolMessage
+from langchain_core.tools import BaseTool, StructuredTool
 from langgraph.types import Command
 from pydantic import BaseModel, Field
 
-if TYPE_CHECKING:
-    from collections.abc import Awaitable, Callable
-
-    from langchain.agents.middleware.types import (
-        ExtendedModelResponse,
-        ModelRequest,
-        ModelResponse,
-    )
-    from langchain_core.embeddings import Embeddings
-    from langchain_core.messages import AIMessage, AnyMessage
-    from langchain_core.tools import BaseTool
+# NOTE: this module intentionally does NOT use `from __future__ import annotations`.
+# `discover_tools` below relies on the `runtime: ToolRuntime` parameter annotation
+# being a real, live type at function-definition time -- the tool node's injection
+# detection (`StructuredTool._injected_args_keys`) inspects `inspect.signature(fn)`
+# directly and only recognizes injected params when the annotation is an actual
+# class, not a postponed (stringified) one. Same reasoning as `async_subagents.py`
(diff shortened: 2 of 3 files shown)


mdrxy:
Thanks for taking the initiative and exploring tool selection in depth—we appreciate the effort here. Dynamically filtering tool schemas, defining the model’s discovery path, and persisting selected tools introduce broader product and middleware design questions. We need to review those tradeoffs and agree on the intended design before considering an implementation, so we’re closing this PR for now. We may reference or reopen it later, depending on the approach we select.

Going forward, please don’t open implementation PRs unless a maintainer has explicitly assigned the work to you. Please begin with an issue or discussion and wait for design alignment and assignment before implementing.

PR #4663, opened 2026-07-12
feat(code): support configured OAuth clients for remote MCP

jkennedyvz:
Closes DCD-42

---

Remote MCP servers that require pre-registered OAuth applications, including Google Gmail MCP, cannot use dynamic client registration. This adds an optional `oauth` client configuration for remote HTTP and SSE servers.

This is useful for providers that require users to create their own OAuth application and register an exact callback URI before authorization. Without configured credentials, dcode can only use the generic dynamic-registration flow, which Google does not support for Gmail MCP.

For Gmail MCP, create an OAuth web client in Google Cloud, register a loopback callback such as `http://localhost:8765/callback`, and configure dcode with environment-backed credentials:

```json
{
  "mcpServers": {
    "gmail": {
      "type": "http",
      "url": "https://gmail.mcp.example/mcp",
      "auth": "oauth",
      "oauth": {
        "clientId": "${GOOGLE_CLIENT_ID}",
        "clientSecret": "${GOOGLE_CLIENT_SECRET}",
        "redirectUri": "http://localhost:8765/callback"
      }
    }
  }
}
```

Run `dcode mcp login gmail` to authorize the account. dcode binds the configured callback port and uses the exact URI for authorization and token exchange. Keep the client secret in an environment variable; it is resolved only when Gmail MCP is activated and is not persisted with tokens.

The configured client is validated before use, resolves environment references only when the selected server activates, binds the registered loopback callback port for interactive login, and uses `client_secret_post` for authorization-code and refresh exchanges. Client secrets remain in memory and configured credentials take precedence over stale dynamic registrations.

## Release note

`dcode` remote MCP configurations can now supply pre-registered OAuth client credentials for providers that do not support dynamic client registration.

### DIFF
--- a/libs/code/deepagents_code/mcp_auth.py
+++ b/libs/code/deepagents_code/mcp_auth.py
@@ -32,6 +32,7 @@ from urllib.parse import parse_qs, urlparse
 import httpx
 from filelock import FileLock, Timeout
 from mcp.client.auth import OAuthClientProvider, TokenStorage
+from mcp.client.auth.oauth2 import get_client_metadata_scopes
 from mcp.client.auth.utils import (
     build_oauth_authorization_server_metadata_discovery_urls,
     build_protected_resource_metadata_discovery_urls,
@@ -1496,7 +1497,10 @@ class _ExpiryAwareOAuthClientProvider(OAuthClientProvider):
             self.context.protocol_version = request.headers.get(MCP_PROTOCOL_VERSION)
             if (
                 not self.context.is_token_valid()
-                and self.context.can_refresh_token()
+                and (
+                    self.context.can_refresh_token()
+                    or self._configured_client_info is not None
+                )
                 and self.context.oauth_metadata is None
             ):
                 # Pre-empt the SDK's 401-path discovery so its refresh branch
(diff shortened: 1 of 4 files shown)


rpelevin:
I independently checked current head `8768921d8fe02260d36cb99430236311a1a80a01` against two configured-client lifecycle transitions that are easy to miss when the token file already contains dynamic client registration (DCR) state.

1. DCR to configured migration, followed by a configured-client token write and process restart. `_initialize()` loads the stale DCR `client_info` before replacing it in memory. Because that DCR identity remains on disk, the next restart sees another client-ID mismatch and clears the fresh configured-client token again.
2. Expired configured-client access token with a usable refresh token and cached resource/OAuth metadata. The proactive configured-client branch runs before the serialized refresh branch, so it starts full authorization even while `can_refresh_token()` is true.

I reproduced both as focused regression tests on that head: `2 failed`. The smallest local patch I found was to:

- remove stale stored DCR `client_info` during migration, without persisting the configured secret;
- gate proactive authorization on `not self.context.can_refresh_token()`;
- add restart-survival and refresh-before-authorization regressions.

With that patch, the focused tests are `2 passed` and the complete `test_mcp_auth.py` file is `143 passed`.

This is local review evidence against the current PR head, not validation of a merged fix or a live Gmail/MCP authorization flow. I can provide the exact patch and tests if useful.

PR #4664, opened 2026-07-12
docs(code): document Gmail MCP OAuth setup

jkennedyvz:
Closes DCD-43

---

Document the configured OAuth-client workflow directly in `dcode mcp login --help`, using Google Gmail MCP as a complete example. The guidance shows environment-variable references for client credentials, requires the same localhost callback URI in Google Cloud, and makes the loopback-only limitation explicit.

This keeps sensitive credentials out of project configuration while preserving the existing `auth: "oauth"` flow for remote servers without a configured client.

The rendered help screen has focused regression coverage.


mdrxy:
I think this is better for [the docs repo](https://github.com/langchain-ai/docs)

Issue #3838, opened 2026-06-10
Add structured payload support for subagent task calls


sirano1004:
### Submission checklist

- [x] This is a feature request, not a bug report.
- [x] I searched existing issues and didn't find this feature.
- [x] I checked the docs and README for existing functionality.
- [x] This request applies to this repo (deepagents) and not an external package.

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] langsmith-sandbox
- [ ] Other / not sure / general

### Feature description

I want subagents to support optional `input_schema`, similar to why agents support `response_schema`.

Right now the supervisor can only pass a string task description to a subagent. That works for simple cases, but it is hard to reliably pass structured context like goal, scope, constraints, references, etc.

This makes subagent handoff less reliable when the subagent needs specific structured input. For example, a supervisor may want to delegate work to a research agent, coding agent, or weather agent with typed structured data, instead of only giving natural language instructions.

It also avoids creating a separate custom task tool schema for every subagent.

It would be useful if a subagent could define something like:

```python
{
    "name": "reviewer",
    "description": "Reviews code changes",
    "input_schema": ReviewTaskInput,
}
```

### Proposed solution (optional)

I already have a local implementation for this.

If assignment is required before opening PR, please assign me and I will open the PR.

The implementation is roughly like this:

```python
class SubAgent(TypedDict):
    name: str
    description: str
    system_prompt: str
    input_schema: NotRequired[type]
```

The task tool gets an optional `payload`:

```python
class TaskToolSchema(BaseModel):
    description: str
    subagent_type: str
    payload: dict[str, Any] | None = None
```

Then before invoking the selected subagent, it checks if that subagent has `input_schema`:

```python
if input_schema is not None:
    if payload is None:
        return "`payload` is required for this subagent"

    adapter = TypeAdapter(input_schema)
    validated = adapter.validate_python(payload)

    content = json.dumps({
        "description": description,
        "payload": adapter.dump_python(validated, mode="json"),
    })
```

If the subagent does not have `input_schema`, the payload can still be forwarded as raw payload.

This keeps the task tool generic. It does not create one custom task tool schema for each subagent. The selected subagent just validates the payload at runtime.

[...]

Issue #2512, opened 2026-04-07
Preserve full message content from `CompiledSubAgent`


SP-John:
## Problem

When a `CompiledSubAgent` returns structured data alongside its final message (e.g., via `additional_kwargs` or by setting a `ToolMessage.artifact`), the upstream `_return_command_with_state_update` in `subagents.py` discards everything except `.text`:

```python
message_text = result["messages"][-1].text.rstrip() if result["messages"][-1].text else ""
return Command(
    update={
        **state_update,
        "messages": [ToolMessage(message_text, tool_call_id=tool_call_id)],
    }
)
```

This means there is no supported way for a `CompiledSubAgent` to return structured data to the parent agent through the `ToolMessage`. The `structured_response` state key is explicitly excluded via `_EXCLUDED_STATE_KEYS`, `additional_kwargs` on the last message are dropped, and -- critically -- if `content` is a list of standard content blocks rather than a plain string, the non-text blocks are also lost.

LangChain Core's [standard content blocks](https://docs.langchain.com/oss/python/langchain/messages#standard-content-blocks) provide a typed, provider-agnostic way to carry structured data alongside text in a single message (e.g., `NonStandardContentBlock` for application-specific payloads). A `CompiledSubAgent` could naturally use this:

```python
content = [
    {"type": "text", "text": "Summary of results..."},
    {"type": "non_standard", "value": {"sql": "SELECT ...", "tier": 2, "rows": 14}},
]
return {"messages": [AIMessage(content=content)]}
```

But because `_return_command_with_state_update` extracts `.text` (which flattens to only `TextContentBlock` entries), the `NonStandardContentBlock` is silently discarded.

The docs say:

> The structured object itself is not returned to the parent agent. When using structured output with subagents, include the structured data in the ToolMessage.

But for `CompiledSubAgent`, users don't control the `ToolMessage` construction -- the framework creates it from `.text` only. The guidance is impossible to follow for compiled subagents, whether using `additional_kwargs`, `artifact`, or standard content blocks.

## Use case

I have a `CompiledSubAgent` that runs a multi-node LangGraph pipeline. The final node produces a human-readable summary as the message content, plus a structured payload containing intermediate results built across the various nodes. The parent agent needs the summary for conversation, but the API consumer needs the structured payload for other purposes.

Today the only options are:

1. Serialize the structured data into the message text (increases parent context token cost, parent LLM may post-process it).
[...]

Issue #606, opened 2025-12-23
[Feature Request] Support for passing file_uri / multimodal context to subagents via the task tool


dcarnelossi:
Description: I am currently working with deep-agents (v0.3.0) using Google Gemini (via langchain-google-genai and google-generativeai) to build a multi-agent system for analyzing complex PDF documents.

My workflow involves uploading a file using the Google File API and passing the file_uri to the main "Orchestrator" agent. While the Main Agent correctly receives and identifies the HumanMessage containing the type: media and file_uri, it is unable to natively pass this context to a subagent.

The Problem: The built-in task tool (used to spawn/delegate to subagents) seems to be designed primarily for text-based instructions (description). It lacks a mechanism to propagate message artifacts, specifically file_uri or full BaseMessage objects, to the subagent's input state.

Currently, the only workaround is to prompt-engineer the Main Agent to manually extract the URI string and paste it into the text description of the task, which is error-prone and loses the semantic meaning of the attachment as a proper message block.

To Reproduce:

Initialize a DeepAgent (Orchestrator) and a Subagent.

Upload a PDF using genai.Client().files.upload to get a file_uri.

Invoke the Main Agent with a message containing the media block:

Python

```
message = HumanMessage(
    content=[
        {"type": "text", "text": "Analyze this file"},
        {"type": "media", "file_uri": "https://...", "mime_type": "application/pdf"}
    ]
)
agent.invoke({"messages": [message]})
```

Observation: The Main Agent receives the file, but when it calls the task tool to delegate to the Subagent, the task schema does not support an attachments or file_uri field. The subagent starts with a fresh state containing only the text instructions, losing access to the file unless explicitly "hacked" into the text prompt.

Expected Behavior: The task tool (and the underlying subagent instantiation logic) should support passing media references or allowing the messages history containing the file_uri to be forwarded to the subagent.

Ideally, the task tool signature could look something like this:

Python

```
def task(description: str, attachments: List[dict] = None, ...):
    # logic to inject attachments into the subagent's state
```

Environment:

deepagents version: 0.3.0

google-generativeai version: 0.8.5

langchain-google-genai version: 4.0.0

Python version: 3.13

Additional Context: This feature is critical for "pass-by-reference" architectures where we want to keep token usage low by avoiding OCR/Text extraction and leveraging the native multimodal capabilities of models like Gemini 1.5 Pro within subagents.

eyurtsev:
Do all of your sub-agents need to work with multi-modal content?

[...]

PR #4669, opened 2026-07-12
feat(sdk): add fork option to task tool so subagents can inherit pare…

### LINKED ISSUE #4668: Add `fork` option to the `task` tool so subagents can inherit parent conversation history
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Feature description

Every subagent invocation through the `task` tool starts from a blank slate. `_validate_and_prepare_state` (`libs/deepagents/deepagents/middleware/subagents.py:658-669`) copies the parent's state but always overwrites `messages`:

```python
subagent_state["messages"] = [HumanMessage(content=description)]
```

(`subagents.py:668`)

The tool description tells the calling model exactly why this matters:

> "Each agent invocation is stateless. You will not be able to send additional messages to the agent, nor will the agent be able to communicate with you outside of its final report. Therefore, your prompt should contain a highly detailed task description..." (`subagents.py:295`)

That isolation is the right default — it's what keeps subagent context windows small and focused. But it means there is no way today to hand a subagent the conversation *so far*. If a user is 40 turns into a debugging session and wants to delegate a parallel investigation, the only option is to hand-summarize everything relevant into `description`, which costs tokens, loses nuance, and duplicates work the parent agent already did.

`_validate_and_prepare_state` already has full access to the parent's history at that exact call site — `runtime.state["messages"]` — before it gets thrown away. This is a small, surgical gap, not a missing subsystem.

This is distinct from two adjacent, already-filed issues:
- #2440 ("Async subagents option to pass certain state vars from the parent") is about forwarding specific *state variables* (e.g. an `access_token`), not conversation history.
- #1359 ("Fork session / rewind command") is a user-facing TUI feature for branching/rewinding the *user's own thread* in `dcode`, not an SDK-level primitive for what a subagent sees when the *model* delegates work via `task`.

Add a `fork: bool = False` parameter to the `task` tool. When `True`, the subagent's initial `messages` is the parent's full history plus the task description, instead of just the description.

**`subagents.py:272-282`** — extend the schema:

```python
class TaskToolSchema(BaseModel):
    """Input schema for the `task` tool."""

    description: str = Field(...)
    subagent_type: str = Field(...)
    fork: bool = Field(
        default=False,
        description=(
            "If True, the subagent starts with the full conversation history so far "
            "instead of just this description. Use when the subagent needs context "
            "from earlier in the conversation that would be costly or lossy to restate. "
            "Only available if the harness has enabled forking for this agent."
        ),
    )
```

**`subagents.py:658-669`** — thread the flag through `_validate_and_prepare_state`:

```python
def _validate_and_prepare_state(
    subagent_type: str,
    description: str,
    runtime: ToolRuntime,
    *,
    fork: bool = False,
) -> tuple[Runnable, dict]:
    subagent = _select_subagent(subagent_type, runtime)
    subagent_state = {k: v for k, v in runtime.state.items() if k not in _EXCLUDED_STATE_KEYS}
    subagent_state = {k: v for k, v in subagent_state.items() if k not in private_state_keys}
    if fork:
        subagent_state["messages"] = [*runtime.state["messages"], HumanMessage(content=description)]
    else:
        subagent_state["messages"] = [HumanMessage(content=description)]
    return subagent, subagent_state
```

**`subagents.py:671-725`** — `task`/`atask` accept `fork: bool = False` and pass it through to `_validate_and_prepare_state`.

**`subagents.py:794-833`** — gate the capability behind a middleware-level flag rather than making it unconditionally available, since some deployments won't want full parent history reaching a subagent with a different tool/permission surface:

```python
def __init__(
    self,
    *,
    subagents: Sequence[SubAgent | CompiledSubAgent],
    private_state_keys: frozenset[str] | None = None,
    state_schema: type | None = None,
    allow_fork: bool = False,
) -> None:
    ...
    task_tool = _build_task_tool(
        self._subagents,
        task_description,
        private_state_keys=self._private_state_keys,
        state_schema=self._state_schema,
        allow_fork=allow_fork,
    )
```

When `allow_fork=False` (the default), `fork=True` on a call either raises a clear tool error or is silently treated as `fork=False` — open question for maintainers on which is friendlier; a clear error seems better since a model asking for `fork` and getting silently ignored is a confusing failure mode.

`_build_task_tool` (`subagents.py:531-537`) gains a matching `allow_fork: bool = False` parameter and forwards it into the closures over `task`/`atask`.

### Implementation

**`libs/deepagents/deepagents/middleware/subagents.py`**
- `TaskToolSchema` (`:272`): add `fork: bool = False` field.
- `TASK_TOOL_DESCRIPTION` (`:285`): document the new usage note (when to use `fork` vs. a fresh description).
- `_build_task_tool` (`:531`): add `allow_fork: bool = False` parameter.
- `_validate_and_prepare_state` (`:658`): add `fork` parameter, branch on it when building `subagent_state["messages"]`.
- `task` / `atask` (`:671`, `:699`): accept `fork` from the tool call, validate against `allow_fork`, forward to `_validate_and_prepare_state`.
- `SubAgentMiddleware.__init__` (`:794`): add `allow_fork: bool = False`, forward to `_build_task_tool`.

### Tests

**`libs/deepagents/tests/unit_tests/test_subagents.py`** (or equivalent existing test module)
- `test_task_fork_inherits_parent_messages` — with `allow_fork=True` and `fork=True`, subagent's initial state includes the parent's prior messages plus the new description.
- `test_task_fork_false_is_isolated_by_default` — `fork=False` (or omitted) preserves current isolated behavior exactly, no regression.
- `test_task_fork_rejected_when_not_allowed` — `fork=True` against a `SubAgentMiddleware` constructed with default `allow_fork=False` produces a clear tool error, not a silent no-op.
- `test_task_fork_respects_private_state_keys` — forked history still has `private_state_keys` stripped, same as the non-fork path.
- Async equivalents for `atask`.

- `runtime.state["messages"]` is the exact source of parent history available at the call site — no new state plumbing is required, only threading the existing value differently.
- Scope is deliberately limited to the synchronous `task` tool (`SubAgentMiddleware`). `start_async_task` (`async_subagents.py:131`, `:279-318`) has the same "description only" limitation and could get an equivalent `fork` flag as a natural follow-up, but async subagents run on a separate remote LangGraph thread, so "fork" there means seeding a new remote thread's initial input with the parent's history rather than reusing local state — a bit more involved, and better scoped as its own PR once the sync-path shape is agreed on.
- Comparable pattern: an agent harness that lets a caller spawn a sub-invocation that inherits the calling context wholesale (vs. one that starts a fresh, isolated context) is a distinction worth having as an explicit opt-in, not a default — this proposal keeps deepagents' current isolated-by-default behavior untouched and adds the inheriting mode purely as an opt-in.

### Demand signal — this is a recognized gap, not a one-off ask

The `task` tool's parent↔subagent boundary is narrow in both directions, and multiple independently-filed issues are hitting different facets of the same root cause:

- **#606** ("Support for passing file_uri / multimodal context to subagents via the task tool", closed but unresolved) — a user building a multimodal PDF-analysis pipeline uploads a file via `file_uri` in a `HumanMessage`. The orchestrator agent sees it, but the `task` tool has no way to forward it to a subagent — only the text `description` crosses the boundary — so the media block has to be hacked into a text string, losing its semantic meaning. `fork=True` fixes this directly: the media block, being part of prior message history, would just flow through with everything else, no new attachment-plumbing needed.
- **#3838** ("Add structured payload support for subagent task calls", open) — same boundary problem from the *input* side: the tool schema only accepts a plain-text `description`, so structured task input (goal/scope/constraints/references) can't be reliably passed either.
- **#2512** ("Preserve full message content from `CompiledSubAgent`", open) — same boundary problem from the *output* side: `_return_command_with_state_update` flattens the subagent's final message to `.text`, dropping content blocks and artifacts on the way back to the parent.

Three issues, three angles, one root cause: the `task` tool's message-passing is narrower than what `BaseMessage`/state already supports. This proposal addresses the *input-history* facet specifically — a fourth angle on a pattern maintainers are already triaging piecemeal.

Other concrete scenarios where `fork` would be used:
- A user 40+ turns into a debugging session wants to delegate a parallel investigation without hand-summarizing everything discussed so far — summarizing is lossy and costs tokens the parent already spent gathering.
- A "second opinion" / review delegation pattern (e.g. a review subagent checking a decision) needs to see the actual prior reasoning, not a paraphrase, since paraphrasing risks losing the exact nuance the reviewer is meant to catch.

Related-but-distinct issues, for triage clarity:
- **#2440** ("Async subagents option to pass certain state vars from the parent") — forwards specific *state variables* (e.g. an `access_token`), not conversation history. Complementary, not overlapping.
- **#1359** ("Fork session / rewind command") — a `dcode` TUI feature for the *user* to branch/rewind their own thread. This proposal is the SDK-level primitive for the analogous case where the *model* delegates work via `task` — worth keeping in mind as a possible future synergy point for dcode's own fork UX, but a separate surface today.

I'm happy to implement this and open a PR — the design above (schema field, `_validate_and_prepare_state` branch, `allow_fork` gate, tests) is fully scoped and ready to go as soon as this issue is assigned to me.

NinaadRao:
Fixes #4668

---

Every subagent invocation through the `task` tool starts from a blank slate: `_validate_and_prepare_state` always overwrites `messages` with just `[HumanMessage(content=description)]`, discarding the parent's history even though `runtime.state["messages"]` has it available at that exact call site. That isolation is the right default — it's what keeps subagent context windows small — but it means a user 40 turns into a debugging session has no way to delegate a parallel investigation without hand-summarizing everything discussed so far, which is lossy and re-spends tokens the parent already spent gathering that context.

This adds an opt-in `fork: bool = False` parameter to the `task` tool. When `True`, the subagent's initial `messages` is the parent's full history plus the task description, instead of just the description. It's gated behind a middleware-level `allow_fork: bool = False` on `SubAgentMiddleware` rather than being unconditionally available, since some deployments won't want full parent history reaching a subagent with a different tool/permission surface — a `fork=True` call against a middleware built with the default `allow_fork=False` is rejected with a clear tool-result error (not a silent no-op and not a raised exception, following the same style the tool already uses for an unknown `subagent_type`).

`private_state_keys` stripping still applies to forked history exactly as it does to the non-fork path — forking changes what `messages` contains, not which other state keys cross the parent/subagent boundary.

This addresses one facet of a gap several independently-filed issues keep hitting from different angles: #606 (closed, unresolved) shows a user unable to forward a `file_uri` multimodal block through the task tool at all, since only the text `description` crosses the boundary; #3838 wants structured *input* beyond plain text; #2512 wants structured *output* preserved on the way back. This proposal is the *history* facet — carrying the conversation so far, not just the latest message.

Scope is deliberately limited to the synchronous `task` tool (`SubAgentMiddleware`). `start_async_task` has the same "description only" limitation and could get an equivalent `fork` flag as a natural follow-up, but async subagents run on a separate remote LangGraph thread, so "fork" there means seeding a new remote thread's initial input rather than reusing local state — better scoped as its own PR.

Fully opt-in — nothing changes unless you explicitly enable it:

~~~python
from deepagents.middleware import SubAgentMiddleware

agent = create_agent(
    model="anthropic:claude-sonnet-5",
    middleware=[
        SubAgentMiddleware(
            backend=my_backend,
            subagents=[...],
            allow_fork=True,  # opt-in; default False keeps today's isolated behavior
        )
    ],
)
```

The model can then call `task(description=..., subagent_type=..., fork=True)` when a subagent genuinely needs context from earlier in the conversation instead of a restated summary.

## Release note

Adds an opt-in `fork` parameter to the `task` tool (gated by `SubAgentMiddleware(allow_fork=True)`). When set, a subagent starts with the parent's full conversation history plus the task description, instead of just the description — useful for delegating a parallel investigation deep into a long session without hand-summarizing everything discussed so far. Defaults to off; existing subagent isolation is unchanged unless explicitly enabled.

Added 6 unit tests in `test_subagent_middleware_init.py` (`TestTaskToolFork`): fork inherits parent messages (sync + async), fork defaults to today's isolated behavior when omitted, fork is rejected with a clear tool error when `allow_fork=False` (sync + async), and forked history still respects `private_state_keys` stripping. `uv run --group test pytest`, `ruff check`, `ruff format --check`, and `ty check` are all clean, and the full existing unit test suite (2252 tests) passes with no regressions — 6 smoke-test snapshots were regenerated to reflect the new `fork` field on the `task` tool's schema and a new usage note in its description; the diffs are exactly those two additions, nothing else changed.

I also ran a 3-scenario script driving a real `create_agent` graph through the actual `ToolNode` (not calling the tool's underlying function directly), with a fake subagent model that reports which prior messages it actually received:

- `fork=True`, `allow_fork=True`: subagent received the parent's full history — a detail from a message several turns before the task call was visible to it.
- `fork=False` (default), `allow_fork=True`: subagent received only the task description, exactly like today — isolation unchanged.
- `fork=True`, `allow_fork=False` (default): rejected with the tool error, subagent never invoked.

Attaching the demo script and the log generated from the run
[fork_context_task_tool_demo.log](https://github.com/user-attachments/files/29943325/fork_context_task_tool_demo.log)
[fork_context_task_tool_demo.py](https://github.com/user-attachments/files/29943327/fork_context_task_tool_demo.py)

## Social handles (optional)
LinkedIn: https://linkedin.com/in/ninaadrao


mdrxy:
Thanks for taking the initiative here—we appreciate the thought and effort you put into this proposal. This change has broader API, security, and architecture implications and requires more design review before we pursue an implementation, so we’re closing this PR for now.

Going forward, please don’t open implementation PRs unless a maintainer has explicitly assigned the work to you. You’re welcome to share ideas on an issue or discussion first and wait for design alignment and assignment before starting implementation.

PR #4695, opened 2026-07-13
feat(code): improve `/goal` clarity

mdrxy:
Potentially not needed

Improved `/goal` criteria review, lifecycle visibility, iteration-limit feedback, and permission-mode terminology. Completed goals no longer affect later turns, and inactive goal tools no longer influence ordinary conversations.

---

As a dcode user pursuing an outcome, I want criteria approval and goal lifecycle feedback to be clear without leaving the review flow or letting finished goals affect unrelated follow-up turns.

The inline criteria review now distinguishes accepting and running, accepting with unrestricted permissions, editing, rejecting with feedback, and cancelling. Goal activation explains persistence and the grading iteration limit, while the status bar keeps an active goal visible without adding a repeated transcript message on every follow-up.

Completed goals are cleared before later turns, and iteration exhaustion keeps unmet goals active while showing remaining criteria and concrete retry, resume, amend, and clear paths. Goal tools are omitted from ordinary model requests until a goal or rubric is active.

`/goal` and `/rubric` now use outcome-focused descriptions. `/criteria` remains executable as a compatibility alias without appearing as a separate completion or help entry. Existing `--auto-approve`, `dangerously-auto`, keybindings, configuration values, and serialized state remain compatible, but user-facing permission copy now describes the mode as unrestricted.

### DIFF
--- a/libs/code/COMMANDS.md
+++ b/libs/code/COMMANDS.md
@@ -23,7 +23,7 @@ aliases, descriptions, visibility, or hidden-command metadata.
 | `/effort` |  | Set reasoning effort for the current model |
 | `/feedback` |  | Send feedback or report an issue |
 | `/force-clear` |  | Stop active work, clear the chat, and start a new thread |
-| `/goal` |  | Set a persistent objective by drafting acceptance criteria |
+| `/goal` |  | Turn an outcome into acceptance criteria and pursue it |
 | `/help` |  | Show help and available commands |
 | `/install` |  | Install an optional integration |
 | `/mcp` |  | Manage MCP servers and authentication |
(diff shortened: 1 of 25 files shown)


mdrxy:
I'm concerned about dynamically removing the goal tools and their system-prompt section when no goal or rubric is active. Tool definitions and system instructions are generally part of the provider's cached prompt prefix, so changing both when entering or leaving goal mode can invalidate cache reuse for a long conversation.

The existing request shape is stable, and inactive goal tools already report that no goal or rubric is active. Could we keep the tools and guidance present consistently and rely on that inactive behavior instead? The reduction in tool noise does not seem worth interrupting prompt-cache continuity whenever goal mode changes.

PR #4745, opened 2026-07-15
feat(evals): flat cross-category shard pool for unified evals

nick-hollon-lc:
## What

Replace the serialized per-provider / per-category eval jobs with one dynamically-drained **per-model pool spanning all three categories**. A shard is a 1-task unit carrying its own `(category, dataset, agent_impl, task)`; `unified_prep` emits a per-model flat matrix and one `eval` job drains it. `max_parallel`/`model_parallel` are **derived** from `concurrency`+`rollouts`; `MAX_SHARDS` 64→200 with proportional per-category packing above the cap. Per-category aggregation keeps the tier-2 scorecard (`aggregate_unified.py`) unchanged. `harbor.yml`'s single-dataset path decouples `shard_parallel` from `n_shards` for dynamic dispatch (no-op at the default `n_shards=10`).

## Validated (CI, docker sandbox, bare harness)

- **Lite single-model** (`gpt-5.6-terra`): complete scorecard across all 3 categories, macro pass@3 0.479 (context 0.0 → 0.625 after the fixes below).
- **Multi-model** (`gpt-5.6-terra` + `anthropic:claude-sonnet-5`): both per-model pools run concurrently under `model_parallel`, total-job guard holds (68 shards), and the non-OpenAI agent runs with the OpenAI judge.

## Fixes folded in (found during the dry-runs)

- Context corpus populate + judge `OPENAI_API_KEY` now read the per-shard `matrix.*` values (they were gated on the leaf inputs, which the flat design leaves empty).
- LangSmith experiment name scoped per category (fixes the cross-dataset run double-count).

## Dependency

Needs the harbor build-lock fix ([harbor-framework/harbor#2340](https://github.com/harbor-framework/harbor/pull/2340)) via `harbor_package_override` until it's released — pin `nick-hollon-lc/harbor@27a6eac8` meanwhile.


mdrxy:
## Review: a few things to harden before merge

Nice refactor — the packing math and drift guard are solid. Four things I'd want addressed, roughly in priority order. The through-line: #2 and #3 create ways to reach an empty/garbage-data state without any red step, and #1 then turns that state into a green run — so #1 is the highest-value fix.

### 1. `aggregate_unified.py:380-388` — all-incomplete now passes green, masking total data loss
Downgrading the all-models-incomplete gate to `::warning::` + `return 0` is right for the "one bad shard shouldn't void the comparison" case. But errored-but-complete trials are already scored `0.0` upstream and never reach this gate — so the gate now *only* fires on genuine data-integrity failure (missing shard artifacts, count mismatch, data loss, or **zero leaves uploaded**). Downgrading that means a run where the infra silently dropped every model's data shows a green check with a leaderboard ranked on nothing.

Suggest keeping the per-model downgrade but restoring `return 1` for the categorical "zero trustworthy data" case — at minimum when `_discover_leaves` found zero usable leaves for the whole run.

### 2. `unified_prep.py:210-236` — empty/missing category silently emits zero jobs
A requested category that resolves to zero tasks produces zero matrix entries with no error; if all are empty the matrix is `[]`, `total_job_guard(n, 0)` passes, `est_tasks`'s `max(..., 1)` floor hides it, and the run launches **no eval jobs** — then rides #1 to green. The full-profile workflow guards externally, but `unified_prep.py` doesn't. Please `raise SystemExit` in `main()` if any requested category produced zero entries or the total matrix is empty.

### 3. `unified_prep.py:315-316` — `json.load(UNIFIED_TASKS_JSON)` has no shape validation
Valid-but-wrong-shape JSON (e.g. `{"autonomous": "taskname"}` — a string, not a list) is truthy, so `pack_tasks("taskname", budget)` iterates the string **by character** and emits shards for tasks `"t"`, `"a"`, `"s"`… silently. Validate `dict[str, list[str]]` the same way `aggregate_unified._load_list_env` (`:344-353`) already does, and `SystemExit` on mismatch.

### 4. `unified_prep.py:167-186` — `_allocate_shard_budgets` trim loop is untested
Confirmed empirically: the existing tests only feed inputs where `excess <= 0`, so the `while excess > 0` loop — the decrement, largest-budget selection, tie-break, and floor-to-1 bump — never executes. A regression that trims the wrong budget or off-by-ones the excess would ship green. Please add a direct unit test with an overflowing input (e.g. `{'big':10000,'s1':1,'s2':1}` at cap 200 → sum 201) asserting `sum(budgets) <= cap`, that the trim came off the largest budget, and that small categories stay at 1; plus a two-equal-largest tie case pinning the category-order winner.

mdrxy (approved):

PR #4752, opened 2026-07-15
style(code): add bottom padding to toast notifications

mdrxy:
Toast notifications now have extra bottom padding so the last line is no longer cramped against the toast edge.

---

Toast notifications inherited Textual's default `padding: 1 1`, which leaves only a single row of padding below the last line of text. Because toasts dock at the bottom of the screen and sit directly on top of the chat input box, that lone row reads as cramped — the final line looks flush against the box edge (visible on the multi-line "Large paste collapsed…" toast). This adds a `Toast` rule to the app stylesheet that bumps the bottom padding to two rows so the last line always has clear breathing room, without changing the top or horizontal insets.

Made by [Open SWE](https://openswe.vercel.app/agents/0b203b6c-63e3-ee91-358d-73d5bbeef6fc)


mdrxy:
need to test

PR #4786, opened 2026-07-16
fix(code): pin filelock below 3.30 to avoid blocking import

mdrxy:
`filelock` 3.30.0 runs a blocking temp-directory probe (`_probe_link_follow_symlinks`) at import time. `deepagents-code` imports `filelock` (via `mcp_auth`) on the async server-graph startup path, so Blockbuster rejects the resulting `os.getcwd` call and the graph fails its readiness check with `BlockingError: Blocking call to os.getcwd`, aborting startup entirely. The broad `>=3.12,<4.0.0` range let a fresh install resolve the just-released 3.30.0, which is why even downgrading `deepagents-code` reproduced it.

Capping at `<3.30` keeps resolution on 3.29.x. A follow-up can make the `filelock` import loop-safe (e.g. warm it up off the event loop) and relax the cap.

## Release Note
Fix `deepagents-code` startup failure (`BlockingError: Blocking call to os.getcwd`) caused by `filelock` 3.30.0.

## Test Plan
- [x] `dcode` launches cleanly with `filelock` resolved to 3.29.x

Made by [Open SWE](https://openswe.vercel.app/agents/a27d2d36-4e59-b0b2-edbb-228bc6e557c9)

### DIFF
--- a/libs/code/pyproject.toml
+++ b/libs/code/pyproject.toml
@@ -78,7 +78,10 @@ dependencies = [
     # MCP
     "langchain-mcp-adapters>=0.3.0,<1.0.0",
     # Cross-process lock serializing MCP OAuth token refreshes.
-    "filelock>=3.12,<4.0.0",
+    # Capped below 3.30: it runs a blocking temp-dir probe at import time
+    # (_probe_link_follow_symlinks), which trips Blockbuster's os.getcwd guard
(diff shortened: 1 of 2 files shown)


aolsenjazz (approved):
LGTM

PR #4826, opened 2026-07-17
feat(cli): tag skill invocations with ls_skill_name trace metadata

paarth-a:
## What

Emit `ls_skill_name` on run metadata for skill invocations in the deepagents CLIs, so skill usage is queryable in LangSmith.

The invoked skill's name already lives in the skill-invocation envelope (`additional_kwargs.__skill.name`, built in `skills/invocation.py`) but was only used for TUI rendering. This promotes it into `config["metadata"]["ls_skill_name"]` for the turn, so it lands on the traced runs.

## Why

Today skill usage isn't aggregatable — the skill name isn't in any queryable metadata field or tag. With `ls_skill_name` you can run:

```
POST /api/v1/runs/stats
{ "session": ["<project>"], "group_by": { "attribute": "metadata", "path": "ls_skill_name" } }
```

for per-skill counts / latency / tokens / cost. The key name matches the Claude Code and Codex tracing plugins, so one query spans all coding agents.

## Scope

- `libs/code/deepagents_code/textual_adapter.py` and `libs/cli/deepagents_cli/textual_adapter.py` (the two CLI twins).
- **Not** the core SDK: there a skill is invoked by the model autonomously `read_file`-ing the `SKILL.md`, so there is no invoke-time hook — that needs a separate design decision (middleware detector vs. dedicated tool).

## Test plan

- [x] `py_compile` both files
- [ ] Invoke a skill in the CLI, confirm `ls_skill_name` on the root run metadata in LangSmith

Draft — paired with sibling PRs in `langsmith-claude-code-plugins` and `langsmith-codex-plugins`.


mdrxy:
I think wrong probably - ping and happy to chat

PR #4831, opened 2026-07-17
perf(code): coordinate async shutdown teardown to cut exit latency

mdrxy:
Interactive shutdown is faster because independent cleanup now overlaps instead of running sequentially, without skipping any cleanup.

---

`session.end` hooks run off the Textual event loop. Interrupted-agent cleanup still finishes before server shutdown; afterward, pending hooks drain while the server stops.

Server stop and restart paths are synchronized to prevent duplicate or racing teardown. Existing timeouts and cleanup guarantees are unchanged, and debug logs record phase timings.

Made by [Open SWE](https://openswe.vercel.app/agents/ce9a3224-140d-08bc-25d2-202982da0203)


mdrxy:
ready; waiting for next release

PR #4868, opened 2026-07-20
refactor(code): centralize JSON type aliases

johannes117:
Centralize recursive JSON aliases in a neutral module so config and plugin code share one definition while preserving plugin import compatibility.

The aliases remain static types only; existing runtime ingress validation is unchanged.

<details>
<summary>Test plan</summary>

- `make -C libs/code format`
- `make -C libs/code lint`
- Focused JSON, hooks, model config, and MCP login tests (546 passed; one unrelated root-permission assertion failed)

</details>

Made by [Open SWE](https://openswe.vercel.app/agents/019f8107-bf8c-7545-9a37-299374653ddc)

### DIFF
--- a/libs/code/deepagents_code/plugins/models.py
+++ b/libs/code/deepagents_code/plugins/models.py
@@ -5,6 +5,11 @@ from __future__ import annotations
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Literal
 
+from deepagents_code.json_types import (  # noqa: TC001
+    JsonObject as JsonObject,
+    JsonValue as JsonValue,
+)
+
 if TYPE_CHECKING:
     from pathlib import Path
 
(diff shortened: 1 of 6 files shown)


libs/code/deepagents_code/plugins/models.py
```diff
@@ -5,15 +5,18 @@
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Literal
 
+from deepagents_code.json_types import (  # noqa: TC001
+    JsonObject as JsonObject,
+    JsonValue as JsonValue,
```
mdrxy:
```suggestion
    JsonValue,
```

why are we doing `X as X` ?

mdrxy (commented):
see ^

libs/code/deepagents_code/plugins/models.py
johannes117: lol

mdrxy (approved):
see ^

PR #4870, opened 2026-07-20
feat(code): add typed hooks data models

johannes117:
Introduces the typed domain and wire contracts that future dcode hook execution will build on. This PR does not integrate lifecycle events or execute the new hook protocol.

Design: https://app.notion.com/p/3a3808527b17808ea4f9e0074d705dc4

---

- Adds discriminated domain events and event-specific decisions for all nine MVP hook events.
- Adds compatible wire input/output, permission update, command configuration, and versioned transport models with cached Pydantic adapters.
- Adds shared recursive JSON validation and exact-key contract coverage.
- Moves the existing fire-and-forget dispatcher to `hooks.legacy` without changing its public imports, making room for the new package.

<details>
<summary>Test plan</summary>

- `make lint`
- `make check_imports`
- Hooks model and legacy compatibility tests: 84 passed
- Full unit suite: 10,105 passed; 3 unrelated terminal-width assertions fail locally because Rich resolves the test consoles to width 80

</details>


libs/code/deepagents_code/hooks/models/config.py
```diff
@@ -0,0 +1,40 @@
...
+
+from pydantic import BaseModel, ConfigDict, Field
+
+from deepagents_code.hooks.models.domain import (  # noqa: TC001 - Pydantic runtime annotation.
+    HookEvent,
+)
+
+
+class _ConfigModel(BaseModel):
+    model_config = ConfigDict(extra="forbid")
```
aolsenjazz: I'm normally a fan of fail-fast but I think this is a case where we don't want fail-fast since we're parsing potentially unknown schema shapes? + if Anthropic adds future fields to handler specs

libs/code/deepagents_code/hooks/models/domain.py
```diff
@@ -0,0 +1,337 @@
...
+from deepagents_code.approval_mode import (  # noqa: TC001 - Pydantic runtime annotation.
+    ApprovalMode,
+)
+from deepagents_code.json_types import (  # noqa: TC001 - Pydantic runtime annotation.
+    JsonObject,
+)
+
+
+class _DomainModel(BaseModel):
+    model_config = ConfigDict(extra="forbid", arbitrary_types_allowed=True)
```
aolsenjazz, line 33: Same issue here w.r.t. reading untrusted data - idt we can reasonably use `extra="forbid"` unless I'm missing something about the intent

libs/code/deepagents_code/hooks/models/domain.py
```diff
@@ -0,0 +1,337 @@
...
+
+class DcodeNotification(_DomainModel):
+    """A notification emitted by a dcode lifecycle owner."""
+
+    type: str
+    message: str
+    title: str | None = None
+
+
+class BackgroundTaskSnapshot(_DomainModel):
```
aolsenjazz, line 103: Do you recall what this represents? iirc claude has background tasks it can execute like crons or something but we don't have an analog. 

I'm also down to just trust

libs/code/deepagents_code/hooks/models/domain.py
```diff
@@ -0,0 +1,337 @@
...
+    status: str
+    description: str
+    command: str | None = None
+    agent_type: str | None = None
+    server: str | None = None
+    tool: str | None = None
+    name: str | None = None
+
+
+class SessionCronSnapshot(_DomainModel):
```
aolsenjazz, line 121: Ok so I _know_ that this isn't represented in dcode. if these data models aren't being used in MVP, let's document. should be able to throw an LLM at addign a note if a data model isn't include in the current hooks workload.

yo ucan do this with a separate PR, this is non-blocking

libs/code/deepagents_code/hooks/models/domain.py
```diff
@@ -0,0 +1,337 @@
...
+    """Domain payload for `Stop`."""
+
+    event: Literal[HookEvent.STOP]
+    continuation_count: int
+    last_assistant_message: str
+    background_tasks: list[BackgroundTaskSnapshot] = Field(default_factory=list)
+    session_crons: list[SessionCronSnapshot] = Field(default_factory=list)
+
+
+class SubagentStartEvent(_DomainModel):
```
aolsenjazz, line 205: I beleive these are also not included in MVP. include in separate clarity docs PR re. "thsi will be used later"

libs/code/deepagents_code/hooks/models/wire.py
```diff
@@ -0,0 +1,390 @@
...
+
+
+class PermissionRule(_WireModel):
+    """Permission rule returned by a hook."""
+
+    tool_name: str = Field(alias="toolName")
+    rule_content: str | None = Field(default=None, alias="ruleContent")
+
+
+class PermissionDestination(StrEnum):
```
aolsenjazz, line 72: This is interesting - what your read on this?

aolsenjazz: I don't remember this as a part of the spec - unless included in the new spec?

libs/code/deepagents_code/hooks/models/wire.py
```diff
@@ -0,0 +1,390 @@
...
+    | ReplaceRulesUpdate
+    | RemoveRulesUpdate
+    | SetModeUpdate
+    | AddDirectoriesUpdate
+    | RemoveDirectoriesUpdate,
+    Field(discriminator="type"),
+]
+
+
+class BackgroundTaskWire(_WireModel):
```
aolsenjazz, line 166: same thing as above re. separate PR to document things that anre't implemnted right away

libs/code/deepagents_code/hooks/models/wire.py
```diff
@@ -0,0 +1,390 @@
...
+    cwd: str
+    hook_event_name: HookEvent
+    prompt_id: UUID | None = None
+    permission_mode: WirePermissionMode | None = None
+    effort: Effort | None = None
+    agent_id: str | None = None
+    agent_type: str | None = None
+
+
+class SessionStartWireInput(BaseHookWireInput):
```
aolsenjazz, line 211: do we need wire definitions for client-only events? my gut says no but maybe I'm missing something

libs/code/deepagents_code/hooks/legacy.py
```diff
@@ -79,7 +79,7 @@
 if TYPE_CHECKING:
     from collections.abc import Mapping
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("deepagents_code.hooks")
```
aolsenjazz, line 90: my TS brain does not understand why

libs/code/deepagents_code/json_types.py
```diff
@@ -2,6 +2,11 @@
 
 from typing import TypeAlias
 
+from pydantic import JsonValue as PydanticJsonValue, TypeAdapter
+
 JsonScalar: TypeAlias = str | int | float | bool | None
-JsonValue: TypeAlias = JsonScalar | list["JsonValue"] | dict[str, "JsonValue"]
+JsonValue: TypeAlias = PydanticJsonValue
```
aolsenjazz, line 8: noice

libs/code/deepagents_code/json_types.py
```diff
@@ -2,6 +2,11 @@
...
 JsonObject: TypeAlias = dict[str, JsonValue]
+
+JSON_VALUE_ADAPTER = TypeAdapter(JsonValue)
```
aolsenjazz, line 11: noice noice

aolsenjazz (approved):
Foregoing review of tests for velocity

Overall LGTM. a few comments

libs/code/deepagents_code/hooks/legacy.py
johannes117: Package-scoped logger so records propagate under `deepagents_code.*` after the hooks package split (legacy module previously lived at hooks.py).

aolsenjazz: ahhhhh nice ty

libs/code/deepagents_code/hooks/models/wire.py
aolsenjazz: sorry edited that was confusing

libs/code/deepagents_code/hooks/models/domain.py
johannes117: switching to "ignore" instead
not using "allow" because we only care about fields that we support, "ignore" is sufficient

libs/code/deepagents_code/hooks/models/config.py
johannes117: same here

libs/code/deepagents_code/hooks/models/domain.py
johannes117: compatibility shape for stop/subagent stop wire context. 
background_tasks are in the system design doc
invocations in the MVP will leave this list empty, same with SessionCronSnapshot. 
Will add docstrings to say that these are not in the MVP

libs/code/deepagents_code/hooks/models/domain.py
johannes117: same as above, will add a docstring to explain

libs/code/deepagents_code/hooks/models/domain.py
aolsenjazz: Yup gotcha. ty

libs/code/deepagents_code/hooks/models/domain.py
johannes117: added docstrings

libs/code/deepagents_code/hooks/models/wire.py
johannes117: external (claude) wire vocab, adding docstring

libs/code/deepagents_code/hooks/models/wire.py
johannes117: added

libs/code/deepagents_code/hooks/models/wire.py
johannes117: yes we need wire definitions for clientonly events. These events still run command handlers on the terminal client, and those handlers get the same external json envelope on stdin

PR #4880, opened 2026-07-21
feat(code): add Hooks v2 execution engine

johannes117:
Depends on #4870

Hooks v2 can now match configured lifecycle handlers, execute command hooks safely, and return typed event-specific decisions.

---

The engine captures immutable configuration snapshots, projects all nine events onto the compatible wire protocol, bounds subprocess time and output, and reduces ordered results without changing the legacy dispatcher. Lifecycle call-site integration remains in a later stack layer.

<details>
<summary>Test plan</summary>

- 99 focused hook/model compatibility tests
- `make lint`
- `make check_imports`

</details>

Made by [Open SWE](https://openswe.vercel.app/agents/019f81f5-89a5-703a-a090-0160871a1091)

## References
- Plan: https://openswe.vercel.app/agents/019f81f5-89a5-703a-a090-0160871a1091/plan


libs/code/deepagents_code/hooks/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+)
+
+
+@dataclass(frozen=True, slots=True)
+class HookHandler:
+    """One ordered command handler in a configuration snapshot."""
+
+    id: str
+    event: HookEvent
+    command: str
```
aolsenjazz, line 58: I'm uneasy about defining this data model to support only the hook case - but I guess this is the play to meet MVP reqs. nothing actionable here

libs/code/deepagents_code/hooks/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+            config: Validated Hooks v2 configuration.
+
+        Returns:
+            A snapshot whose handler order and matchers cannot change in flight.
+        """
+        expanded: dict[HookEvent, tuple[HookHandler, ...]] = {}
+        for event, groups in config.hooks.items():
+            handlers: list[HookHandler] = []
+            for group_index, group in enumerate(groups):
+                matcher, error = _compile_matcher(group.matcher)
```
aolsenjazz, line 98: Based on my read, if a matcher fails compilation, we still add a hook handler to the handlers array with the error passed in the constructor

Is this definitely what we want? I would expect to reject failed matchers and warn

aolsenjazz: this part of the tech spec appears to be in violation https://app.notion.com/p/Hooks-System-Design-v2-3a3808527b17808ea4f9e0074d705dc4?source=copy_link#322a966267924cdb99ede9a95292cb48

libs/code/deepagents_code/hooks/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+
+        Returns:
+            Matching handlers and diagnostics for invalid configured matchers.
+        """
+        event = invocation.event.event
+        target = _match_target(invocation)
+        matched: list[HookHandler] = []
+        diagnostics: list[HookDiagnostic] = []
+        for handler in self.handlers.get(event, ()):
+            if handler.matcher_error is not None:
```
aolsenjazz: Related to comment above - do we really want to be emitting diagnostics on bad matchers here? or catch at compilation time and warn

libs/code/deepagents_code/hooks/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+def _compile_matcher(value: str | None) -> tuple[Pattern[str] | None, str | None]:
+    if not value:
+        return None, None
+    try:
+        return re.compile(value), None
+    except re.error as exc:
+        return None, f"Invalid hook matcher {value!r}: {exc}"
+
+
+def _match_target(invocation: HookInvocation) -> str | None:
```
aolsenjazz, line 196: I haven't read far enough - but since we're using a class inheritance approach here, we should be able to wrap this branching logic into a base class + overrides as necessary?

not a big deal, but that might scale better if systems are set up for it

libs/code/deepagents_code/hooks/runner.py
```diff
@@ -0,0 +1,240 @@
...
+    except TimeoutError:
+        await _terminate(process)
+        return _failure(
+            handler.id,
+            "timeout",
+            f"Hook exceeded its {timeout:g} second timeout",
+        )
+    except asyncio.CancelledError:
+        await _terminate(process)
+        raise
```
aolsenjazz, line 95: This doesn't seem right to me - will revisit

libs/code/deepagents_code/hooks/runner.py
```diff
@@ -0,0 +1,240 @@
...
+        max_output_bytes: Maximum retained bytes for each output stream.
+
+    Returns:
+        Validated protocol output and structured diagnostics.
+
+    Raises:
+        asyncio.CancelledError: If the caller cancels command execution.
+    """
+    try:
+        argv = shlex.split(handler.command)
```
aolsenjazz: Sol thinks this may be an issue `The runner instead applies shlex.split() and calls create_subprocess_exec(), causing pipes, redirects, variable expansion, &&, and globs to be passed as literal arguments. Such hooks fail open rather than executing as configured.`

Verifying on my end

libs/code/deepagents_code/hooks/runner.py
```diff
@@ -0,0 +1,240 @@
...
+    if stderr_truncated:
+        diagnostics.append(
+            _diagnostic(
+                handler.id,
+                "stderr_truncated",
+                f"Hook stderr exceeded {max_output_bytes} bytes",
+            )
+        )
+
+    if process.returncode == _BLOCKING_EXIT_CODE:
```
aolsenjazz, line 118: Don't we need event-specific handling of exit code 2? E.g. `SessionStart` cannot block but reducer still stop processing events right?

aolsenjazz: @open-swe please verify this claim

libs/code/deepagents_code/hooks/runner.py
aolsenjazz: yeah: • 
Claude defines hook command as a shell command with no args field (cc_src/schemas/hooks.ts:32-46) and executes the whole string through a shell (cc_src/utils/hooks.ts:940-983), preserving pipes, redirects, $VAR, &&, and globs.
• 
PR #4880 likewise exposes only command, but incorrectly uses shlex.split() plus create_subprocess_exec() (runner.py:62-78), treating shell operators as literal argv.
• 
The fix should use asyncio.create_subprocess_shell(handler.command, ...). If args support is later added, use create_subprocess_exec(command, *args) only when args is explicitly present. The review finding is correct; tests cover simple commands only and miss shell syntax.

libs/code/deepagents_code/hooks/engine.py
```diff
@@ -0,0 +1,71 @@
...
+            )
+            return reduce_hook_results(
+                invocation,
+                (),
+                diagnostics=(*match.diagnostics, diagnostic),
+            )
+
+        results: list[HandlerResult] = []
+        for handler in match.handlers:
+            result = await run_command_handler(
```
aolsenjazz: Correct me if I'm wrong - this is executing handlers in serial right? I believe design doc says that we'll execute in parallel - and I think we want to do this so that we don't end up with several potentially long timeouts/long units of work delaying execution needlessly

libs/code/deepagents_code/hooks/reducer.py
```diff
@@ -0,0 +1,276 @@
...
+        ):
+            state.diagnostics.append(_loop_guard_diagnostic())
+        else:
+            state.context.append(message)
+    else:
+        state.stop_reason = message
+        state.continue_processing = False
+
+
+def _merge_specific(
```
aolsenjazz, line 177: It's not really explicitly in this PR, but Sol brought up a great point -

If a customer is relying on `updatedInput` and we're silently ignoring it, we're so significantly changing the behavior of the hook that I would consider it "breakage"

It's all hypothetical but I'm leaning toward walking back this strategy, adding a blacklist for this field, and informing the user at hook load.

I don't like that, but I think it's better than the alternative

libs/code/deepagents_code/hooks/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+                            status_message=spec.status_message,
+                            matcher=matcher,
+                            matcher_text=group.matcher,
+                            matcher_error=error,
+                        )
+                    )
+            expanded[event] = tuple(handlers)
+        return cls(handlers=MappingProxyType(expanded))
+
+    def match(self, invocation: HookInvocation) -> HookMatch:
```
aolsenjazz, line 131: Another Sol comment that I think is likely true - it looks like we're matching on dcode tool names like execute/write_file/edit_file instead of the Claude-compatible tool names

Maybe we should pass data thru the wire adapters before running matchers?

aolsenjazz: The example given is `A configured Bash policy hook therefore never runs for an actual execute call. Tool names and inputs should pass through the planned ToolWireAdapter before both matcher evaluation and payload projection.`

libs/code/deepagents_code/hooks/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+                        field="matcher",
+                    )
+                )
+                continue
+            if (
+                event not in _MATCHABLE_EVENTS
+                or handler.matcher is None
+                or (target is not None and handler.matcher.search(target) is not None)
+            ):
+                matched.append(handler)
```
aolsenjazz: Another Sol thing that i think is correct -

We're just evaluating hook matches based on regex right? Claude branched matching implementation e.g. Exact matches for "Edit|Write" or wildcard ("*") matches

aolsenjazz: https://code.claude.com/docs/en/hooks#matcher-patterns

libs/code/deepagents_code/hooks/snapshot.py
```diff
@@ -0,0 +1,147 @@
...
+
+    from deepagents_code.hooks.models.config import HooksConfig
+    from deepagents_code.hooks.models.domain import HookInvocation
+
+_MATCHABLE_EVENTS = frozenset(
+    {
+        HookEvent.PERMISSION_REQUEST,
+        HookEvent.NOTIFICATION,
+        HookEvent.PRE_TOOL_USE,
+        HookEvent.POST_TOOL_USE,
```
aolsenjazz, line 42: Sure I'm missing what's doing on here - only 4 matchable events? Also confused here because pre/post tool use are implemented server-side and I'm not certain that both notification/permission_request are? Can you clarify?

libs/code/deepagents_code/hooks/runner.py
aolsenjazz: :(

libs/code/deepagents_code/hooks/runner.py
aolsenjazz: I don't see this being handled upstream - maybe I'm missing something?

Also maybe this is because I'm a js guy - but are we raising a generic exception here? or would we want to propagate CancelledError/some other custom error

aolsenjazz (changes requested):

libs/code/deepagents_code/hooks/snapshot.py
johannes117: noted, can expand later

libs/code/deepagents_code/hooks/snapshot.py
johannes117: okay slight refactor, removed matcher_error from HookHandler. _compile_matcher failure now means the whole group is skipped, handlers don't enter the snapshot

should now match the spec

libs/code/deepagents_code/hooks/snapshot.py
johannes117: same fix as above, refactored. bad matchers are caught in from_config, skipped, logged and added to snapshot.diagnostics

libs/code/deepagents_code/hooks/snapshot.py
aolsenjazz: nice, resolving

libs/code/deepagents_code/hooks/snapshot.py
aolsenjazz: nice, resolving

libs/code/deepagents_code/hooks/snapshot.py
johannes117: left as a flat helper for now since the dispatch is small, but happy to revisiit later when we add more matchable fields?

aolsenjazz: based

libs/code/deepagents_code/hooks/runner.py
johannes117: left as is:
this kills the hook subprocess then it re-raises the same cancelled error
the cancel propagates to the lifecycle owner, which is the intended asyncio contract

libs/code/deepagents_code/hooks/runner.py
johannes117: fixed, switched to shell execution to match claudes contract

libs/code/deepagents_code/hooks/runner.py
aolsenjazz: Ah nice ty! you'll make me a python guy, yet :)

libs/code/deepagents_code/hooks/runner.py
johannes117: addressed in the reducer

libs/code/deepagents_code/hooks/engine.py
johannes117: fixed, handlers now run concurrently

libs/code/deepagents_code/hooks/reducer.py
johannes117: we now emit an unsupported_field diagnostic, which warns the user when seeing updatedInput

libs/code/deepagents_code/hooks/snapshot.py
johannes117: fixed, added a shared tool wire adapter in hooks/tools.py

libs/code/deepagents_code/hooks/snapshot.py
johannes117: fixed, implemented claude-compatible matcher branching

libs/code/deepagents_code/hooks/snapshot.py
johannes117: uh yeah this was incomplete, apologies, updated!

aolsenjazz (approved):
It seems like previous round of review requirements have been met, stamping for velocity

PR #4912, opened 2026-07-21
chore(code): update dependency minimums

mdrxy:
Refreshes Deep Agents Code's dependency lower bounds and lockfile to keep minimum-version installs on current supported releases.

The NVIDIA extra assertion now matches the updated `aiohttp` minimum, and Ruff's newer selector rules are accounted for in the lint configuration.


mdrxy:
@SocketSecurity ignore pypi/numpy@2.5.1

PR #4916, opened 2026-07-21
feat(code): add Hooks v2 capability snapshots

johannes117:
Related DCD-70

Hooks v2 now validates and freezes configuration into deterministic session snapshots backed by one authoritative event capability registry.

---

This is 1/3 in the DCD-70 stack.

- Defines event ownership, matching, timeout, output, and aggregation capabilities.
- Loads project and user configuration in stable precedence order with semantic-only legacy migration.
- Computes canonical snapshot hashes and rejects unsupported matcher or handler combinations before execution.

<details>
<summary>Test plan</summary>

- `uv run --group test pytest tests/unit_tests/hooks` — 81 passed at this stack layer
- `make lint`

</details>

-------------------------
- main
  - **johannes117/code/dcd-70-hooks-capabilities** 👈
    - [johannes117/code/dcd-70-hooks-execution](https://github.com/langchain-ai/deepagents/pull/4917)
      - [johannes117/code/dcd-70-hooks-transcripts](https://github.com/langchain-ai/deepagents/pull/4918)


libs/code/deepagents_code/hooks/loading.py
```diff
@@ -0,0 +1,241 @@
...
+        Frozen load result with canonical `snapshot_id`.
+    """
+    sources = (
+        tuple(paths)
+        if paths is not None
+        else (
+            project_hooks_path(cwd),
+            user_hooks_path(config_dir),
+        )
+    )
```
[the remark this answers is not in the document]
johannes117: addressed

libs/code/deepagents_code/hooks/models/config.py
```diff
@@ -28,6 +29,18 @@ class CommandHandlerSpec(_ConfigModel):
...
+    @model_validator(mode="after")
+    def _reject_async_commands(self) -> Self:
+        if self.async_:
+            msg = "async command hooks are not supported in MVP"
+            raise ValueError(msg)
+        # Normalize explicit `async: false` to omitted so equivalent configs
+        # share a snapshot hash.
+        if self.async_ is False:
+            return self.model_copy(update={"async_": None})
+        return self
```
aolsenjazz:
```suggestion
    @field_validator("async_", mode="after")
    @classmethod
    def _normalize_async(cls, value: bool | None) -> None:
        if value:
            raise ValueError("async command hooks are not supported in MVP")
        return None
```

aolsenjazz: Really not a big deal but it looks like this is a cleaner and more correct pattern - returning `self.model_copy` would be a violation of the validation contract.

Ordinarily wouldn't care but let's get the good patterns in for LLMs to copy moving forward

libs/code/deepagents_code/hooks/capabilities.py
```diff
@@ -0,0 +1,267 @@
...
+
+
+@dataclass(frozen=True, slots=True)
+class HookEventSpec:
+    """Immutable capability description for one hook event."""
+
+    event: HookEvent
+    owner: HookOwner
+    event_model: type[BaseModel]
+    decision_model: type[BaseModel]
```
aolsenjazz: Using `type[BaseModel]` isn't really getting us anything here besides know that this is a pydantic base model

Surely we have a more useful type that we can use e.g. `BaseHookDecision`

libs/code/deepagents_code/hooks/capabilities.py
```diff
@@ -0,0 +1,267 @@
...
+)
+
+
+@dataclass(frozen=True, slots=True)
+class HookEventSpec:
+    """Immutable capability description for one hook event."""
+
+    event: HookEvent
+    owner: HookOwner
+    event_model: type[BaseModel]
```
aolsenjazz: Looks like we should be able to use `HookDomainEvent` instead of `BaseModel`

libs/code/deepagents_code/hooks/capabilities.py
aolsenjazz: And if we can't use narrower types then something smells a lot with our impl/typedefs

libs/code/deepagents_code/hooks/capabilities.py
```diff
@@ -0,0 +1,267 @@
...
+            exit_code_policy=ExitCodePolicy.DIAGNOSE,
+            plain_output_policy=PlainOutputPolicy.IGNORE,
+            aggregation_policy=AggregationPolicy.CONTEXT,
+            supported_handler_types=frozenset({HandlerType.COMMAND}),
+        ),
+    }
+)
+
+
+def assert_hook_event_registry_complete() -> None:
```
aolsenjazz: I am wracking my brain for why we need this

if the hook event registry incomplete, surely that is a programmer error

aolsenjazz: I mean - if this is just a runtime check to keep typedefs in sync with implementations, surely we can derive a type from `HookEvent` that will complain if an event spec is missing

libs/code/deepagents_code/hooks/capabilities.py
```diff
@@ -0,0 +1,267 @@
...
+    field = model.model_fields.get("event")
+    if field is None or get_args(field.annotation) != (event,):
+        msg = f"{kind.title()} model for {event.value} has an invalid discriminator"
+        raise RuntimeError(msg)
+
+
+assert_hook_event_registry_complete()
+
+
+def get_event_spec(event: HookEvent) -> HookEventSpec:
```
aolsenjazz, line 208: I'm not convinced that we should expose both this and the `HOOK_EVENT_SPECS`. What is the correct way to get an event spec? Do we need to export the full SPECS collection? Or can we constrain the public API better?

libs/code/deepagents_code/hooks/loading.py
```diff
@@ -0,0 +1,245 @@
...
+Precedence (highest first, earlier in reduction order):
+
+1. Project: `{cwd}/.deepagents/hooks.json`
+2. User: `~/.deepagents/hooks.json` (or `config_dir/hooks.json` in tests)
+
+Sources are concatenated per event. Project groups precede user groups so a
+project `continue: false` wins before lower-precedence handlers run.
+
+Legacy list-shaped documents are migrated only for events whose lifecycle
+semantics genuinely match Hooks v2. `tool.use` is never treated as `PreToolUse`.
```
aolsenjazz:
```suggestion
semantics genuinely match Hooks v2.
```

aolsenjazz: Looks like an LLM changed its mind

libs/code/deepagents_code/hooks/loading.py
```diff
@@ -0,0 +1,245 @@
...
+class LoadedHooksConfig:
+    """Validated configuration plus load diagnostics and source paths."""
+
+    config: HooksConfig
+    diagnostics: tuple[HookDiagnostic, ...]
+    sources: tuple[Path, ...]
+    snapshot_id: str
+
+
+def project_hooks_path(cwd: Path) -> Path:
```
aolsenjazz: Variable name is smelly - if this is supposed to be a project root let's name this project_root

libs/code/deepagents_code/hooks/loading.py
```diff
@@ -0,0 +1,245 @@
...
+            return None, (
+                HookDiagnostic(
+                    code="invalid_config",
+                    severity="warning",
+                    message=f"Legacy hooks list missing at {path}",
+                    field=str(path),
+                ),
+            )
+        legacy_entries = [item for item in hooks if isinstance(item, Mapping)]
+        migrated = migrate_legacy_hooks(legacy_entries)
```
aolsenjazz, line 227: nice

libs/code/deepagents_code/hooks/migration.py
```diff
@@ -0,0 +1,90 @@
...
+from deepagents_code.hooks.models.domain import HookEvent
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping, Sequence
+
+# Only legacy `session.end` has the same lifecycle boundary and side-effect-only
+# behavior as its Hooks v2 counterpart. Legacy `session.start` runs once per
+# prompt execution, `context.compact` runs before the offload operation, and
+# `permission.request` is a batched observation that cannot return a decision.
+_LEGACY_EVENT_MAP: dict[str, tuple[HookEvent, str | None]] = {
```
aolsenjazz, line 27: I'd be keen to throw an LLM at better legacy event migration. Surely we can support more than just session.end even if they end up getting migrated to different event names (e.g. session start -> on prompt submit or whatever the event is)

libs/code/deepagents_code/hooks/migration.py
aolsenjazz, line 1: Can we build in a deprecation warning to users? let's sunset this on Sept 1st cc @mdrxy

libs/code/deepagents_code/hooks/loading.py
```diff
@@ -0,0 +1,245 @@
...
+        diagnostic = HookDiagnostic(
+            code="legacy_migrated" if migrated.hooks else "legacy_unmapped",
+            severity="debug",
+            message=message,
+            field=str(path),
+        )
+        return migrated, (diagnostic,)
+
+    try:
+        config = HOOKS_CONFIG_ADAPTER.validate_python(data)
```
aolsenjazz: This would reject the entire file if one bad hook fails validation - possible to do per-hook validation to preserve good hooks?

aolsenjazz: Probably should have a unit test to catch this

libs/code/deepagents_code/hooks/loading.py
```diff
@@ -0,0 +1,245 @@
...
+            followed by user hooks.
+
+    Returns:
+        Frozen load result with canonical `snapshot_id`.
+    """
+    sources = (
+        tuple(paths)
+        if paths is not None
+        else (
+            (project_hooks_path(cwd), user_hooks_path(config_dir))
```
aolsenjazz: We should canonicalize and dedupe paths

libs/code/deepagents_code/hooks/models/config.py
```diff
@@ -28,6 +29,18 @@ class CommandHandlerSpec(_ConfigModel):
     command: str
     timeout: float | None = None
```
aolsenjazz: Sol keeps complaining that this can be infinity, negative numbers, etc

I'm tempted to add stricter validation just to keep sol quiet :)

libs/code/deepagents_code/hooks/snapshot.py
```diff
@@ -180,10 +203,10 @@ def _compile_matcher(
 
 def _handler_matches(
```
aolsenjazz, line 204: Sol is complaining about the mismatched behavior between TS regex and python regex which is.... annoying but also maybe valid enough.

Suppose we roll with python matching for now

aolsenjazz (changes requested):

libs/code/deepagents_code/hooks/migration.py
johannes117: how would we surface that? like if they invoke a legacy hook they get a toast message?

aolsenjazz: Let's wait till we're at UI impl to figure this out. Maybe we expose a narrow callback (`on_legacy_hook_executed`) that the UI layer can subscribe to for displaying notifications.

libs/code/deepagents_code/hooks/models/config.py
johannes117: yep switched to use a field_validator

libs/code/deepagents_code/hooks/capabilities.py
johannes117: updated to:
```
event_model: type[HookDomainEvent]
decision_model: type[BaseHookDecision]
matcher_field: MatcherField | None
```

libs/code/deepagents_code/hooks/capabilities.py
johannes117: addressed above

libs/code/deepagents_code/hooks/capabilities.py
johannes117: removed these

libs/code/deepagents_code/hooks/capabilities.py
johannes117: addressed this by making the mapping private: _HOOK_EVENT_SPECS

leading underscore marks as internal, this constrains the public API to just get_event_spec(event)

libs/code/deepagents_code/hooks/loading.py
johannes117: applied

libs/code/deepagents_code/hooks/loading.py
johannes117: changed to project_root, although cwd does make sense to me, but I think project_root is more appropriate here

libs/code/deepagents_code/hooks/migration.py
johannes117: could we revisit in a follow up potentially?

libs/code/deepagents_code/hooks/loading.py
aolsenjazz: I guess another concern I have here is when running from I non-project root - wdyt? any concerns here?

aolsenjazz: e.g. I often run dcode from ~/workspace

libs/code/deepagents_code/hooks/migration.py
johannes117: okay added:
```
_LEGACY_HOOKS_REMOVAL_DATE = "September 1, 2026"
```

when legacy config is loaded it now produces a warning diagnostic

libs/code/deepagents_code/hooks/loading.py
johannes117: yep refactored to do per-hook validation

libs/code/deepagents_code/hooks/loading.py
johannes117: yep added normalization step before loading
now these will resolve to one source:
```
/config/hooks.json
/config/../config/hooks.json
```

libs/code/deepagents_code/hooks/models/config.py
johannes117: tightened to:
```
timeout: float | None = Field(default=None, gt=0, allow_inf_nan=False)
```

libs/code/deepagents_code/hooks/snapshot.py
johannes117: rolling with python matching for now

aolsenjazz (approved):
remaining questiosn were answered in slack

PR #4917, opened 2026-07-21
fix(code): harden Hooks v2 command execution

johannes117:
Depends on #4916
Related DCD-70

Hooks v2 command handlers now execute with event-aware policies, bounded terminal output, sanitized environments, and complete process cleanup.

---

This is 2/3 in the DCD-70 stack.

- Applies plain-output, exit-code, continuation, permission, and deferred-field policies without discarding sibling diagnostics.
- Validates terminal sequences and removes ambient secret and telemetry environment values.
- Cleans up complete process groups on timeout or cancellation and maps native/MCP tool calls only from trustworthy metadata.
- Keeps materialized filesystem paths outside domain and transport state.

<details>
<summary>Test plan</summary>

- `uv run --group test pytest tests/unit_tests/hooks` — 95 passed at this stack layer
- `make lint`

</details>

-------------------------
- main
  - **fix(code): harden Hooks v2 command execution** :point_left:
    - https://github.com/langchain-ai/deepagents/pull/4918
      - https://github.com/langchain-ai/deepagents/pull/4971


libs/code/deepagents_code/hooks/env.py
```diff
@@ -0,0 +1,55 @@
...
+
+import os
+from typing import TYPE_CHECKING
+
+from deepagents_code.config_manifest import _SECRET_NAME_MARKERS
+
+if TYPE_CHECKING:
+    from collections.abc import Mapping
+
+_OTEL_PREFIX = "OTEL_"
```
aolsenjazz: otel??

Will try to come back to this during review but sorry if I leave this in

libs/code/deepagents_code/hooks/env.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def is_secret_env_name(name: str) -> bool:
```
aolsenjazz: should be able to reuse `_is_secret_env`

libs/code/deepagents_code/hooks/runner.py
```diff
@@ -137,6 +156,14 @@ async def run_command_handler(
     try:
         decoded = json.loads(stdout)
     except (json.JSONDecodeError, UnicodeDecodeError):
+        plain = _decode(stdout).strip()
+        policy = get_event_spec(resolved_event).plain_output_policy
```
aolsenjazz: To me this is doing the job of the reducer - it should be up to the reducer to implement policies - this violates SoC

libs/code/deepagents_code/hooks/env.py
```diff
@@ -0,0 +1,55 @@
...
+    and strips values whose names look like secrets. Hooks are user-authored
+    trusted code, but secret values should not be ambiently available.
+
+    Args:
+        source: Environment to sanitize. Defaults to `os.environ`.
+
+    Returns:
+        A new environment mapping suitable for `asyncio.create_subprocess_shell`.
+    """
+    env = dict(os.environ if source is None else source)
```
aolsenjazz: Not blocking - but it might be preferable to rewrite this to inject env variables for testability. not confirmed but I could also see us injecting env vars into the hook execution subprocess - checking if CC does this now

aolsenjazz: Yeah actually it looks like it does:

```
• 
CLAUDE_PROJECT_DIR; plugin hooks also get CLAUDE_PLUGIN_ROOT, CLAUDE_PLUGIN_DATA, and CLAUDE_PLUGIN_OPTION_<KEY>.
• 
CLAUDE_ENV_FILE only for SessionStart, Setup, CwdChanged, and FileChanged.
• 
Context-dependent variables including CLAUDE_EFFORT, CLAUDE_CODE_REMOTE=true, and CLAUDE_CODE_BRIDGE_SESSION_ID.
```

aolsenjazz: Suppose we don't need to sanitize injected value - but the point remains

libs/code/deepagents_code/hooks/runner.py
```diff
@@ -214,9 +241,22 @@ async def _read_bounded(
 
 
 async def _terminate(process: Process) -> None:
-    if process.returncode is None:
-        process.kill()
-    await process.wait()
+    """Kill the hook process group, then reap the direct child."""
+    if process.returncode is not None:
+        return
```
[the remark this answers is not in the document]
aolsenjazz: Honestly not sure how to handle - I'm not convinced that this is the right fix

libs/code/deepagents_code/hooks/reducer.py
```diff
@@ -87,10 +87,10 @@ def reduce_hook_results(
     state = _Reduction(diagnostics=list(diagnostics))
     for result in results:
         state.diagnostics.extend(result.diagnostics)
+        if result.plain_output is not None:
```
aolsenjazz, line 97: As mentioned in runner.py - can we return always return plain_output from runner.py and apply policies here? we're already applying exit code policy in this file

libs/code/deepagents_code/hooks/reducer.py
```diff
@@ -127,6 +164,26 @@ def _merge_output(
     _merge_specific(invocation, state, handler_id, specific)
 
 
+def _diagnose_extra_fields(
```
aolsenjazz, line 184: nice

libs/code/deepagents_code/hooks/reducer.py
aolsenjazz: Surely we can do better here?

aolsenjazz: e.g. this should be `WireModel` following typeadapter right?

libs/code/deepagents_code/hooks/reducer.py
```diff
@@ -135,43 +192,73 @@ def _merge_block(
...
+                    "consecutive attempts"
+                ),
+            )
+        )
+        return
+    state.continue_loop = True
+    state.feedback.append(message)
 
 
 def _merge_specific(
```
aolsenjazz, line 280: I fear we are reaching the scale where branching on discriminators fails

libs/code/deepagents_code/hooks/reducer.py
```diff
@@ -182,33 +269,67 @@ def _merge_specific(
...
         _append(state.context, specific.additional_context)
+        for attr, wire_name in _DEFERRED_SESSION_START_FIELDS:
+            value = getattr(specific, attr)
+            if value in (None, False, [], ""):
+                continue
+            state.diagnostics.append(
+                HookDiagnostic(
+                    code="unsupported_field",
+                    severity="warning",
+                    message=f"{wire_name} is not supported and was ignored",
```
aolsenjazz: Maybe it's because of the warning message, the behavior for this branch is unclear to me

libs/code/deepagents_code/hooks/reducer.py
```diff
@@ -182,33 +269,67 @@ def _merge_specific(
...
+                    code="unsupported_field",
+                    severity="warning",
+                    message="permissionDecision defer is not supported and was ignored",
+                    handler_id=handler_id,
+                    field="permissionDecision",
+                )
+            )
+            behavior = None
         if specific.updated_input is not None:
             _diagnose_unsupported_updated_input(state, handler_id)
```
aolsenjazz, line 348: Just to clarify on `updated_input` handling - 

With the PR yesterday, I suggested that we explicitly complain and omit on hooks that use `updateInput` or whatever the field is

My concern here is that if we ended up implementing (which I believe we did), this will confound future human and LLM readers

libs/code/deepagents_code/hooks/reducer.py
```diff
@@ -182,33 +269,67 @@ def _merge_specific(
...
     elif isinstance(specific, PermissionRequestSpecificOutput):
         decision = specific.decision
         if decision.behavior == "allow":
-            if (
+            has_updated_input = (
                 isinstance(decision, PermissionAllow)
                 and decision.updated_input is not None
-            ):
+            )
+            if has_updated_input:
```
aolsenjazz: same comment as above

libs/code/deepagents_code/hooks/reducer.py
```diff
@@ -221,29 +342,36 @@ def _merge_specific(
...
+            state.diagnostics.append(
+                HookDiagnostic(
+                    code="unsupported_field",
+                    severity="warning",
+                    message="updatedToolOutput is not supported and was ignored",
+                    handler_id=handler_id,
+                    field="updatedToolOutput",
+                )
+            )
+        if specific.updated_mcp_tool_output is not None:
```
aolsenjazz: I will start a convo on this sooner than later (and updated tool input). I'm just not sure of the implications still

libs/code/deepagents_code/hooks/projection.py
```diff
@@ -49,53 +50,65 @@
...
-    prompt_id: NotRequired[UUID | None]
-    permission_mode: WirePermissionMode
-    effort: NotRequired[Effort | None]
+    permission_mode: NotRequired[WirePermissionMode]
+    prompt_id: NotRequired[UUID]
+    effort: NotRequired[Effort]
 
 
-def project_hook_input(invocation: HookInvocation) -> HookWireInput:
+class _CommonWireFields(_CoreWireFields):
```
aolsenjazz: The intent of this isn't very clear due to the name

libs/code/deepagents_code/hooks/projection.py
```diff
@@ -49,53 +50,65 @@
...
+    agent_id: NotRequired[str]
+    agent_type: NotRequired[str]
+
+
+def project_hook_input(
```
aolsenjazz, line 68: Another area of approaching ugly scaling w.r.t. branching logic

libs/code/deepagents_code/hooks/projection.py
```diff
@@ -174,51 +190,99 @@ def project_hook_input(invocation: HookInvocation) -> HookWireInput:
...
 
 
-def _common_fields(invocation: HookInvocation) -> _CommonWireFields:
+def _core_fields(
+    invocation: HookInvocation,
+    transcript_path: Path | None,
+) -> _CoreWireFields:
     context = invocation.context
-    return {
+    if transcript_path is None:
```
aolsenjazz: Ok so I might get the answer with the next PR but I believe transcript_path and agent_transcript_path are required here, but marked optional in the engine. This is adding compexity and needless validation if the type contract shoudl in reality guarantee the existence of transript_path

Maybe this is fixed in next PR

libs/code/deepagents_code/hooks/projection.py
```diff
@@ -174,51 +190,99 @@ def project_hook_input(invocation: HookInvocation) -> HookWireInput:
...
+    if permission_mode is not None:
+        fields["permission_mode"] = permission_mode
+    if context.prompt_id is not None:
+        fields["prompt_id"] = context.prompt_id
+    if context.effort is not None:
+        fields["effort"] = Effort(level=context.effort)
+    return fields
+
+
+def _common_fields(
```
aolsenjazz: Yeah reading the implementation of this and the core_fields function this is very odd

aolsenjazz: Who does what? who is the real base class?

libs/code/deepagents_code/hooks/projection.py
```diff
@@ -174,51 +190,99 @@ def project_hook_input(invocation: HookInvocation) -> HookWireInput:
...
+        fields["agent_id"] = agent.id
+        fields["agent_type"] = agent.name
+    return fields
 
 
-def serialize_hook_input(invocation: HookInvocation) -> bytes:
+def serialize_hook_input(
+    invocation: HookInvocation,
+    *,
+    transcript_path: Path | None = None,
```
aolsenjazz: Here's another case where transcript_path is none
What's going on here? Should we build these fields into HookInvocation?

libs/code/deepagents_code/hooks/projection.py
```diff
@@ -174,51 +190,99 @@ def project_hook_input(invocation: HookInvocation) -> HookWireInput:
...
+            ),
+        )
+    return ()
+
+
+def _permission_mode(mode: ApprovalMode) -> WirePermissionMode | None:
     return {
         ApprovalMode.MANUAL: WirePermissionMode.DEFAULT,
-        ApprovalMode.AUTO: WirePermissionMode.AUTO,
+        ApprovalMode.AUTO: None,
```
aolsenjazz: I'm not convinced that this is what we want to do. why would we omit auto permission mode?

libs/code/deepagents_code/hooks/tools.py
```diff
@@ -36,56 +36,48 @@ def _edit_input(args: JsonObject) -> JsonObject:
...
-            else offset
-        )
-    return result
-
-
-def _glob_input(args: JsonObject) -> JsonObject:
-    return _select(args, "pattern", "path")
-
-
-def _grep_input(args: JsonObject) -> JsonObject:
```
aolsenjazz, line 55: Are we really comfortable breaking back compat for these tools? I wouldn't be so eager to rip these out - makes more sense to me to preserve until Sept 1 sunset

libs/code/deepagents_code/hooks/env.py
aolsenjazz: Unclear to me why we're stripping OTEL_ prefixed vars. The implementation details _what_ but not _why_ (maybe I am a fool for not knowing _why_)

aolsenjazz: Ok looks like a claude remnant which may be relevant if we ever use otel but idt this is reelvant atm

libs/code/deepagents_code/hooks/validate_terminal_sequence.py
aolsenjazz, line 1: Trying not to be a stickler about this stuff but `terminal.py` isn't an ideal file name - `validate_terminal_seqence.py` makes it clearer for humans and LLMs

libs/code/deepagents_code/hooks/runner.py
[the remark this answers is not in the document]
aolsenjazz: From Sol review: 

```
Suggested fix: on POSIX, attempt killpg(process.pid, SIGKILL) even when the direct process has a return code; only use the early return for non-POSIX direct-child cleanup. Add a regression test where the direct shell exits before a descendant performs a delayed side effect.
```

wdyt?

libs/code/tests/unit_tests/hooks/test_execution.py
```diff
@@ -0,0 +1,227 @@
...
+    )
+    for _ in range(50):
+        if grandchild_pid.is_file():
+            break
+        await asyncio.sleep(0.01)
+
+    task.cancel()
+    with pytest.raises(asyncio.CancelledError):
+        await task
+    if grandchild_pid.is_file():
```
aolsenjazz: Valid point from Sol - here's the full dump

```
The new test expects os.kill(pid, 0) to raise immediately after process-group termination. On POSIX, that call also succeeds for a terminated process that remains briefly as a zombie, so it does not prove the descendant is still executing. This failed on the Python 3.12 Ubuntu job while the other Python jobs passed:

FAILED test_runner_kills_process_group_on_cancellation
Failed: DID NOT RAISE ProcessLookupError
The test passed locally on macOS/Python 3.14, confirming the platform/race sensitivity.

Suggested fix: poll with a bounded deadline and distinguish a live process from Z state on Linux, or assert that a delayed descendant side effect never occurs. Keep a separate regression test for finding 1, where the shell exits before timeout/cancellation.
```

These subprocess tests generally smell. Honestly I don't have solutions but maybe we can throw an LLM at a more robust solution for these

libs/code/deepagents_code/hooks/runner.py
aolsenjazz, line 103: This still doesn't feel right to me - if I'm understanding correctly this will not result in a HandlerResult/decision, which I don't think is what we want

libs/code/deepagents_code/hooks/runner.py
aolsenjazz: We would also presumably want to move the plan decoding outside of an except block to return unconditionally ofc

aolsenjazz (changes requested):

libs/code/deepagents_code/hooks/runner.py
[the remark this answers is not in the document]
johannes117: okay removed the unconditional returncode early return.
on POSIX, _terminate() now always attempts killpg, even if the shell already exited. Non-POSIX only kills when direct process is still running.

libs/code/deepagents_code/hooks/env.py
johannes117: removed OTEL specific behavior

libs/code/deepagents_code/hooks/env.py
johannes117: removed

libs/code/deepagents_code/hooks/runner.py
johannes117: moved policy out of the runner into the reducer

johannes117: plain_output should only return when it isn't JSON, not unconditionally

libs/code/deepagents_code/hooks/env.py
johannes117: I think we should do in a follow up if non-blocking

aolsenjazz: this should be included in MVP but yes we can add these as a follow up PR to this

libs/code/deepagents_code/hooks/reducer.py
johannes117: as mentioned above returning plain_output is only if non-JSON stdout
this matches CC reference implementation
can change if you feel strongly

aolsenjazz: Nah sounds good

libs/code/deepagents_code/hooks/reducer.py
johannes117: tightened the type and replaced the isinstance blocks with dispatch on the wire models

libs/code/deepagents_code/hooks/reducer.py
johannes117: addressed in the same refactor as above

libs/code/deepagents_code/hooks/reducer.py
johannes117: to clarify, renamed _DEFERRED_SESSION_START_FIELDS to _UNSUPPORTED_SESSION_START_FIELDS

we iterate through these:
```
_UNSUPPORTED_SESSION_START_FIELDS = (
    ("initial_user_message", "initialUserMessage"),
    ("session_title", "sessionTitle"),
    ("watch_paths", "watchPaths"),
    ("reload_skills", "reloadSkills"),
)
```

behavior is:
applied addtionalContext to session context
for each unsupported session start field, if present and non-empty: emit a warning and do not apply them
session start otherwise continues, fail open

aolsenjazz: Right great, "deferred" threw me for a loop

libs/code/deepagents_code/hooks/reducer.py
johannes117: okay should we change the diagnostic string to be "not supported yet" or something? leaving this one open. 
Sounds like to be less confusing we should just implement? depends on whats necessary to meet the minimum requirements etc

libs/code/deepagents_code/hooks/reducer.py
johannes117: replied above, will leave open

libs/code/deepagents_code/hooks/reducer.py
aolsenjazz: Yeah, let's do a fast follow on this so I can check with SDK folks

aolsenjazz: Committing to have this figured out by 12 PT tomorrow

libs/code/deepagents_code/hooks/reducer.py
johannes117: noted, will close just to keep this pr tidy

libs/code/deepagents_code/hooks/projection.py
johannes117: collapsed _CommonWireFields and _CoreWireFields into _BaseWireFields
agent= optional for subagent events

libs/code/deepagents_code/hooks/projection.py
johannes117: refactored with @singledispatch same as the previous refactor

libs/code/deepagents_code/hooks/projection.py
johannes117: okay tightened the type contract for the client transcript path
now required on engine.run, serialize_hook_input, project_hook_input, and _base_fields
agent_transcript_path is still optional because only SubagentStop needs it

libs/code/deepagents_code/hooks/projection.py
johannes117: similar fix to previous: collapsed _core_fields and _common_fields into one typeddict _BaseWireFields with optional agent= when subagent identity is needed

libs/code/deepagents_code/hooks/projection.py
johannes117: addressed above

libs/code/deepagents_code/hooks/projection.py
johannes117: yeah I agree, reverting

libs/code/deepagents_code/hooks/tools.py
johannes117: reverted

libs/code/deepagents_code/hooks/validate_terminal_sequence.py
johannes117: renamed to validate_terminal_sequence.py

libs/code/tests/unit_tests/hooks/test_execution.py
johannes117: addressed

libs/code/deepagents_code/hooks/runner.py
johannes117: from solbro:
```
Cancel is the caller aborting the work (user interrupt, agent abort, teardown). Swallowing it and returning a normal HandlerResult would break asyncio cancellation: a cancelled task would look like it completed successfully, and asyncio.gather in the engine would keep going instead of propagating cancel.
```

As opposed to timeout which is a hook-policy failure, which is why we reduce into a decision

aolsenjazz (approved):

PR #4918, opened 2026-07-21
feat(code): add Hooks v2 session transcripts

johannes117:
Depends on #4917
Closes DCD-70

Hooks v2 now materializes private versioned transcript projections and exposes them through a session-scoped client runtime.

---

This is 3/3 in the DCD-70 stack.

- Adds deterministic per-thread and subagent JSONL projections with redaction, path-safe identities, private permissions, atomic writes, revisions, and retention.
- Adds a client-owned runtime that freezes one snapshot and materializes real transcript paths immediately before command execution.
- Keeps lifecycle call-site and server/client dispatch wiring out of scope for the following tickets.

<details>
<summary>Test plan</summary>

- `uv run --group test pytest tests/unit_tests/hooks` — 101 passed
- `make lint`
- `make check_imports`

</details>

-------------------------
- main
  - https://github.com/langchain-ai/deepagents/pull/4917
    - **feat(code): add Hooks v2 session transcripts** :point_left:
      - https://github.com/langchain-ai/deepagents/pull/4971


libs/code/deepagents_code/hooks/transcript.py
```diff
@@ -0,0 +1,463 @@
...
+    r"(?<![A-Za-z0-9_-])eyJ[A-Za-z0-9_-]{6,}\."
+    r"[A-Za-z0-9_-]{6,}\.[A-Za-z0-9_-]{6,}"
+)
+_URL_RE = re.compile(r"https?://[^\s<>\"']+", re.IGNORECASE)
+_SAFE_PREFIX_RE = re.compile(r"[^a-z0-9]+")
+_SAFE_PREFIX_LENGTH = 32
+_EMPTY_REVISION = hashlib.sha256(b"").hexdigest()
+
+
+class TranscriptRecord(BaseModel):
```
aolsenjazz, line 74: Could we get some docstrings for some of these attrs?

I don't think its clear to readers whats going on witth sequence or record_id etc

libs/code/deepagents_code/hooks/runtime.py
aolsenjazz, line 1: very happy with the system as designed

aolsenjazz: nice job

libs/code/deepagents_code/hooks/runtime.py
```diff
@@ -0,0 +1,162 @@
...
+    async def invoke(self, invocation: HookInvocation) -> HookDecision:
+        """Materialize transcripts, execute matching handlers, and return a decision.
+
+        Args:
+            invocation: Domain lifecycle invocation.
+
+        Returns:
+            Event-specific decision with notices, sequences, and diagnostics.
+        """
+        prepared = self.prepare_invocation(invocation)
```
aolsenjazz, line 120: I am a _little_ concerned about perf overhead introduced by the I/O here

We should throw an agent at profiling execution time just as a sanity check at some point

libs/code/deepagents_code/hooks/runtime.py
```diff
@@ -0,0 +1,162 @@
...
+            workspace_trusted=workspace_trusted,
+            config_dir=config_dir,
+        )
+        snapshot = HooksSnapshot.from_config(
+            loaded.config,
+            diagnostics=loaded.diagnostics,
+            snapshot_id=loaded.snapshot_id,
+        )
+        store = TranscriptStore(
+            transcript_root or (cwd / ".deepagents" / "transcripts")
```
aolsenjazz: I don't think this is what we want? Maybe I'm mistaken but I would assume that we would want to use `~/.deepagents` right?

aolsenjazz: Yeah everything here is built off of the assumption that we have a `cwd/.deepagents` dir which I think is a flawed assumption - I'm not sure that anyone would even want this to materialize under normal conditions

aolsenjazz (changes requested):

libs/code/deepagents_code/hooks/runtime.py
johannes117: agree, this should definitely be in the ~/.deepagents

libs/code/deepagents_code/hooks/transcript.py
johannes117: added!

libs/code/deepagents_code/hooks/runtime.py
johannes117: agreed, I think we should profile once everything is wired up and we should test with really large sessions
going to close for now
here are the results of some simple local profiling:
(~0.05 ms clean, <5 ms p95 for a 1 MB rewrite)

aolsenjazz (approved):
nice

PR #4946, opened 2026-07-22
fix(code): authorize Auto actions from `ask_user` selections

mdrxy:
In Auto mode, answering an `ask_user` prompt now authorizes the exact action the user selected, so the agent no longer stalls asking the user to retype a choice they just made.

---

Auto mode's action classifier previously accepted only literal chat-input text as evidence of user consent. A selection made through an `ask_user` prompt became an ordinary tool result, which the classifier is required to treat as untrusted context. As a result, after the user picked an option such as "Rebase my new local commit onto the remote branch, then push," the follow-up action was still denied and the agent looped back asking the user to explicitly re-authorize the same command.

When the user answers an `ask_user` prompt, `dcode` now records a server-owned authorization receipt bound to the current thread, turn, and originating tool call. The classifier admits these validated same-turn answers as consent evidence, resolving references such as "run it" or a selected option without requiring the user to restate the command.

The grant is deliberately narrow:

- Only answers recorded by `dcode` for the current turn count; model- or tool-authored text cannot forge one, and the model-authored question is never treated as consent.
- Receipts are bound to the LangGraph execution thread, the active turn, and the exact `ask_user` call, so stale, cross-turn, cross-thread, duplicated, or replayed receipts are ignored.
- An answer authorizes only the action and target it names. Chained or broader operations, riskier variants (for example, force-push when the user approved a normal push), and destructive escalations still require their own explicit authorization.
- Cancelled, errored, or malformed responses record no receipt.


mdrxy:
Superseded by #4993.

The replacement carries forward this PR's server-owned, same-turn `ask_user` authorization receipts and combines them with #4947's deterministic handling of built-in conversation compaction. The combined PR makes the Auto-mode decision order explicit while keeping verified prior consent separate from operations that intrinsically require no approval.

Closing this PR in favor of #4993.

PR #4952, opened 2026-07-22
fix(code): apply harness profile suffix on runtime model swap

mdrxy:
When a model is selected at runtime (via `/model`), its harness profile system-prompt suffix is now applied. Previously the GLM-5.2 `<execution>` guidance was silently dropped because profiles were only resolved against the default model at agent construction time.

---

`create_deep_agent` resolves harness profiles against the *default* configurable model at construction time. In `deepagents-code`, the user-selected model (e.g. `baseten:zai-org/GLM-5.2`) is swapped in later by `ConfigurableModelMiddleware`, so a profile whose only effect is a `system_prompt_suffix` never reached the model.

The fix re-resolves the harness profile for the swapped-in model inside `_build_overrides` and appends its `system_prompt_suffix`, layering on top of the existing `### Model Identity` patch so both edits coexist. A substring guard prevents double-appending, spec resolution falls back to `_model_spec_from_result`, and models without a suffix (the common case) are left untouched. The change is contained to `configurable_model.py`; no construction-time SDK behavior is altered.

Made by [Open SWE](https://openswe.vercel.app/agents/1fa41b84-f267-88fe-5183-e77087fb6e30)


mdrxy:
Need to revisit probably at SDK level

PR #4957, opened 2026-07-22
fix(code): make Auto mode generally available

aolsenjazz:
Auto approval mode is now available in the local interactive TUI without setting `DEEPAGENTS_CODE_EXPERIMENTAL`; sandbox and headless sessions still fall back to Manual.

---

This removes the experimental environment gate across CLI startup, app eligibility, server graph construction, and agent middleware while preserving the local-only safety constraints. Verified with focused unit tests plus package lint, type, formatting, and import checks.

Made by [Open SWE](https://openswe.vercel.app/agents/e0c9c9a9-85ca-7dc8-d759-d89431bc5276)


libs/code/deepagents_code/tui/widgets/approval.py
```diff
@@ -152,8 +152,8 @@ def __init__(
                 file-operation previews.
             id: Optional widget ID. Defaults to 'approval-menu'.
             auto_mode_eligible: Whether Auto mode can be enabled in this session.
-                When `False` (e.g. the experimental opt-in is off), the "Enable
-                Auto for this thread" option is not offered.
+                When `False` (e.g. a sandbox is active), the "Enable Auto for this
```
aolsenjazz, line 155: Flagging @mdrxy - why would this be false when a sandbox is active? Surely there are actions we potentially want to deny even when in a sandbox?

Another question would be - do we really need this `auto_mode_eligible` attr once we're in GA?

aolsenjazz: Ok looks like this is a deliberate decision - would you provide some info so I understand?

libs/code/deepagents_code/main.py
```diff
@@ -1991,10 +1991,7 @@ def help_parent(help_fn: Callable[[], None]) -> list[argparse.ArgumentParser]:
         "--auto-approve",
         action="store_true",
         default=None,
-        help=(
-            "Interactive local TUI only: enable beta classifier-backed Auto mode. "
-            "Requires DEEPAGENTS_CODE_EXPERIMENTAL=1."
-        ),
+        help="Interactive local TUI only: enable beta classifier-backed Auto mode.",
```
aolsenjazz: Is "beta" really helping us here? Sure it's not the final implementation we'd be looking for, but I see this as a confounding word

libs/code/tests/unit_tests/test_agent.py
```diff
@@ -4811,23 +4811,14 @@ def _build_mock_settings(tmp_path: Path) -> Mock:
         mock_settings.interpreter_ptc_acknowledge_unsafe = False
         return mock_settings
 
-    @pytest.mark.parametrize(
-        ("experimental", "expected"), [(False, False), (True, True)]
-    )
-    def test_auto_mode_requires_experimental_flag(
+    def test_auto_mode_does_not_require_experimental_flag(
         self,
         tmp_path: Path,
```
aolsenjazz: Needless test

libs/code/tests/unit_tests/test_app.py
```diff
@@ -24059,7 +24059,7 @@ async def aput_store_item(
 class TestLiveApprovalModeWrites:
     """Verify live approval-mode write and toggle failure behavior."""
 
-    def test_auto_startup_requires_experimental_flag(
+    def test_auto_startup_does_not_require_experimental_flag(
```
aolsenjazz: needless test

libs/code/tests/unit_tests/test_main_args.py
```diff
@@ -341,7 +341,7 @@ def test_accepts_auto_approve_in_interactive_mode(self) -> None:
         with (
             patch.object(sys, "argv", ["deepagents", "--auto-approve", "-m", "hello"]),
             patch.object(sys, "stdin", mock_stdin),
-            patch.dict(os.environ, {"DEEPAGENTS_CODE_EXPERIMENTAL": "1"}),
+            patch.dict(os.environ, {"DEEPAGENTS_CODE_EXPERIMENTAL": "0"}),
```
aolsenjazz: needless

libs/code/tests/unit_tests/test_server_graph.py
```diff
@@ -270,7 +270,7 @@ async def cleanup(self) -> None:
             system_prompt=None,
             interactive=True,
             auto_approve=False,
-            auto_mode_enabled=False,
+            auto_mode_enabled=True,
```
aolsenjazz, line 273: cc @mdrxy

libs/code/deepagents_code/tui/widgets/approval.py
mdrxy: Auto is local-only for now on purpose.

Under the hood Auto does more than “skip the approve prompt.” It makes judgment calls using context about *this machine’s* project: where the repo root is, what the git remote looks like, and whether a path is inside that trusted tree. Those checks are built for a normal local worktree on the host.

A sandbox breaks that assumption. Tools aren’t writing to the same FS the host process is looking at — paths, symlinks, and “where is the project root?” can mean something different inside Daytona/Runloop/etc. If we naively reused the host-side checks, Auto could allow or deny for the wrong reasons. Same issue for the temporary files Auto has the agent use for scratch: those are allocated in the host temp dir today, which isn’t the same place sandbox tools run.

Important distinction: sandbox sessions still get normal Manual approvals in the TUI. We’re only holding back the classifier auto-approve path until it’s honest there.

Making Auto work with sandboxes later isn’t a new product mode. Same Shift+Tab / fallback UI. We’d just need the policy to reason about the *sandbox's* workspace (remote root, remote path semantics, scratch that actually lives where tools run — or host-temp disabled under sandbox). Once that binding is solid, Auto + sandbox can ship together.

libs/code/tests/unit_tests/test_server_graph.py
mdrxy: Yep — intentional.

With the experimental gate gone, interactive + no sandbox enables Auto at graph build:

`auto_mode_enabled = config.interactive and sandbox_backend is None`

So this expectation flips `False → True` for the default local interactive path. Sandbox (or non-interactive) still keeps it off.

PR #4995, opened 2026-07-23
fix(code): inject OpenAI `prompt_cache_key` for any OpenAI-provider endpoint

mdrxy:
`dcode` now attaches the per-thread `prompt_cache_key` to every model whose provider resolves to `openai`, regardless of base URL, so reliable prompt-cache prefix routing also applies through the LangSmith gateway and other OpenAI-compatible endpoints — not just `api.openai.com`. A default-on `models.openai_prompt_cache_key` opt-out is added for endpoints that reject unknown request fields.

---

Previously `ConfigurableModelMiddleware` only injected `prompt_cache_key` when the resolved `ChatOpenAI` base URL was `api.openai.com` (or an official regional subdomain). Any OpenAI-compatible proxy — including the managed LangSmith gateway (`gateway.smith.langchain.com/openai/v1`) that dcode is provisioned to use — was classified as a custom endpoint and silently dropped the key, so consecutive turns over a stable ~58-59K-token prefix missed prompt caching and were billed at full input price (observed ~7x per-turn cost spikes).

`prompt_cache_key` is an optional, additive OpenAI request field: it sharpens prefix-cache routing on model families that support it and is otherwise inert. Rather than maintaining a hostname allowlist, `_is_openai_model` now treats any model reporting `ls_provider == "openai"` as eligible and attempts the key unconditionally. Endpoints that reject unknown fields can opt out via the new `models.openai_prompt_cache_key` config option (`[models].openai_prompt_cache_key` in `config.toml`, or `DEEPAGENTS_CODE_OPENAI_PROMPT_CACHE_KEY`, default on). A user-supplied key (on the model or per-invocation) is still preserved, and the Fireworks session-affinity path is unchanged.

Verified end to end that a model created with the gateway base URL now serializes `prompt_cache_key` in the Responses payload, and that the opt-out suppresses it.

Co-authored-by: open-swe[bot] <open-swe@users.noreply.github.com>

Made by [Open SWE](https://openswe.vercel.app/agents/09eb0978-379d-0b80-9beb-73aa2ef6ab79)


libs/code/deepagents_code/_env_vars.py
```diff
@@ -274,6 +274,20 @@
 flow. Parsed by `is_env_truthy`: accepts `1`, `true`, `yes`, `on` as enabled.
 """
 
+OPENAI_PROMPT_CACHE_KEY = "DEEPAGENTS_CODE_OPENAI_PROMPT_CACHE_KEY"
```
aolsenjazz, line 285: I wonder if a better long-term solution is to just use a generic `PROMPT_CACHE_KEY` env variable for multi-provider coverage

I don't think decomposing this into provider-specific variables is useful to the end user

aolsenjazz: Also - I could see orgs wanting to set org- or team-specific cache keys to maximal static prefix sharing. imo better to support str | falsy

libs/code/deepagents_code/config.py
```diff
@@ -3132,6 +3132,28 @@ def is_memory_auto_save_enabled() -> bool:
...
+    is still forwarded). This is the opt-out for OpenAI-compatible endpoints that
+    reject unknown request fields.
+    """
+    from deepagents_code.config_manifest import (
+        get_option,
+        load_config_toml,
+        resolve_scalar,
+    )
+
+    option = get_option("models.openai_prompt_cache_key")
```
aolsenjazz, line 3245: This will change if we support str | falsy

libs/code/deepagents_code/configurable_model.py
```diff
@@ -104,45 +103,20 @@ def _is_fireworks_model(model: object) -> bool:
 
 
 def _is_openai_model(model: object) -> bool:
```
aolsenjazz, line 105: This function can just go away if this is how we're rewriting it

I see this as a good thing - inspecting the base url feels like overloading this function

libs/code/deepagents_code/configurable_model.py
```diff
@@ -218,9 +192,11 @@ def _with_openai_prompt_cache_key(
     `model_settings` reaches the wire unchanged. `prompt_cache_key` is an
     optional, additive request field: it sharpens prefix-cache routing on model
     families that support it (GPT-5.6 and later) and is otherwise inert, so the
-    same key is sent to every OpenAI model without a version gate. This mirrors
-    the Fireworks path (`_with_fireworks_session_settings`), which sets
-    `prompt_cache_key` the same way and additionally sends an
+    same key is sent to every OpenAI-provider model — including custom base URLs
```
aolsenjazz: these docs feel out of scope of this variable

libs/code/deepagents_code/configurable_model.py
```diff
@@ -346,7 +342,7 @@ def _build_overrides(
         if _is_fireworks_model(effective_model):
             updated_settings = _with_fireworks_session_settings(settings, ctx.thread_id)
             injected = "Fireworks session settings"
-        elif _is_openai_model(effective_model):
+        elif _is_openai_model(effective_model) and _openai_prompt_cache_key_enabled():
```
aolsenjazz: If we do str | falsy we do not need this additional check

libs/code/deepagents_code/_env_vars.py
mdrxy: Keeping this OpenAI-scoped and as a bool + per-thread default for this PR.

**Naming:** Injection is not one shared behavior across providers. Fireworks already has its own path (`prompt_cache_key` + `x-session-affinity`), and Anthropic uses a different caching model entirely. A generic `PROMPT_CACHE_KEY` name would overpromise common semantics and get ambiguous when those providers disagree. Happy to revisit a shared name later if multiple providers genuinely share one setting and one meaning.

**Value shape:** Org/team keys for shared static prefixes are a real product use case, but I'm not expanding this into `str | falsy` here. The bool matches the bug fix (restore key injection behind non-`api.openai.com` endpoints) without changing the config mental model. Power users can already supply a custom `prompt_cache_key` on the model / invocation today. If we want first-class org/team keys in config, I'd rather add a dedicated value setting later than overload one knob that means both enable and value.

libs/code/deepagents_code/config.py
mdrxy: Yep — only if we take the `str | falsy` shape. Staying with the bool toggle here, so this resolver remains `enabled -> bool`.

libs/code/deepagents_code/configurable_model.py
mdrxy: Keeping the thin helper.

It now only checks `ls_provider == "openai"` (base-URL allowlisting is gone, which I agree was overload), but that mirrors `_is_anthropic_model` / `_is_fireworks_model` and keeps call sites readable. The docstring here is also the natural home for the eligibility policy (any OpenAI-provider endpoint, not just `api.openai.com`).

I'd only inline if we collapsed all of the provider helpers the same way — out of scope for this PR.

libs/code/deepagents_code/configurable_model.py
mdrxy: Good call — trimmed this docstring to injection-only behavior (add key if missing; preserve user-supplied keys). Provider eligibility / endpoint policy / opt-out stay on the caller helpers instead.

libs/code/deepagents_code/configurable_model.py
mdrxy: True under a key-or-off config shape. Keeping the separate enabled gate for the bool opt-out: resolve once at construction, then `if openai_prompt_cache_key:` before inject. That only collapses if we redesign the setting as in the `str | falsy` suggestion.

libs/code/deepagents_code/_env_vars.py
aolsenjazz: hmmmmm I think first-class org/team keys routed to cache keys is a logical disconnect and canonicalizing differences in nomenclature under one key (PROMPT_CACHE_KEY or whatever) is better UX. at some point dcode needs to be opinionated rather than saying "here are 5 different levels you can pull

but won't let this get in the way of shipping a fix

aolsenjazz: curious what _your_ thoughts are vs this LLMs thoughts

mdrxy: The job of this PR is restoring `prompt_cache_key` injection.

On the longer-term shape: I’m aligned that we shouldn’t end up with a pile of provider-specific “almost the same” knobs. I’d rather unify cache-key semantics when we actually introduce generalizable prompt-caching middleware — i.e. one opinionated setting/path that different providers can plug into, instead of inventing a fake-generic `PROMPT_CACHE_KEY` ahead of shared behavior existing.

PR #4997, opened 2026-07-23
feat(code): integrate Hooks v2 server lifecycle events

johannes117:
Depends on #4971
Closes DCD-69

Server-owned Hooks v2 events now pause the agent over LangGraph interrupts, run on the client `HooksRuntime`, and resume with typed decisions.

---

This starts the DCD-69 stack on top of the DCD-70 leaf (#4971).

- Adds a versioned `hook_invocation` interrupt/resume transport around `HookInvocationRequest` / `HookInvocationResponse`.
- Installs `ServerHooksMiddleware` for `PreToolUse`, `PostToolUse`, `Stop`, `SubagentStart`, and `SubagentStop`, gated by the session's configured server events so idle sessions skip the round-trip.
- Creates a session-scoped client `HooksRuntime` and fulfills server interrupts from both the Textual adapter and headless runner.
- Applies `deny`/`ask` as blocking tool errors for now; `ask` approval UX remains for DCD-71.

## Review Guide

1. `hooks/server_middleware.py` — middleware, gating, interrupt raise, decision apply
2. `hooks/interrupt.py` — `hook_invocation` pause/resume contract
3. `hooks/client.py` — client fulfillment + resume ledger
4. `agent.py` — middleware install (main + subagents)
5. `tui/textual_adapter.py`, `client/non_interactive.py` — interrupt fulfillment call sites
6. Skim: `envelope.py`, `context.py`, `runtime.py`, `_cli_context.py`, small `app`/`auto_mode`/`main`/`snapshot`/`engine` touches
7. Tests: `test_server_lifecycle.py`, then `test_agent` / `test_non_interactive` / `test_engine`

<details>
<summary>Test plan</summary>

- `uv run --group test pytest tests/unit_tests/hooks` — 133 passed
- `make lint` in `libs/code`

</details>

-------------------------
- main
  - **feat(code): integrate Hooks v2 server lifecycle events** :point_left:
    - https://github.com/langchain-ai/deepagents/pull/5010
      - https://github.com/langchain-ai/deepagents/pull/5044


libs/code/deepagents_code/app.py
```diff
@@ -4211,10 +4213,25 @@ async def _init_session_state(self) -> None:
...
             )
+            try:
+                # Interactive sessions keep project hooks off until a dedicated
+                # workspace-trust prompt lands (design-doc security follow-up).
+                state.hooks_runtime = HooksRuntime.create(
+                    cwd=Path(self._cwd),
+                    workspace_trusted=False,
+                )
+            except Exception:
+                logger.exception("Failed to create HooksRuntime; server hooks disabled")
```
aolsenjazz, line 4312: I don't know how our notification system works, but I have to assume that we need to notify the user if their server hooks aren't working beyond a logger print

libs/code/deepagents_code/app.py
```diff
@@ -4211,10 +4213,25 @@ async def _init_session_state(self) -> None:
...
-            return TextualSessionState(
+            from pathlib import Path
+
+            from deepagents_code.hooks.runtime import HooksRuntime
+
+            state = TextualSessionState(
                 approval_mode=self._approval_mode,
                 thread_id=self._lc_thread_id,
             )
+            try:
```
aolsenjazz, line 4304: A couple more thoughts in addition to comment below -

- if we're going to show a notification on failure, we will need to run this when the UI exists
- that will require putting this in its own closure, which I think is probably good for SoC anyways

libs/code/deepagents_code/hooks/server_middleware.py
```diff
@@ -0,0 +1,711 @@
...
+        thread_id=thread_id,
+        cwd=cwd,
+        prompt_id=prompt_id,
+        approval_mode=(
+            approval if isinstance(approval, ApprovalMode) else ApprovalMode.MANUAL
+        ),
+    )
+
+
+def _context_mapping(runtime_context: object) -> dict[str, Any]:
```
aolsenjazz: We should narrow this return type; we control the fields we're mapping to and their correlated types. return `dict[str, Any]` will only make life harder for ourselves downstream

libs/code/deepagents_code/hooks/server_middleware.py
```diff
@@ -0,0 +1,711 @@
...
+        msg = f"Expected {expected.__name__}, got {type(decision).__name__}"
+        raise TypeError(msg)
+    return decision
+
+
+def _session_gate(runtime_context: object) -> _SessionHookGate | None:
+    fields = _context_mapping(runtime_context)
+    snapshot_id = fields.get("hooks_snapshot_id")
+    events = fields.get("hooks_server_events")
+    if not isinstance(snapshot_id, str) or not snapshot_id:
```
aolsenjazz: I mentioned this in a comment on `_context_mapping` but we're not making life easier for ourselves by returning `dict[str, Any]` from the context mapping fn - case in point this

libs/code/deepagents_code/hooks/server_middleware.py
```diff
@@ -0,0 +1,711 @@
...
+_TASK_TOOL_NAME = "task"
+
+
+class ServerHooksState(AgentState[Any]):
+    """Agent state extensions for server-owned hook middleware."""
+
+    _hooks_stop_continuation_count: NotRequired[int]
+
+
+class _SessionHookGate(TypedDict):
```
aolsenjazz: _SessionHookGate is so much more cognitive complexity than something like HookEventsForSnapshot or something

I'm early into reading so there's a chance that I'm missing something fundamental, but I feel like I've spent too much time reading the variable `gate` and thinking "ok no so that's actually just {snapshot_id, events}"

libs/code/deepagents_code/hooks/server_middleware.py
```diff
@@ -0,0 +1,711 @@
...
+    raw = interrupt(build_hook_interrupt_payload(request))
+    response = parse_hook_resume_value(
+        raw,
+        invocation_id=request.invocation_id,
+        snapshot_id=request.snapshot_id,
+    )
+    return response.decision
+
+
+def _hook_context(
```
aolsenjazz: This feels like a repeat of the plugins function/var naming issue - there are several things in this file which are named `_hook_context` or `call` which require much more cognitive work to figure out what's actually going on

My theory is that this applies to LLMs as well

libs/code/deepagents_code/hooks/server_middleware.py
```diff
@@ -0,0 +1,711 @@
...
+            return request
+        agent = _task_agent_identity(call)
+        decision = _invoke_hook(
+            context,
+            SubagentStartEvent(event=HookEvent.SUBAGENT_START, agent=agent),
+            gate=gate,
+            config=request.runtime.config,
+            deadline=self._default_deadline,
+        )
+        decision = _require_decision(decision, SubagentStartDecision)
```
aolsenjazz, line 326: is `require` the preferred term for this in pythonland? in js I'm used to `ensure`

idc which, but if `require` is pythonic I'll try to stick to that

aolsenjazz (commented):
Started reviewing this over the weekend, apparently I didn't press submit sorry.

These may just be noise by now

libs/code/deepagents_code/hooks/server_middleware.py
aolsenjazz, line 1: If we we have the time while I'm reviewing, it would be nice to clean up the data validation layer. E.g. like 40% ish of this file is just validating graph state/context attributes. Would be really nice to move that stuff to a separate module and use schema validation

libs/code/deepagents_code/hooks/interrupt.py
```diff
@@ -0,0 +1,121 @@
...
+            f"Hook resume invocation_id mismatch: expected {invocation_id}, "
+            f"got {response.invocation_id}"
+        )
+        raise ValueError(msg)
+    if response.snapshot_id != snapshot_id:
+        msg = (
+            f"Hook resume snapshot_id mismatch: expected {snapshot_id}, "
+            f"got {response.snapshot_id}"
+        )
+        raise ValueError(msg)
```
aolsenjazz, line 114: Do we know what the failure mode is here? How do systems recover when a ValueError is raised?

libs/code/deepagents_code/hooks/interrupt.py
```diff
@@ -0,0 +1,121 @@
...
+
+    model_config = ConfigDict(extra="forbid")
+
+    type: Literal["hook_invocation"] = HOOK_INVOCATION_INTERRUPT_TYPE
+    request: HookInvocationRequest
+
+
+HOOK_INVOCATION_INTERRUPT_ADAPTER = TypeAdapter(HookInvocationInterrupt)
+
+HookResumeValue: TypeAlias = dict[str, Any]
```
aolsenjazz: Nit: can't we use those json types that we defined earlier?
happy to wrong about this

libs/code/deepagents_code/hooks/interrupt.py
```diff
@@ -0,0 +1,121 @@
...
+        value: Raw interrupt value from LangGraph.
+
+    Returns:
+        The embedded request, or `None` when `value` is not a hook interrupt.
+    """
+    if (
+        not isinstance(value, dict)
+        or value.get("type") != HOOK_INVOCATION_INTERRUPT_TYPE
+    ):
+        return None
```
aolsenjazz, line 66: Do we know how this failure mode is treated upstream? E.g. I notice that in the parse resume value we're throwing on bad invocation_id/snapshot_id (which I realize is a little different) but just want to make sure we're considering failure modes here

libs/code/deepagents_code/hooks/client.py
```diff
@@ -0,0 +1,172 @@
...
+        """Return one shared result for concurrent and repeated delivery."""
+        async with self._lock:
+            completed = self._completed.get(key)
+            if completed is not None:
+                return completed
+            task = self._in_flight.get(key)
+            if task is None:
+                task = asyncio.create_task(self._run(key, operation))
+                self._in_flight[key] = task
+        return await asyncio.shield(task)
```
aolsenjazz, line 68: How does this impact timeout cancellation?

libs/code/deepagents_code/agent.py
```diff
@@ -1929,6 +1929,14 @@ def _should_interrupt_tool_call(
     Returns:
         `True` to interrupt, or `False` for Auto/YOLO bypass.
     """
+    from deepagents_code.hooks.server_middleware import pre_tool_behavior
+
+    tool_call = getattr(request, "tool_call", None)
+    tool_call_id = str(tool_call.get("id") or "") if isinstance(tool_call, dict) else ""
+    hook_behavior = pre_tool_behavior(getattr(request, "state", None), tool_call_id)
+    if hook_behavior in {"allow", "deny"}:
```
aolsenjazz: This is an interesting point that @mdrxy should be aware of -

Effectively we are allowing hook behavior to override existing HITL + other control systems yes?

Kimi says this is how claude does it - just tagging for vis

libs/code/deepagents_code/agent.py
aolsenjazz, line 1: It would be really good to have a concrete idea of the relationship between all of these interrupt systems. Not sure that there's a unit of work that needs to be done here, but at least in my mind, it's very unclear which systems are at play at any given notable lifecycle checkpoint (pretool etc), which take precedence, ordering

Part of that is skill issue since i'm sort of out of the code ofc

aolsenjazz (approved):
Stamping, but left a few questions which are worth ensuring we know the answer to

Didn't get too into depth but checked out the core files + systems

libs/code/deepagents_code/app.py
johannes117: the 4th PR in this stack deals with surfacing feedback to the user in the TUI, will keep in mind for that PR

libs/code/deepagents_code/app.py
johannes117: noted

libs/code/deepagents_code/hooks/server_middleware.py
johannes117: narrowed

libs/code/deepagents_code/hooks/server_middleware.py
johannes117: addressed

libs/code/deepagents_code/hooks/server_middleware.py
johannes117: addressed

libs/code/deepagents_code/hooks/server_middleware.py
johannes117: okay after some research:
_ensure* is a lazy initialization / idempotent setup. for example make sure x exists or initialized etc, often a no-op if already done
_require* is a validation that raises if the condition doesn't hold

so this is consistent with existing convention

libs/code/deepagents_code/hooks/server_middleware.py
johannes117: have done a small rename pass:
- _hook_context --> _build_hook_context
- local context (HookContext) --> hook_context
- call (ToolCallData) --> tool_call
- _SessionHookGate / _session_gate / gate --> _HookEventsForSnapshot / _hook_events_for_snapshot / enabled_events
- pre --> pre_tool_outcome

libs/code/deepagents_code/hooks/server_middleware.py
johannes117: I think out of scope for this PR, might be able to in a follow up

libs/code/deepagents_code/hooks/interrupt.py
johannes117: basically this can occur when a stale checkpoint is resumed after a hooks config change
doing a small refactor to catch at the 2 callsites and raise  a typed exception, which will prompt the user to start a new session

libs/code/deepagents_code/hooks/interrupt.py
johannes117: yes was able to use those! addressed

libs/code/deepagents_code/hooks/interrupt.py
johannes117: answer from kimi:
•  None ≠ failure — it's the routing path. LangGraph multiplexes interrupt kinds (hook_invocation, HITL, ask_user) over one channel; both call sites use is_hook_interrupt_payload to route, and None just means "not mine, try the next decoder."
•  Actual schema violation (type matches but body fails pydantic) → raises ValidationError, aborts the run. Deliberate: graph is paused mid-tool awaiting a decision we can't generate, no safe local recovery. Matches existing HITL behavior (textual_adapter.py:1250).
•  Resume-side asymmetry is intentional: None = routing (not our interrupt), ValueError = identity mismatch (our interrupt, wrong answer — something's broken, abort).
•  As of last commit, both mismatch cases surface as typed HooksSnapshotChangedError at the fulfillment boundary instead of bare exceptions.
•  No silent drops anywhere. Left ValidationError untyped to match HITL;

libs/code/deepagents_code/hooks/client.py
johannes117: no impact on timeout cancellation
handler timeouts fire unchanged inside the shielded task

PR #5002, opened 2026-07-23
feat(evals): per-leaf token & cost analysis for unified evals

davibinboi:
## What this PR does

- Adds **per-variant token & cost reporting** to the unified eval workflow. For every variant tested — a (branch × agent config × model) combination — it shows **total input tokens, total output tokens, and total cost (USD)**.
- Renders these in a new **"Token usage and cost"** table in the run summary: one row per variant, right beside the existing accuracy leaderboard.
- Reports cost in **two ways** per variant: **completed-only** (trials that finished cleanly) and **true spend** (every trial), plus a **`completed / expected (N errored)`** count — so a variant that erred out is obvious and not silently compared as if it were clean. Expected counts come from **tasks × rollouts** (not shard count), so coverage is reported correctly.
- Pulls the numbers from **LangSmith** (the only place token/cost data exists — local result files only carry pass/fail), then joins them back to each variant.
- **Keeps the API key isolated**: only the one usage-collection job (scoped to the `evals` Environment, `contents: read` only) can read from LangSmith; the chart-publishing job never sees the key. If LangSmith is unavailable, the columns degrade to `—` instead of failing the run.
- **Leaves the accuracy leaderboard and all existing metrics unchanged.**

### Built on stock Harbor (no fork, no custom plugin)

- Runs on **released Harbor `0.20.0`** (`harbor[langsmith]>=0.20.0,<0.21.0` + `harbor-langsmith>=0.3.0,<0.4.0`) — **no `harbor_package_override` fork**. Harbor 0.20's stock `langsmith` plugin already logs each leaf to exactly the LangSmith experiment name it is given, so the custom `unified_langsmith_plugin` is **deleted**.
- **LangSmith experiment names are computed up front** from a single shared helper (`.github/scripts/experiment_name.py`), used by the run step, the prep step, and the aggregator — so the name can never drift between where a leaf is logged and where its cost is read back. This replaces the previous marker-file round-trip.
- **Standard exception-based retry** (`--max-retries`) — the fork-only `--retry-if-reward-below` reward-gate is removed.

## Tests
- `.github/scripts` helper-script suite passes (experiment-name helper, prep, collector, shard/row aggregation, workflow wiring). `libs/evals` unit tests pass (Harbor + LangSmith integration). `ruff format`, `ruff check`, and `ty` type-check are clean over the eval package; workflow YAML parses.
- Verified **end to end on a live stock-Harbor 0.20 Docker run** (no override) comparing two branches (`main` vs `ss/benchmark-unified-evals-no-todos`), model `gpt-5.6-terra`, 2 context tasks × 2 rollouts. The usage job pulled real per-variant token/cost from LangSmith and rendered the table:
  - `main / bare`: **208,526 tokens · $0.19** (2/4 completed, 2 errored)
  - `ss/benchmark-unified-evals-no-todos / bare`: **345,562 tokens · $0.30** (2/4 completed, 2 errored)

  (The errored rollouts are the eval subjects themselves failing on those tasks — the report surfaces them via the `completed / expected (N errored)` count rather than hiding them, which is exactly the feature.)
- Standard `--max-retries` retry verified on a separate run: a failed trial is retried and the "actual retries" summary still reads `result.json` `stats.n_retries`.

## Eval Run with Changes
https://github.com/langchain-ai/deepagents/actions/runs/30129067425


.github/scripts/collect_langsmith_usage.py
```diff
@@ -0,0 +1,403 @@
...
+    rollouts = summary.get("rollouts_per_task")
+    if (
+        isinstance(expected_shards, int)
+        and not isinstance(expected_shards, bool)
+        and expected_shards >= 0
+        and isinstance(rollouts, int)
+        and not isinstance(rollouts, bool)
+        and rollouts > 0
+    ):
+        return expected_shards * rollouts
```
nick-hollon-lc: one shard can have more than 1 task so i don't think this is right

libs/evals/deepagents_harbor/unified_langsmith_plugin.py
nick-hollon-lc, line 1: i don't think we need this file - if you use harbor main this should be fixed

.github/scripts/aggregate_shards.py
nick-hollon-lc, line 1: i think most of this langsmith experiment code is unnecessary with the changes now on main plus harbor supporting a single experiment

nick-hollon-lc: actually unless this is needed so you know what experiments to look in? but i would guess that we don't need to push experiment name into the shards since we should know it up front somewhere

nick-hollon-lc (changes requested):
tested this and it works but i see some things i don't love in the actual code

.github/scripts/collect_langsmith_usage.py
```diff
@@ -0,0 +1,377 @@
...
+    expected = coverage["expected_rollouts"]
+    if expected is None:
+        return usage["status"] == "complete"
+    return all(
+        cast(int, coverage[field]) >= expected
+        for field in ("observed_rollouts", "token_rollouts", "priced_rollouts")
+    )
+
+
+def collect_experiment(
```
nick-hollon-lc: is this used on the production path? if not pls remove

.github/scripts/evals/collect_langsmith_usage.py
```diff
@@ -0,0 +1,377 @@
...
+    one over rollouts that reached a terminal result (``error`` unset), so a
+    leaf's cost can be compared without being skewed by its failure rate.
+    """
+    observed = token_covered = priced = completed = errored = 0
+    completed_token_covered = completed_priced = 0
+    prompt_tokens = completion_tokens = total_tokens = 0
+    completed_prompt = completed_completion = completed_total = 0
+    total_cost = Decimal(0)
+    completed_cost = Decimal(0)
+    for run in runs:
```
nick-hollon-lc, line 160: could we collapse some of the locals in this loop into a status dict that we update so it's a bit easier to track?

nick-hollon-lc (changes requested):

nick-hollon-lc (approved):
lgtm now! thank you!!

PR #5016, opened 2026-07-24
fix(code): trim `web_search` and `fetch_url` tool descriptions

mdrxy:
Shortens the model-facing `web_search` and `fetch_url` tool descriptions by dropping provider branding, response-field inventories, and generic "synthesize for the user" instructions that do not change tool behavior.

---

Tavily remains visible in user-facing setup and approval copy; only the LLM tool schema text was trimmed.

### DIFF
--- a/libs/code/deepagents_code/tools.py
+++ b/libs/code/deepagents_code/tools.py
@@ -324,32 +324,17 @@ def web_search(  # noqa: ANN201  # Return type depends on dynamic tool configura
     topic: Literal["general", "news", "finance"] = "general",
     include_raw_content: bool = False,
 ):
-    """Search the web using Tavily for current information and documentation.
-
-    This tool searches the web and returns relevant results. After receiving results,
-    you MUST synthesize the information into a natural, helpful response for the user.
+    """Search the web for current information.
 
     Args:
-        query: The search query (be specific and detailed)
-        max_results: Number of results to return (default: 5)
-        topic: Search topic type - "general" for most queries, "news" for current events
-        include_raw_content: Include full page content (warning: uses more tokens)
+        query: Specific search query.
+        max_results: Number of results to return.
+        topic: `"general"`, `"news"`, or `"finance"`.
+        include_raw_content: Include full page text (large; prefer `fetch_url` for a
+            single URL).
 
     Returns:
-        Dictionary containing:
-        - results: List of search results, each with:
-            - title: Page title
-            - url: Page URL
-            - content: Relevant excerpt from the page
-            - score: Relevance score (0-1)
-        - query: The original search query
-
-    IMPORTANT: After using this tool:
-    1. Read through the 'content' field of each result
-    2. Extract relevant information that answers the user's question
-    3. Synthesize this into a clear, natural language response
-    4. Cite sources by mentioning the page titles or URLs
-    5. NEVER show the raw JSON to the user - always provide a formatted response
+        Search hits with title, URL, snippet, and score.
     """
     try:
         import requests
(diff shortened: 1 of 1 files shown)


libs/code/deepagents_code/tools.py
```diff
@@ -324,32 +324,17 @@ def web_search(  # noqa: ANN201  # Return type depends on dynamic tool configura
...
+    """Search the web for current information.
 
     Args:
-        query: The search query (be specific and detailed)
-        max_results: Number of results to return (default: 5)
-        topic: Search topic type - "general" for most queries, "news" for current events
-        include_raw_content: Include full page content (warning: uses more tokens)
+        query: Specific search query.
+        max_results: Number of results to return.
+        topic: `"general"`, `"news"`, or `"finance"`.
```
mdrxy: why do we have this?

libs/code/deepagents_code/tools.py
```diff
@@ -324,32 +324,17 @@ def web_search(  # noqa: ANN201  # Return type depends on dynamic tool configura
...
-    you MUST synthesize the information into a natural, helpful response for the user.
+    """Search the web for current information.
 
     Args:
-        query: The search query (be specific and detailed)
-        max_results: Number of results to return (default: 5)
-        topic: Search topic type - "general" for most queries, "news" for current events
-        include_raw_content: Include full page content (warning: uses more tokens)
+        query: Specific search query.
+        max_results: Number of results to return.
```
mdrxy: will model still know/see default? yes right?

libs/code/deepagents_code/tools.py
```diff
@@ -324,32 +324,17 @@ def web_search(  # noqa: ANN201  # Return type depends on dynamic tool configura
     topic: Literal["general", "news", "finance"] = "general",
     include_raw_content: bool = False,
 ):
-    """Search the web using Tavily for current information and documentation.
-
-    This tool searches the web and returns relevant results. After receiving results,
-    you MUST synthesize the information into a natural, helpful response for the user.
+    """Search the web for current information.
```
mdrxy, line 347: do we need to bring back this?

```txt
After receiving results, you MUST synthesize the information into a natural, helpful response for the user.
```

libs/code/deepagents_code/tools.py
```diff
@@ -324,32 +324,17 @@ def web_search(  # noqa: ANN201  # Return type depends on dynamic tool configura
...
-            - score: Relevance score (0-1)
-        - query: The original search query
-
-    IMPORTANT: After using this tool:
-    1. Read through the 'content' field of each result
-    2. Extract relevant information that answers the user's question
-    3. Synthesize this into a clear, natural language response
-    4. Cite sources by mentioning the page titles or URLs
-    5. NEVER show the raw JSON to the user - always provide a formatted response
+        Search hits with title, URL, snippet, and score.
```
mdrxy, line 350: do we need to bring back the `IMPORTANT: After using this tool` guidelines?

libs/code/deepagents_code/tools.py
```diff
@@ -394,29 +379,14 @@ def web_search(  # noqa: ANN201  # Return type depends on dynamic tool configura
...
-    This tool fetches web page content and converts it to clean markdown text,
-    making it easy to read and process HTML content. After receiving the markdown,
-    you MUST synthesize the information into a natural, helpful response for the user.
+    """Fetch a URL and return the page content as markdown.
 
     Args:
-        url: The URL to fetch (must be a valid HTTP/HTTPS URL)
-        timeout: Request timeout in seconds (default: 30)
+        url: HTTP or HTTPS URL to fetch.
+        timeout: Request timeout in seconds.
```
mdrxy: same Q re: default

libs/code/deepagents_code/tools.py
```diff
@@ -394,29 +379,14 @@ def web_search(  # noqa: ANN201  # Return type depends on dynamic tool configura
...
-        - markdown_content: The page content converted to markdown
-        - status_code: HTTP status code
-        - content_length: Length of the markdown content in characters
-
-    IMPORTANT: After using this tool:
-    1. Read through the markdown content
-    2. Extract relevant information that answers the user's question
-    3. Synthesize this into a clear, natural language response
-    4. NEVER show the raw markdown to the user unless specifically requested
+        Fetched page markdown plus status metadata.
```
mdrxy, line 407: same Q as `web_search`

libs/code/deepagents_code/tools.py
mdrxy: Dropped. After #5019, arg docs live on `Field(description=...)`, so the docstring `Args` block (including the `topic` enum restatement) just duplicated schema and inflated the tool description.

libs/code/deepagents_code/tools.py
mdrxy: Yes. Defaults stay on the JSON schema properties (`max_results` → 5, `topic` → `"general"`, `include_raw_content` → false). Only required is `query`.

libs/code/deepagents_code/tools.py
mdrxy: No — leaving it out. That line is a generic UX instruction, not tool behavior, which is what this PR is trimming.

libs/code/deepagents_code/tools.py
mdrxy: No — same reason. The numbered IMPORTANT block only restated “read results, synthesize, cite, don’t dump JSON,” which does not change how the tool works.

libs/code/deepagents_code/tools.py
mdrxy: Yes — `timeout` still has `"default": 30` on the parameter schema; only `url` is required.

libs/code/deepagents_code/tools.py
mdrxy: No for `fetch_url` too. Keeping the short summary + return note; online response-style guidance stays out of the tool description.

PR #5045, opened 2026-07-24
feat(code): surface Hooks v2 runtime feedback

johannes117:
Hook execution now reports progress, warnings, notices, and permission decisions consistently in the TUI and non-interactive client. Configured `statusMessage` values appear while handlers run, and server-owned hook output is no longer confined to logs.

---

A shared `HookFeedback` presenter keeps event ownership out of the UI layer while preserving ordered engine reduction. It deduplicates visible diagnostics, tracks concurrent handler statuses without clearing a still-running hook, emits validated terminal sequences once, and attributes hook-owned allow/deny decisions in both clients.

Re-anchored onto `main` after #5105 squash-merged. Call sites keep the parent `WorkspaceTrust` API and layer `HookFeedback` on top.

## Review Guide

1. `hooks/feedback.py` — `HookFeedback` presenter (progress / notices / decisions)
2. `hooks/engine.py` — handler progress reporting
3. `hooks/client_lifecycle.py`, `hooks/client.py` — feedback attachment vs log-only
4. `client/non_interactive.py`, `app.py`, `tui/widgets/status.py` — UI/status sinks
5. Skim: `HOOKS.md`, tiny adapter/runtime/threat-model edits
6. Tests: `test_feedback.py`, then engine / client-lifecycle / status

<details>
<summary>Test plan</summary>

- Added focused presenter, concurrent progress, `statusMessage`, server-notice, and permission-attribution coverage.
- Targeted feedback / engine / lifecycle / trust / status suite passed on the prior stack tip; CI re-runs on this re-anchor.

</details>

-------------------------
- main
  - **feat(code): surface Hooks v2 runtime feedback** :point_left:

<sup>[Stack](https://www.git-town.com/how-to/proposal-breadcrumb.html) generated by [Git Town](https://www.git-town.com)</sup>


libs/code/deepagents_code/hooks/presenter.py
aolsenjazz, line 1: I think the code is good in this file, but this file points to another awkward system design thing. I think right now the ownership hierarchy is

App -> SessionState -> Hook runtime -> HookFeedback -> App (UI callbacks)

Imo ideally we have one stable presenter owned by a parent hooks structure (some sort of coordinator)

libs/code/deepagents_code/hooks/engine.py
```diff
@@ -3,22 +3,34 @@
...
 
     from deepagents_code.hooks.models.domain import HookDecision, HookInvocation
-    from deepagents_code.hooks.snapshot import HooksSnapshot
+    from deepagents_code.hooks.snapshot import HookHandler, HooksSnapshot
+
+logger = logging.getLogger(__name__)
+
+
+def _default_progress_message(handler: HookHandler) -> str:
+    from deepagents_code.config import get_glyphs
```
aolsenjazz: this is really a nit but can we replace `get_glyphs` with `...`? This has more to do with principle than implementation - but I'd like to keep presentation logic out of the engine

libs/code/deepagents_code/hooks/engine.py
```diff
@@ -36,6 +48,7 @@ async def run(
         *,
         transcript_path: Path,
         agent_transcript_path: Path | None = None,
+        progress: Callable[[HookProgress], None] | None = None,
```
aolsenjazz: nit: `on_progress`

libs/code/deepagents_code/hooks/engine.py
```diff
@@ -100,3 +116,57 @@ async def run(
...
+
+async def _run_handler(
+    handler: HookHandler,
+    payload: bytes,
+    *,
+    cwd: Path,
+    default_timeout: float,
+    max_output_bytes: int,
+    operation_id: str,
+    progress: Callable[[HookProgress], None] | None,
```
aolsenjazz: nit: `on_progress`

libs/code/deepagents_code/hooks/client_lifecycle.py
```diff
@@ -172,7 +170,28 @@ class ClientHookService:
...
+
+    def __post_init__(self) -> None:
+        """Resolve one presenter shared with the session runtime when available."""
+        if self.feedback is not None:
+            self._feedback = self.feedback
+        elif self.notice is not None:
+            notice = self.notice
+
+            def sink(message: str, severity: HookFeedbackSeverity) -> None:
+                del severity
```
aolsenjazz: python is not a serious language

libs/code/deepagents_code/hooks/client_lifecycle.py
```diff
@@ -172,7 +170,28 @@ class ClientHookService:
 
     runtime: _ClientHooksRuntime
     notice: Callable[[str], None] | None = None
+    feedback: HookFeedback | None = None
```
aolsenjazz: If we end up going with a larger structure to govern the hooks subsystems, would be great to hoist this

libs/code/deepagents_code/hooks/presenter.py
aolsenjazz: Also nit - would be great to rename HookFeedback to something more true-to-form. I would expect HookFeedback to be a simple domain object like { message: string, severity: typeof SEVERITY.... }

libs/code/deepagents_code/hooks/client_lifecycle.py
aolsenjazz, line 1: Composition gets muddy in this file in particular:

`ClientHookService` own:
- `notice`
- _feedback` and `feedback` which both hold references to their own `notice`
- `runtime` which holds a reference to `feedback` which holds a reference to `notice`

I feel strongly that we need to think more about these relationships

libs/code/deepagents_code/hooks/presenter.py
johannes117: HooksManager is now the coordinator and owns one stable HookPresenter

libs/code/deepagents_code/hooks/engine.py
johannes117: moved presentation default out of the engine

libs/code/deepagents_code/hooks/engine.py
johannes117: renamed

libs/code/deepagents_code/hooks/engine.py
johannes117: renamed

libs/code/deepagents_code/hooks/client_lifecycle.py
johannes117: removed sink wrapper

libs/code/deepagents_code/hooks/client_lifecycle.py
johannes117: notice and feedback were removed and ownership was hoisted to HooksManager

libs/code/deepagents_code/hooks/client_lifecycle.py
johannes117: addressed with refactor

libs/code/deepagents_code/client/non_interactive.py
```diff
@@ -225,6 +243,14 @@ def stop(self) -> None:
             finally:
                 self._live = None
 
+    @staticmethod
+    def _renderable(message: str) -> RichSpinner:
```
aolsenjazz: nit: Not a very helpful function name

libs/code/deepagents_code/client/non_interactive.py
```diff
@@ -329,6 +355,27 @@ def _inert_hooks() -> HooksManager:
     return HooksManager.inert()
 
 
+def _plain_hook_presenter(console: Console) -> HookPresenter:
```
aolsenjazz: nice, I like this pattern

libs/code/deepagents_code/client/non_interactive.py
```diff
@@ -1342,24 +1389,67 @@ async def _run_agent_loop(
...
-        identity=lambda: identity,
-        notice=lambda notice: console.print(Text(notice), highlight=False),
-        # Project hooks require an explicit opt-in, matching `--trust-project-mcp`.
-        # Persisted interactive trust deliberately does not carry into headless
-        # runs, so CI never inherits a grant made at someone's terminal.
-        trust=WorkspaceTrust.explicit_only(Path.cwd(), granted=trust_project_hooks),
-    )
+    if hooks is not None:
+        state.hooks = hooks
+        state.hooks.attach_output(notice=present_hook_notice, status=hook_status)
```
aolsenjazz: Based on my read, it looks like we've carved out an exception to the presentation layer by doing this. is there a reason we don't leverage the hooks-manager-internal presenter?

libs/code/deepagents_code/hooks/engine.py
```diff
@@ -82,12 +90,14 @@ async def run(
         )
         results = await asyncio.gather(
             *(
-                run_command_handler(
+                _run_handler(
```
aolsenjazz, line 94: LLMs have pointed this out a few times in my review, I think they've complained about this that it may be worth addressing in a fast follow:

`1. Load-time config diagnostics re-toast on every hook invocation
HookEngine.run folds self.snapshot.diagnostics into every decision (engine.py:109, and again at :79). present_decision → present_diagnostics now surfaces warnings/errors to the user; before this PR _apply_common_effects only logged them, so nobody noticed they were in every decision.`

No-op for now is fine

aolsenjazz (approved):
Just a few nits, but I'm ready to stamp. I think the design is more inline with what we want. just check in with the comment I made re. the exception made for the noninteractive presentation layer

libs/code/deepagents_code/client/non_interactive.py
johannes117: renamed to _build_spinner

libs/code/deepagents_code/client/non_interactive.py
johannes117: removed

libs/code/deepagents_code/hooks/engine.py
johannes117: addressed

Issue #3664, opened 2026-05-29
Empty binary reads emit an empty content block instead of the empty-file warning


mdrxy:
### Submission checklist

- [x] This is a bug, not a usage question.
- [x] I added a clear and descriptive title.
- [x] I searched existing issues and didn't find this.
- [x] I can reproduce this with the latest released version.
- [x] I included a minimal reproducible example and steps to reproduce.

### Area (Required)

- [x] deepagents (SDK)

### Related Issues / PRs

* #3657 — surfaced this during review; this is **pre-existing** and not caused by that fix.

### Reproduction Steps / Example Code (Python)

```python
"""Empty binary reads emit an empty content block instead of the empty-file
warning that empty text reads produce.

Run:
    python demo_empty_binary_read.py
"""

from deepagents.backends.protocol import ReadResult
from deepagents.backends.state import StateBackend
from deepagents.middleware.filesystem import FilesystemMiddleware, FilesystemState
from langchain.tools import ToolRuntime


def run(path: str, content: str, encoding: str):
    class B(StateBackend):
        def read(self, p, *, offset=0, limit=100):
            return ReadResult(file_data={"content": content, "encoding": encoding})

    middleware = FilesystemMiddleware(backend=B())
    tool = next(t for t in middleware.tools if t.name == "read_file")
    runtime = ToolRuntime(
        state=FilesystemState(messages=[], files={}),
        context=None,
        tool_call_id="x",
        store=None,
        stream_writer=lambda _: None,
        config={},
    )
    return tool.invoke({"file_path": path, "runtime": runtime})


# Empty TEXT file -> clear "empty contents" warning
m = run("/a/empty.txt", "", "utf-8")
print("text  :", repr(m.content))

# Empty BINARY file -> empty content block, status="success", no signal
m = run("/a/empty.pdf", "", "base64")
print("binary:", repr(m.content))
```

### Error Message and Stack Trace (if applicable)

```shell
text  : 'System reminder: File exists but has empty contents'
binary: [{'type': 'file', 'base64': '', 'mime_type': 'application/pdf'}]
```

### Description

* When `read_file` reads an **empty text** file, `_handle_read_result` runs `check_empty_content` and returns a clear message — `System reminder: File exists but has empty contents`.
[...]

PR #5050, opened 2026-07-24
fix(sdk): keep blank read windows out of the empty-file warning

### LINKED ISSUE #4955: `read_file` reports a non-empty file as "File exists but has empty contents" when the requested line range is blank
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

Not a duplicate of any open issue. Related work touches the same `check_empty_content` helper but for a different trigger (truly empty binary reads, not a blank window of a non-empty file):

- #3664 — Empty binary reads emit an empty content block instead of the empty-file warning (closed).
- #3675 — fix(sdk): guard empty binary reads with empty-content warning (closed PR). Fix for #3664.

```python
Requires only `pip install deepagents` — no model or API key needed. The repro drives `FilesystemBackend.read` (any backend behaves the same) and then applies the exact emptiness check the middleware runs in `_handle_read_result`.

import tempfile

from deepagents.backends.utils import check_empty_content

root = tempfile.mkdtemp()

# A NON-empty file whose lines 2-5 are blank.
backend.write("/f.txt", "line1\n\n\n\n\nline6\n")

# Read the blank window (0-indexed offset=1, limit=4 -> lines 2..5).
read_result = backend.read("/f.txt", offset=1, limit=4)
window = read_result.file_data["content"]
print("window content:", repr(window))   # '\n\n\n\n' — a blank slice of a non-empty file

# This is exactly what FilesystemMiddleware._handle_read_result does
# (libs/deepagents/deepagents/middleware/filesystem.py:1756):
print("middleware verdict:", check_empty_content(window))
# Observed: 'System reminder: File exists but has empty contents'
# Expected: the four blank lines rendered with line numbers 2-5.

---
```

```shell
No exception is raised — the failure is silent. `read_file` returns the "File exists but has empty contents" warning instead of the (blank) requested lines.
```

### Description

`FilesystemMiddleware._handle_read_result` re-applies the *whole-file* empty check to the *paginated window* returned by the backend:

```python
content = read_result.file_data["content"]   # this is the SLICED window (offset..limit)
empty_msg = check_empty_content(content)
if empty_msg:
    return ToolMessage(content=empty_msg, ...)  # "File exists but has empty contents"
```

`check_empty_content` (`libs/deepagents/deepagents/backends/utils.py:254-265`) is designed for whole files, and the backends already short-circuit genuinely empty files *before* slicing (`utils.py:449`, `filesystem.py:483`). So the only way to reach this branch is a whitespace-only window of a file that has content elsewhere — e.g. `read_file(path, offset=99, limit=20)` landing on blank lines.

**Expected:** a blank window renders the (blank) lines with their line numbers; the "empty file" warning is reserved for genuinely empty files.

**Currently happening:** the model is told the entire file is empty and never sees the blank lines it requested, which can make it wrongly conclude the file has no content and stop. Affects all backends via the shared middleware.

**Suggested fix:** do not re-run `check_empty_content` on a sliced window in the middleware; rely on the backend's whole-file empty handling and render the window normally.

---

### Environment / System Info

- OS: Windows 11 (also reproducible on Linux/macOS — logic is platform-independent)
- Python: 3.11
- deepagents: 0.6.12
- langchain: >=1.3.12,<2.0.0
- langchain-core: >=1.4.9,<2.0.0
- Relevant env vars: none (repro needs no model/API key)

rajarshidattapy:
# Description

Fixes #4955

`read_file` with an offset/limit that lands on blank lines reported "System reminder: File exists but has empty contents" for a file that has content elsewhere — the model was told the whole file was empty and never saw the lines it asked for. Blank windows now render as blank numbered lines, and the warning is reserved for files that really are empty.

---

`_handle_read_result` ran `check_empty_content` on `read_result.file_data["content"]`, which is the *sliced window*, not the whole file. Backends short-circuit genuinely empty files before slicing, so the only way to reach that branch is a whitespace-only window of a non-empty file.

Removing the check outright (as the issue suggested) would regress #3664: #3675 deliberately moved it ahead of type/encoding routing so empty binary reads return the warning instead of a degenerate empty content block, and State/Store backends rely on it for empty text content.

So instead the check is gated on `read_result.start_line is None`. Pagination metadata is set only when a backend actually sliced a window out of a file that has content — both whole-file short-circuits (`backends/utils.py:449`, `backends/filesystem.py:483`) leave those fields unset, and binary reads never set them. A paginated window is therefore never "this file is empty", while empty text files and empty binary reads (mapped and unmapped extensions) keep their warning.

The repro from the issue now returns:

```text
2
3
4
5

[Read 4 lines (lines 2-5 of 6 total). 1 line remaining from offset 5.]
```

One regression test added alongside the existing empty-file tests in `tests/unit_tests/test_middleware.py`. The unit suite passes locally; the two failures I see are pre-existing Windows symlink-privilege failures that also fail on `main`.

Out of scope but worth a follow-up: `FilesystemBackend` returns an empty file's warning *as* `file_data["content"]`, so it renders as `1  System reminder: ...` rather than a bare warning. That is pre-existing and the convention is asserted by the filesystem/sandbox/langsmith-sandbox backend tests, so changing it is a separate cross-backend change.


rajarshidattapy:
can you take a check @mdrxy ?

rajarshidattapy:
Hey @mdrxy , I've updated the branch, no merge conflicts now!

ccurme (approved):
Thanks for this! I slimmed this down a bit and removed the sandbox changes. Those changes were adding overhead just to pick up a pathological case where we page into a file that is completely blank (there's already a check for zero byte files in sandboxes).

PR #5060, opened 2026-07-24
fix(sdk): stop the overflow clip from dropping media and over-clipping

### LINKED ISSUE #4954: Summarization overflow fallback (`_slice_read_file_tm`) silently drops image blocks from `read_file` tool results
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

Not a duplicate of any open issue. This is the same *class* of bug that was fixed previously, but in a different code path that the fix missed:

- #2873 — Summarization middleware image loss (closed). Original report of media loss during summarization.
- #2929 — fix(sdk): preserve image blocks in summarization archive (closed PR). Fixed media loss in the archive/eviction path.
- #3990 — fix(sdk): preserve media references in summarization archives (closed PR). Further hardened the archive path.

Those fixes covered `_message_eviction` / archive handling. The `_overflow_clip._slice_read_file_tm` fallback path was never updated and still discards image blocks.

```python
Requires only `pip install deepagents` — no model or API key needed. The repro calls the overflow-clip helper directly with a `read_file` tool result that carries an image block.

from langchain_core.messages import AIMessage, ToolMessage

from deepagents.middleware._overflow_clip import _clip_overflow_tail

import tempfile

root = tempfile.mkdtemp()
backend.write("/pic.png", "placeholder")  # the real file lives on the backend

# An assistant turn that called read_file on an image, and the ToolMessage
# carrying the returned image content block.
ai = AIMessage(
    content="",
    tool_calls=[{"id": "call_1", "name": "read_file", "args": {"file_path": "/pic.png"}}],
)
tm = ToolMessage(
    tool_call_id="call_1",
    name="read_file",
    content=[{"type": "image", "base64": "aGVsbG8=", "mime_type": "image/png"}],
)
messages = [ai, tm]

new_messages, _replacements = _clip_overflow_tail(
    messages,
    backend,
    keep=("tokens", 1),               # tiny threshold so clipping engages
    max_input_tokens=1000,
    token_counter=lambda msgs: 10_000,  # force the tail batch over the threshold
    large_tool_results_prefix="/large_tool_results",
)

clipped = new_messages[-1]
print("type of content:", type(clipped.content).__name__)
print("content:", clipped.content)

# Observed: content is a plain string ("...[Output was truncated...]") — the
# {"type": "image", ...} block is gone.
# Expected: the image block is preserved (as the archive/eviction path does),
# and no truncation notice is appended to an already-small result.
```

```shell
No exception is raised — the failure is silent. The image content block is dropped from the message and replaced with a text-only truncation notice.
```

### Description

When `SummarizationMiddleware` hits a `ContextOverflowError`, it calls `_clip_overflow_tail` to shrink the trailing `ToolMessage` batch. For any `read_file` result, `_slice_read_file_tm` (`libs/deepagents/deepagents/middleware/_overflow_clip.py:76-93`) rebuilds the message from `_extract_text_from_message(msg)`, which returns only text blocks:

```python
def _slice_read_file_tm(msg, original_path):
    content = _extract_text_from_message(msg)          # text blocks only
    notice = "\n\n[Output was truncated due to context window size limits. ...]"
    return msg.model_copy(update={"content": content[:4_000] + notice})
```

When the `read_file` result is an image (or other multimodal) block, `content` is `""`, so the message becomes a text-only notice and the image block is discarded from the conversation. This is inconsistent with `_message_eviction._build_evicted_content`, which explicitly preserves non-text blocks.

A secondary problem: the truncation is unconditional. A small `read_file` result that happens to sit in an over-threshold tail batch still gets a misleading "Output was truncated" notice appended even though nothing was actually truncated.

**Expected:** non-text (image/audio/video) blocks are preserved, and the truncation notice is only added when the content actually exceeds the slice size (`len(content) > 4_000`).

**Currently happening:** image blocks are silently dropped; small results get a false truncation notice.

**Suggested fix:** reuse `_build_evicted_content` to retain media blocks, and gate the notice on `len(content) > 4_000`.

### Environment / System Info

- OS: Windows 11 (also reproducible on Linux/macOS — logic is platform-independent)
- Python: 3.11
- deepagents: 0.6.12
- langchain: >=1.3.12,<2.0.0
- langchain-core: >=1.4.9,<2.0.0
- Relevant env vars: none (repro needs no model/API key)

rajarshidattapy:
Fixes #4954

When summarization's overflow fallback clips the trailing tool-result batch, a `read_file` result that returned an image became a text-only "Output was truncated" notice — the image was dropped with nothing pointing at where it went. Clipped media now leaves an explicit pointer to the original `file_path`, and small results that happen to sit in an over-budget batch are left alone instead of being rewritten and labelled truncated.

---

`_slice_read_file_tm` rebuilt the message from `_extract_text_from_message`, which returns text blocks only, so a media-only result collapsed to `""` plus a notice. It also appended that notice unconditionally, including to results that were never truncated.

Two changes:

- The truncation notice is now emitted only when the text actually exceeds the slice size, and a separate notice covers removed media. Media payloads are still not carried over — this path runs *because* the batch overflowed the context window, so keeping the inline base64 would defeat the clip — but the notice names the original path, which still holds the file, so the agent can re-read it.
- `_is_worth_clipping` skips results already smaller than the slice size. Rewriting those cannot shrink the batch: on the generic path the replacement is a `TOO_LARGE_TOOL_MSG` stub with a head+tail preview, which is frequently longer than the result it replaces. Everything big enough to matter is still clipped, so the aggregate reduction the post-clip retry depends on is unchanged.

On the aggregate guarantee: I first tried the narrower reading of "leave results untouched when possible" — clip largest-first and stop once the batch is back under the threshold. That breaks `test_summarization_clips_read_file_batch_on_overflow` and its siblings, because the `keep`-derived threshold is not the model's real input limit; stopping early left the retry still overflowing. Skipping only what is too small to shrink gets the "don't rewrite what wasn't the problem" behavior without weakening the reduction.

New unit tests in `tests/unit_tests/middleware/test_overflow_clip.py` cover the media pointer, small siblings surviving intact, no false truncation notice, and async parity. The existing end-to-end clip tests still pass unchanged.

Not addressed here: on the generic (non-`read_file`) path, `_build_evicted_content` still retains media blocks inline, so an image in some other tool's result won't shrink either. That helper is shared with the archive/eviction path where retaining media is deliberate (#3990), so changing it is a separate decision rather than something to fold into this fix.


rajarshidattapy:
can you take a check @mdrxy ?

rajarshidattapy:
Hey @mdrxy , I've updated the branch, no merge conflicts now!

mdrxy:
Would you mind re-opening this rewritten against #5563 ?

rajarshidattapy:
@mdrxy sure on it!

rajarshidattapy:
Reopened as #5638, based on `mdrxy/sdk/preview-note-conditional` (#5563). cc @mdrxy

PR #5104, opened 2026-07-27
feat(code): integrate Hooks v2 client lifecycle events

johannes117:
Supersedes #5010 (auto-closed when #4997 squash-merged; GitHub seals force-pushed closed PRs).
Related: [DCD-71](https://linear.app/langchain/issue/DCD-71/hooks-v2-integrate-client-owned-lifecycle-events)

dcode now runs client-owned Hooks v2 lifecycle events consistently in interactive and headless sessions, including hook-driven approval decisions before permission prompts are shown.

---

- Invokes `SessionStart` and `SessionEnd` at startup, resume, clear, compact, switch, and exit boundaries.
- Applies `PermissionRequest` allow, deny, ask, interrupt, notice, terminal, and stop effects before client approval resolution.
- Routes supported dcode notifications through one typed service with explicit wire mappings while avoiding migrated legacy duplicates.
- Preserves session-start context for the next model turn and adds TUI/headless decision-parity coverage.
- Ignores generated local transcript state through `.gitignore`.

This branch was re-anchored onto `main` after #4997 squash-merged; the diff is identical to the approved #5010 head (`f4e8ee672`), minus a stale resurrection of `todo_list_prompt.md` that #5098 deleted on main.

## Review Guide

1. `hooks/client_lifecycle.py` — `ClientHookService`, context, permission outcomes
2. `app.py` — SessionStart/End boundaries + service wiring
3. `tui/textual_adapter.py`, `client/non_interactive.py` — interactive vs headless parity
4. Skim: `transcript.py`, `server_middleware.py` (PermissionRequest), `runtime.py` / `projection.py`
5. Tests: `test_client_lifecycle.py`, then adapter / non-interactive / server-lifecycle deltas

<details>
<summary>Test plan</summary>

- 266 hooks unit tests pass on the rebuilt tip
- 412 non-interactive / Textual adapter / status widget tests pass

</details>

-------------------------
- main
  - **feat(code): integrate Hooks v2 client lifecycle events** :point_left:
    - https://github.com/langchain-ai/deepagents/pull/5105


libs/code/deepagents_code/hooks/client_lifecycle.py
```diff
@@ -0,0 +1,451 @@
...
+        return PermissionHookOutcome({"type": "approve"})
+    if permission.behavior == "deny":
+        denied = PermissionReviewDecision(type="reject")
+        if permission.reason:
+            denied["message"] = permission.reason
+        return PermissionHookOutcome(denied, interrupt=permission.interrupt)
+    return PermissionHookOutcome(None)
+
+
+def permission_review_payload(
```
aolsenjazz: Not a big deal but any reason we wouldn't just `dict(decision)`

libs/code/deepagents_code/hooks/client_lifecycle.py
```diff
@@ -0,0 +1,451 @@
...
+        )
+
+
+@dataclass(slots=True)
+class ClientHookService:
+    """Execute client-owned events and apply their common side effects."""
+
+    runtime: _ClientHooksRuntime
+    notice: Callable[[str], None] | None = None
+    _session_context: dict[str, list[str]] = field(default_factory=dict)
```
aolsenjazz: Can we type this better? Or name the var differently? This is just a bit too ambiguous imo

libs/code/deepagents_code/app.py
```diff
@@ -4290,45 +4307,155 @@ async def _post_paint_init(self) -> None:
...
-                state.hooks_runtime = HooksRuntime.create(
-                    cwd=Path(self._cwd),
-                    workspace_trusted=False,
+                state = TextualSessionState(
+                    approval_mode=self._approval_mode,
+                    thread_id=self._lc_thread_id,
                 )
+                try:
+                    # Interactive sessions keep project hooks off until a dedicated
+                    # workspace-trust prompt lands (design-doc security follow-up).
```
aolsenjazz:
```suggestion
                    # workspace-trust prompt lands
```

libs/code/deepagents_code/app.py
```diff
@@ -4212,6 +4228,7 @@ async def _post_paint_init(self) -> None:
             group="startup-skill-discovery",
         )
 
+        self._session_init_started = True
```
aolsenjazz: We're also running this inside of `_init_session_state`

It also looks like we're not reading this var, just writing

libs/code/deepagents_code/app.py
```diff
@@ -3560,6 +3573,9 @@ def __init__(
         Lazily constructed by the session-init worker so we don't block
         startup on it.
         """
+        self._session_state_ready = asyncio.Event()
+        self._session_init_started = False
+        self._session_init_lock = asyncio.Lock()
```
aolsenjazz: All of these effective locks are smelly to me. I don't _think_ these are doing anything meaningful

aolsenjazz: Kimi agrees - please verify:

_session_init_lock: prevents the background initializer (app.py:4232) and inline startup fallback (app.py:8204) from constructing two session states. Since construction is now synchronous, it is mostly defensive.
• 
_session_state_ready and _session_init_started: accomplish nothing currently—they are only set, never read or awaited. They are leftovers from the prior worker/to_thread design and can be removed.

libs/code/deepagents_code/app.py
aolsenjazz, line 1: Would you take a look at this file again and consider how we can further extract hook logic and state ownership to some other structure? Ideally, we have a `HooksManager` or some other high-level coordinator that can own the hooks runtime, clients hooks etc. The ideal result (though probably idealistic) would be that `app.py` can call `HooksManager.on_session_end()` or something similar

Instead there a _lot_ of logic surrounding these callsites e.g. `SessionStart` hook callsite has a lot of runtime/hooks inspection before invoking - I would assume that we want the hooks structures to manage this logic

Something else which suggests a composition issues is that both `app.py` and the client lifecycle own the hooks runtime. What is the correct composition here?

ok I made a pic to demonstrate what I'm trying to communicate

Image

aolsenjazz: It should also be able to run its own initialization as much as possible outside of app.py

aolsenjazz (commented):
Please take a look at app.py and lmk what you think. Left a number of comments re. composition + ownership

libs/code/deepagents_code/tui/textual_adapter.py
aolsenjazz, line 1: This file is exhibiting the same issues as app.py imo. Even if we decide to leave things mostly unchanged, we can't be throwing domain specific utility functions into a massive core class

libs/code/deepagents_code/tui/textual_adapter.py
```diff
@@ -44,6 +52,13 @@ class _TokensShowCallback(Protocol):
 
         def __call__(self, *, approximate: bool = False) -> None: ...
 
+    class _ClientHookSessionState(Protocol):
```
aolsenjazz: Not the right place for this

libs/code/deepagents_code/tui/textual_adapter.py
```diff
@@ -94,6 +113,122 @@ def __call__(self, *, approximate: bool = False) -> None: ...
 _ASK_USER_UNSUPPORTED_ERROR = "ask_user not supported by this UI"
 
 
+def _client_hook_context(session_state: _ClientHookSessionState) -> ClientHookContext:
+    from deepagents_code.approval_mode import ApprovalMode, coerce_approval_mode
+    from deepagents_code.hooks.client_lifecycle import ClientHookContext
+
+    return ClientHookContext.create(
+        thread_id=session_state.thread_id,
+        approval_mode=coerce_approval_mode(
```
aolsenjazz: No need to coerce; `ClientHookSessionState` ensures this is type `ApprovalMode`

We shouldn't need get attr either; `ApprovalMode` is not none-able

libs/code/deepagents_code/tui/textual_adapter.py
```diff
@@ -44,6 +52,13 @@ class _TokensShowCallback(Protocol):
 
         def __call__(self, *, approximate: bool = False) -> None: ...
 
+    class _ClientHookSessionState(Protocol):
+        thread_id: str
+        approval_mode: ApprovalMode
+        turn_id: str | None
+        client_hooks: ClientHookService | None
```
aolsenjazz: `client_hooks` suggests something different than the entire service

libs/code/deepagents_code/tui/textual_adapter.py
```diff
@@ -94,6 +113,122 @@ def __call__(self, *, approximate: bool = False) -> None: ...
...
+        ),
+        prompt_id=getattr(session_state, "turn_id", None),
+    )
+
+
+def _has_client_hook_handlers(
+    session_state: _ClientHookSessionState,
+    event: HookEvent,
+) -> bool:
+    service = getattr(session_state, "client_hooks", None)
```
aolsenjazz: logical disconnect between client_hooks name and service var name

libs/code/deepagents_code/tui/textual_adapter.py
```diff
@@ -94,6 +113,122 @@ def __call__(self, *, approximate: bool = False) -> None: ...
...
+    return service is not None and service.has_handlers(event)
+
+
+async def _notify_client_hook(
```
aolsenjazz: Injecting the hooks service into a utility function to manage the execution of the hooks service `notification` function imo further points to the composition issue here - I feel strongly that we don't want to delegate all of this extra work to util functions/core classes

aolsenjazz (changes requested):
one more callout I should make here: the way the system is designed, everything is super defensive. I don't see a world where a client hooks service is ever `None`. Imo it doesn't matter if init fails + the service is unusable; consumers of the hooks system shouldn't ever need to do a defined-check on a service

libs/code/deepagents_code/hooks/client_lifecycle.py
johannes117: renamed _session_context to _pending_context to make it more clear that these are context strings pending injection into the next model turn. also added a comment for more clarity

libs/code/deepagents_code/app.py
johannes117: addressed

libs/code/deepagents_code/app.py
johannes117: addressed

libs/code/deepagents_code/app.py
johannes117: addressed

libs/code/deepagents_code/app.py
johannes117: addressed

libs/code/deepagents_code/tui/textual_adapter.py
johannes117: addressed

libs/code/deepagents_code/tui/textual_adapter.py
johannes117: addressed

libs/code/deepagents_code/tui/textual_adapter.py
johannes117: addressed

libs/code/deepagents_code/tui/textual_adapter.py
johannes117: addressed

libs/code/deepagents_code/tui/textual_adapter.py
johannes117: addressed

libs/code/deepagents_code/tui/textual_adapter.py
johannes117: addressed

aolsenjazz (approved):
we ball

PR #5105, opened 2026-07-27
feat(code): add project hooks workspace trust

johannes117:
Supersedes #5044 (auto-closed when #4997 squash-merged; GitHub seals force-pushed closed PRs).

Project-level hooks now require an explicit workspace trust decision before their commands run. Interactive users can allow once, remember the workspace, or skip project hooks; headless runs remain opt-in through `--trust-project-hooks`.

---

This closes the execution gap that left interactive project hooks permanently disabled while preserving fail-closed behavior. Trust is keyed to the canonical repository root, saved atomically, and enforced again at runtime if a snapshot is ever constructed inconsistently. `config path` and the threat model now expose the relevant project, user, and trust locations.

Re-anchored onto `main` after #5104 squash-merged, so the PR diff contains only this branch's commits.

## Review Guide

1. `hooks/trust.py` — store, keys, atomic write, trust APIs
2. `hooks/loading.py` — project source ingest vs skip
3. `hooks/runtime.py` — fail-closed runtime guard
4. `main.py` — `--trust-project-hooks` + interactive prompt
5. Skim: `app.py`, `client/commands/config.py`, `THREAT_MODEL.md`
6. Tests: `test_trust.py`, then config/app/main touches

<details>
<summary>Test plan</summary>

- Full hooks suite passes on the rebuilt stack tip. `test_trust.py` keeps one case per regressable behavior: canonical persistence, fail-closed corrupt-store handling, project-source provenance, the runtime trust guard, the three prompt outcomes, prompt suppression on persisted trust, and Textual wiring.

</details>

-------------------------
- main
  - **feat(code): add project hooks workspace trust** :point_left:

### DIFF
--- a/libs/code/deepagents_code/app.py
+++ b/libs/code/deepagents_code/app.py
@@ -2777,6 +2777,7 @@ class DeepAgentsApp(App):
         model_explicitly_set: bool = False,
         interpreter_arg: bool | None = None,
         defer_server_start: bool = False,
+        trust_project_hooks: bool = False,
         title: str | None = None,
         sub_title: str | None = None,
         **kwargs: Any,
@@ -2844,6 +2845,7 @@ class DeepAgentsApp(App):
                 `server_kwargs`.
             defer_server_start: Whether to keep app-owned server startup paused
                 until the user configures credentials or explicitly picks a model.
+            trust_project_hooks: Whether project-scoped hook commands may run.
             title: Override the Textual `App.title` shown in the optional
                 header bar (shown when `DEEPAGENTS_CODE_SHOW_HEADER` is set or
                 the installation is stale).
@@ -2867,6 +2869,7 @@ class DeepAgentsApp(App):
             self.sub_title = sub_title
 
         self._register_custom_themes()
+        self._trust_project_hooks = trust_project_hooks
 
         self.theme = _load_theme_preference()
         """Active Textual theme name.
@@ -4385,12 +4388,12 @@ class DeepAgentsApp(App):
             return
         # Loading stays on the event loop deliberately: it is a cheap config
         # read, and `to_thread` races server-ready startup tests that only
-        # yield a few event-loop turns. Interactive sessions keep project hooks
-        # off until a dedicated workspace-trust prompt lands.
+        # yield a few event-loop turns.
         session_state.hooks = HooksManager.create(
             cwd=Path(self._cwd),
             identity=session_state.hook_identity,
             notice=lambda message: self.notify(message, markup=False),
+            workspace_trusted=self._trust_project_hooks,
         )
         # Re-read the app-owned selection last so a mode change during
         # construction cannot be overwritten by the freshly built state.
@@ -4426,7 +4429,10 @@ class DeepAgentsApp(App):
         """Reload hook configuration after the session working directory moves."""
         from pathlib import Path
 
-        await self._hooks.reload(cwd=Path(self._cwd))
+        await self._hooks.reload(
+            cwd=Path(self._cwd),
+            workspace_trusted=self._trust_project_hooks,
+        )
 
     async def _run_session_start_hook(self, cause: SessionStartCause) -> bool:
         """Run `SessionStart`, surfacing a stop as a chat message.
@@ -23208,6 +23214,7 @@ async def run_textual_app(
     model_explicitly_set: bool = False,
     interpreter_arg: bool | None = None,
     defer_server_start: bool = False,
+    trust_project_hooks: bool = False,
     title: str | None = None,
     sub_title: str | None = None,
 ) -> AppResult:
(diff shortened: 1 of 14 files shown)


libs/code/deepagents_code/app.py
```diff
@@ -2727,6 +2727,7 @@ def __init__(
         model_explicitly_set: bool = False,
         interpreter_arg: bool | None = None,
         defer_server_start: bool = False,
+        trust_project_hooks: bool = False,
```
aolsenjazz: It's not great imo that we're storing this piece of state here when the function of this file w.r.t .`trust_project_hooks` is really to just pass this to a different structure to govern + apply

If you end up creating a higher level manager, we should be able to construct here to pass fields + init during the rest of the initialization block?

That should also allows us to remove these none-checks when trying to dispatch hooks

libs/code/deepagents_code/app.py
```diff
@@ -23154,6 +23155,7 @@ async def run_textual_app(
     model_explicitly_set: bool = False,
     interpreter_arg: bool | None = None,
     defer_server_start: bool = False,
+    trust_project_hooks: bool = False,
```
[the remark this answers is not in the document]
aolsenjazz: Reasonable-ish-enough comment

libs/code/deepagents_code/main.py
```diff
@@ -2941,11 +2944,16 @@ def _project_mcp_picker_has_terminal() -> bool:
 
 def _run_project_mcp_trust_action_picker(
```
aolsenjazz: This function is either overloaded or just incorrect w.r.t. hooks usage.

If we're going to execute something similar for hooks confirmation, we need to either rewrite this function to generify failure modes to create a separate functions for hook confirmation

libs/code/deepagents_code/main.py
```diff
@@ -3541,6 +3556,97 @@ def _check_mcp_project_trust(
...
+        "Only allow hooks for projects you trust. Future edits to this file "
+        "will run without asking again if you always allow.",
+        style="yellow",
+        highlight=False,
+    )
+    action = _select_project_mcp_trust_action(
+        prompt_console,
+        remember_label=_PROJECT_HOOKS_REMEMBER_LABEL,
+    )
+    if action is _ProjectMcpTrustPromptOutcome.INTERRUPTED:
```
aolsenjazz: Need to change the `select_project_mcp_trust_action` function as outlined above. Reusing `_ProjectMcpTrustPromptOutcome` doesn't feel great here

aolsenjazz (changes requested):
Hopping on a meeting, submitting this feedback for now before I inevitably forget

libs/code/deepagents_code/app.py
aolsenjazz: also tbf that's the only way to do it rn with the system as designed so it checks out why you did it

libs/code/deepagents_code/hooks/trust.py
aolsenjazz, line 1: Not a big thing but based on my read, this is more complex than it needs to be? It looks like we've built support for locks keyed on individual paths - but there will only ever be one store right?

If so I would assume that this `_TRUST_STORE_THREAD_LOCKS: dict[str, threading.Lock] = {}` is overly complex plumbing

aolsenjazz: i.e. we should be able to get away with a single lock

libs/code/deepagents_code/hooks/runtime.py
```diff
@@ -97,7 +102,9 @@ def create(
             snapshot=snapshot,
             transcripts=store,
             engine=engine,
-            cwd=cwd,
+            cwd=project_context.user_cwd,
+            workspace_trusted=workspace_trusted,
```
aolsenjazz, line 115: This isn't an issue now but putting in a pin in this -

If in the future we allow changing working dir (which I could see happening), we'll may want to revisit this; a user could start in a trusted dir, navigate to untrusted, and `workspace_trusted` would still be true

aolsenjazz (commented):
Overall pretty good! I think there are a couple things that would be worth taking a look at (I think mostly surfaced in the last review submission) but generally very good!

libs/code/deepagents_code/hooks/runtime.py
johannes117: addressing this

libs/code/deepagents_code/app.py
johannes117: yep refactoring this

libs/code/deepagents_code/main.py
johannes117: addressed this by generifying

libs/code/deepagents_code/main.py
johannes117: addressed, same as above

libs/code/deepagents_code/hooks/trust.py
johannes117: yep refactored to use single lock

aolsenjazz (approved):
LGTM

PR #5107, opened 2026-07-28
fix(acp): scope cancel() to the requested session

### LINKED ISSUE #5084: bug(acp): cancel() ignores session_id and cancels every in-flight prompt
### Submission checklist

### Area (Required)

- [ ] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [x] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
# AgentServerACP multiplexes many sessions on one instance, but the cancellation
# flag is a single instance-wide boolean rather than per-session state.
#
# Cancelling session A also terminates session B's in-flight prompt.

import asyncio

from deepagents_acp.server import AgentServerACP

agent = AgentServerACP(agent=my_compiled_graph)   # any CompiledStateGraph
agent.on_connect(my_client)

session_a = await agent.new_session(cwd="/tmp", mcp_servers=[])
session_b = await agent.new_session(cwd="/tmp", mcp_servers=[])

task_a = asyncio.create_task(
    agent.prompt([text_block("A")], session_id=session_a.session_id)
)
task_b = asyncio.create_task(
    agent.prompt([text_block("B")], session_id=session_b.session_id)
)
await asyncio.sleep(0)

await agent.cancel(session_id=session_a.session_id)   # cancel ONLY session A

print((await task_a).stop_reason)   # "cancelled"  — correct
print((await task_b).stop_reason)   # "cancelled"  — WRONG, expected "end_turn"
```

```shell

```

### Description

```
`AgentServerACP.cancel()` accepts a `session_id` but never reads it. It sets a single
instance-wide boolean:

    async def cancel(self, session_id: str, **kwargs: Any) -> None:
        """Cancel the current execution."""
        self._cancelled = True

`prompt()` then checks that same shared flag in two places, with no comparison against
the session it is actually serving:

    while current_state is None or current_state.interrupts:
        if self._cancelled:
            self._cancelled = False
            return PromptResponse(stop_reason="cancelled")
        ...
        async for stream_chunk in agent.astream(...):
            if self._cancelled:
                self._cancelled = False
                return PromptResponse(stop_reason="cancelled")

Every other piece of per-conversation state on this class is a dict keyed by session id —
`_session_modes`, `_session_models`, `_session_cwds`, `_session_plans`,
`_session_mcp_servers`, `_session_mode_states`, `_allowed_command_types`. `_cancelled` is
the sole exception, and it is the one piece of state whose scope actually changes
behaviour mid-run.

`new_session()` mints a fresh `session_id` per call and stores its state alongside the
others, so multiple concurrent sessions on one `AgentServerACP` instance are a supported
shape. When two are in flight and a client cancels one, the other observes
`self._cancelled == True` on its very next loop check and returns
`stop_reason="cancelled"` — work nobody asked to stop, silently discarded and reported as
a normal cancellation.

There is a second, subtler consequence of the same root cause. Because whichever prompt
notices the flag first also resets it to `False`, a cancel can be consumed by the wrong
session and the intended target then runs to completion. So the failure is not merely
"too much gets cancelled" — which prompt actually honours the cancel is a race.

Expected: `cancel(session_id)` should only stop the prompt for that session; concurrent
prompts for other sessions should be unaffected.

Suggested direction: replace the boolean with `set[str]` (or a dict) of cancelled session
ids, have `cancel()` add its `session_id`, and have `prompt()` test and clear membership
for its own `session_id`. That matches how the rest of the class already scopes state and
keeps the change inside `server.py`.

Note on scope: `self._agent` is currently a single shared instance too, so heavy
concurrent use has other rough edges. I have deliberately kept this issue to the
cancellation flag, which is self-contained and fixable on its own.

Verified on `main` at commit 8da0ccb. The existing test
(`test_agent.py::test_acp_agent_cancel_stops_prompt`) exercises only one session, which is
why this has not surfaced. Happy to open a PR with a failing regression test if a
maintainer would like to assign it.
```

### Environment / System Info

OS: Linux
Python: 3.13
deepagents-acp: (version from libs/acp/pyproject.toml)
deepagents: 0.6.12
Verified on main @ 8da0ccb

Mohith26:
Fixes #5084

## What

`AgentServerACP.cancel()` ignored its `session_id` argument and set a shared `self._cancelled` boolean. On a multiplexed server, cancelling one session killed every in-flight prompt, and a race decided which prompt actually consumed the cancel.

The boolean is replaced with a `set[str]` of cancelled session ids, scoped exactly like the class's other per-session state (`_session_modes`, `_session_cwds`). `cancel()` adds the id; `prompt()` clears its own id on entry and only honors a cancel for its own session at both check points.

## Tests

`test_acp_agent_cancel_only_cancels_target_session`: two concurrent sessions gated mid-stream with asyncio events; cancel session A, let session B finish first, assert B ends with `end_turn` and A with `cancelled`. On main this fails with B wrongly cancelled. Full libs/acp suite: 84 passed. ruff clean.

Credit to @RagavRida for the report in #5084.


hntrl (approved):
LGTM, thanks! will do some fixups

PR #5130, opened 2026-07-28
fix(code): re-attach media when undo or redo restores its placeholder

mdrxy:
Fix pasted image/video attachments being lost when a deleted `[image N]` placeholder is restored with ctrl+z or ctrl+y.

---

Backspacing a pasted `[image N]` token removed it atomically, but `MediaTracker.sync_to_text` also discarded the attachment permanently. A following ctrl+z restored placeholder-shaped *text* with nothing behind it, so `_find_placeholder_span` no longer treated it as bound — backspace chewed it one character at a time (`[image 1`) and the image was silently dropped from the submitted message.

`MediaTracker` now *detaches* payloads whose token left the draft instead of discarding them, and re-attaches a detached item when a history reversal restores its token. This mirrors the existing collapsed-paste design, where `ChatInput._pasted_contents` is deliberately retained until submit so undo can restore a paste placeholder with its content.

## Only the edit that removed a payload may bring it back

A placeholder token can reappear in the draft for reasons that have nothing to do with undo: the user typing `[image 1]` as prose, pasting text that quotes it, or pressing ↑ to recall an earlier prompt (history stores the display token verbatim). Re-attaching on token presence alone would bind a deleted image to that text and strip the user's own words from the model-facing message.

Re-attachment is therefore keyed on the identity of the concrete `TextArea` edit batch, not on token text. When an edit removes a token, the payload is filed under a marker on that batch; a reversal re-attaches only payloads filed under the batch it is reversing. Textual preserves the batch object when moving it between the undo and redo stacks, so the marker survives.

Token text alone is not sufficient evidence even *within* a real undo. Undoing an edit that falls inside a token — typing a stray character into `[image 1]` and undoing it, say — makes that occurrence look newly inserted to a text diff, which would rebind a deleted image to prose the user typed. Edit identity distinguishes the two.

## Undo and redo are mirror images

Textual binds `undo` and `redo` in the same `BINDINGS` block (`ctrl+z`/`super+z` and `ctrl+y`/`super+y`). Both are routed through one helper, because either direction can be the operation that restores a token: undoing a delete puts it back, and so does redoing a paste. Handling only one leaves the other restoring a placeholder as bare text while its payload sits unreachable in the detached pool — the original bug, on the other key.

Each direction passes its batch identity as both the detaching and the restoring marker. Whichever way a batch runs, it is the only edit that can have removed the tokens it puts back, so one marker serves both roles: redoing a deletion re-detaches, and undoing it again re-attaches.

## IDs held for undo are not reused

Deleting the only placeholder used to restart the counter, so the next paste reused `[image 1]`. Undoing back past that paste then restored a draft written about the *old* image while the message carried the new one — a wrong attachment rather than a missing one.

`add_media` now skips IDs still held by a detached payload. **This is user-visible: after deleting `[image 1]`, the next paste is `[image 2]`, so numbering can look like it skips.** That is the cost of never binding a restored token to a different image.

## The pool is bounded, and running out is not silent

Detached payloads hold full base64 data and a single item can approach `MAX_MEDIA_BYTES` (~27 MB encoded), so the pool is bounded by bytes (`MAX_DETACHED_MEDIA_BYTES`, 32 MB per kind) as well as item count — a count-only cap of ten would admit roughly 270 MB per kind. Eviction is oldest-first and stops at the first payload that does not fit rather than best-fitting smaller ones around it; the newest entry is always retained so the most recent delete stays undoable even when its payload alone exceeds the budget.

Past the cap, a reversal can restore a token whose payload is gone. That token would look attached and reach the model as bare `[image N]` text, so it is logged and surfaced to the user (`[image 1] no longer has media attached — too many deletions to undo`) rather than failing silently. The log is written before the pending set is drained, so a failure to deliver the toast still leaves a record.

## Lifecycle

Detached payloads are released when a message is dispatched to the agent. Slash and shell submits never consume the draft's media, so they clear the tracker outright at the submission boundary: releasing only the detached pool would miss media still *attached* at that moment, which the next keystroke would then detach into a pool nothing releases again — pinning the payload for the rest of the session and reserving its ID.

Detached payloads are deliberately excluded from the per-message snapshot taken at submit: submission replaces the draft via `clear_text`, which resets the undo history, so no undo could reach them — carrying them would pin base64 payloads on every transcript message for the life of the session. `restore()` releases them for the same reason, since the restored text also arrives with a fresh history.

## Undo cost tracks the edit, not the draft

Deciding which occurrences a reversal introduced needs a diff, and a character-level `SequenceMatcher` over a whole draft is quadratic — on a 30 KB draft a single ctrl+z blocked the UI for over seven seconds. The shared prefix and suffix are unchanged by construction, so only the differing middle is diffed, and the result is computed once and shared across both media kinds instead of once per kind. The same 30 KB draft now costs ~0.4 ms.

## Per-kind state lives in one object

Images and videos differ only in their token pattern, so the tracker held two parallel sets of fields and two near-identical ~56-line sync bodies. The invariants above — attached and detached never sharing a token, an ID never reused while a payload still holds it — were upheld by agreement across ten methods with nothing checking them.

`_MediaKindState` now owns one kind's attached list, detached pool, detaching-edit identities, and ID counter. `_rebind` is the sole writer of all four, so the invariants are established in one place, and `_partition` computes the split without mutating anything. The pattern is `self._pattern`, so pairing the wrong pattern with a list is no longer expressible. `MediaTracker`'s existing attributes remain as properties over that state, backed by a single copy of the data.

Media payloads also arrive from `media_utils` genuinely unbound. Those loaders used to fill the required `placeholder` field with a token-shaped `[image]` that fails the numbered-token regex, relying on the tracker to overwrite it; it now defaults to empty, so an unbound payload cannot be mistaken for a numbered one.

## Test Plan

- [ ] Paste an image, backspace the `[image 1]` token, press ctrl+z, then backspace again — the whole token is removed in one keypress.
- [ ] Paste an image, backspace, ctrl+z, then submit — the image is still attached to the message.
- [ ] Paste an image, press ctrl+z to undo the paste, then ctrl+y — the image is attached again and the token deletes atomically.
- [ ] Paste an image, backspace, ctrl+z, then ctrl+y — the token is removed again and no image is attached; a further ctrl+z re-attaches it.
- [ ] Paste an image, backspace, then type `[image 1]` by hand — it stays ordinary text, edits character by character, and no image is attached.
- [ ] Paste an image, backspace, type `[image 1]` by hand, type a stray character inside the token, then ctrl+z — the text is restored and no image is attached.
- [ ] Paste an image, backspace, then press ↑ to recall an earlier prompt containing `[image 1]` — no image is attached.
- [ ] Paste an image, backspace, paste a different image — the new image is `[image 2]` and only it is attached.
- [ ] Paste two images, delete both in one edit, ctrl+z — both are re-attached, and each token carries its own image.
- [ ] Paste an image, submit a `/help` command, then keep typing — no attachment lingers and the next paste is `[image 1]` again.
- [ ] Paste a long block of text (or disable paste collapsing), then press ctrl+z — the undo is instant.

Made by [Open SWE](https://openswe.vercel.app/agents/aaf21c55-c6fb-4325-11c2-aeb60a8f141c)


mdrxy:
Coordination note: #5542 adds atomic cursor navigation for real media placeholders and needs a way to distinguish the concrete attached occurrence from manually typed identical text. Your edit-identity approach appears compatible. Is this PR still intended to land, and if so, will it expose a small stable occurrence/range primitive that #5542 can consume? #5542 will remain independently correct rather than depending on this large lifecycle change.

PR #5162, opened 2026-07-29
fix(infra): harden release-notes draft model configuration

mdrxy:
Release-note drafting now rejects OpenAI Responses-API-only models up front and raises the shared output-token ceiling so reasoning models can finish large changelogs.

---

`RELEASE_BOT_MODEL` is an operator-facing Actions variable, so docs and the helper need to stay honest about which models can actually work. The drafter only calls Chat Completions, but `RELEASING.md` told people to pick any JSON-Schema model — including OpenAI `*-pro` / `*codex*` IDs that only work on Responses. That used to surface as a confusing HTTP failure on the release PR.

The previous 4096 token cap is also too tight for reasoning models: OpenAI counts reasoning against `max_completion_tokens`, and Anthropic extended thinking has the same shape. `responseText()` already fails closed on non-stop finishes, so large releases would post a failure comment instead of truncating silently — still not what we want on the first big multi-package changelog.

This change:
- rejects known Responses-only OpenAI names (mirroring `langchain-openai`'s prefix/`codex` check) with an error that names `RELEASE_BOT_MODEL`
- raises one shared `MAX_OUTPUT_TOKENS` ceiling used by all three providers
- documents the Chat Completions constraint for operators

Out of scope: implementing Responses API support.


.github/RELEASING.md
```diff
@@ -85,14 +85,16 @@ APP_SLUG=<app-slug>
...
+Create the `release-bot` environment without required reviewers or other approval rules, because approval would block automatic drafting. Add `RELEASE_BOT_MODEL` as an environment variable, using an explicit `provider:model` value with one of the supported providers and a model that supports JSON Schema structured output. The model must also accept an output-token limit of at least 32,768, since the helper requests that ceiling on every provider; a model whose own limit is lower rejects the request outright. Every current OpenAI, Anthropic, and Gemini model an operator would reasonably pick clears it — the exceptions are older small models such as the Gemini 2.0 line, which cap at 8,192. Add the matching provider's API key as an environment secret (only the configured provider's key is required). The workflow reads a fixed secret name per provider:
```
mdrxy:
```suggestion
Create the `release-bot` environment without required reviewers or other approval rules, because approval would block automatic drafting. Add `RELEASE_BOT_MODEL` as an environment variable, using an explicit `provider:model` value with one of the supported providers and a model that supports JSON Schema structured output. The model must also accept an output-token limit of at least 32,768, since the helper requests that ceiling on every provider; a model whose own limit is lower rejects the request outright. Every current OpenAI, Anthropic, and Gemini model an operator would reasonably pick clears it — the exceptions are older small models. Add the matching provider's API key as an environment secret (only the configured provider's key is required). The workflow reads a fixed secret name per provider:
```

.github/RELEASING.md
```diff
@@ -85,14 +85,16 @@ APP_SLUG=<app-slug>
...
 
 | `RELEASE_BOT_MODEL` provider | Environment secret name |
 | ------------------------------ | ----------------------- |
 | `openai`                       | `OPENAI_API_KEY`        |
 | `anthropic`                    | `ANTHROPIC_API_KEY`     |
 | `google_genai`                 | `GOOGLE_API_KEY`        |
 
+For `openai:…`, pick a Chat Completions model (for example `openai:gpt-5.5`). This helper only calls Chat Completions, so Responses-API-only models cannot be used. It rejects two patterns up front, mirroring `langchain-openai`: exactly the prefixes `gpt-5-pro`, `gpt-5.2-pro`, `gpt-5.4-pro`, and `gpt-5.5-pro` (matched as prefixes, so date-suffixed variants are covered), and any model name containing `codex`. That list is a fast path, not a general `*-pro` rule — a future `-pro` release it does not name will pass validation and then fail at the API, where the draft-failure comment reports OpenAI's own explanation.
```
mdrxy:
```suggestion
For `openai:…`, pick a Chat Completions model (for example `openai:gpt-5.5`). This helper only calls Chat Completions, so Responses-API-only models cannot be used.
```

.github/SECRETS.md
```diff
@@ -103,7 +103,7 @@ managed package.
 
 | Actions variable | Value | Type | Purpose |
 | --- | --- | --- | --- |
-| `RELEASE_BOT_MODEL` | Variable | Environment | Model used for changelog generation. |
+| `RELEASE_BOT_MODEL` | Variable | Environment | Model used for changelog generation (`provider:model`). Must support JSON Schema structured output and a 32,768-token output ceiling. For `openai:…`, use a Chat Completions model — the four `gpt-5*-pro` prefixes and any name containing `codex` are rejected. See [RELEASING.md](RELEASING.md) for the exact list. |
```
mdrxy:
```suggestion
| `RELEASE_BOT_MODEL` | Variable | Environment | Model used for changelog generation (`provider:model`). Must support JSON Schema structured output and a 32,768-token output ceiling. For `openai:…`, use a Chat Completions model. See [RELEASING.md](RELEASING.md) for the exact list. |
```

PR #5181, opened 2026-07-30
feat(code): register MiniMax as a first-class model provider

octo-patch:
Reason: the first-class credential and base-URL registries omit a MiniMax provider, so its native API key and official OpenAI-compatible endpoints cannot be selected.

## What

Register `minimax` as a recognized model provider so a MiniMax provider (configured as an OpenAI-compatible `class_path` provider) resolves its native credentials and endpoint without the user re-declaring env var names.

- `model_config.py`
  - Add `"minimax": "MINIMAX_API_KEY"` to `PROVIDER_API_KEY_ENV` (the credential registry). This also auto-populates the credential manifest, which is generated from this map.
  - Add `"minimax": ("MINIMAX_API_BASE",)` to `PROVIDER_BASE_URL_ENV` (the base-URL registry). MiniMax is OpenAI-compatible and has no dedicated integration, so the app resolves this provider-specific variable and passes it as `base_url`; the shared `OPENAI_BASE_URL` is intentionally never reused so a real OpenAI endpoint can't be clobbered. Documented in the adjacent comment block.
- `config_manifest.py`
  - Add `"minimax": ("langchain_openai", "openai")` to `_PROVIDER_DEPENDENCIES` so the provider maps to its integration module and existing extra (MiniMax is served through `langchain_openai`, already a core dependency).
- `tui/widgets/auth.py`
  - Add a `PROVIDER_DISPLAY_NAMES` entry (`"MiniMax"`) and a `PROVIDER_API_KEY_URLS` entry (the interface-key page) so `/auth` renders a branded label and a key-acquisition link.
- Tests
  - Extend `TestProviderApiKeyEnv` and add `TestProviderBaseUrlEnv::test_minimax_uses_dedicated_openai_compatible_base` to pin the new registry entries.

The existing exhaustiveness gates (`test_manifest_covers_every_provider_credential`, `test_provider_dependency_metadata_is_exhaustive`, `test_every_known_provider_has_a_display_name`, `test_providers_without_key_url_are_intentionally_omitted`) already require every credential provider to appear across these maps; the additions keep them green.

## Checks

- `ruff format --check` and `ruff check` on the changed files — pass.
- `ty check` on `deepagents_code/model_config.py`, `config_manifest.py`, and `tui/widgets/auth.py` — pass.
- `uv run --group test pytest` for `tests/unit_tests/test_model_config.py`, `tests/unit_tests/test_config_manifest.py`, `tests/unit_tests/test_extras_info.py`, `tests/unit_tests/tui/widgets/test_auth_widgets.py`, and `tests/unit_tests/client/commands/test_extras.py` — 1,027 passed, 1 unrelated pre-existing failure in `test_model_config.py::TestModelConfigLoad::test_unreadable_file_returns_empty_config`, which relies on `chmod 000` denying reads and does not hold when the suite runs as root.


octo-patch:
Bumping this PR: the MiniMax provider registration is complete and tests pass. Reopening for review.

Issue #4397, opened 2026-06-30
Strip or convert unsupported multimedia formats


mdrxy:
### Area

- Deep Agents Code (`libs/code`)

### Summary

Deep Agents Code can send unsupported image attachments (for example HEIC from macOS Photos pasteboard exports) to the model, causing provider `BadRequestError`s before the agent has a chance to respond. The unsupported image block can also remain in conversation state, so later text-only turns fail the same way.

Related: #3864 covers invalid/oversized multimodal blocks emitted by SDK `read_file`; this issue is specifically about Deep Agents Code local media ingestion and persisted request history.

### Observed behavior

A user pasted/dropped a macOS Photos temporary HEIC export path into Deep Agents Code:

```text
/Users/.../Library/Containers/com.apple.Photos/Data/tmp/TemporaryItems/PasteboardItemExports/... .heic
```

The root trace failed in the `model` node / `ChatOpenAI` call with:

```text
BadRequestError: Error code: 400 - The image data you provided does not represent a valid image. Please check your input and try again with one of the supported image formats: ['image/jpeg', 'image/png', 'image/gif', 'image/webp'].
```

On the next turn, the user asked:

```text
can you convert from heic to jpg
```

That text-only request failed with the same provider error before the agent could act, because the prior unsupported HEIC attachment was still present in the messages sent to the model.

### Expected behavior

Deep Agents Code should not send provider-unsupported media blocks to the model. It should either:

- convert readable unsupported images (for example HEIC/TIFF/BMP/ICO) into a provider-supported format such as JPEG or PNG before attaching them, or
- strip/replace unsupported media blocks with a textual placeholder before each model call.

Follow-up text-only turns should not keep failing because a previous unsupported media block was checkpointed into the conversation.

### Possible fixes

- Add an outbound allow-list for image MIME subtypes before constructing multimodal message content. For OpenAI, the safe list is `jpeg`, `png`, `gif`, and `webp`.
- In `get_image_from_path`, if PIL can open an unsupported image format, re-encode it to PNG or JPEG rather than sending original bytes and MIME type.
- Add a request-history sanitizer before model invocation that walks `request.messages` and replaces unsupported image/video content blocks from previous turns with text placeholders, e.g. `[Unsupported image omitted: image/heic]`.
- Make sure unsupported media is not persisted in a form that can repeatedly poison later turns.

### Code pointers

- `libs/code/deepagents_code/media_utils.py`
[...]

Issue #3864, opened 2026-06-11
Filesystem `read_file` can checkpoint invalid or oversized multimodal image blocks


mdrxy:
### Area (Required)

- [x] deepagents (SDK)

### Related Issues / PRs

Related to #3660, but distinct: #3660 tracks authoritative backend-declared file metadata (`encoding`, `mime_type`, etc.) instead of extension guessing. This issue tracks validating/policy-routing binary read results before `read_file` emits multimodal content blocks that can be checkpointed into message history.

### Reproduction Steps / Example Code (Python)

```python
import base64

from deepagents.backends import StateBackend
from deepagents.backends.protocol import ReadResult
from deepagents.middleware.filesystem import FilesystemMiddleware, FilesystemState
from langchain.tools import ToolRuntime


class InvalidBase64ImageBackend(StateBackend):
    def read(self, path, *, offset=0, limit=100):
        return ReadResult(file_data={"content": "not-valid-base64!!!", "encoding": "base64"})


class OversizedImageBackend(StateBackend):
    def read(self, path, *, offset=0, limit=100):
        payload = base64.b64encode(b"x" * (5 * 1024 * 1024 + 1)).decode("ascii")
        return ReadResult(file_data={"content": payload, "encoding": "base64"})


def run(backend):
    middleware = FilesystemMiddleware(backend=backend)
    runtime = ToolRuntime(
        state=FilesystemState(messages=[], files={}),
        context=None,
        tool_call_id="read-1",
        store=None,
        stream_writer=lambda _: None,
        config={},
    )
    read_file = next(tool for tool in middleware.tools if tool.name == "read_file")
    return read_file.invoke({"file_path": "/app/screenshot.png", "runtime": runtime})


invalid = run(InvalidBase64ImageBackend())
oversized = run(OversizedImageBackend())

print(invalid.status, invalid.content[0]["type"], invalid.content[0]["mime_type"], invalid.content[0]["base64"])
print(oversized.status, oversized.content[0]["type"], oversized.content[0]["mime_type"], len(base64.b64decode(oversized.content[0]["base64"])))
```

With `deepagents==0.6.8`, this returns `ToolMessage(status="success")` with image content blocks for both invalid and oversized payloads:

```text
invalid_base64_image: status=success, type=image, mime_type=image/png, base64='not-valid-base64!!!'
oversized_image: status=success, type=image, mime_type=image/png, decoded_len=5242881
```

### Error Message and Stack Trace (if applicable)

[...]

Issue #3660, opened 2026-05-29
Let backends declare file type/mime on `ReadResult` instead of guessing from extension


mdrxy:
### Submission checklist

- [x] This is a feature request, not a bug report.
- [x] I searched existing issues and didn't find this feature.
- [x] I checked the docs and README for existing functionality.
- [x] This request applies to this repo (deepagents) and not an external package.

### Area (Required)

- [x] deepagents (SDK)

### Feature description

Today the read pipeline classifies a file's type **solely from its path extension**, via `_get_file_type` / `_EXTENSION_TO_FILE_TYPE` (`backends/utils.py`). That table is derived from Google's Gemini multimodal supported-formats docs and only covers a fixed subset of extensions; anything else falls through to `"text"`.

This is fragile for **custom backends**, which are a supported extension point. A backend that already knows a file's true type (e.g. a document store that records `application/vnd.openxmlformats-officedocument.wordprocessingml.document` for a `.docx`) has no way to communicate that to the middleware. It can set `FileData["encoding"] = "base64"`, but `_handle_read_result` ignores `encoding` and re-derives the type from the extension — so an unrecognized extension is treated as text and the base64 payload is line-numbered into the LLM context. (That leak is the bug tracked separately in #3657; this request is the structural enhancement that removes the guessing altogether.)

The request: add an **optional, backend-declared** type/mime field to the read return type so backends can state the type authoritatively, and have the middleware honor it — falling back to the existing extension heuristic only when the field is absent (full backward compatibility).

### Proposed solution (optional)

Add an optional `mime_type` to `FileData` (and surface it through `ReadResult`):

```python
class FileData(TypedDict):
    content: str
    encoding: str                  # "utf-8" | "base64"
    mime_type: NotRequired[str]    # backend-declared, e.g. "application/pdf"
    # ...existing created_at / modified_at...
```

`_handle_read_result` prefers the declared value, then the current guess:

```python
mime_type = (
    read_result.file_data.get("mime_type")
    or mimetypes.guess_type("file" + Path(validated_path).suffix)[0]
    or "application/octet-stream"
)
```

The text-vs-binary decision should key off `encoding` (authoritative) rather than the extension; the declared `mime_type` (or, when absent, the extension map) is then used only to pick the multimodal block type (`image` / `video` / `audio` / `file`). Backends that don't set the field behave exactly as today.

Open questions for maintainers:

[...]

Issue #1313, opened 2026-02-12
Enable image reads from files based on model profile


sydney-runkle:
### Privileged issue

- [x] I am a Deep Agents maintainer.

### Issue Content

some models don't support multimodal results from tools, we track this in profiles

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] Other / not sure / general

1Ninad:
Hi @sydney-runkle @mdrxy, For models that don’t support vision, what should the fallback be – explicit error, auto caption, semantic conversion, model routing, something else?

PR #5194, opened 2026-07-30
fix(sdk): scrub multimodal content blocks unsupported by the model's profile

### LINKED ISSUE #37576: anthropic: base64 file blocks with non-PDF text mime types fail with media_type validation error
### Submission checklist

- [x] I added a clear and descriptive title that summarizes this issue.
- [x] I used the GitHub search to find a similar question and didn't find it.
- [x] I am sure that this is a bug in LangChain rather than my code.
- [x] The bug is not resolved by updating to the latest stable version of LangChain (or the specific integration package).
- [x] This is not related to the langchain-community package.
- [x] I posted a self-contained, minimal, reproducible example. A maintainer can copy it and run it AS IS.

### Package (Required)

- [x] langchain-anthropic

### Related Issues / PRs

- #36939 — related core-level bug (`_convert_openai_format_to_data_block` hard-codes `mime_type="application/pdf"`). This issue is distinct: it lives in `langchain-anthropic` and concerns the outbound translation of standard `file` blocks to Anthropic `document` blocks.

```python
import base64
import os

assert os.environ.get("ANTHROPIC_API_KEY")

csv = "a,b,c\n1,2,3\n"
b64 = base64.b64encode(csv.encode()).decode()

model = ChatAnthropic(model="claude-sonnet-4-5")
model.invoke(
    [
        HumanMessage(
            content=[
                {"type": "text", "text": "summarize"},
                {
                    "type": "file",
                    "data": b64,
                    "mime_type": "text/csv",
                    "metadata": {"filename": "data.csv"},
                },
            ]
        )
    ]
)
```

```
anthropic.BadRequestError: Error code: 400 - {'type': 'error', 'error': {'type': 'invalid_request_error', 'message': "messages.0.content.1.document.source.base64.media_type: Input should be 'application/pdf'"}}
```

### Description

- I am sending a standard `type: "file"` content block with `mime_type: "text/csv"` and a base64 source in a `HumanMessage` to `ChatAnthropic`.
- I expect the request to succeed — the same content blocks work when routed through `langchain-aws` / Bedrock Converse, which accepts `csv, doc, docx, html, md, txt, xls, xlsx, pdf` as document source types.
- Instead, the Anthropic Messages API rejects the request because `ChatAnthropic` forwards `mime_type` verbatim into the `document.source.base64.media_type` field, and Anthropic's API only accepts `application/pdf` there.

The relevant code path is the `elif block["type"] == "file":` branch in [`libs/partners/anthropic/langchain_anthropic/chat_models.py`](https://github.com/langchain-ai/langchain/blob/master/libs/partners/anthropic/langchain_anthropic/chat_models.py) at approximately line 366:

```python
elif "base64" in block or block.get("source_type") == "base64":
    formatted_block = {
        "type": "document",
        "source": {
            "type": "base64",
            "media_type": block.get("mime_type") or "application/pdf",
            "data": block.get("base64") or block.get("data", ""),
        },
    }
```

`mime_type` is forwarded with no check against the Anthropic-accepted set.

Anthropic's Files API documentation is explicit that non-PDF text formats (`.csv`, `.txt`, `.md`, `.docx`, `.xlsx`) must be converted to plain text and inlined: https://platform.claude.com/docs/en/docs/build-with-claude/files — section "Working with other file formats".

**Suggested fix:** in the same branch, when the mime type is `text/*` and not `application/pdf`, decode the base64 to UTF-8 and emit a `text`-source document block (which Anthropic does accept) instead of `base64`:

```python
elif "base64" in block or block.get("source_type") == "base64":
    mime_type = block.get("mime_type") or "application/pdf"
    data = block.get("base64") or block.get("data", "")
    if mime_type.startswith("text/") and mime_type != "application/pdf":
        try:
            text = base64.b64decode(data).decode("utf-8")
            formatted_block = {
                "type": "document",
                "source": {"type": "text", "media_type": "text/plain", "data": text},
            }
        except (binascii.Error, UnicodeDecodeError):
            formatted_block = {
                "type": "document",
                "source": {"type": "base64", "media_type": mime_type, "data": data},
            }
    else:
        formatted_block = {
            "type": "document",
            "source": {"type": "base64", "media_type": mime_type, "data": data},
        }
```

PDF behavior is unchanged. Happy to submit a PR if this approach is acceptable.

### System Info

```
System Information
------------------
> OS:  Darwin
> OS Version:  Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:45 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T6030
> Python Version:  3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:25:12) [Clang 14.0.6 ]

Package Information
-------------------
> langchain_core: 1.2.17
> langchain: 1.2.15
> langchain_anthropic: 1.3.4
> langchain_aws: 1.0.0
> langchain_classic: 1.0.1
> langsmith: 0.7.6
> langgraph: 1.1.6
> langgraph-checkpoint: 4.0.0
```

imnishitha:
fixes https://github.com/langchain-ai/langchain/issues/37576

Reading a file that `read_file` can't attach to the current model (e.g. a `.docx`) no longer fails the thread with a non-retryable provider error. Instead, the unsupported attachment is replaced with a text placeholder before the request is sent.

Reported error: calling `read_file` on a `.docx` caused Anthropic to reject the request with a `media_type` validation error, leaving the agent unable to recover.

**Root cause:** `read_file` returns non-text files as multimodal content blocks. Anthropic only accepts inline `file` blocks for `application/pdf`; other document types (e.g. `.docx`, `.pptx`) are rejected server-side.

**Fix:** `FilesystemMiddleware.wrap_model_call` / `awrap_model_call` now scrub outgoing requests, replacing unsupported multimodal blocks with a text placeholder before they reach the provider.

- Images, audio, video, and PDFs are filtered using `ModelProfile` capabilities (`image_inputs`, `audio_inputs`, `video_inputs`, `pdf_inputs`, plus `*_tool_message` variants). Missing profile fields default to supported; only an explicit `False` strips a block.
- Non-PDF file types currently use a small provider allowlist (`openai`, `google_genai`). Other providers (including Anthropic) receive a placeholder instead of an unsupported attachment.

Note: This is a hot fix to resolve the error. Broader fix is to extend the model profiles.

### Before fix
image

### After fix
Screenshot 2026-08-03 at 10 21 05 AM

---

Related issues: https://github.com/langchain-ai/deepagents/issues/4397, https://github.com/langchain-ai/deepagents/issues/1313, https://github.com/langchain-ai/deepagents/issues/3864, https://github.com/langchain-ai/deepagents/issues/3660, https://github.com/langchain-ai/langchain/issues/36939, https://github.com/langchain-ai/langchain/issues/38872


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -112,6 +115,20 @@
 _VIDEO_SAMPLING_RATE: Final = 0.5
 """Seconds between sampled frames when extracting stills from a video."""
 
+_MULTIMODAL_BLOCK_TYPES: Final = frozenset({"image", "audio", "video", "file"})
```
ccurme: (nit) can we derive this from somewhere? (e.g., `_EXTENSION_TO_FILE_TYPE.values()` if logically related)

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -154,6 +171,135 @@ def _move_media_results_after_tool_results(messages: list[AnyMessage]) -> list[A
...
+    if callable(get_ls_params):
+        with contextlib.suppress(Exception):
+            ls_params = get_ls_params()
+            if isinstance(ls_params, dict) and isinstance(ls_params.get("ls_provider"), str):
+                return ls_params["ls_provider"].replace("-", "_").lower()
+    return _PROVIDER_BY_CLASS_NAME.get(type(model).__name__)
+
+
+def _multimodal_block_supported(
+    block: Mapping[str, Any],
```
ccurme: why is this not `ContentBlock`, given it looks like we're passing in `message.content_blocks` below? there's some defensive code here that I don't think we need.

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -154,6 +171,135 @@ def _move_media_results_after_tool_results(messages: list[AnyMessage]) -> list[A
...
+    mime_type = block.get("mime_type", "unknown")
+    path = message.additional_kwargs.get("read_file_path", "the requested file")
+    return cast(
+        "ContentBlock",
+        {
+            "type": "text",
+            "text": (
+                f"[read_file: {path} was not attached because this model does not support "
+                f"{block_type} content ({mime_type}). Ask the user to share the contents "
+                "directly, or use a model that supports this file type.]"
```
ccurme: Let's keep this minimal, e.g., drop the second sentence (there isn't always a user to ask).

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -154,6 +171,135 @@ def _move_media_results_after_tool_results(messages: list[AnyMessage]) -> list[A
...
+    A `model` with no `profile` (including `None` `model`, e.g. in tests) is
+    treated as an empty profile rather than skipped: `ModelProfile` is often
+    absent for models `langchain_anthropic` doesn't have a static entry for
+    (e.g. `ChatAnthropic(model="claude-3-5-sonnet-latest")`), and the
+    provider-based non-PDF `file` gate doesn't depend on profile data at all —
+    skipping the whole scrub in that case would silently leave the exact
+    `.docx`-on-Anthropic bug this fixes unfixed for those models. An empty
+    profile still defaults every per-field check to "supported."
+    """
+    profile = getattr(model, "profile", None)
```
ccurme: why getattr

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -154,6 +171,135 @@ def _move_media_results_after_tool_results(messages: list[AnyMessage]) -> list[A
...
 
 
+_PROVIDER_BY_CLASS_NAME: Final = {
+    "ChatOpenAI": "openai",
+    "AzureChatOpenAI": "openai",
+    "ChatAnthropic": "anthropic",
+    "ChatGoogleGenerativeAI": "google_genai",
+    "ChatVertexAI": "google_genai",
+}
+"""Fallback provider lookup for models whose `_get_ls_params` isn't usable (e.g. in tests)."""
```
ccurme: shouldn't need this, can we use `model._llm_type` and delete this mapping + `_infer_model_provider`?

libs/deepagents/tests/unit_tests/middleware/test_filesystem_multimodal_profile.py
```diff
@@ -0,0 +1,304 @@
+"""Tests for scrubbing multimodal content blocks unsupported by `model.profile`.
```
ccurme: We almost always need end-to-end tests. This lets us define behavior without worrying about internals. If we refactor the internal functions, these tests can completely change and it will be hard to know if any user-facing behavior is changing.

libs/deepagents/deepagents/middleware/filesystem.py
imnishitha: because `model` isn't always a real `BaseChatModel` like model=_FakeModel() in unit tests.

imnishitha: modified the tests to use a BaseChatModel

libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -154,6 +177,114 @@ def _move_media_results_after_tool_results(messages: list[AnyMessage]) -> list[A
...
+
+
+def _model_tolerates_non_pdf_files(model: "BaseChatModel | None") -> bool:
+    """Whether `model` is known to accept non-PDF `file` blocks.
+
+    Checks `_llm_type`, a property every `BaseChatModel` implements
+    """
+    if model is None:
+        return False
+    return any(marker in model._llm_type for marker in _NON_PDF_FILE_TOLERANT_LLM_TYPE_MARKERS)
```
ccurme: why check for substrings here vs. exact match? I'd rather point at specific chat models than make inferences about support from substrings.

example: [this](https://github.com/langchain-ai/langchain-aws/blob/e27684c58342aa28a7642b8babe5b23035a1a977/libs/aws/langchain_aws/chat_models/openai.py#L157-L160) chat model (born yesterday) has `openai` in `_llm_type` because it inherits ChatOpenAI, but it serves a bunch of different models.

libs/deepagents/tests/unit_tests/middleware/test_filesystem_multimodal_profile.py
```diff
@@ -0,0 +1,196 @@
+"""End-to-end tests for scrubbing multimodal content blocks unsupported by `model.profile`.
```
ccurme: Can we follow the patterns in https://github.com/langchain-ai/deepagents/blob/main/libs/deepagents/tests/unit_tests/test_end_to_end.py exercising the whole flow through `create_deep_agent`?

There is a fake chat model there you can use.

libs/deepagents/deepagents/middleware/filesystem.py
imnishitha: thanks for pointing this out

ccurme (approved):

PR #5195, opened 2026-07-30
feat(code): add cache and context status row

aolsenjazz:
The status footer now uses a second row for cache reads/writes, context percentage and token totals, and cumulative USD cost.

---

Keeping workspace/model state on the first row prevents live usage metrics from competing with cwd and branch space. Cache metadata is validated through the pricing path and scoped to the active thread.

Verified with focused status, session-stat, cost-tracking, adapter, and thread-reset tests plus package lint/type checks.

Made by [Open SWE](https://openswe.vercel.app/agents/7a831dda-3c65-a0f6-35ce-62af0c34f84f)


aolsenjazz:
closing; opened for a demo

PR #5198, opened 2026-07-30
feat(code): load hooks from installed plugins

johannes117:
Enabled plugins can now contribute Hooks v2 handlers through `hooks/hooks.json`, manifest paths, or inline configuration. Their hooks receive plugin path variables, appear in the plugin manager, and activate on startup or `/reload`.

---

Plugin hook sources retain provenance for deterministic snapshots and per-plugin environments. Invalid plugin documents are isolated and surfaced as diagnostics without disabling user, project, or other plugin hooks.

## Review Guide
1. Plugin declarations:
- `manifest.py`
- `hooks.py`

2. Hook loading and execution:
- `loading.py`
- `snapshot.py`
- `engine.py`

3. Reload behavior:
- `app.py`


libs/code/deepagents_code/app.py
aolsenjazz, line 1: let's run hooks discovery sync. I'm not convinced that we should add a lock to cover startup just for one async execution - let's stick to the convention

aolsenjazz: This isn't a unit of work for right now - but whenever we have some time to refactor - we should extract all required-but-async-capable and run those in parallel, awaiting on the pool

libs/code/deepagents_code/tui/modals/plugin_manager/content.py
aolsenjazz, line 1: nice!

libs/code/deepagents_code/hooks/loading.py
```diff
@@ -37,10 +40,54 @@
...
 
+if TYPE_CHECKING:
+    from deepagents_code.json_types import JsonValue
+
 logger = logging.getLogger(__name__)
 _LEGACY_HOOKS_REMOVAL_DATE = "September 1, 2026"
 
 
+@dataclass(frozen=True, slots=True)
+class HooksSource:
```
aolsenjazz: nit but also not a big deal: I'd love to see this expressed as a union of FileHooksSource/PluginHooksSource or something, and just pass around the base class

aolsenjazz (approved):
stamping, we should at least fix the sync vs async init thing

PR #5210, opened 2026-07-30
fix(sdk): only explain truncation markers when preview omits lines

mdrxy:
Large offloaded results now tell the model exactly what a preview omitted instead of always describing a head-and-tail truncation marker.

---

When the agent saves a large tool result, message, or `execute` output to the filesystem, it puts a stub in the context. The stub has a sentence about the preview, then the preview itself. A preview can lose content in several ways, but the stub showed this one sentence each time:

> Here is a preview showing the head and tail of the result (lines of the form `... [N lines truncated] ...` indicate omitted lines in the middle of the content):

That sentence is correct only when the preview dropped whole lines from the middle. The stub now has a `preview_note` placeholder that names only the losses that occurred. One of three notes appears:

```txt
Here is a preview of the result:
Here is a preview showing the head and tail of the result (lines of the form `... [N lines truncated] ...` indicate omitted lines in the middle of the content):
Here is a preview of the result (the output contains lines longer than 1000 characters; this preview shows only their first 1000 characters):
```

The first appears when the preview lost nothing. The second appears when the preview dropped whole lines from the middle. The third appears when the preview clipped long lines. When both losses occur, the note shows both caveats. For an evicted `HumanMessage`, the note says `content` instead of `result`.

Every stub goes through a single renderer that derives the note from the preview it is rendering, so a note cannot describe losses some other preview had.

These cases gave no warning before:

- **Clipped lines.** `_create_content_preview` clips each line at 1000 characters. A 5000-character line (a `.jsonl` dump, a minified bundle) made a 1003-character preview. The preview promised a marker it did not insert. It did not mention the approximately 4000 lost characters. Now the stub says:

  ```txt
  Here is a preview of the result (the output contains lines longer than 1000 characters; this preview shows only their first 1000 characters):
  ```

- **The end of `execute` output.** The capture wrapper makes a head-and-tail preview only when the output has more lines than the head-plus-tail budget. If not, it returned a bare 4000-byte prefix. The prefix cut off in the middle of a line. It had no marker and no `truncated` flag. Now the excerpt ends with an in-band notice. The wording is careful about the capture cap: when the output hit `max_capture_bytes`, the notice does not report the cap as the real output size, and it does not say the saved file is complete:

  ```txt
  ... [output clipped here; 120015 bytes total, full output at the path above] ...
  ... [output clipped here; capture stopped at its 5000-byte limit, so the file at the path above is also incomplete] ...
  ```

- **Byte-capped head and tail excerpts.** The wrapper's `head -c`/`tail -c` caps bite before its line caps, so output with very long lines (JSONL logs, minified JS, base64 blobs) shows fewer lines than the budget and cuts the outermost ones mid-line. The truncation marker counts only whole lines dropped from the middle, so nothing mentioned that cut. The wrapper now says so where it happens:

  ```txt
  ... [the excerpts above and below are byte-capped, so the lines bordering this marker may be cut mid-line] ...
  ```

Whether a preview carries a truncation marker is reported by the code that built it, never by scanning the rendered text — command output can contain a line that looks exactly like a marker.

`TOO_LARGE_TOOL_MSG` and `TOO_LARGE_HUMAN_MSG` imports from `middleware.filesystem` still work and keep their released wording. Each import emits `LangChainDeprecationWarning`. Both are deprecated in `deepagents==0.7.7` and will be removed in `deepagents==0.9.0`.

Made by [Open SWE](https://openswe.vercel.app/agents/b1d3d30f-a21e-d9f1-a834-2091339a02a5)


mdrxy:
#5563

PR #5224, opened 2026-07-31
fix(code): skip manual HITL for the `/offload` seeded compaction

mdrxy:
`/offload` (and `/compact`) no longer pops a spurious approval request for the compaction it was just asked to perform.

---

`/offload` compacts server-side by injecting a synthetic assistant message that already carries a forced `compact_conversation` call. Because no model produced that message, the Auto-mode HITL middleware never saw a decision plan for it, treated it as corrupt state, and raised a HITL interrupt. The `/offload` driver caught and self-approved that interrupt, so the command worked — but via a fragile catch-and-resume that left an errored middleware run in every trace.

The fix recognizes the seeded call using trust signals a model cannot forge (client-set run context, the deterministic offload seed ID, exact `force=True` args) and lets it through. If any signal is missing, the batch falls through to the existing manual-review path, which stays intact. The real-model path is unchanged.

<details>
<summary>Test plan</summary>

- New unit tests cover the pass-through plus every way a call fails the trust check.
- The server-side `/offload` integration test asserts no interrupts surface; it fails on `main` and passes with this change.

</details>

Made by [Open SWE](https://openswe.vercel.app/agents/2d273875-739a-ed79-b7d3-fe3e76942fe7)


mdrxy:
Closing in favor of an explicit server-side forced-compaction operation for /offload. That design will retain this PR’s regression coverage while removing the synthetic assistant tool call and its HITL exception.

PR #5229, opened 2026-07-31
fix(sdk): resolve exact-file `delete` targets with first-match-wins o…

### LINKED ISSUE #5113: delete ignores first-match-wins permission ordering for workspace allowlists
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
import tempfile
from pathlib import Path
from types import SimpleNamespace

from deepagents.backends import FilesystemBackend
from deepagents.middleware.filesystem import FilesystemMiddleware, FilesystemPermission

with tempfile.TemporaryDirectory() as tmp:
    root = Path(tmp)
    (root / "workspace").mkdir()

    permissions = [
        FilesystemPermission(
            operations=["read", "write"],
            paths=["/workspace/**"],
            mode="allow",
        ),
        FilesystemPermission(
            operations=["read", "write"],
            paths=["/**"],
            mode="deny",
        ),
    ]

    # This mirrors the FilesystemMiddleware configured internally by
    # create_deep_agent(..., permissions=permissions).
    middleware = FilesystemMiddleware(
        backend=backend,
        tools=["read_file", "write_file", "delete"],
        _permissions=permissions,
    )
    tools = {tool.name: tool for tool in middleware.tools}
    runtime = SimpleNamespace(tool_call_id="repro")

    write_result = tools["write_file"].func(
        file_path="/workspace/allowed.txt",
        content="ok",
        runtime=runtime,
    )
    delete_result = tools["delete"].func(
        file_path="/workspace/allowed.txt",
        runtime=runtime,
    )

    print(write_result.status, write_result.content)
    print(delete_result.status, delete_result.content)
    print((root / "workspace" / "allowed.txt").exists())
```

```shell
success Updated file /workspace/allowed.txt
error Error: permission denied for write on /workspace/allowed.txt (matches deny rule(s): /**)
True
```

### Description

The permissions documentation defines declaration-order, first-match-wins semantics and uses this exact pattern to isolate an agent to a workspace:

1. allow read/write on `/workspace/**`
2. deny read/write on `/**`

Under those rules, `write_file` and `edit_file` correctly use the first matching allow rule. However, `delete` rejects even an exact file under `/workspace` because it calls `_find_delete_deny_patterns`, which scans every write-deny rule and treats the trailing catch-all `/**` as effective even after an earlier allow matched.

This makes `delete` unusable with the documented workspace isolation configuration.

Expected behavior:

- deleting `/workspace/allowed.txt` succeeds because the first matching write rule allows it;
- recursively deleting a directory still fails if any target or descendant is effectively denied after applying rule order.

The recursive all-or-nothing safety goal is valid. The inconsistency is that the current overlap check ignores effective first-match rule ordering. A fix could evaluate effective permissions for the target and affected descendants, or the public contract/documentation should explicitly define different ordering semantics for `delete` and provide a supported workspace-isolation configuration.

I reproduced this on the latest released version and on current `main` at commit `f1727c785c44809e7b1b7ed8c5d0b31c560ae3ad`.

Documentation: https://docs.langchain.com/oss/python/deepagents/permissions

### Environment / System Info

OS: macOS 26.5 (host); Debian 12 container
Python: 3.12.13
deepagents: 0.7.0b2
langchain: 1.3.14
langchain-core: 1.5.0
Also reproduced against deepagents main commit f1727c785c44809e7b1b7ed8c5d0b31c560ae3ad

imnishitha:
Closes #5113

`delete` now uses the same first-match-wins permission check as `write_file` and `edit_file` when deleting a confirmed leaf target: a plain file, including cases where an exact-match `ls` looks like an empty directory on flat/virtual backends until checked against the parent listing. This makes the documented workspace isolation pattern (`allow /workspace/**`, then `deny /**`) work consistently for deletes.

For targets that may have descendants (directories, including empty ones, or uncertain paths), the existing conservative behavior is unchanged: any denied descendant still blocks the recursive delete. To distinguish the two cases, `delete` uses the backend's existing `ls` to detect whether the target has nested entries.


libs/deepagents/deepagents/middleware/filesystem.py
```diff
@@ -342,15 +342,74 @@ def _wildcard_delete_overlap(pattern: str, anchor: str, target: str) -> bool:
...
+    instead of `ls(target)`, since only the former distinguishes files from
+    directories consistently across all backends.
+    """
+    if not permissions_configured:
+        return False
+    try:
+        ls_result = backend.ls(str(PurePosixPath(target).parent))
+    except NotImplementedError:
+        return True
+    if ls_result.error is not None:
```
nick-hollon-lc, line 508: `StateBackend.ls()` and `StoreBackend.ls()` do not return `not_a_directory` for file paths; they return a successful empty `LsResult` instead. Consequently, an exact stored/state-backed file is classified as `has_descendants=True`, so it still takes the conservative deny-overlap path and remains blocked by `allow /work/**` followed by `deny /**`.

The added regression tests use `FilesystemBackend`, whose `ls()` does emit `not_a_directory`, so they do not expose this. Could we either make leaf detection work with the state/store `ls()` semantics or add coverage for those backends? This needs to include sync and async paths.

nick-hollon-lc (changes requested):
i think one issue with state and store bakend could you tal?

libs/deepagents/deepagents/middleware/filesystem.py
imnishitha: Fixed: `ls(target)` is now the primary signal; only the ambiguous `entries=[], error=None` case falls back to checking `FileInfo.is_dir` from the parent listing, which works consistently across all backends.

nick-hollon-lc (approved):
lgtm now!

PR #5232, opened 2026-07-31
fix(code): return Ctrl+C status from CLI startup

johannes117:
Ctrl+C at the project-hooks trust prompt printed Interrupted but exited 0 instead of conventional status 130. Return the explicit status from `cli_main` and propagate it through the module entrypoint so both launch paths preserve the interruption status.

Verified with the targeted startup tests and the hooks/main test suites (402 tests).

### DIFF
--- a/libs/code/deepagents_code/__init__.py
+++ b/libs/code/deepagents_code/__init__.py
@@ -21,7 +21,7 @@ __all__ = [
 ]
 
 
-def __getattr__(name: str) -> Callable[[], None]:
+def __getattr__(name: str) -> Callable[[], int | None]:
     """Lazy import for `cli_main` to avoid loading `main.py` at package import.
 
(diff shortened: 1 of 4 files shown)


johannes117:
Closing - the bug does not reproduce.

Verified against the real trust prompt in an untrusted scratch project (the deepagents repo itself is already in the hooks trust store, so the prompt never renders there - that is why earlier local checks were inconclusive). On main, Ctrl+C at the project-hooks trust prompt exits 130 via both the dcode console script and python -m deepagents_code.main.

The console script wrapper is sys.exit(cli_main()), so the existing sys.exit(130) already propagates. Converting it to return 130 kept the console script at 130 but made python -m deepagents_code.main exit 0, because that module guard discards the return value - so the change introduced the failure it was meant to fix, on a path test_version.py exercises.

PR #5238, opened 2026-08-01
fix(code): collapse finished dynamic subagents

aolsenjazz:
Dynamic subagent panels now collapse to a compact status header when the final task ends, hide when the next turn begins, and automatically reopen for a later workflow.

---

The expanded panel can consume a large part of the terminal after its work is complete. Completion, failure, and cancellation totals remain visible in the header, while a manual collapse stays respected during the current workflow.

The collapsed summary is only relevant while the user can still act on it, so the panel clears eagerly at the next turn boundary (`prepare_turn`) instead of lingering until another workflow starts. This replaces the deferred `_pending_reset` clear, which no longer has a purpose once the panel collapses itself.

Made by [Open SWE](https://openswe.vercel.app/agents/f0ce2762-8448-3d73-edff-e3dd83696f6f)

### DIFF
--- a/libs/code/deepagents_code/app.py
+++ b/libs/code/deepagents_code/app.py
@@ -15755,10 +15755,9 @@ class DeepAgentsApp(App):
             self._inflight_turn_start = time.monotonic()
             self._inflight_thread_id = self._lc_thread_id
 
-            # Arm the subagent fan-out panel for this turn, seeding the session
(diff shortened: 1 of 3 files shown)


mdrxy:
Height should also match the # of subagents

Screenshot 1

aolsenjazz:
no action taking on this. closing

PR #5260, opened 2026-08-03
fix(code): make reject-with-feedback discoverable in approval menu

mdrxy:
Makes the free-text rejection path discoverable, and stops the quick keys resolving an approval while the reason field is open.

---

**Discoverability.** The `Tab` hint only rendered while Reject was highlighted, and `Tab` was a no-op from every other row — so users deciding with `y`/`n` from the default Approve row never saw it. The hint is now unconditional and reads `Tab reject with feedback` (`amend` read like `git commit --amend`), and `Tab` moves the cursor to Reject before opening the field. Adds a splash tip, and mirrors `tab` at App level like the other approval keys — with `priority=True`, since `Screen` already binds `tab -> focus_next` and would otherwise win. `check_action` steps it aside outside an approval so focus traversal is unaffected.

**Quick keys.** Arrow keys were gated on `_reason_input_active`; `y`/`a`/`1`/`2`/`3` and Enter were not. Clicking the menu body moves focus off the reason `Input` without closing it, and there `y` approved the call being rejected while Enter sent a bare reject that dropped the typed text. Now `on_focus` hands focus back (with `select_on_focus=False`, or the next keystroke would replace the reason), `_handle_selection` refuses to decide while the field is open, and Enter submits what was typed. `Esc` still cancels without deciding.

Made by [Open SWE](https://openswe.vercel.app/agents/068a2454-0b8d-ef97-96b3-b4d9cf28cd7e)

## References
- Plan: https://openswe.vercel.app/agents/068a2454-0b8d-ef97-96b3-b4d9cf28cd7e/plan

### DIFF
--- a/libs/code/deepagents_code/tui/widgets/startup_tip.py
+++ b/libs/code/deepagents_code/tui/widgets/startup_tip.py
@@ -29,7 +29,6 @@ _TIPS: dict[str, int] = {
     "Use /model to switch models mid-conversation": 2,
     "Use /effort to change the current model's reasoning effort": 1,
     "Press ctrl+x to compose prompts in your external editor": 1,
-    "Press Tab at an approval prompt to reject with feedback": 2,
     "Use /skill:<name> to invoke a skill directly": 1,
     "Use /theme to customize the TUI's colors": 1,
     "Use /skill-creator to build reusable agent skills": 1,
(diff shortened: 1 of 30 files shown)


libs/code/deepagents_code/tui/widgets/startup_tip.py
```diff
@@ -29,6 +29,7 @@
     "Use /model to switch models mid-conversation": 2,
     "Use /effort to change the current model's reasoning effort": 1,
     "Press ctrl+x to compose prompts in your external editor": 1,
+    "Press Tab at an approval prompt to reject with feedback": 2,
```
mdrxy:
```suggestion
```

PR #5262, opened 2026-08-03
feat(code): refresh price catalog hourly in the background

mdrxy:
Cost estimates can now pick up new model prices without waiting for a dcode release. Once cost estimates are first used, dcode checks for updated pricing data in the background every hour.

To use only the pricing data bundled with dcode, set `DEEPAGENTS_CODE_PRICES_AUTO_UPDATE=0` or `[update].prices_auto_update = false` in `config.toml`. Offline mode also disables these downloads.

---

New model prices can be unavailable until a dcode release updates its bundled pricing data. This change refreshes that data in the background after pricing is first needed, so it does not add work at startup.

If an update fails or appears incomplete, dcode keeps using the pricing data it already has.


mdrxy:
Superseded by #5264 (branch renamed to match repo naming convention; same commits).

PR #5348, opened 2026-08-06
fix(sdk): disambiguate `read_file` gutter separator from source inden…

### LINKED ISSUE #5343: `read_file` two-space gutter separator is ambiguous with space indentation, breaking `edit_file` round-trips on space-indented code
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
Source file (spaces, 4-wide indent):


function foo() {
    return Promise.reject(errorObj);
}


`read_file` output (gutter = right-justified number + 2 spaces):


 1  function foo() {
 2      return Promise.reject(errorObj);
 3  }


Focus on line 2: ` 2` + `<2 sep spaces>` + `<4 indent spaces>` + `return ...;`.

What the LLM produces for `edit_file.old_string` (observed repeatedly):


          return Promise.reject(errorObj);


i.e. **6 leading spaces** = `2 (separator) + 4 (indent)`, vs. the true source
**4 leading spaces**. `old_string` does not match the file; the edit silently
no-ops or fails. The error is **deterministic given the model's parsing** and
scales with indentation depth (deeper nesting -> larger / more confusing gap).
```

```shell

```

### Description

`format_content_with_line_numbers` separates the line-number gutter from source
content with **two spaces**. For languages that use **spaces** for indentation,
this makes the gutter separator visually indistinguishable from source
indentation. Some LLMs cannot reliably tell the two apart and, when constructing
`edit_file`'s `old_string` from `read_file` output, fold the 2-space separator
into the indentation, causing systematic match failures.

This breaks the `read_file` -> `edit_file` round-trip that deepagents itself
defines and relies on.

The current implementation (0.7.4):

```python
return "\n".join(f"{marker:>{marker_width}}  {line}" for marker, line in rows)
```

The docstring explicitly explains the choice of a space-based separator:

> Line markers are separated from source content with two spaces so **source
> tabs cannot be confused with a gutter separator**.

This closes half of the ambiguity (TAB-indented code vs. a TAB gutter) but opens
the symmetric half: **space-indented code vs. a space gutter**. When source uses
spaces for indentation (the default in JS/TS/Python/Go formatting, Prettier,
gofmt-compatible editors, etc.), there is no lexical boundary between "gutter
spaces" and "indentation spaces", and weaker/medium LLMs have no reliable way to
reconstruct the exact original indentation from the rendered line.

### Environment / System Info

- deepagents: `0.7.4`
- Relevant code: `deepagents/backends/utils.py::format_content_with_line_numbers`
- Python: 3.11
- OS: Linux (also reproduced on Windows)
- Affected models (space-indented code): glm-5.2

NinaadRao:
Closes #5343

`read_file`'s line-number gutter now separates the marker from source content with ` | ` instead of two spaces, so the gutter can never be confused with space-indented source. Weaker models were folding the old two-space separator into leading indentation when constructing `edit_file`'s `old_string`, producing a systematic mismatch (deeper indentation made the gap larger and easier to misread) that silently no-op'd or failed the edit.

---

The two-space separator was chosen specifically to disambiguate from tab-indented source (`#4561`), but that left the mirror-image problem: no lexical boundary exists between "two gutter spaces" and "the first two columns of 2/4/8-space-indented source," which is the default indentation style for JS/TS/Python/Go formatters. A single non-whitespace separator removes the ambiguity for both tab- and space-indented source, matching the convention tools like `bat`/`less -N` already use.

This is a producer/consumer contract spanning two packages, so both sides moved together:

- `format_content_with_line_numbers` (the producer, in `deepagents`) now emits ` | `.
- `_truncate_paginated_read` (same package) parses the new separator when computing truncation boundaries.
- `ReadFileContinuationNoticeMiddleware._is_numbered_read_file_row` (the nemotron harness profile) and `ToolCallMessage._compact_line_gutter` (the `deepagents-code` TUI) both now match the current ` | ` separator, while continuing to tolerate the two legacy forms (the two-space gutter, and the original `cat -n` tab) so cached or persisted transcripts written by an older `deepagents` version still parse and display correctly. Only newly-produced output uses the new separator.

Since this changes a wire format shared across packages with independent release cadences, the SDK and CLI changes are bundled in one branch here for review, but may need to land as two separate PRs (SDK first, then `deepagents-code` once its `deepagents` pin is bumped) to satisfy this repo's one-release-PR-per-package convention — happy to split however maintainers prefer.

<details>
<summary>Test plan</summary>

- Existing `format_content_with_line_numbers`, `_is_numbered_read_file_row`, and `_compact_line_gutter` unit tests updated to the new separator.
- New coverage: a space-indented source round-trip demonstrating the gutter is now unambiguous, and legacy-format tolerance tests (two-space and `cat -n` tab) for both consumers.
- `make lint` and `make test` clean in `libs/deepagents` (2453 passed) and targeted runs in `libs/code` (436 passed in the touched TUI module, plus the one other file referencing the format).

</details>

### DIFF
--- a/libs/code/deepagents_code/tui/widgets/messages.py
+++ b/libs/code/deepagents_code/tui/widgets/messages.py
@@ -206,16 +206,18 @@ _SUCCESS_EXIT_RE = re.compile(r"\n?\[Command succeeded with exit code 0\]\s*$")
 """Strip the SDK's `[Command succeeded with exit code 0]` trailer from tool output."""
 
 
-_READ_FILE_GUTTER_RE = re.compile(r"^ *(\d+(?:\.\d+)?)(?:  |\t)(.*)$")
+_READ_FILE_GUTTER_RE = re.compile(r"^ *(\d+(?:\.\d+)?)(?: \| |  |\t)(.*)$")
 """Match a `read_file` gutter row into (marker, source).
 
 The marker is a bare `N` or `N.M` (the latter a wrapped-line continuation) —
 both sides of the dot required, so a stray `.5` head is not a gutter. The
-separator is exactly two spaces (current format) or a single tab (legacy
-`cat -n`). Only the separator is consumed and leading padding is spaces-only, so
-source indentation — including leading tabs — after the gutter stays put. Kept in
-sync with the separator emitted by deepagents' `format_content_with_line_numbers`
-(the authoritative producer). See `ToolCallMessage._compact_line_gutter`.
+separator is ` | ` (current format), or a legacy separator — two spaces, or a
+single tab (`cat -n`) — from output persisted by an older deepagents version.
+Only the separator is consumed and leading padding is spaces-only, so source
+indentation — including leading tabs and spaces — after the gutter stays put.
+Kept in sync with the separator emitted by deepagents'
+`format_content_with_line_numbers` (the authoritative producer). See
+`ToolCallMessage._compact_line_gutter`.
 """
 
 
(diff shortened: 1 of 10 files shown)


aolsenjazz:
Hey @NinaadRao, I see this issue is closed but wanted to let you know that this is something we're actively solving internally. We see the same issue with typescript/go files.

I also wanted to ask - since you seem to be enthusiastic dcode user, I'd love to chat. Would love to hop on an email thread to learn about what you like/don't like/want to see from dcode in the future. my email is alex.olsen@langchain.dev if you don't mind reaching out!

PR #5405, opened 2026-08-10
feat(code): warn above configurable session cost threshold

johannes117:
Deep Agents Code now warns once when a thread's estimated cost exceeds a configurable soft limit, defaulting to $50, and suggests `/offload` or `/clear`.

---

Cost warnings use the graph's authoritative cumulative total rather than provisional estimates, avoiding alerts on spend that may later be corrected downward. Configure `[warnings].session_cost_threshold_usd`; non-positive values disable the warning.

## Screenshots
Screenshot 2026-08-10 at 12 03 15 PM


aolsenjazz (approved):
nice

PR #5422, opened 2026-08-11
fix(code): preserve rubric grader tool runtime

mdrxy:
Rubric graders can use read-only repository tools during `/goal` acceptance checks again.

---

The grader's `read_file`, `ls`, `glob`, and `grep` wrappers reuse the SDK filesystem tools' `args_schema`, so LangChain identifies each wrapper's injected `ToolRuntime` from the wrapper's own annotations. Those annotations were strings under postponed evaluation, so the runtime was not recognized as injected and was dropped during input validation — every grader repository tool call then failed before execution, ending the acceptance check with a grader error.

Rather than augmenting the reused schemas, the grader's tool factory now lives in its own module that keeps annotations concrete. That confines the constraint to the tools that need it, leaves `agent.py` free to rely on postponed evaluation for its `TYPE_CHECKING`-only imports, and keeps the SDK's per-argument descriptions in the model-facing schema. The move is otherwise byte-for-byte, and the module is only imported alongside `agent.py`, so startup fast paths such as `dcode -v` are unaffected.

The underlying `StructuredTool` behavior is a latent issue in `langchain-core`; the general fix is proposed separately in langchain-ai/langchain#39602. This change makes the grader correct regardless of when that lands.

A regression test drives all four wrappers through a real `ToolNode`, and a second test pins the no-postponed-annotations requirement for the new module so the failure mode cannot silently return.

Made by [Open SWE](https://openswe.vercel.app/agents/ebca0faf-2546-8060-544b-d5d970d027a4)

### DIFF
--- a/libs/code/deepagents_code/agent.py
+++ b/libs/code/deepagents_code/agent.py
@@ -55,6 +55,7 @@ from langchain.tools import (
     ToolRuntime,  # LangChain inspects this annotation for runtime injection.
 )
 from langchain_core.tools import StructuredTool, tool
+from pydantic import BaseModel, ConfigDict, create_model
 
 from deepagents_code import theme
 from deepagents_code._cli_context import CLIContextSchema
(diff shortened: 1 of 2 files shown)


mdrxy:
Closing in favor of an upstream `langchain-core` fix for postponed `ToolRuntime` annotations with custom `args_schema` values.

Once released, we’ll require the fixed version and retain integration coverage without carrying the schema-level workaround here.

PR #5430, opened 2026-08-11
fix(sdk): patch invalid partial tool calls

### LINKED ISSUE #4662: `invalid_tool_calls` end the run silently and deferred repair is marked successful
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli
- [ ] code
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
from collections.abc import Sequence
from typing import Any

from langchain_core.language_models import BaseChatModel
from langchain_core.messages import AIMessage, BaseMessage, HumanMessage
from langchain_core.outputs import ChatGeneration, ChatResult
from langchain_core.tools import BaseTool, tool


class StubModel(BaseChatModel):
    responses: list[AIMessage]
    calls: int = 0

    @property
    def _llm_type(self) -> str:
        return "stub"

    def bind_tools(
        self, tools: Sequence[dict[str, Any] | type | BaseTool], **kwargs: Any
    ) -> "StubModel":
        return self

    def _generate(self, messages: list[BaseMessage], **kwargs: Any) -> ChatResult:
        response = self.responses[self.calls]
        self.calls += 1
        return ChatResult(generations=[ChatGeneration(message=response)])


@tool
def echo(value: str) -> str:
    """Echo a value."""
    return value


malformed = AIMessage(
    content="",
    additional_kwargs={
        "tool_calls": [
            {
                "id": "call_malformed",
                "type": "function",
                "function": {"name": "echo", "arguments": '{"value":'},
            }
        ]
    },
)
agent = create_deep_agent(
    model=StubModel(responses=[malformed, AIMessage(content="Recovered")]),
    tools=[echo],
    subagents=[],
)

first = agent.invoke({"messages": [HumanMessage(content="Call echo.")]})
print(first["messages"][-1].content)  # ""
print(first["messages"][-1].invalid_tool_calls)  # contains call_malformed

second = agent.invoke(
    {"messages": [*first["messages"], HumanMessage(content="What happened?")]}
)
for message in second["messages"]:
    print(message.type, getattr(message, "status", None), message.content)
```

## Actual behavior

1. The first run returns an `AIMessage` with `content == ""` and a populated
   `invalid_tool_calls` list.
2. Because there are no valid `tool_calls`, the graph ends rather than
   continuing to the model with corrective feedback.
3. The next invocation's `PatchToolCallsMiddleware.before_agent` appends:

   
   Tool call echo with id call_malformed could not be executed - arguments were malformed or truncated.
   

4. The appended `ToolMessage.status` is `"success"` (the default), not
   `"error"`.

## Expected behavior

- An `invalid_tool_call` with an ID should be treated as an actionable failed
  tool attempt in the same agent run.
- Deepagents should append an error `ToolMessage` and route back to the model,
  allowing it to retry with valid arguments or explain the failure to the user.
- The synthetic message must use `status="error"`.
- If an in-run retry cannot be supported, the current turn should at least
  return a visible error rather than an empty assistant response.

### Description

When a model returns malformed tool arguments, LangChain places the call in
`AIMessage.invalid_tool_calls` instead of `AIMessage.tool_calls`. A deepagents
run then routes to `END`, producing an empty assistant turn with no explanation
or in-run retry.

`PatchToolCallsMiddleware` repairs the dangling invalid call only at the start
of a later invocation. Its synthetic `ToolMessage` is constructed without an
explicit `status`, so LangChain reports it as `status="success"` even though it
states that the call could not be executed.

This means a user must send another message (for example, “What happened?”)
before the agent sees the corrective feedback and can recover.

### Environment / System Info

- `deepagents==0.6.12`
- `langchain==1.3.12`
- `langchain-core==1.4.9`
- `langgraph==1.2.8`
- Python 3.13

hntrl:
Fixes #4662

Marks deferred synthetic tool-call repairs as errors without retrying malformed calls in the same invocation. The original `AIMessage` remains the final message for that invocation.

Made by [Open SWE](https://openswe.vercel.app/agents/b1e599d9-4ad2-5932-a0ed-c6da6d0fd5c6) · openai:gpt-5.6-sol (high)

### DIFF
--- a/libs/deepagents/deepagents/middleware/patch_tool_calls.py
+++ b/libs/deepagents/deepagents/middleware/patch_tool_calls.py
@@ -42,7 +42,6 @@ class PatchToolCallsMiddleware(AgentMiddleware):
 
         answered_ids = {msg.tool_call_id for msg in messages if msg.type == "tool"}
         patched_messages = []
-        patched_any = False
         for msg in messages:
             patched_messages.append(msg)
             if not isinstance(msg, AIMessage):
@@ -51,9 +50,8 @@ class PatchToolCallsMiddleware(AgentMiddleware):
                 error_message = self._error_tool_message(tool_call)
                 if error_message is not None and error_message.tool_call_id not in answered_ids:
                     patched_messages.append(error_message)
-                    patched_any = True
 
-        if not patched_any:
+        if len(patched_messages) == len(messages):
             return None
         return {"messages": [RemoveMessage(id=REMOVE_ALL_MESSAGES), *patched_messages]}
 
--- a/libs/deepagents/deepagents/middleware/patch_tool_calls.py
+++ b/libs/deepagents/deepagents/middleware/patch_tool_calls.py
@@ -1,49 +1,86 @@
 """Middleware to patch dangling tool calls in the messages history."""
 
+from collections.abc import Mapping
 from typing import Any
 
-from langchain.agents.middleware import AgentMiddleware, AgentState, TracePolicy, omit_payload
-from langchain_core.messages import AIMessage, AnyMessage, RemoveMessage, ToolMessage
+from langchain.agents.middleware import (
+    AgentMiddleware,
+    AgentState,
+    TracePolicy,
+    hook_config,
+    omit_payload,
+)
+from langchain_core.messages import AIMessage, RemoveMessage, ToolMessage
 from langgraph.graph.message import REMOVE_ALL_MESSAGES
 from langgraph.runtime import Runtime
 
 
 class PatchToolCallsMiddleware(AgentMiddleware):
-    """Middleware to patch dangling tool calls in the messages history."""
+    """Middleware to repair incomplete tool calls in the messages history."""
+
+    @staticmethod
+    def _error_tool_message(tool_call: Mapping[str, Any]) -> ToolMessage | None:
+        """Create an error result for an identified tool call."""
+        tool_call_id = tool_call["id"]
+        if tool_call_id is None:
+            return None
+        name = tool_call["name"] or "unknown"
+        if tool_call.get("type") == "invalid_tool_call":
+            content = f"Tool call {name} with id {tool_call_id} could not be executed - arguments were malformed or truncated."
+        else:
+            content = f"Tool call {name} with id {tool_call_id} was cancelled - another message came in before it could be completed."
+        return ToolMessage(content=content, name=name, tool_call_id=tool_call_id, status="error")
 
     trace_policy = TracePolicy(process_inputs=omit_payload)
     """Omit hook inputs from traces by default; set a `TracePolicy` to override."""
 
     def before_agent(self, state: AgentState, runtime: Runtime[Any]) -> dict[str, Any] | None:  # noqa: ARG002
-        """Before the agent runs, handle dangling tool calls from any AIMessage."""
+        """Before the agent runs, repair dangling tool calls from prior turns."""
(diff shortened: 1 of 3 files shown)


mdrxy:
!keep-open

libs/deepagents/deepagents/middleware/patch_tool_calls.py
```diff
@@ -1,49 +1,86 @@
...
-                else:
-                    content = f"Tool call {name} with id {tool_call_id} was cancelled - another message came in before it could be completed."
-                patched_messages.append(ToolMessage(content=content, name=name, tool_call_id=tool_call_id))
+                error_message = self._error_tool_message(tool_call)
+                if error_message is not None and error_message.tool_call_id not in answered_ids:
+                    patched_messages.append(error_message)
+                    patched_any = True
 
+        if not patched_any:
+            return None
```
hntrl: why not just use patched_messages length, why are we adding patched_any

colifran (approved):

libs/deepagents/deepagents/middleware/patch_tool_calls.py
```diff
@@ -1,49 +1,84 @@
...
+        for tool_call in last_message.invalid_tool_calls:
+            error_message = self._error_tool_message(tool_call)
+            if error_message is not None:
+                error_messages.append(error_message)
+
+        if not error_messages:
+            return None
+        if last_message.tool_calls:
+            return {"messages": error_messages}
+        return {"messages": error_messages, "jump_to": "model"}
```
ccurme, line 74: Might be worth a quick check in partner packages that invalid tool calls are sent back to the model; otherwise here we'd get ToolMessages without matching tool calls.

libs/deepagents/deepagents/middleware/patch_tool_calls.py
```diff
@@ -1,49 +1,84 @@
...
+        """Create an error result for an identified tool call."""
+        tool_call_id = tool_call["id"]
+        if tool_call_id is None:
+            return None
+        name = tool_call["name"] or "unknown"
+        if tool_call.get("type") == "invalid_tool_call":
+            content = f"Tool call {name} with id {tool_call_id} could not be executed - arguments were malformed or truncated."
+        else:
+            content = f"Tool call {name} with id {tool_call_id} was cancelled - another message came in before it could be completed."
+        return ToolMessage(content=content, name=name, tool_call_id=tool_call_id, status="error")
```
ccurme, line 32: Confirming use of `is_error` here makes sense https://platform.claude.com/docs/en/agents-and-tools/tool-use/handle-tool-calls#handling-errors-with-is-error

libs/deepagents/deepagents/middleware/patch_tool_calls.py
ccurme, line 74: Are we comfortable with this looping?

ccurme (approved):

libs/deepagents/deepagents/middleware/patch_tool_calls.py
ccurme: This isn't new; it turns out it was an issue for anthropic (fixed in https://github.com/langchain-ai/langchain/pull/40372). Will canvass other integrations as well.

PR #5464, opened 2026-08-13
fix(sdk): pass parent state to async subagent invoke

### LINKED ISSUE #2440: Async subagents option to pass certain state vars from the parent
### Checked other resources

### Area (Required)

- [x] deepagents (SDK)
- [ ] cli

### Feature description

When spawning async subagents, its useful for the child tasks to be able to inherit certain state variables from the parent.

For example, we set an access token in the state variable `access_token` for our deep agents in order to give them the ability to authenticate with our API. Child async tasks that are created by a deep agent should have the same authentication as the parent agent. It seems like the simplest way to achieve this is to have all async tasks inherit the `access_token` state var from their parent.

Additionally it would be useful to have all child async tasks have a reference to the thread_id of their calling thread, perhaps in a `parent_thread_id` key.

Include a list of state keys that should get automatically passed to an async subagent when it is created.

_No response_

hntrl:
fixes #2440

* refactors the state propagation logic for sync subagents to a shared utility
* makes it so async subagents pass that state + the additional message as input to the graph run through langgraph sdk

This is technically a behavior change, but we're patching this over as a correctness issue since this is how sync subagents are also propagating state.

### DIFF
--- a/libs/deepagents/deepagents/middleware/_state.py
+++ b/libs/deepagents/deepagents/middleware/_state.py
@@ -12,11 +12,16 @@ if TYPE_CHECKING:
     from typing import Any
 
 
+_FORKED_CONTEXT_KEY = "_deepagents_forked_context"
+"""State flag that prevents a forked subagent from delegating again."""
+
+
 _EXCLUDED_STATE_KEYS = {
     "messages",
     "todos",
     "structured_response",
     "async_tasks",
+    _FORKED_CONTEXT_KEY,
 }
 """State keys that are excluded when passing state to subagents and when
 returning updates from subagents.
--- a/libs/deepagents/deepagents/middleware/_state.py
+++ b/libs/deepagents/deepagents/middleware/_state.py
@@ -37,7 +37,6 @@ def prepare_subagent_state(
     state: Mapping[str, Any],
     *,
     private_state_keys: frozenset[str] = frozenset(),
-    excluded_keys: frozenset[str] = frozenset(),
 ) -> dict[str, Any]:
     """Copy the public portion of parent state for a child subagent.
 
@@ -45,7 +44,7 @@ def prepare_subagent_state(
     returns. This allows synchronous and remote subagents to use their native
     message representations while sharing the same visibility policy.
     """
-    excluded = _EXCLUDED_STATE_KEYS | private_state_keys | excluded_keys
+    excluded = _EXCLUDED_STATE_KEYS | private_state_keys
     return {key: value for key, value in state.items() if key not in excluded}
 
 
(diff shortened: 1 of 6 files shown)


libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -395,19 +396,7 @@ def _fork_messages(
...
-
-1. The messages key is handled explicitly to ensure only the final message
-    is included
-2. The todos and `structured_response` keys are excluded as they do not have
-    a defined reducer and no clear meaning for returning them from a subagent
-    to the main agent.
-3. Agent-private fields on middleware state schemas are excluded from both
-    subagent output and subagent inputs.
-"""
+"""State keys excluded when passing state to isolated subagents."""
```
hntrl: this is now duplicated in _state.py. can we just upstream and defer this to state?

libs/deepagents/deepagents/middleware/_state.py
```diff
@@ -3,10 +3,52 @@
...
+3. Agent-private fields on middleware state schemas are excluded from both
+    subagent output and subagent inputs.
+"""
+
+
+def prepare_subagent_state(
+    state: Mapping[str, Any],
+    *,
+    private_state_keys: frozenset[str] = frozenset(),
+    excluded_keys: frozenset[str] = frozenset(),
```
hntrl: we have two parameters for the same thing now? do we just append into private_state_keys if we need to add values here? why do we destructure into two kwargs?

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -389,26 +390,6 @@ def _fork_messages(
 The calling agent only sees your final assistant message, not your intermediate work, tool results, or status tracking. Ensure your final
 response contains the complete answer."""
 
-_EXCLUDED_STATE_KEYS = {
-    "messages",
-    "todos",
-    "structured_response",
-    _FORKED_CONTEXT_KEY,
```
hntrl, line 396: you're dropping _FORKED_CONTEXT_KEY in the new excluded list?

hntrl:
!keep-open

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -749,7 +730,10 @@ def _validate_and_prepare_state(
             else:
                 # A compiled runnable is opaque -- it may not declare these
                 # channels, and internal state isn't ours to hand it.
-                inherited = {key: value for key, value in runtime.state.items() if key not in _EXCLUDED_STATE_KEYS | private_state_keys}
+                inherited = prepare_subagent_state(
+                    runtime.state,
+                    private_state_keys=private_state_keys | {_FORKED_CONTEXT_KEY},
```
hntrl: why not just have this in the _EXCLUDED_STATE_KEYS list? thats how it was before

colifran (approved):

PR #5513, opened 2026-08-14
fix(sdk): throw tool error with root path input in virtual mode

### LINKED ISSUE #5469: `LocalShellBackend(virtual_mode=True)` silently nests host paths returned by `execute` when reused by file tools
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

Related to #3050 and the merged fix #3762, which added virtual-to-host path guidance for CompositeBackend routes.

The direct LocalShellBackend case is still uncovered because `_route_host_path_prompt()` returns an empty string for non-CompositeBackend backends.

Also related to #1791, an earlier report about incompatible absolute/virtual paths between file tools and `execute`.

#3796 discusses the same shell/file-tool namespace split, but its scope is shell confinement and security rather than silent path nesting.

```python
from importlib.metadata import version
from pathlib import Path
from tempfile import TemporaryDirectory

from deepagents.backends import LocalShellBackend


with TemporaryDirectory() as tmp:
    root = Path(tmp).resolve()
    backend = LocalShellBackend(root_dir=root, virtual_mode=True)

    # execute() exposes the real host working directory.
    host_cwd = Path(backend.execute("pwd").output.strip())

    # Reuse that host path with a file tool.
    requested = host_cwd / "fib.py"
    result = backend.write(
        str(requested),
        "def fib(n: int) -> int:\n    return n\n",
    )

    expected = root / "fib.py"
    nested = root / str(requested).lstrip("/")

    print(f"deepagents={version('deepagents')}")
    print(f"write_error={result.error!r}")
    print(f"returned_path={result.path}")
    print(f"expected_exists={expected.exists()}")
    print(f"nested_exists={nested.exists()}")
    print(f"expected={expected}")
    print(f"nested={nested}")
```

```shell
deepagents=0.7.5
write_error=None
returned_path=/private/var/folders/jj/j2w9ynv54lgglkwc1jcpk9s8664bj7/T/tmpjd_28mqe/fib.py
expected_exists=False
nested_exists=True
expected=/private/var/folders/jj/j2w9ynv54lgglkwc1jcpk9s8664bj7/T/tmpjd_28mqe/fib.py
nested=/private/var/folders/jj/j2w9ynv54lgglkwc1jcpk9s8664bj7/T/tmpjd_28mqe/private/var/folders/jj/j2w9ynv54lgglkwc1jcpk9s8664bj7/T/tmpjd_28mqe/fib.py
```

### Description

I am using a direct `LocalShellBackend` with both shell execution and filesystem tools:

    LocalShellBackend(root_dir=workspace, virtual_mode=True)

This exposes two different path namespaces to the same agent:

- `execute()` runs a real host shell with `root_dir` as its working directory. Therefore, `execute("pwd")` returns the real host path, for example `/private/.../workspace`.
- File tools such as `write`, `read`, and `edit` interpret every leading-slash path as a virtual path under `root_dir`. For example, `/fib.py` maps to `<root_dir>/fib.py`.

If the host path returned by `execute` is reused with `write`, the operation reports success and returns the requested host-looking path. Internally, however, `_resolve_path()` strips the leading slash and joins the entire host path to `root_dir`.

As a result, this:

    write("/private/.../workspace/fib.py", ...)

silently creates:

    <root_dir>/private/.../workspace/fib.py

instead of:

    <root_dir>/fib.py

This is the reverse direction of the namespace mismatch addressed for `CompositeBackend` in #3050 and #3762. The mapping prompt introduced there explicitly returns an empty string for non-CompositeBackend backends, so a direct `LocalShellBackend` still provides no explanation of the relationship between shell paths and virtual file-tool paths.

Expected behavior:

- At minimum, the agent-facing prompt should explain that file tools use virtual paths such as `/fib.py`, while shell commands run with `root_dir` as their host working directory and should normally use cwd-relative paths such as `fib.py`.
- A file-tool call using a host path that points inside this backend's own `root_dir` should preferably return an actionable error or normalize it to the corresponding virtual path.
- The backend should not report success while silently duplicating its host root path inside itself.

Actual behavior:

- `write()` reports success.
- The returned path looks like the requested host path.
- The expected file is absent.
- A nested duplicate of the host path is created under `root_dir`.

The issue is reproducible without a model or API key, so it is not solely an unsuccessful model trajectory. In an agent run, it caused a correct `fib.py` implementation to be written to the nested location while the verifier reported that `fib.py` was missing.

Suggested regression coverage:

1. `write("/fib.py", ...)` still creates `<root_dir>/fib.py`.
2. Reusing `execute("pwd") + "/fib.py"` with a file tool does not silently create `<root_dir>/<absolute-host-path>/fib.py`.
3. The direct `LocalShellBackend` system prompt explains shell paths versus virtual file-tool paths.
4. `virtual_mode=False` behavior remains unchanged.

### Environment / System Info

OS: macOS 26.5.2 (arm64)
Python: 3.13.8
deepagents: 0.7.5; also reproduced on 0.6.12
langchain: transitive dependency; no LangChain APIs are used directly by the reproducer
Relevant environment variables: none
Model/API key: none; the reproducer is backend-only

hntrl:
fixes #5469

Adds a condition when `virtual_mode=True` for `FileSystemBackend` (and subsequently `LocalShellBackend`) receives the root path when the filesystem tools are invoked

* works by relying on the error logic of `.relative_to` path method to surface if the input path is a subpath of the root path. If it is we throw `OSError` which gives the agent guidance

---

Assume:

```text
root_dir = /foo/bar
virtual_mode = True
```

| Input to file tool | Detected as existing workspace path? | Result | On-disk target / error |
|---|---:|---|---|
| `fib.py` | No — relative input | Allowed | `/foo/bar/fib.py` |
| `/fib.py` | No — not beneath `/foo/bar` as an OS path | Allowed virtual workspace path | `/foo/bar/fib.py` |
| `/src/app.py` | No — not beneath `/foo/bar` as an OS path | Allowed virtual workspace path | `/foo/bar/src/app.py` |
| `/foo/bar/fib.py` | Yes | Rejected | Suggests `/fib.py` |
| `/foo/bar/src/app.py` | Yes | Rejected | Suggests `/src/app.py` |
| `/foo/bar` | Yes | Rejected | Suggests `/.` |
| `/baz/bax.txt` | No — outside `/foo/bar` | Allowed under existing virtual-path semantics | `/foo/bar/baz/bax.txt` |
| `/foo/other.txt` | No — outside `/foo/bar` | Allowed under existing virtual-path semantics | `/foo/bar/foo/other.txt` |
| `../secrets.txt` | N/A | Rejected by existing traversal check | `Path traversal not allowed` |
| `/../secrets.txt` | N/A | Rejected by existing traversal check | `Path traversal not allowed` |
| `/foo/bar/../other.txt` | N/A | Rejected by existing traversal check | `Path traversal not allowed` |

The rejected case raises because `.relative_to` doesn't throw an error if the input can be recognized as a subpath to `root_path`

```text
root_dir = /
virtual_mode = True
```

| Input to file tool | Result | Why |
|---|---|---|
| `/fib.py` | Allowed; writes `/fib.py` | The workspace is already filesystem root, so there is no duplicate workspace prefix to remove. |
| `/foo/bar/fib.py` | Allowed; writes `/foo/bar/fib.py` | Every absolute path is necessarily under `/`; rejecting it would break the virtual-path contract. |

Co-authored-by: krakenalt <44535722+krakenalt@users.noreply.github.com>

### DIFF
--- a/libs/deepagents/deepagents/backends/filesystem.py
+++ b/libs/deepagents/deepagents/backends/filesystem.py
@@ -198,9 +198,25 @@ class FilesystemBackend(BackendProtocol):
         Raises:
             ValueError: If path traversal is attempted in `virtual_mode` or if the
                 resolved path escapes the root directory.
-            OSError: If the path is a symlink loop (`ELOOP`).
+            OSError: If a host path is supplied beneath the virtual root or the path is a
+                symlink loop (`ELOOP`).
         """
         if self.virtual_mode:
+            host_path = Path(key)
+            # A shell paired with this backend reports paths on the host filesystem,
+            # while file tools consume virtual paths. Reject a host path that points
+            # inside the virtual root rather than silently nesting it below that root.
+            if host_path.is_absolute() and self.cwd != self.cwd.parent:
+                try:
+                    # `relative_to` raises ValueError outside the workspace and returns
+                    # the workspace-relative path when the input is inside it.
+                    virtual_path = "/" + host_path.resolve().relative_to(self.cwd).as_posix()
+                except ValueError:
+                    pass
+                else:
+                    msg = f"Use '{virtual_path}' instead: file-tool paths are already relative to the workspace"
+                    raise OSError(msg)
+
             vpath = key if key.startswith("/") else "/" + key
             if ".." in vpath or vpath.startswith("~"):
                 msg = "Path traversal not allowed"
--- a/libs/deepagents/tests/unit_tests/backends/test_local_shell_backend.py
+++ b/libs/deepagents/tests/unit_tests/backends/test_local_shell_backend.py
@@ -152,6 +152,22 @@ def test_local_shell_backend_integration_shell_and_filesystem() -> None:
         assert "Shell created" in content.file_data["content"]
 
 
+def test_local_shell_backend_rejects_host_path_from_execute(tmp_path: Path) -> None:
+    """Host paths returned by the shell must not silently become virtual paths."""
+    backend = LocalShellBackend(root_dir=tmp_path, virtual_mode=True)
+
+    shell_result = backend.execute("pwd")
+    assert shell_result.exit_code == 0
+    requested_path = Path(shell_result.output.strip()) / "fib.py"
+
+    write_result = backend.write(str(requested_path), "def fib(n: int) -> int:\n    return n\n")
+
+    assert write_result.error is not None
+    assert "Use '/fib.py' instead: file-tool paths are already relative to the workspace" in write_result.error
+    assert not (tmp_path / "fib.py").exists()
+    assert not (tmp_path / str(requested_path).lstrip("/")).exists()
+
+
 def test_local_shell_backend_ls_info() -> None:
     """Test listing directory contents."""
     with tempfile.TemporaryDirectory() as tmpdir:


hntrl:
prompting guidance will be more appropriate here

PR #5542, opened 2026-08-17
fix(code): jump over input placeholder tokens with arrow keys

mdrxy:
Arrow keys and word jumps in the chat input now treat attached-media and collapsed-paste placeholders (`[image N]`, `[video N]`, `[Pasted text #N]`) as single units. When you press the left arrow at the right edge of a placeholder, the cursor moves to the left edge. When you press the right arrow at the left edge, the cursor moves to the right edge. The cursor can only stop at a placeholder edge, never inside it. Shift+arrow keys select the text by whole placeholders. Word jumps (Option/Alt+arrow keys) also skip a whole placeholder now. Before this change, a word jump into `[Pasted text #1 +3 lines]` could stop on `lines]` because the placeholder contains spaces.

---

This change makes cursor movement the same as deletion, which already removes a bound placeholder with one keypress. The same condition applies: only placeholders that are backed by a tracked attachment or stored paste content are treated as one unit. If you type placeholder text by hand (for example, `[image 2]`), the cursor still moves through it character by character.

One detail to look at carefully: a space after a placeholder stays ordinary text for cursor movement. So when you press the left arrow from `x[image 1] |`, the cursor lands at `x[image 1]| `. But backspace removes a trailing space together with the placeholder. This difference is on purpose: the trailing space is text that the user typed. But it is easy to change if it feels wrong.


mdrxy:
Follow-up to review: the current text-based placeholder lookup incorrectly treats a manually typed duplicate (for example, a second `[image 1]`) as the attached image. I’ll change cursor navigation to recognize the concrete tracked attachment occurrence/range instead of matching rendered token text, and add a regression test. This remains scoped to cursor semantics; attachment lifecycle and undo/redo stay out of this PR.

mdrxy:
Finish #5130 after

PR #5553, opened 2026-08-17
fix(sdk): propagate subagent state keys for provided subagent middleware

### LINKED ISSUE #5189: `SubAgentMiddleware.private_state_keys` is assigned after construction, so a caller-supplied instance keeps an empty key set
**Package:** `deepagents` · **Severity:** medium, not reachable from `deepagents-code` · **Type:** latent correctness / API design

## Summary

`private_state_keys` is what stops agent-private state from crossing the subagent boundary. It defaults to `frozenset()` on `SubAgentMiddleware` and is assigned by `create_deep_agent` *after* the instance is built. Any code path that ends up with a different `SubAgentMiddleware` instance installed in the graph gets the empty default, and every private channel is then forwarded into subagents and merged back out — silently, with no warning and no error.

`deepagents-code` never constructs a `SubAgentMiddleware`, so it is not exposed. Library users who pass their own are.

## Mechanism

Three facts combine:

1. `SubAgentMiddleware.private_state_keys` defaults to `frozenset()`.
2. `create_deep_agent` derives the real key set from every middleware's `state_schema` and assigns it onto the instance it created, via a property setter — not through the constructor.
3. `_apply_custom_middleware` in `deepagents/graph.py` can **replace** a default middleware instance with a caller-supplied one matched by `.name`.

So a caller who passes their own `SubAgentMiddleware` in `middleware=` can have their instance installed in place of the internally-constructed one. The derivation in step 2 targets the instance from step 2's own construction, so the installed replacement retains `frozenset()`.

The consequence is not a crash. `SubAgentMiddleware` filters nothing, so private keys flow both ways: parent state leaks into subagents, and subagent bookkeeping merges back into the parent. The latter is exactly the class of bug fixed in #5164 — where two parallel `task` results writing the same single-value channel produced `InvalidUpdateError` — so in the parallel case this manifests as a confusing crash attributed to the wrong cause.

## Why `deepagents-code` is unaffected

dcode passes its middleware list through `create_deep_agent` and never constructs a `SubAgentMiddleware` itself, so the internally-created instance is always the installed one and always receives the derived key set. The #5164 fix is therefore effective for dcode. This issue is about the general library contract.

## Suggested fix

Preferred: make the key set a constructor argument rather than a post-construction assignment, so an instance cannot exist in an under-configured state. This is the "make the invalid state unrepresentable" version and removes the failure mode rather than detecting it.

If the setter must stay for compatibility: re-derive and assign after `_apply_custom_middleware` has settled which instance is installed, so the assignment targets the real one.

Either way, a cheap backstop: warn when a `SubAgentMiddleware` is installed with an empty `private_state_keys` while some middleware in the stack *does* declare private fields. That combination is almost certainly a misconfiguration, and it is exactly the state this issue describes.

## References

Surfaced during review of #5164. Symbols: `SubAgentMiddleware.private_state_keys` and its setter in `libs/deepagents/deepagents/middleware/subagents.py`; the derivation and `_apply_custom_middleware` in `libs/deepagents/deepagents/graph.py`; `private_state_field_names` in `libs/deepagents/deepagents/middleware/_state.py`.

hntrl:
fixes #5189

We recently added support to override the provided subagent middleware when using `create_deep_agent` (matched by name).
* some details in the internal graph constructor got missed as apart of this
* this fixes one of them: we now propagate private state keys (state keys annotated with `PrivateStateAttr`) to any instance of SubAgentMiddleware in the final middleware stack

Made by [Open SWE](https://openswe.vercel.app/agents/8788a543-9ff7-53f1-b3d1-55e146c5d597)

### DIFF
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -941,7 +941,7 @@ def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly
     private_state_keys = private_state_field_names(*state_schemas)
     # Apply the keys after caller middleware has replaced the task middleware.
     for middleware_instance in deepagent_middleware:
-        if isinstance(SubAgentMiddleware, type) and isinstance(middleware_instance, SubAgentMiddleware):
+        if isinstance(middleware_instance, SubAgentMiddleware):
             middleware_instance.private_state_keys = middleware_instance.private_state_keys | private_state_keys
     # Verify every main-profile exclusion matched at least one middleware in
     # either the main agent stack or the GP subagent stack. An entry that
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -30,7 +30,6 @@ from langgraph.store.base import BaseStore
 from langgraph.types import Checkpointer
 from langgraph.typing import ContextT
 
-import deepagents.middleware.subagents as subagents_module
 from deepagents._api.deprecation import deprecated, warn_deprecated
 from deepagents._excluded_middleware import (
     _apply_excluded_middleware,
@@ -942,7 +941,7 @@ def create_deep_agent(  # noqa: C901, PLR0912, PLR0915  # Complex graph assembly
     private_state_keys = private_state_field_names(*state_schemas)
     # Apply the keys after caller middleware has replaced the task middleware.
     for middleware_instance in deepagent_middleware:
-        if isinstance(middleware_instance, subagents_module.SubAgentMiddleware):
+        if isinstance(SubAgentMiddleware, type) and isinstance(middleware_instance, SubAgentMiddleware):
             middleware_instance.private_state_keys = middleware_instance.private_state_keys | private_state_keys
     # Verify every main-profile exclusion matched at least one middleware in
     # either the main agent stack or the GP subagent stack. An entry that
--- a/libs/deepagents/deepagents/graph.py
+++ b/libs/deepagents/deepagents/graph.py
@@ -57,6 +57,7 @@ from deepagents.middleware.subagents import (
     CompiledSubAgent,
     SubAgent,
     SubAgentMiddleware,
+    SubAgentMiddleware as _SubAgentMiddleware,
     _is_compiled_subagent,
     _is_forked_subagent,
 )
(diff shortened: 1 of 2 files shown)


libs/deepagents/deepagents/graph.py
```diff
@@ -57,6 +57,7 @@
     CompiledSubAgent,
     SubAgent,
     SubAgentMiddleware,
+    SubAgentMiddleware as _SubAgentMiddleware,
```
hntrl: why are we aliasing this? why not just use SubAgentMiddleware

libs/deepagents/deepagents/graph.py
```diff
@@ -30,6 +30,7 @@
 from langgraph.types import Checkpointer
 from langgraph.typing import ContextT
 
+import deepagents.middleware.subagents as subagents_module
```
hntrl: why can't we just use the existing SubAgentMiddleware import?

libs/deepagents/deepagents/graph.py
```diff
@@ -939,8 +939,10 @@ class MyState(DeepAgentState):
     state_schemas = [state_schema] if state_schema is not None else []
     state_schemas.extend(mw.state_schema for mw in deepagent_middleware if getattr(mw, "state_schema", None) is not None)
     private_state_keys = private_state_field_names(*state_schemas)
-    if sub_agent_middleware is not None:
-        sub_agent_middleware.private_state_keys = private_state_keys
+    # Apply the keys after caller middleware has replaced the task middleware.
+    for middleware_instance in deepagent_middleware:
+        if isinstance(SubAgentMiddleware, type) and isinstance(middleware_instance, SubAgentMiddleware):
```
hntrl: isnt is just 
```suggestion
        if isinstance(middleware_instance, SubAgentMiddleware):
```
?

libs/deepagents/deepagents/graph.py
```diff
@@ -939,8 +939,10 @@ class MyState(DeepAgentState):
     state_schemas = [state_schema] if state_schema is not None else []
     state_schemas.extend(mw.state_schema for mw in deepagent_middleware if getattr(mw, "state_schema", None) is not None)
     private_state_keys = private_state_field_names(*state_schemas)
-    if sub_agent_middleware is not None:
-        sub_agent_middleware.private_state_keys = private_state_keys
+    # Apply the keys after caller middleware has replaced the task middleware.
+    for middleware_instance in deepagent_middleware:
+        if isinstance(middleware_instance, SubAgentMiddleware):
+            middleware_instance.private_state_keys = middleware_instance.private_state_keys | private_state_keys
```
hntrl, line 945: I could be convinced we don't want this --
* rationale for appending existing private keys with middleware derived private keys is because if I supply SubAgentMiddleware I can technically add my own private state keys.
* We still don't want to ignore `PrivateStateAttr` keys on middleware

colifran (approved):

PR #5562, opened 2026-08-18
fix(sdk): surface sandbox glob failures instead of reporting no matches

mdrxy:
`SandboxBackend.glob` can report an empty search when the remote helper failed. The agent then concludes that the files do not exist. The agent can write a file again that is already there.

## The three failure paths

`_parse_glob_output` reads JSON lines from a helper script. The script runs in the sandbox with `2>&1`.

- The code did not examine `exit_code`. If the system stopped the helper, or if `python3` was not there, the function returned `matches=[], truncated=False`. `_parse_grep_output` examines `exit_code`. The glob function now does the same.
- If the output was empty, the function returned before it read `result.truncated`. An output clipped to zero bytes thus looked like a correct empty result.
- The test for an unreadable last line used the wrong variable. The parse loop changes that variable when the walk reports a limit. The test thus also removed a traceback. The code now uses `result.truncated`.

## The truncation cause

One boolean showed four conditions: the time limit, the match limit, an unreadable directory, and a clipped output. The note to the model always told it to make the search more narrow. For an unreadable directory that advice is wrong. A more narrow pattern cannot show files in a directory that the sandbox cannot read. The agent thus retries without end.

`GlobResult` now has a `truncation_reason` field. The values are `budget`, `unreadable` and `transport`. The `unreadable` note tells the model that a more narrow search does not help.

## Traversal patterns

`glob("../*.py")` gave four different results. `FilesystemBackend` raised an error in virtual mode, but returned `[]` in the other mode. `StateBackend` and `StoreBackend` returned `[]`. The sandbox returned a structured error. The protocol says that a refusal is an `error`, not an exception.

The rejection moves into `compile_grep_include_glob`. All backends use that function. All five backends thus give `GlobResult(error=..., matches=None)`. A refusal raises the new `InvalidGlobPatternError`. This class is a subclass of `ValueError`, thus the current handlers continue to operate.

## Other changes

- `ASYNC_GLOB_TIMEOUT` limits `aglob`. The script limit applies only to the walk. A stopped sandbox thus held the caller with no limit.
- A refused `grep(glob=...)` is now a result. It went through `sync_grep`, which has no `try` block.
- A failure during the walk keeps the matches that the script found before the failure.
- The script does not walk `/proc`, `/sys` and `/dev` when the root is `/`.
- The sandbox search root is now absolute. `_check_fs_permission` compares `deny` rules with absolute paths only.
- The code writes remote errors, walk errors and refusals to the log.
- In `CompositeBackend`, a root-anchored pattern does not go to the routed backends. Their results always have the route as a prefix, thus they are deeper than the anchor permits.
- `compile_recursive_glob` is removed. It was an alias of `compile_grep_include_glob` with one caller.

## Compatibility

`truncation_reason` is a new field with the default `None`. `InvalidGlobPatternError` is a subclass of `ValueError`. `compile_recursive_glob` was internal and was never exported.

### DIFF
--- a/libs/deepagents/deepagents/backends/composite.py
+++ b/libs/deepagents/deepagents/backends/composite.py
@@ -18,6 +18,7 @@ from deepagents.backends.protocol import (
     FileInfo,
     FileUploadResponse,
     GlobResult,
+    GlobTruncationReason,
     GrepMatch,
     GrepResult,
     LsResult,
@@ -70,7 +71,8 @@ def _strip_route_from_pattern(pattern: str, route_prefix: str) -> str:
         route_prefix: The route prefix (e.g. `/memories/`).
 
     Returns:
-        The pattern with the route prefix stripped, or the original pattern
+        The pattern with the route prefix replaced by a leading `/`
+            (e.g. `/memories/**/*.md` -> `/**/*.md`), or the original pattern
             if it doesn't match the route.
     """
     bare_pattern = pattern.lstrip("/")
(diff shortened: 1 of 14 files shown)


mdrxy:
Superseded by #5566, which carries the same commit on a branch that follows the `<username>/<scope>/<description>` convention.

PR #5563, opened 2026-08-18
fix(sdk): explain truncation markers only when preview omits lines

mdrxy:
Large tool-result and evicted human-message previews always included the sentence "lines of the form `... [N lines truncated] ...` indicate omitted lines in the middle of the content", even when the preview fit within the head+tail budget and inserted no marker. The note now describes only losses the preview actually has, so the model is never told to look for a marker that was not inserted.

---

`_create_content_preview` now returns a `ContentPreview` that records, out of band, whether whole lines were dropped from the middle — rather than inferring it by scanning rendered text, which would mistake a literal marker line in the content for real truncation. `_preview_note` selects a plain "Here is a preview of the …" sentence or the head/tail sentence with the marker explanation based on that flag, and both the tool-result stub and the evicted human-message stub render through one helper so the note always describes the preview it introduces.

This is intentionally narrow: per-line clipping disclosure, sandbox execute-offload preview accuracy, and deprecation compatibility for the legacy template constants are split into follow-up PRs.


mdrxy:
!keep-open

PR #5564, opened 2026-08-18
fix(sdk): disclose per-line clipping in large-result previews

mdrxy:
Stacked on #5563.

When a previewed line exceeds the per-line character budget, the preview silently showed only its leading characters — presenting an excerpt as the whole result. Previews now add a disclosure when any displayed line is clipped, alongside (and independently of) the omitted-middle-lines marker explanation.

---

`_create_content_preview` clips each shown line to `PREVIEW_LINE_CHAR_LIMIT` and records whether any line was actually cut, surfacing it as a second `ContentPreview` flag alongside `lines_omitted`. The two flags track independent kinds of loss — a preview can omit middle lines, clip long displayed lines, both, or neither — so `_preview_note` reports each only when it occurred. A line exactly at the limit is not reported as clipped.

The capture-offload path in `FilesystemMiddleware` has no per-line budget (its losses are byte-based and disclosed in-band), so it passes `lines_clipped=False` explicitly rather than leaving the flag to be inferred.

### DIFF
--- a/libs/deepagents/deepagents/middleware/_message_eviction.py
+++ b/libs/deepagents/deepagents/middleware/_message_eviction.py
@@ -38,12 +38,16 @@ You can do this by specifying an offset and limit in the read_file tool call. Fo
 {content_sample}
 """
 
-_PREVIEW_LINE_CHAR_LIMIT: Final = 1000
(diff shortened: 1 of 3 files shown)


mdrxy:
!keep-open

PR #5567, opened 2026-08-18
fix(sdk): disclose byte-cap losses in capture-offload execute previews

mdrxy:
Stacked on #5564.

Capture-offload previews of large `execute` output misreported what they dropped in three ways: a preview of few-but-huge lines (a `.jsonl` dump) ended mid-output with no note; a capture that hit its byte cap advertised the cap as the output's true total and claimed the saved file held the full output; and a head/tail excerpt whose 2000-byte caps cut a shown line mid-line disclosed nothing, since the truncation marker only counts whole lines. Each preview now discloses its actual byte-based loss, and the marker presence that drives the "head and tail" note is carried out of band rather than inferred by scanning output text.

---

The capture wrapper emits a fifth meta field — the line surplus `captured_lines - head - tail` — so `_parse_capture_execute_output` sets `ExecuteOffloadResult.preview_has_truncation_marker` from producer state instead of scanning for a marker that command output could literally contain. Three in-band notices close the byte-loss gaps:

- a leading byte excerpt that loses the end of the output closes with `... [output clipped here; N bytes total, full output at the path above] ...`;
- when capture hit its cap, that swaps to `capture stopped at its N-byte limit, so the file at the path above is also incomplete` — the cap is not the output's real size and the saved file is incomplete;
- a head/tail preview whose excerpts hit their byte caps inserts `... [the excerpts above and below are byte-capped ...] ...` before the marker.

The surplus is parsed separately from the exit code so a garbled fifth field degrades only the note's wording, never the exit code or cap flag. Malformed/absent meta lines now log a warning, since the fallback silently reframes a preview as complete output.


mdrxy:
!keep-open

PR #5568, opened 2026-08-18
fix(sdk): preserve legacy large-result templates behind deprecation

mdrxy:
Stacked on #5567.

`TOO_LARGE_TOOL_MSG` (re-exported from `deepagents.middleware.filesystem`) and `TOO_LARGE_HUMAN_MSG` shipped through `deepagents==0.7.6` as importable module attributes. The preceding PRs in this stack rename them to private `_TOO_LARGE_*` and change the template wording, which would break any downstream `from deepagents.middleware.filesystem import TOO_LARGE_TOOL_MSG`. This preserves the released formatting contract behind a deprecated compatibility path.

---

Importing either name from `deepagents.middleware.filesystem` still succeeds and returns a frozen copy of the released template text (placeholders and wording unchanged), but now emits a `LangChainDeprecationWarning` via a module-level `__getattr__`, noting deprecation in `deepagents==0.7.7` and removal in `deepagents==0.9.0`. The frozen copies deliberately keep the old fixed wording rather than the shared `TRUNCATION_MARKER_TEMPLATE`, so the compatibility surface does not drift as the live templates evolve.

### DIFF
--- a/libs/deepagents/deepagents/backends/protocol.py
+++ b/libs/deepagents/deepagents/backends/protocol.py
@@ -828,13 +828,29 @@ class ExecuteOffloadResult:
     offloaded: bool
     """Whether the output was left at the capture path.
 
-    When `True`, `response.output` holds only a head/tail preview and the full
-    output lives at the capture path on the sandbox filesystem. When `False`,
-    `response.output` is the complete output.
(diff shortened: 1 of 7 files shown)


mdrxy:
!keep-open

PR #5590, opened 2026-08-18
fix(code): make `/offload` interruptible

aolsenjazz:
`/offload` and `/compact` can now be interrupted with Escape as their progress hint promises.

---

Previously, the interface showed `esc to interrupt` while offloading, but pressing Escape did nothing. On a slow or stuck offload, users remained on the `Offloading` state with no way to regain control without waiting for the request to finish or fail.

Now Escape cancels an in-progress offload and returns the interface to an idle, usable state. If the offload had already completed when cancellation arrived, its result is preserved; otherwise, the pending operation is cleaned up so the conversation can continue normally.

Made by [Open SWE](https://openswe.vercel.app/agents/d5b34378-8e7e-57e7-8e22-82cb74c789f6)


mdrxy:
Will rebase #5261 after merging

PR #5603, opened 2026-08-18
feat(code,talon): require Python 3.12 or greater

mdrxy:
`dcode` now requires Python 3.12 or newer (previously 3.11). Users on 3.11 will need to upgrade their interpreter to install or update to this release.

---

The floor bump lets the codebase use 3.12-only syntax and drop 3.11 compatibility shims:

- PEP 695 `type` statements replace `TypeAlias` annotations, and generic functions/classes use type-parameter syntax instead of `TypeVar`/`Generic[...]`
- `Protocol` / `override` import from `typing` instead of `typing_extensions`
- The pre-PEP-706 `tarfile` extraction fallback (needed only for 3.11.0–3.11.3) is gone; extraction always uses `filter="data"`
- `uv.lock` re-locks without 3.11-only backports (`tomli`, `backports-tarfile`)
- CI matrices, `ty`, and ruff `target-version` all move to 3.12; integration test defaults and release builds are updated to match

`deepagents-talon` depends on `deepagents-code`, so its floor moves in the same commit — it cannot bump independently without a broken window where talon allows an interpreter its dependency rejects. This is the deliberate coordinated multi-package bump called out in RELEASING.md, so the `allow-lockfile-release` label is applied to acknowledge the talon release-please fan-out.

<details>
<summary>Test plan</summary>

- `ruff check` / `ruff format --check` clean
- `ty check` clean
- Full `libs/code` unit suite passes (two pre-existing failures in `test_app.py` / `test_server_graph.py` reproduce on the base commit and are unrelated)
- COMMANDS.md catalog check and lockfile freshness hooks pass

</details>


.github/workflows/integration_tests.yml
```diff
@@ -6,7 +6,7 @@
 # or type a custom path in the override field.
 
 name: "⏰ Integration Tests"
-run-name: "integration tests — ${{ inputs.working-directory-override || (inputs.working-directory == 'all' && 'all libs') || inputs.working-directory }} (py ${{ inputs.python-version || '3.11, 3.13' }})"
+run-name: "integration tests — ${{ inputs.working-directory-override || (inputs.working-directory == 'all' && 'all libs') || inputs.working-directory }} (py ${{ inputs.python-version || (contains(format('{0} {1}', inputs.working-directory, inputs.working-directory-override), 'libs/code') && '3.12, 3.13') || '3.11, 3.13' }})"
```
mdrxy, line 9: what's this change?

libs/code/deepagents_code/hooks/models/config.py
```diff
@@ -59,7 +59,10 @@ def _normalize_async(cls, value: bool | None) -> None:
             raise ValueError(msg)
 
 
-HandlerSpec: TypeAlias = CommandHandlerSpec
+# Extension point for future handler kinds, kept as a plain assignment rather
+# than a `type` alias: a `type` alias becomes the schema identity and renames
+# the generated `$defs` entry from `CommandHandlerSpec` to `HandlerSpec`.
+HandlerSpec = CommandHandlerSpec
```
mdrxy, line 65: what's up with this?

libs/code/deepagents_code/plugins/adapters/skills.py
```diff
@@ -4,19 +4,22 @@
...
 
-SkillPath: TypeAlias = str
-SkillLabel: TypeAlias = str
-SkillNamespace: TypeAlias = str
-DirectorySkillSource: TypeAlias = tuple[SkillPath, SkillLabel]
-PluginSkillSource: TypeAlias = tuple[SkillPath, SkillLabel, SkillNamespace]
-CodeSkillSource: TypeAlias = DirectorySkillSource | PluginSkillSource
+# Documentation aliases: these name the role each `str` plays in the source
+# tuples below, but they are plain aliases, not `NewType`s. Any `str` satisfies
+# all three, so swapping a label for a namespace still type-checks.
```
mdrxy, line 16: is this comment needed?

libs/code/deepagents_code/skills/merge.py
```diff
@@ -12,20 +12,22 @@
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, TypeVar
+
+# Runtime (not TYPE_CHECKING) import: PEP 695 type-parameter bounds are lazy but
+# are evaluated on access, so a TYPE_CHECKING-only `Mapping` raises `NameError`.
```
mdrxy, line 17: ?

libs/code/deepagents_code/json_types.py
```diff
@@ -1,12 +1,16 @@
...
 
-JsonScalar: TypeAlias = str | int | float | bool | None
-JsonValue: TypeAlias = PydanticJsonValue
-JsonObject: TypeAlias = dict[str, JsonValue]
+type JsonScalar = str | int | float | bool | None
+
+# Plain re-export, not a `type` alias: `pydantic.JsonValue` is already a
+# `TypeAliasType`, and wrapping it in a second one defeats pydantic's recursive
+# alias collapsing, which more than doubles the length of validation-error
+# titles on every JSON payload this module validates.
```
mdrxy, line 10: ?

libs/code/pyproject.toml
```diff
@@ -236,6 +235,9 @@ division-by-zero = "error"
 # across environments.
 unused-ignore-comment = "ignore"
 
+[tool.ruff]
+target-version = "py312"
+
```
mdrxy: not needed i believe

.github/workflows/integration_tests.yml
mdrxy: The workflow's display name was hardcoded to say `py 3.11, 3.13` regardless of which package was being tested. Now that `libs/code` runs on a 3.12 floor, the run-name picks up the actual versions being tested: if the target directory (or override) is `libs/code`, show `3.12, 3.13`; otherwise show the default `3.11, 3.13`. The `format('{0} {1}', ...)` builds a string from both inputs so `contains(..., 'libs/code')` matches whichever one is set.

libs/code/deepagents_code/hooks/models/config.py
mdrxy: `HandlerSpec` was previously declared as `HandlerSpec: TypeAlias = CommandHandlerSpec`, which makes Pydantic treat the alias as the schema identity — the generated JSON Schema's `$defs` entry gets named `HandlerSpec` instead of `CommandHandlerSpec`, and any downstream consumer (or generated docs / OpenAPI client) sees the wrong name. Switching to a plain assignment (`HandlerSpec = CommandHandlerSpec`) keeps the alias as a forward-compatible extension point without renaming the schema. The comment above it documents why we don't use `type HandlerSpec = CommandHandlerSpec` (PEP 695) either — same problem.

libs/code/deepagents_code/plugins/adapters/skills.py
mdrxy: Yes — without it, the three aliases `SkillPath = str`, `SkillLabel = str`, `SkillNamespace = str` look interchangeable and easy to "simplify" back into a single alias or bare `str` annotations. The comment documents the actual constraint: these are *documentation* aliases that name the role each `str` plays, not `NewType`s, so the type checker won't catch a swapped label/namespace. Anyone touching the tuple types below should know that before assuming the names enforce anything.

libs/code/deepagents_code/skills/merge.py
mdrxy: The PEP 695 generic syntax (`def merge_skill[SkillT: Mapping[str, object]](...)`) makes the bound lazy, but it still has to resolve `Mapping` at runtime when the function is defined. Under `from __future__ import annotations` + `TYPE_CHECKING`, `Mapping` only exists during static analysis and would raise `NameError` the moment the module is imported. The import has to be a real runtime import; the comment explains why so nobody moves it back under `TYPE_CHECKING`.

libs/code/deepagents_code/json_types.py
mdrxy: `pydantic.JsonValue` is already a `TypeAliasType`. Wrapping it in a second PEP 695 `type` alias breaks Pydantic's recursive-alias collapsing and makes validation-error titles more than twice as long (they'd print the full expanded alias at every nesting level). Keeping it as a plain re-export (`JsonValue = PydanticJsonValue`) preserves the terse error messages. The comment exists because it looks like an obvious candidate to convert alongside the other two aliases.

libs/code/pyproject.toml
mdrxy: `libs/evals/pyproject.toml` already sets `target-version = "py312"` and ruff doesn't infer it from `requires-python`. Without this line, ruff defaults to `py311` parsing for the whole `libs/code` tree, so it would either (a) fail to parse PEP 695 `type` alias statements (added in 3.12) or (b) flag them, depending on the rule. The `type` aliases introduced in this PR (`json_types.py`, `plugins/adapters/skills.py`, `skills/merge.py`) require `py312` for ruff to analyze them correctly.

mdrxy: Correction on my previous reply — you're right, this isn't needed. Ruff infers a missing `target-version` from `requires-python` in the same `pyproject.toml` (see [Inferring the Python version](https://docs.astral.sh/ruff/configuration/#inferring-the-python-version)). Verified locally: with the `[tool.ruff] target-version = "py312"` block removed, `ruff check --show-settings` still reports `unresolved_target_version = 3.12` because `requires-python = ">=3.12"` is set. I'll drop it.

libs/talon/pyproject.toml
```diff
@@ -7,15 +7,14 @@ name = "deepagents-talon"
 version = "0.0.3"
```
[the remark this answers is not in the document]
mdrxy: Accurate — thanks for catching this. Fixed in ad1913021: converted `_parse_number` in `channels/base.py` to PEP 695 type parameters (`def _parse_number[T](...)`) and `FakeCodeLoaderResult` in `tests/test_mcp.py` to a `type` alias. `make -C libs/talon lint` now passes cleanly (ruff check + format + ty all green).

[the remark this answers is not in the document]
mdrxy: Accurate — the gate is correct that `feat(code): ...` is bump-worthy while the diff also touches `libs/talon/pyproject.toml` (the `requires-python` floor, `ty` version, and the lint fixes in ad1913021). Talon's floor *has* to move in lockstep because it depends on `deepagents-code`, so splitting this into two PRs would leave talon broken on 3.11-only syntax until the second PR landed.\n\nOptions: (1) retitle the PR to `feat(code, talon): require Python 3.12 or greater` if multi-scope titles are allowed here; (2) keep the `allow-lockfile-release` label as the acknowledge-bypass and accept that release-please will open a talon release PR too; or (3) split talon's `pyproject.toml` + lint fixes into a follow-up PR that merges immediately after this one. Preference?

PR #5617, opened 2026-08-18
fix(code): render stale-PATH hint as next step, not warning

mdrxy:
When the installer successfully updates the shell profile, the running shell still lacks `~/.local/bin` on PATH until it's reloaded. That case was rendered with the same yellow warning styling as a genuine profile setup failure, even though the install succeeded and new terminals work immediately. Now it prints a cyan info line with a paste-ready command instead:

    ▸ To use dcode in this shell, run:
      > source ~/.local/bin/env

---

`ensure_path_setup` returns distinct codes for "profile setup failed" (rc=1, already accompanied by a specific warning) and "profile written, running shell just stale" (rc=2). The footer previously lumped both under one `log_warn "Restart your shell, or run:"` block, so a fully successful install could end on a warning. The rc=2 path now renders as a `log_info` next-step hint styled like a prompt (`> command`), matching how other installers (e.g. fx) present the same affordance. The rc=1 path keeps the warning styling since something genuinely went wrong.

Adapted from the fx install script (https://releases.fx.sh).


mdrxy:
Superseded by #5618 (branch renamed to match convention).

PR #5631, opened 2026-08-19
feat(code): define and discover Python extensions

johannes117:
Related #5556

Defines the narrow Python extension API, registry, provenance model, and deterministic source discovery. Extension discovery remains disabled unless `DEEPAGENTS_CODE_EXPERIMENTAL` is truthy.

---

Layer 1 of 4. This layer owns registration validation and collision handling, transactional rollback primitives, plugin manifest declarations, and discovery from user, configured, temporary CLI, plugin, entry-point, and trusted project sources. Configuration, loading, graph hosting, and user-facing trust controls are deliberately excluded.

## Stack

### DIFF
--- a/libs/code/deepagents_code/extensions/api.py
+++ b/libs/code/deepagents_code/extensions/api.py
@@ -1,0 +1,180 @@
+"""Public registrar passed to Python extension factories."""
+
+from __future__ import annotations
+
+import re
+from typing import TYPE_CHECKING, Any
+
+from deepagents_code.extensions.registry import ExtensionError
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from pathlib import Path
+
+    from deepagents.backends.protocol import BackendProtocol
+    from langchain.agents.middleware.types import AgentMiddleware
+    from langchain_core.tools import BaseTool
+
+    from deepagents_code.extensions.registry import ExtensionRegistry, SourceInfo
+
+_ROUTE_PREFIX = re.compile(r"^/(?:[a-z0-9][a-z0-9_-]*/)+$")
+
+
+class ExtensionAPI:
+    """Factory-scoped registrar and read-only session context.
+
+    Each extension receives a dedicated instance, which attributes every
+    registration to the extension's source without exposing mutable dcode
+    application state.
+    """
+
+    def __init__(
+        self,
+        registry: ExtensionRegistry,
+        source: SourceInfo,
+        *,
+        cwd: Path,
+        mode: str,
+    ) -> None:
+        """Bind a registrar to one extension's provenance.
+
+        Args:
+            registry: Shared registry receiving extension units.
+            source: Provenance recorded on each registration.
+            cwd: Working directory for this session.
+            mode: Runtime mode, either `interactive` or `headless`.
+        """
+        self._registry = registry
+        self._source = source
+        self._cwd = cwd
+        self._mode = mode
+        self._active = True
+
+    def _deactivate(self) -> None:
+        """Close this registrar after failed initialization or shutdown."""
+        self._active = False
+
+    def _ensure_active(self) -> None:
+        if not self._active:
+            msg = "Extension registration is closed for this session"
+            raise ExtensionError(msg)
+
+    @property
+    def cwd(self) -> Path:
+        """Working directory for this session."""
+        return self._cwd
+
+    @property
+    def mode(self) -> str:
+        """Runtime mode, either `interactive` or `headless`."""
+        return self._mode
+
+    @property
+    def has_ui(self) -> bool:
+        """Whether this session has an interactive terminal UI."""
+        return self._mode == "interactive"
+
+    @property
+    def path(self) -> Path:
+        """Entry file for this extension."""
+        return self._source.path
+
+    def register_middleware(
+        self,
+        middleware: AgentMiddleware[Any, Any] | type[AgentMiddleware[Any, Any]],
+    ) -> None:
+        """Install LangChain middleware on the agent.
+
+        A class is instantiated without arguments. Pass an instance when
+        construction requires configuration.
+
+        Args:
+            middleware: Middleware class or instance.
+
+        Raises:
+            ExtensionError: If `middleware` is invalid or cannot be constructed.
+        """
+        self._ensure_active()
+        from langchain.agents.middleware.types import AgentMiddleware
+
+        try:
+            instance = middleware() if isinstance(middleware, type) else middleware
+        except Exception as exc:
+            msg = f"Could not construct middleware {middleware!r}: {exc}"
+            raise ExtensionError(msg) from exc
+        if not isinstance(instance, AgentMiddleware):
+            kind = type(instance).__name__
+            msg = f"Registered middleware must be an AgentMiddleware, got {kind}"
+            raise ExtensionError(msg)
+        self._registry.add_middleware(instance, self._source)
+
+    def register_tool(self, tool: BaseTool | Callable[..., Any]) -> None:
+        """Expose an LLM-callable tool.
+
+        Plain callables are converted using LangChain's tool schema inference.
+
+        Args:
+            tool: Tool instance or plain callable.
+
+        Raises:
+            ExtensionError: If a callable cannot be converted into a tool.
+        """
+        self._ensure_active()
+        from langchain_core.tools import BaseTool, tool as as_tool
+
+        if isinstance(tool, BaseTool):
+            self._registry.add_tool(tool, self._source)
+            return
+        try:
+            converted = as_tool(tool)
+        except Exception as exc:
+            name = getattr(tool, "__name__", repr(tool))
+            msg = f"Could not convert {name} into a tool: {exc}"
+            raise ExtensionError(msg) from exc
+        self._registry.add_tool(converted, self._source)  # type: ignore[arg-type]  # callable overload returns BaseTool
+
+    def register_backend_route(self, prefix: str, backend: BackendProtocol) -> None:
+        """Mount a backend under a virtual filesystem path.
+
+        File operations under `prefix` are routed to `backend` by the agent's
+        `CompositeBackend`. Shell execution remains on the default backend and
+        cannot access routed virtual content.
+
+        Args:
+            prefix: Lowercase absolute path ending in `/`, such as `/memories/`.
+            backend: Backend serving file operations under the prefix.
+
+        Raises:
+            ExtensionError: If the prefix or backend is invalid.
+        """
+        self._ensure_active()
+        from deepagents.backends.protocol import BackendProtocol
+
+        if _ROUTE_PREFIX.fullmatch(prefix) is None:
+            msg = (
+                f"Invalid backend route prefix {prefix!r}: use lowercase path "
+                "segments and include leading and trailing slashes"
+            )
+            raise ExtensionError(msg)
+        if not isinstance(backend, BackendProtocol):
+            msg = (
+                f"Backend route {prefix!r} got {type(backend).__name__}, "
+                "which is not a BackendProtocol"
+            )
+            raise ExtensionError(msg)
+        self._registry.add_backend_route(prefix, backend, self._source)
+
+    def on_shutdown(self, hook: Callable[[], Any]) -> None:
+        """Register a deterministic session teardown callback.
+
+        Args:
+            hook: Sync or async zero-argument callback.
+
+        Raises:
+            ExtensionError: If `hook` is not callable.
+        """
+        self._ensure_active()
+        if not callable(hook):
+            msg = "Shutdown hook is not callable"
+            raise ExtensionError(msg)
+        self._registry.add_shutdown_hook(hook, self._source)
--- a/libs/code/deepagents_code/extensions/api.py
+++ b/libs/code/deepagents_code/extensions/api.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import re
+from enum import StrEnum
 from typing import TYPE_CHECKING, Any
 
 from deepagents_code.extensions.registry import ExtensionError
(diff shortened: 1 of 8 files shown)


libs/code/deepagents_code/extensions/discovery.py
```diff
@@ -0,0 +1,198 @@
...
+
+def _scan(directory: Path, scope: SourceScope) -> tuple[list[SourceInfo], list[str]]:
+    try:
+        entries = sorted(directory.expanduser().iterdir())
+    except FileNotFoundError:
+        return [], []
+    except OSError as exc:
+        return [], [f"Could not scan extension directory {directory}: {exc}"]
+
+    sources: list[SourceInfo] = []
```
johannes117:
```suggestion
    sources: list[SourceInfo] = []
<<<<<<< HEAD
```

libs/code/deepagents_code/extensions/discovery.py
```diff
@@ -0,0 +1,198 @@
...
+    errors: list[str] = []
+    for entry in entries:
+        try:
+            source = _entry_source(entry, scope)
+        except (OSError, RuntimeError) as exc:
+            errors.append(f"Could not inspect extension entry {entry}: {exc}")
+            continue
+        if source is not None:
+            sources.append(source)
+    return sources, errors
```
johannes117:
```suggestion
    return sources, []
=======
    errors: list[str] = []
    for entry in entries:
        try:
            source = _entry_source(entry, scope)
        except (OSError, RuntimeError) as exc:
            errors.append(f"Could not inspect extension entry {entry}: {exc}")
            continue
        if source is not None:
            sources.append(source)
    return sources, errors


def _entry_source(entry: Path, scope: SourceScope) -> SourceInfo | None:
    if entry.is_file() and entry.suffix == ".py":
        return _source(entry, scope)
    for filename in ("__init__.py", "extension.py"):
        candidate = entry / filename
        if candidate.is_file():
            return _source(candidate, scope, package=True)
    return None
>>>>>>> d1fdc30ed81885f491520bc79505f38de08ae61b
```

libs/code/deepagents_code/extensions/discovery.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+def _entry_point_sources() -> DiscoveryResult:
+    sources: list[SourceInfo] = []
+    errors: list[str] = []
+    entries = sorted(
+        importlib.metadata.entry_points(group=ENTRY_POINT_GROUP),
+        key=lambda entry: (entry.name, entry.value),
+    )
+    for entry in entries:
```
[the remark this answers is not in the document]
johannes117: Fixed in `347c965`: entry point enumeration is now guarded, so malformed distribution metadata returns a non-fatal discovery error and the remaining sources still resolve. Added a regression test covering this path.

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class SourceScope(StrEnum):
+    """Authority scope that supplied an extension."""
+
+    USER = "user"
+    PROJECT = "project"
+    TEMPORARY = "temporary"
```
aolsenjazz, line 37: making a note for myself - what is temporary?

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+@dataclass(frozen=True, slots=True)
+class SourceInfo:
+    """An extension entry file, ownership, and import shape."""
+
+    path: Path
+    is_package: bool = False
```
aolsenjazz, line 45: What's going on here?

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+    plugin_id: str | None = None
```
aolsenjazz: Not immediately clear why we need a plugin_id, my gut says that SourceInfo should be mechanism-agnostic; it should only care about the relevant path

my gut has been wrong before

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+    scope: SourceScope = SourceScope.USER
+    version: str | None = None
```
aolsenjazz, line 48: Curious to see how this fits in

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+    installed_root: Path | None = None
+
+    @property
+    def label(self) -> str:
```
aolsenjazz, line 52: I suppose this suggest the invariant that there is a one-to-one relationship between an extension file and a single plugin

I'm not convinced that we want to create that invariant at this point

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+    def __init__(self) -> None:
+        """Create an empty registry."""
+        self.middleware: list[RegisteredUnit[AgentMiddleware[Any, Any]]] = []
+        self.tools: list[RegisteredUnit[BaseTool]] = []
+        self.backend_routes: list[RegisteredUnit[BackendProtocol]] = []
+        self.shutdown_hooks: list[RegisteredUnit[Callable[[], Any]]] = []
+        self._lock = threading.RLock()
+        self._listeners: list[Callable[[str, RegisteredUnit[Any]], None]] = []
+        self._apis: list[ExtensionAPI] = []
+        self._restart_required = False
```
aolsenjazz, line 100: I thought we nixed this?

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+
+class ExtensionRegistry:
+    """Hold extension registrations in load order."""
+
+    def __init__(self) -> None:
+        """Create an empty registry."""
+        self.middleware: list[RegisteredUnit[AgentMiddleware[Any, Any]]] = []
+        self.tools: list[RegisteredUnit[BaseTool]] = []
+        self.backend_routes: list[RegisteredUnit[BackendProtocol]] = []
+        self.shutdown_hooks: list[RegisteredUnit[Callable[[], Any]]] = []
```
aolsenjazz, line 95: Interesting, curious to see how this is tied into the extant hooks system. ideally they use the same engine and we basically just have two diff flush locations so-to-speak

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+        self._lock = threading.RLock()
+        self._listeners: list[Callable[[str, RegisteredUnit[Any]], None]] = []
+        self._apis: list[ExtensionAPI] = []
+        self._restart_required = False
+
+    @property
+    def restart_required(self) -> bool:
```
aolsenjazz, line 103: per comment above - I thought we nixed dynamic registration + therefore shouldn't need restart? I won't comment on this again

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+        """Whether a late registration needs a new agent graph."""
+        with self._lock:
+            return self._restart_required
+
+    def require_restart(self) -> None:
+        """Record that registrations have changed graph construction inputs."""
+        with self._lock:
+            self._restart_required = True
+
+    def _snapshot(self) -> tuple[int, int, int, int]:
```
aolsenjazz: I don't _love_ the idea of using a `tuple[int x 4]` for this - no semantic meaning to the positions here. would rather see either a dict or just collapsing all of these into a single RegisteredUnit[]

Generally imo given that we're passing around the RegisteredUnit abstraction, if we get this right, we should generally be able to operate on registered units as one collection rather than decomposed collections of Middleware, Tools etc

aolsenjazz: non blocking tho

libs/code/deepagents_code/extensions/registry.py
aolsenjazz, line 1: generally looks good! couple of comments re. how we removed dynamic loading. also not convinced that we need to decompose storage of RegisteredUnits into collections of their component types, but it's fine

libs/code/deepagents_code/extensions/registry.py
aolsenjazz: If this is the most correct solve for whatever it's solving, maybe a quick docstring. otherwise it reads odd

libs/code/deepagents_code/extensions/registry.py
aolsenjazz: Under what condition is using plugin_id as a label more correct than the path?

libs/code/deepagents_code/extensions/api.py
```diff
@@ -0,0 +1,180 @@
...
+    application state.
+    """
+
+    def __init__(
+        self,
+        registry: ExtensionRegistry,
+        source: SourceInfo,
+        *,
+        cwd: Path,
+        mode: str,
```
aolsenjazz: Possible to define this as an enum?

libs/code/deepagents_code/extensions/api.py
```diff
@@ -0,0 +1,180 @@
...
+    ) -> None:
+        """Bind a registrar to one extension's provenance.
+
+        Args:
+            registry: Shared registry receiving extension units.
+            source: Provenance recorded on each registration.
+            cwd: Working directory for this session.
+            mode: Runtime mode, either `interactive` or `headless`.
+        """
+        self._registry = registry
```
aolsenjazz, line 55: Dependencies are a little hairy here according to my read. the extension registry maintains a collection of ExtensionAPI, which also contain a reference to their parent registry? this indicates a design issue to me

libs/code/deepagents_code/extensions/registry.py
```diff
@@ -0,0 +1,225 @@
...
+    """Hold extension registrations in load order."""
+
+    def __init__(self) -> None:
+        """Create an empty registry."""
+        self.middleware: list[RegisteredUnit[AgentMiddleware[Any, Any]]] = []
+        self.tools: list[RegisteredUnit[BaseTool]] = []
+        self.backend_routes: list[RegisteredUnit[BackendProtocol]] = []
+        self.shutdown_hooks: list[RegisteredUnit[Callable[[], Any]]] = []
+        self._lock = threading.RLock()
+        self._listeners: list[Callable[[str, RegisteredUnit[Any]], None]] = []
```
aolsenjazz: It's not immediately clear what the purpose of _listeners is. can we get a better var name or docstring?

aolsenjazz (changes requested):

libs/code/deepagents_code/extensions/registry.py
johannes117: temporary is for extensions passed explicitly with --extension for one run, can add a docstring

libs/code/deepagents_code/extensions/registry.py
johannes117: we support 2 shapes: a standalone file *.py and a directory containing __init__.py or extension.py
is_package=False is for standalone and is_package=True is for directory

libs/code/deepagents_code/extensions/registry.py
johannes117: agree, will remove and replace with a mechanism agnostic field like source_id or something

libs/code/deepagents_code/extensions/registry.py
johannes117: version has no effect, its just metadata to determine the version of a package that the extension came from

libs/code/deepagents_code/extensions/registry.py
johannes117: on the first comment, we aren't introducing a 1:1 relationship here, one plugin can provide many extension files, we record one source on a discovered extension

johannes117: I've renamed the plugin_id variable to source_id to be more clear. but yes the canonical path is more correct, its just less human readable in a warning, for example:
~/.deepagents/plugins/cache/.../extension.py vs calendar@official

libs/code/deepagents_code/extensions/registry.py
johannes117: I think we may have misunderstood eachother, I don't think that requiring a restart for an extension loaded counts as dynamic registration

libs/code/deepagents_code/extensions/registry.py
johannes117: keeping them separate is easier, they need stuff like reverse registration order, ownership by the server process that loaded the extension
existing hooks system has different dispatch and reporting behavior.

libs/code/deepagents_code/extensions/registry.py
johannes117: replied above but yeah this is not what I meant by dynamic registration. 
What I thought you meant by lazy loading was the ability to reload without restarting

libs/code/deepagents_code/extensions/registry.py
johannes117: super agree, refactoed to use RegistrySnapshot dataclass with named fields

libs/code/deepagents_code/extensions/registry.py
johannes117: separating RegisteredUnits is useful for type safety, I think a unified collection would be more complex

libs/code/deepagents_code/extensions/api.py
johannes117: yes implemented

libs/code/deepagents_code/extensions/api.py
johannes117: okay im going to move the api lifetime ownership to the extension runtime/loader and leave the registry respoinsble for registered units and notifications

libs/code/deepagents_code/extensions/registry.py
johannes117: renaming _listeners to _registration_listeners
subscribe() to subscribe_to_registrations()

libs/code/deepagents_code/extensions/registry.py
aolsenjazz: Sure

libs/code/deepagents_code/extensions/registry.py
aolsenjazz: Hm

aolsenjazz (approved):

PR #5632, opened 2026-08-19
feat(code): configure and load Python extensions

johannes117:
Related #5556

Configures, authorizes, loads, and tears down Python extensions with failure isolation.

---

Layer 2 of 4. Configuration supports TOML and environment overrides, explicit trusted source paths, and project trust controls. Factories load deterministically off-loop, failed factories roll back partial registrations, discovery and load errors remain isolated, and shutdown callbacks run in reverse order. Agent graph integration and user-facing prompts are deliberately excluded.

## Stack

Made by [Open SWE](https://openswe.vercel.app/agents/01a03f51-8274-7281-b722-10ff3de50a11)


libs/code/deepagents_code/extensions/settings.py
```diff
@@ -0,0 +1,115 @@
...
+class TrustPolicy(StrEnum):
+    """Default decision for project-authored extension code."""
+
+    ASK = "ask"
+    ALWAYS = "always"
+    NEVER = "never"
+
+
+@dataclass(frozen=True, slots=True)
+class ExtensionSettings:
```
aolsenjazz, line 28: What resource is this `ExtensionSettings` dataclass tied to? A file? a single-unit-of-extension vis-a-vis a package _or_ file?

libs/code/deepagents_code/extensions/settings.py
```diff
@@ -0,0 +1,115 @@
...
+    enabled: bool = True
+    """Whether extension discovery is enabled."""
+
+    trust: TrustPolicy = TrustPolicy.ASK
+    """Fallback policy for project extension code."""
+
+    extra_files: tuple[Path, ...] = ()
+    """Explicit files from trusted user configuration."""
+
+    extra_dirs: tuple[Path, ...] = ()
```
aolsenjazz: I don't know this system at all, but my gut says that we should flatten extra_dirs into extra_files - what does this distinction buy us?

libs/code/deepagents_code/extensions/settings.py
```diff
@@ -0,0 +1,115 @@
...
+
+    Returns:
+        Effective settings with safe defaults for malformed values.
+
+    Raises:
+        RuntimeError: If the extension manifest options are unavailable.
+    """
+    from deepagents_code.config_manifest import get_option
+
+    resolver = get_config_resolver()
```
aolsenjazz, line 97: hell yes, love the see the resolver API in action

libs/code/deepagents_code/extensions/settings.py
```diff
@@ -0,0 +1,115 @@
...
+    enabled_option = get_option("extensions.enabled")
+    trust_option = get_option("extensions.trust")
+    if enabled_option is None or trust_option is None:
+        msg = "Extension options are missing from the config manifest"
+        raise RuntimeError(msg)
+    enabled = cast("bool", resolver.get(enabled_option).value)
```
aolsenjazz: Mannn this is a disappointing API. in typescript the return type would be auto inferred via associating the `extensions.enabled` key with the specified generic type at registry construction time

I guess this is more of a question for @mdrxy - is such an API not possible in python or are there impl constraints blocking this implementation?

libs/code/deepagents_code/extensions/settings.py
aolsenjazz: Ok looks like this is related to an extension config filed in the config resolver right?

libs/code/deepagents_code/extensions/settings.py
```diff
@@ -0,0 +1,142 @@
...
+        return {}
+    except (OSError, tomllib.TOMLDecodeError):
+        logger.warning(
+            "Could not read extensions config from %s",
+            DEFAULT_CONFIG_PATH,
+            exc_info=True,
+        )
+        return {}
+    section = data.get("extensions")
+    return section if isinstance(section, dict) else {}
```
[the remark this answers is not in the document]
aolsenjazz: @johannes117 I think this feedback is good. is there good reason to bypass the resolver? ideally the resolver is the SSoT for config settings

aolsenjazz (commented):
Submitting a few comment before a meeting. plenty more review to do

libs/code/deepagents_code/extensions/trust.py
```diff
@@ -0,0 +1,59 @@
...
+    """Return whether extension execution is trusted for a project.
+
+    Args:
+        project_root: Project root to inspect.
+        store_path: Alternate extension trust store.
+
+    Returns:
+        Whether the canonical project root has a persisted grant.
+    """
+    return is_project_hooks_trusted(
```
aolsenjazz, line 36: Reusing the `is_project_hooks_trusted` function without refactoring out a dedicated project trust system doesn't feel _great_ to me

Depending on the lift, I'd rather do this sooner than later to keep systems more pure

libs/code/deepagents_code/extensions/trust.py
```diff
@@ -0,0 +1,59 @@
...
+        project_root,
+        store_path=store_path or _store_path(),
+    )
+
+
+def trust_project_extensions(
```
aolsenjazz, line 42: Yeah idk there's just something about wrapping the trust system of project extensions around the trust system of hooks which smells just a little too much for me

aolsenjazz: Maybe a standalone project trust system is the psuedo-trust-system-consolidation we touch on a couple weeks ago

libs/code/deepagents_code/extensions/runtime.py
```diff
@@ -0,0 +1,183 @@
...
+            apis.append(api)
+    return ExtensionLoadResult(
+        registry,
+        tuple(errors),
+        active=True,
+        _apis=tuple(apis),
+    )
+
+
+def bind_server_extensions(extensions: ExtensionLoadResult) -> None:
```
aolsenjazz, line 131: this function seems more load bearing than it is - would you just document that this is only used for a test? or otherwise move this to a test/test util file?

libs/code/deepagents_code/extensions/runtime.py
```diff
@@ -0,0 +1,183 @@
...
+    """Retain the server-owned extension runtime for lifespan teardown."""
+    global _server_extensions  # noqa: PLW0603
+    _server_extensions = extensions
+
+
+def server_extension_report() -> dict[str, object]:
```
aolsenjazz, line 137: unused? or maybe used in a later PR?

libs/code/deepagents_code/extensions/runtime.py
aolsenjazz, line 1: Looks like there's some extra plumbing in this file re. server extensions that is only used for tests - are these units of code used in later in the PR stack? or can we move out of this file (if possible) or at least clearly demarcate testing code from production?

libs/code/deepagents_code/extensions/runtime.py
```diff
@@ -0,0 +1,183 @@
...
+    """Registrations and isolated errors from one load pass."""
+
+    registry: ExtensionRegistry = field(default_factory=ExtensionRegistry)
+    errors: tuple[str, ...] = ()
+    active: bool = False
+    """Whether at least one authorized extension source activated the runtime."""
+    _apis: tuple[ExtensionAPI, ...] = field(default=(), repr=False)
+
+
+_server_extensions: ExtensionLoadResult | None = None
```
aolsenjazz, line 40: Maybe I'll have a complete understanding of this when I'm done reading - what is the distinction between a server extension and a not-server extension? looks like this stuff is mostly just used in tests rn?

libs/code/deepagents_code/extensions/loader.py
```diff
@@ -0,0 +1,110 @@
...
+from deepagents_code.extensions.registry import ExtensionError
+
+if TYPE_CHECKING:
+    from collections.abc import Awaitable, Callable
+    from pathlib import Path
+
+    from deepagents_code.extensions.registry import ExtensionRegistry, SourceInfo
+
+
+def _module_name(path: Path) -> str:
```
aolsenjazz: nit: _module_name doesn't really encapsulate the work being done

nonblocking/nbd tho

libs/code/deepagents_code/extensions/loader.py
```diff
@@ -0,0 +1,110 @@
...
+        else None,
+    )
+    if spec is None or spec.loader is None:
+        msg = f"Could not import extension {source.path}"
+        raise ExtensionError(msg)
+    module = importlib.util.module_from_spec(spec)
+    sys.modules[name] = module
+    try:
+        spec.loader.exec_module(module)
+    except (KeyboardInterrupt, SystemExit, Exception) as exc:
```
aolsenjazz, line 45: I'm not really sure what the consideration/concern here is but it's interest to see a `KeyboardInterrupt` in the catch - do you think this is the right elevation at which to handle interrupts?

aolsenjazz: oop looks like you have the correct elevation later in the file

libs/code/deepagents_code/extensions/trust.py
aolsenjazz: actually fast follow is more appropriate

libs/code/deepagents_code/_server_config.py
aolsenjazz, line 1: Difficult to glean why we need to expose this to the server. I'm certain there's a reason - would you mind a sentence on this?

aolsenjazz (approved):

libs/code/deepagents_code/extensions/settings.py
johannes117: yes

libs/code/deepagents_code/extensions/settings.py
johannes117: I've combined into extra_paths which supports files and paths

libs/code/deepagents_code/extensions/settings.py
johannes117: resolving this one as I don't think there is anything actionable for this pr

libs/code/deepagents_code/extensions/trust.py
johannes117: leaving for now for simplicity, can fast follow later in a separate pr

libs/code/deepagents_code/extensions/trust.py
johannes117: agree, I think we should definitely do this, but again deferring until later for now

libs/code/deepagents_code/extensions/runtime.py
johannes117: this is actually wired into the production server lifecycle in the next stack layer, bit of an awkward split apologies

libs/code/deepagents_code/extensions/runtime.py
johannes117: yes later PR sorry

libs/code/deepagents_code/extensions/runtime.py
johannes117: apologies this was an awkward split, these are used in the next layer in the stack, I'll try to prevent splitting like this in the future

libs/code/deepagents_code/extensions/runtime.py
johannes117: again this would make more sense in the follow up PR but basically _server_extensions is the particular result owned by the persistent agent server, and ExtensionLoadResult is a generic loaded extension runtime owned by its caller. 
this allows a caller to shut it down later using shutdown_extensions(result)

libs/code/deepagents_code/extensions/loader.py
johannes117: renamed to _extension_module_name

libs/code/deepagents_code/_server_config.py
johannes117: the server loads the extensions so we need to pass it the user's approval and selected extension paths

PR #5633, opened 2026-08-19
feat(code): host Python extensions in the agent runtime

johannes117:
Related #5556

Hosts extension tools, middleware, and virtual filesystem backends in the agent runtime.

---

Layer 3 of 4. Static registrations are applied during graph construction, late tools remain callable, and late graph-bound registrations signal that a restart is required. Backend routes are validated against internal and sandbox boundaries, shell execution remains on the default backend, and the persistent server lifespan owns teardown and a loopback-only provenance endpoint. User-facing trust and inspection UX are deliberately excluded.

## Stack


libs/code/deepagents_code/server_lifespan.py
```diff
@@ -0,0 +1,54 @@
...
+    """Release extensions before their server event loop closes."""
+    try:
+        yield
+    finally:
+        from deepagents_code._env_vars import EXPERIMENTAL, is_env_truthy
+
+        if is_env_truthy(EXPERIMENTAL):
+            from deepagents_code.extensions.runtime import shutdown_server_extensions
+
+            await shutdown_server_extensions()
```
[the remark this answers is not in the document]
johannes117: Fixed in 11cdecf1d. Lifespan teardown is now unconditional and remains safe when no extension runtime was bound; the regression test clears the experimental flag during the lifespan and verifies shutdown still runs.

libs/code/deepagents_code/extensions/hosting.py
```diff
@@ -0,0 +1,170 @@
...
+class ExtensionRuntimeMiddleware(AgentMiddleware):
+    """Expose tools registered after the agent graph was built."""
+
+    name = "__deepagents_extension_runtime__"
+
+    def __init__(self, registry: ExtensionRegistry) -> None:
+        """Bind dynamic model and tool dispatch to `registry`."""
+        self._registry = registry
+
+    def _tools(
```
aolsenjazz, line 47: I'm a little surprised by this delivery mechanism for tools - I would expect that we would add tools at graph creation time (vis-a-vis the tools array passed into create_deep_agent)

Is this to support lazy (mid-runtime) registration?

libs/code/deepagents_code/extensions/hosting.py
aolsenjazz, line 1: ordinarily I might be concerned about SRP for this file, but we're ripping it

aolsenjazz (approved):

libs/code/deepagents_code/extensions/hosting.py
johannes117: yes this is to support mid-runtime registration

PR #5649, opened 2026-08-19
feat(code): enforce configured model allowlists

mdrxy:
`dcode` can now be restricted to an approved set of models. Add a `models.allowed` list to `config.toml` (or push it via managed config) and every model the app tries to use — the launch default, `--model` flag, `/model` selections, saved defaults, and subagent model declarations — is checked against it. Entries are exact `provider:model` specs or `provider:*` wildcards that admit a whole provider's lineup. Unset means all models are allowed; an empty list means none are.

---

**Why:** organizations rolling out `dcode` need to guarantee agents only run on approved models (cost, compliance, data routing). Previously any model with a credential was usable, and there was no enforcement point.

**How:**

- A new `models.allowed` config key (user or managed layer) parses into an ordered, deduplicated allowlist of exact `provider:model` specs and `provider:*` wildcards. A malformed list fails closed (deny-all) with an error naming the defect, so a typo can't silently disable the guardrail.
- Every model-resolution path is gated: CLI flags, the model selector (blocked entries are hidden from recommendations; typing a blocked spec explains the policy and lists what's allowed), saved defaults (a blocked stored default is ignored with a log, not advertised as `(default)`), and subagent declarations (the error names the declaring file).
- Bare model names are canonicalized the same way `create_model` resolves them before the policy check, so `gpt-5.6-terra` matches `openai:gpt-5.6-terra` — or an `openai:*` wildcard — instead of being spuriously rejected; a name whose provider can't be established stays unmatchable, so inference is never a bypass.
- A `provider:*` wildcard admits that provider's whole lineup where an exact list would prune unlisted models. The wildcard names no model itself, so default resolution expands it to the provider's discovered models — the registry profile lineup merged with any configured list — rather than selecting it literally; a wildcard for a provider with no discovered or configured models fails closed with a "no discoverable models" message instead of a phantom credential prompt.
- Blocked selections raise `ModelNotAllowedError`, whose message names the policy layer (user file vs. administrator-managed), quotes the offending spec, and lists the allowed entries — instead of the previous generic I/O-style failure. When the allowlist is active but no allowed model has credentials, a distinct `NoAllowedModelCredentialsError` keeps `/auth` from accepting a key and appearing to do nothing.

Edge cases covered: remote no-auth providers (e.g., a LAN Ollama endpoint) remain valid allowlist fallback candidates; bare Bedrock IDs must be written `bedrock:<id>` in the list since they otherwise split at the version colon and could never match; `dcode tools list` skips enforcement because it only compiles graphs it never invokes.


mdrxy:
This should be merged after rebasing on #5755

PR #5659, opened 2026-08-19
fix(code): make tool arg validation errors recoverable

mdrxy:
A model that passes malformed arguments to `ask_user` (for example an empty questions list, or a choice with a blank value) previously crashed the whole run: the tool raises `ValueError`, `ToolNode` never converted it, and the error bubbled up as a fatal `Agent error`. The model never saw the failure and got no chance to reissue corrected arguments, so the user's turn dead-ended. Now those validation errors surface to the model as an error `ToolMessage` it can read and fix in one retry, and the run continues.

---

This wires langchain's `ToolErrorMiddleware` (available since `langchain>=1.3.14`; this package already requires `>=1.3.15`) into the agent's middleware stack, scoped to `ask_user` — the tool that validates model-authored arguments by raising `ToolArgumentError`, a `ValueError` subclass introduced here so recovery keys off intent rather than a shared built-in type. (`read_file` is deliberately out of scope: it already catches its own argument errors and returns an error `ToolMessage`, and its remaining `ValueError`s are backend invariants that must stay fatal.) The `on_error` handler returns a message naming the tool and the validation detail (which the existing error text already carries) for `ToolArgumentError`, and returns `None` for everything else so unexpected errors still propagate and halt the run. `ToolErrorMiddleware` re-raises LangGraph control-flow signals (`GraphBubbleUp` / interrupts) unchanged, so `ask_user`'s `interrupt()`-based flow is unaffected.

The tool itself still raises; only the run-level handling changes. This is the approach decided externally and discussed in langchain-ai/langchain#38781.


libs/code/deepagents_code/ask_user.py
```diff
@@ -110,14 +111,14 @@ def _validate_choices(
...
             msg = (
                 f"multi_select question {question_text!r} has a choice value "
                 f"containing {MULTI_SELECT_FORBIDDEN_IN_VALUE!r}, which would "
                 f"make the joined answer ambiguous: {value!r}"
             )
-            raise ValueError(msg)
+            raise ToolArgumentError(msg)
 
 
 def _validate_questions(questions: list[Question]) -> None:
```
sydney-runkle, line 124: shouldn't the burden be on pydantic to validate the Question type, we shouldn't need a custom error?

mdrxy: #5682

Issue #1359, opened 2026-02-17
Fork session / rewind command


mdrxy:
### Privileged issue

- [x] I am a Deep Agents maintainer.

### Issue Content

Unified spec for conversation rewind and session forking — two commands, shared infrastructure. Subsumes #1526 (credit @Illuminated2020 for the architectural research there).

**Core idea**: a checkpoint-aware turn picker that powers two write modes:
- `/rewind` — destructive rollback within the current thread
- `/fork` — non-destructive branch into a new thread

---

#### Trigger

| Surface | Command | Notes |
|---------|---------|-------|
| Slash command | `/rewind`, `/fork` | Opens shared interactive picker |
| Keybind | Configurable (TBD) | Same as `/fork` |
| CLI subcommand | `deepagents fork [--last] [--at <turn>] [--blank]` | `--last` = fork most recent session; `--at` = specify turn number; `--blank` = empty editor (no pre-fill) |

#### Picker

- Modal screen (same pattern as `ThreadSelectorScreen`)
- Displays all turns (user + assistant) with:
  - Turn number
  - Role indicator (user / assistant)
  - Message preview (truncated)
  - Timestamp
- Arrow-key navigation, enter to select
- User messages are the primary "decision points" but assistant turns are selectable too (e.g., "that tool call was wrong, redo from here")

#### State

- **Rewind (`/rewind`)**: rolls checkpoint back to selected turn, discards everything after. Confirmation prompt: _"Discard N messages? (y/n)"_ — single keystroke since `/fork` exists as the safe alternative.
- **Fork (`/fork`)**: copies full history up to selected turn into a new thread. Inherits model config, MCP servers, and agent name. By default, pre-fills the editor with the original user message at the fork point; `--blank` flag for empty editor.

#### Storage & relationships

- Forked threads store `parent_thread_id` + `fork_checkpoint_id` in thread metadata (enables navigation)
- Unlimited tree depth (fork a fork is fine)
- Deletion of parent thread **orphans** children (they become root-level threads)

#### Navigation

- **Sibling-cycle**: keybinds to move between sibling forks sharing the same parent + fork point
- **Parent-return**: keybind to jump back to the parent thread at the fork point
- **`/threads`**: flat list with "forked from {parent}" badge (v1); tree indentation possible in future

#### Signal

- Status bar shows `"Forked from {parent_thread_id}"` when in a child thread
- Toast notification on fork creation

#### Extension points

[...]

PR #5714, opened 2026-08-21
feat(sdk): add subagent conversation forking

### LINKED ISSUE #4668: Add `fork` option to the `task` tool so subagents can inherit parent conversation history

thushanth-bengre-langchain:
Closes #4668.

Adds `ForkedSubAgent`, a subagent spec that inherits the parent's effective conversation history and exact system prompt instead of only seeing the task description — useful when a subagent is delegated deep into an investigation and shouldn't have to re-derive context the parent already gathered.

`mode="fork"` is required rather than optional on `ForkedSubAgent`, since `SubAgent.system_prompt` is also optional — making both optional would let the two specs collapse into the same shape and get silently misrouted between forking and isolated behavior.

Reconstructs history using the same summarization-cutoff logic as isolated subagents (now shared via a common helper), drops the parent's still-unresolved trailing tool-call message so the forked prefix matches what was actually cached, and mirrors in a fresh memory middleware instance when `memory=` is configured. Experimental, flagged as such in the docstrings.

Note: #4668 proposed a per-call `fork: bool` flag on the `task` tool; this implements the capability as a spec-level `ForkedSubAgent` type instead. A per-call flag can be layered on as a follow-up if the tool-level ergonomics are wanted.


libs/deepagents/deepagents/graph.py
```diff
@@ -682,6 +701,14 @@ class MyState(DeepAgentState):
             subagent_middleware.extend(_subagent_profile.materialize_extra_middleware())
 
             append_prompt_caching_middleware(subagent_middleware)
+            if is_forked and memory is not None:
+                subagent_middleware.append(
+                    MemoryMiddleware(
+                        backend=backend,
+                        sources=memory,
+                        add_cache_control=True,
+                    )
```
hntrl: Is it possible to reuse this middleware instead of recreating?

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -243,6 +280,35 @@ def node(state):
...
+
+def _validate_subagent_mode(spec: _SubAgentSpec) -> None:
+    """Reject unsupported context modes before a subagent can run."""
+    mode = spec.get("mode")
+    if mode not in (None, "handoff", "fork"):
+        msg = f"SubAgent '{spec['name']}' has invalid mode '{mode}'; expected 'handoff' or 'fork'"
+        raise ValueError(msg)
+
+
+def _is_forked_subagent(spec: _SubAgentSpec) -> bool:
```
hntrl: can/ should use TypeGuard here

libs/deepagents/deepagents/graph.py
```diff
@@ -731,16 +758,20 @@ class MyState(DeepAgentState):
...
+            if not is_forked:
+                processed["system_prompt"] = _apply_profile_prompt(_subagent_profile, spec.get("system_prompt", ""))
             if subagent_interrupt_on is not None:
-                processed_spec["interrupt_on"] = subagent_interrupt_on
-            inline_subagents.append(processed_spec)
+                processed["interrupt_on"] = subagent_interrupt_on
+            if is_forked:
+                inline_subagents.append(cast("ForkedSubAgent", processed))
+            else:
+                inline_subagents.append(cast("SubAgent", processed))
```
hntrl, line 788: Don't love these type casts

mdrxy:
Closes #4668 — with a shape difference worth noting:

- **#4668 proposed:** per-call `fork: bool` on the `task` tool, gated by an `allow_fork` middleware flag.
- **This PR:** a separate `ForkedSubAgent` spec with required `mode="fork"` — forking is a property of the subagent definition, not a per-call flag. `mode` is required so the two specs can't collapse into the same shape and get silently misrouted.

Same underlying capability: forked subagent gets the parent's effective history (same summarization-cutoff logic as isolated subagents) plus the task description. Isolated-by-default is unchanged.

@NinaadRao — if the per-call `fork=` flag itself matters beyond the capability, that can be layered on as a follow-up.

Unaffected related issues: #1359, #2440, #3838, #2512.

(Also editing the PR description to include the `Closes #4668` line.)

mdrxy (changes requested):
* Forks probably shouldn't run prompt-producing middleware. Today a fork builds its middleware stack like any other subagent. `MemoryMiddleware`, `SkillsMiddleware`, and any profile extras each append to the system prompt. But then `_ForkSystemMessageMiddleware` runs last and replaces the whole thing with the captured parent message. So that work happens on every call and is thrown away every time. Since a fork's prompt is definitionally the parent's captured prompt, I think we can skip prompt-producing middleware at fork construction and set the captured message as the fork's static prompt?
* A forked spec with `system_prompt` should raise, not silently drop it. The `Never` annotation only helps typed users; at runtime `_resolved_declarative_spec` filters the key out, so `ForkedSubAgent(system_prompt="You are a SQL expert...")` runs without complaint and the prompt is silently discarded (the kind of misrouting `required mode="fork"` seems to be designed to prevent). A `ValueError` in `_validate_subagent_mode` (or a sibling validator) would make bad config loud.
* Reject duplicate subagent names. The context leak Open SWE flagged is currently avoided only because `subagents_by_name` and `subagent_graphs` both happen to resolve duplicates last-write-wins. That agreement is incidental: a `ValueError` on duplicate names removes the whole class of bug instead of relying on two maps staying in lockstep.

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -243,6 +280,35 @@ def node(state):
...
+    """Return whether a subagent inherits parent conversation history."""
+    return spec.get("mode") == "fork"
+
+
+def _fork_messages(state: Mapping[str, object], description: str) -> list[AnyMessage]:
```
hntrl: i think we don't have to do `cast` in here if this is typed properly

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -530,13 +658,19 @@ def _validate_and_prepare_state(
...
         description: str,
         runtime: ToolRuntime,
+        effective_system_message: SystemMessage | None,
     ) -> tuple[Runnable, dict]:
         """Prepare state for invocation."""
         subagent = _select_subagent(subagent_type, runtime)
-        # Create a new state dict to avoid mutating the original
-        subagent_state = {k: v for k, v in runtime.state.items() if k not in _EXCLUDED_STATE_KEYS}
+        forked = subagent_type in fork_mode_names
+        subagent_state = {k: v for k, v in runtime.state.items() if k not in _EXCLUDED_STATE_KEYS and k != _PARENT_SYSTEM_MESSAGE_KEY}
```
hntrl: instead of adding new condition just add to `_EXCLUDED_STATE_KEYS`

libs/deepagents/tests/unit_tests/middleware/test_subagent_middleware_init.py
```diff
@@ -107,6 +115,183 @@ def test_public_init_type_hints_are_runtime_resolvable(self) -> None:
         assert "state_schema" in create_hints
         assert "return" in create_hints
 
+    def test_forked_subagent_is_exported_with_required_mode(self) -> None:
+        assert ExportedForkedSubAgent is ForkedSubAgent
+        assert "mode" in ForkedSubAgent.__required_keys__
+        assert "system_prompt" in ForkedSubAgent.__optional_keys__
+
```
hntrl:
```suggestion
```

thushanth-bengre-langchain:
> * Forks probably shouldn't run prompt-producing middleware. Today a fork builds its middleware stack like any other subagent. `MemoryMiddleware`, `SkillsMiddleware`, and any profile extras each append to the system prompt. But then `_ForkSystemMessageMiddleware` runs last and replaces the whole thing with the captured parent message. So that work happens on every call and is thrown away every time. Since a fork's prompt is definitionally the parent's captured prompt, I think we can skip prompt-producing middleware at fork construction and set the captured message as the fork's static prompt?
> * A forked spec with `system_prompt` should raise, not silently drop it. The `Never` annotation only helps typed users; at runtime `_resolved_declarative_spec` filters the key out, so `ForkedSubAgent(system_prompt="You are a SQL expert...")` runs without complaint and the prompt is silently discarded (the kind of misrouting `required mode="fork"` seems to be designed to prevent). A `ValueError` in `_validate_subagent_mode` (or a sibling validator) would make bad config loud.
> * Reject duplicate subagent names. The context leak Open SWE flagged is currently avoided only because `subagents_by_name` and `subagent_graphs` both happen to resolve duplicates last-write-wins. That agreement is incidental: a `ValueError` on duplicate names removes the whole class of bug instead of relying on two maps staying in lockstep.


- Fixed — forks now skip building `MemoryMiddleware`/`SkillsMiddleware` at construction instead of computing then discarding them.
- Fixed — a forked spec with `system_prompt` set now raises `ValueError` in `_validate_subagent_mode`
- Fixed — duplicate subagent names now raise ValueError before anything gets compiled

libs/deepagents/deepagents/graph.py
thushanth-bengre-langchain: Fixed — that block (`mirrored MemoryMiddleware`) is gone entirely, so there's nothing left to recreate or reuse.

libs/deepagents/deepagents/middleware/subagents.py
thushanth-bengre-langchain: Used `TypeIs[ForkedSubAgent]` instead of `TypeGuard`

libs/deepagents/deepagents/graph.py
thushanth-bengre-langchain: Leaving as-is. We are using type cast elsewhere in this file. Fixing this would be a much bigger change. Don't see any functional improvements by changing it.

libs/deepagents/deepagents/middleware/subagents.py
thushanth-bengre-langchain: Fixed — folded` _PARENT_SYSTEM_MESSAGE_KEY` into `_EXCLUDED_STATE_KEYS`

libs/deepagents/tests/unit_tests/middleware/test_subagent_middleware_init.py
thushanth-bengre-langchain: Removed the test

libs/deepagents/deepagents/middleware/subagents.py
thushanth-bengre-langchain: Removing this cast would mean generically typing ToolRuntime throughout task/atask, too big a change for a type-only nicety. so leaving this as it is.

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -243,6 +283,57 @@ def node(state):
...
+def _is_forked_subagent(spec: _SubAgentSpec) -> TypeIs[ForkedSubAgent]:
+    """Return whether a subagent inherits parent conversation history."""
+    return spec.get("mode") == "fork"
+
+
+def _fork_messages(state: Mapping[str, object], description: str) -> list[AnyMessage]:
+    """Build fork input without replaying the in-flight tool-call message."""
+    messages = list(cast("Sequence[AnyMessage]", state.get("messages", [])))
+    if messages and isinstance(messages[-1], AIMessage) and messages[-1].tool_calls:
+        messages.pop()
```
ccurme: does this mutate in-memory mesages

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -164,6 +170,40 @@ class Findings(BaseModel):
...
+
+    A forked subagent receives the parent's effective conversation history and
+    exact system prompt. It cannot define a separate `system_prompt` or
+    `skills`, since either would diverge from the parent's exact message.
+    """
+
+    mode: Literal["fork"]
+    """Required discriminator that enables conversation forking."""
+
+    system_prompt: NotRequired[Never]
```
ccurme: (nit) why is this needed

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -164,6 +170,40 @@ class Findings(BaseModel):
...
+    """
+
+    system_prompt: NotRequired[str]
+    """Instructions for the subagent. Uses an empty prompt when omitted."""
+
+    mode: NotRequired[Literal["handoff"]]
+    """Context mode. Declarative `SubAgent` instances are always isolated."""
+
+
+class ForkedSubAgent(_SubAgentBase):
```
ccurme: Regarding the supervisor model deciding forked or not: the PR description says
> A per-call flag can be layered on as a follow-up if the tool-level ergonomics are wanted

what do our types look like in that case? do we still have `ForkedSubAgent` or would it collapse into one type?

libs/deepagents/deepagents/middleware/subagents.py
thushanth-bengre-langchain: Not strictly needed, we can remove this.

libs/deepagents/deepagents/middleware/subagents.py
thushanth-bengre-langchain: Depends on if we want to allow users to still configure 3 different subagents(handoff, fork, dynamic).
 If yes, we need the `ForkedSubAgent` since users will be configuring that type.

But if we go the direction of letting the model decide if the subagent should be forked or not depending on that task (the users do not configure this), then I can see this collapsing into one type.

libs/deepagents/deepagents/middleware/summarization.py
```diff
@@ -778,39 +804,8 @@ def _apply_event_to_messages(
...
-                cutoff_idx,
-                len(messages),
-            )
-            return [summary_msg]
-
-        result: list[AnyMessage] = [summary_msg]
-        result.extend(messages[cutoff_idx:])
-        return result
+        """Reconstruct effective messages from raw state and a summary event."""
+        return _apply_summarization_event(messages, event)
```
mdrxy: why helper?

mdrxy (approved):

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -166,6 +183,44 @@ class Findings(BaseModel):
...
+
+    A forked subagent receives the parent's effective conversation history and
+    exact system prompt. It cannot define a separate `system_prompt` or
+    `skills`, since either would diverge from the parent's exact message.
+
+    `tools` isn't restricted the same way -- a forked subagent's own tools
+    work normally. The tradeoff is cache cost, not correctness: on
+    exact-match caching providers (confirmed for Anthropic), any tools
+    difference invalidates the whole cached prefix, so a fork with its own
+    tools generally can't share the parent's cache there. Context
```
ccurme: we explicitly prohibit separate system messages. why do we permit different sets of tools?

ccurme (approved):

hntrl (approved):

libs/deepagents/deepagents/middleware/subagents.py
```diff
@@ -534,18 +735,41 @@ def _validate_and_prepare_state(
...
+        subagent = subagents_by_name[subagent_type]
+        subagent_runnable = _select_subagent(subagent_type, runtime)
+
+        if _is_forked_subagent(subagent) or _is_forked_compiled_subagent(subagent):
+            # The event is folded into the message list above to support compiled subagents
+            if _is_forked_subagent(subagent):
+                # A declarative fork runs the same graph shape as its parent, so
+                # private channels carry over and its own middleware can rebuild
+                # what the parent built from them.
+                inherited = {key: value for key, value in runtime.state.items() if key not in _FORK_EXCLUDED_STATE_KEYS}
```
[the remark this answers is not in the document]
hntrl: This is intentional -- we added PrivateStateAttr specifically as a marker to say "don't propagate this value to subagents". Forked subagents are meant to be a contextual copy of the parent agent which means we need to retain this information

PR #5731, opened 2026-08-21
feat(code): add advisory project `.env` trust prompt with per-project skip

mdrxy:
When you launch `dcode` interactively in a directory that has a project `.env` you haven't seen before, it now tells you the file will be loaded and lets you change that — without blocking the launch. A short prompt notes that `.env` values reach subprocesses (and can run code via shell startup hooks or git config) before any approval prompt, then offers: **Continue (load .env)** (the default — Enter, Esc, or Ctrl+C all load it), **Don't load .env this session**, or **Never load .env in this project**. "Never" remembers the choice and skips the file silently on future launches, including from subdirectories of the directory that owns the file. Headless, piped-stdin, and subcommand launches are never prompted and load the file normally.

---

The `.env`→process-env path is the trust boundary tracked as T12 in `libs/code/THREAT_MODEL.md`: a cloned repo's `.env` is attacker-controlled input that lands in `os.environ` and reaches the startup `git` detection and `execute` shells. #4288 and #5723 denylisted known execution-hook keys, and #5726 added a hard opt-out (`read_project_dotenv=false`). This adds the *interactive* layer: a visible, advisory confirmation so a user launching in an untrusted tree is told the file exists and can decline it, rather than loading it silently.

Design decisions:

- **Opt-out, never a gate.** The default action (Enter, Esc, Ctrl+C, Ctrl+D) loads the file, so the prompt can only *reduce* trust. It runs before the settings bootstrap (which loads the project `.env` at the `config.py` settings import) so the decision applies to the current launch. To support a non-deny default, the shared pre-TUI picker gained a `default_action` parameter (existing callers keep the deny-first, fail-closed default).
- **Headless is untouched.** Gated on `interactive_tui and _trust_picker_has_terminal()`, matching the dep-floor prompt's reasoning: piped-stdin launches mount the TUI but have no TTY for the picker, and non-interactive/subcommand launches must never block. Those load the file normally; the deliberate suppression path there is the #5726 flag.
- **Keyed to the `.env`'s parent directory.** Skip decisions are stored against the directory that actually owns the discovered `.env` (`skip_key_for_start_path`), not the invocation directory or `ProjectContext.project_root`. A non-Git project has no `project_root`, so keying on `user_cwd` would store the launch subdirectory and miss an ancestor `.env` on a later launch from the root or a sibling. Since `_find_dotenv_from_start_path` walks up from cwd, a subdirectory launch resolves to the same key and is covered.
- **Preview and runtime agree.** `_preview_dotenv_environ` consults the same store as `_load_dotenv`, so the cwd-switch `/reload` preview and tracing diagnostics never report settings from a `.env` that startup skipped.
- **Own store, independent of hooks.** `dotenv_skip.py` (`~/.deepagents/.state/dotenv_skip.json`) mirrors `hooks/trust.py` (versioned JSON, atomic write, `FileLock` + threading lock) but is separate — skipping a `.env` is an independent decision from trusting hooks.
- **Two independent skip sources, explicit flag wins.** `_load_dotenv` consults both `resolve_read_project_dotenv()` (the #5726 managed/env/global/config.toml opt-out) and the persisted skip store. Either one skips; neither forces a load the other disabled.
- **Skip mechanism.** The prompt's session/persistent "skip" answers set `DEEPAGENTS_CODE_READ_PROJECT_DOTENV=0` for the process (denied from every `.env`, so no dotenv file can re-enable it) and, for "never", persist the key. The loader's store check covers future headless launches where the prompt isn't shown.
- **Terminal-injection hardening.** The path shown in the prompt and the "will be skipped" confirmation are run through `sanitize_control_chars` (the helper already used in approval dialogs) before Rich markup escaping, so a crafted directory name carrying ANSI escapes or BiDi/zero-width Unicode cannot spoof which `.env` the prompt is about. Only the displayed strings are sanitized; the persisted key keeps the real path so lookups still match.

Tests cover the skip store (persistence, canonical keying, corrupt/unsupported-version fail-closed, missing store), the skip key following the `.env` parent across subdirectory launches (including non-Git projects), the loader and preview both consulting the store, and the existing trust-picker deny-default preserved. The full main / ACP / reload / config / dotenv-skip / hooks-trust suites pass; lint and type checks clean.


mdrxy:
May re-visit

Issue #5030, opened 2026-07-24
Upstream `ReliableRubricMiddleware` hooks into `RubricMiddleware`


mdrxy:
## Feature description

Upstream the dcode-specific pieces of `ReliableRubricMiddleware` into SDK `RubricMiddleware` so apps can plug the generic grading engine into their own tools, runtime context, transcript hygiene, and reliability needs without subclassing.

### Why this exists today

SDK `RubricMiddleware` already owns the product core:

> When someone hands me a checklist, grade the agent’s work against it, and maybe send the agent back if it’s not good enough.

Dcode still wraps it because the app grader is not a bare model with a couple tools. The nested grader needs dcode-specific integration that the SDK constructor/runtime does not expose:

1. **Let the grader inspect the repo safely** — dcode middleware, budgets, path rules, and runtime context. The SDK grader builder does not take `grader_middleware` / `grader_context_schema`, and does not pass outer runtime `context` into the nested grader.
2. **Don’t grade on secret internal chat notes** — strip goal/rubric control messages before the grader reads transcript evidence.
3. **If the grader’s network hiccups mid-response, try grading once more** — one retry of the grader only, never the whole coding turn; only mid-response transport failures.
4. **If the grader pauses for approval, bubble that up cleanly** — re-raise nested `GraphBubbleUp` instead of collapsing it into `grader_error`.

In plain terms:

- **SDK** = generic grading engine
- **dcode wrapper** = plug that engine into dcode’s tools, safety rails, UI/control-message world, and reliability needs

### Desired outcome

Most of the wrapper becomes stock SDK surface. Dcode keeps app policy (tool lists, budgets, roots, control-message detector) and stops owning a parallel grading engine. Success when `ReliableRubricMiddleware` is gone or is a thin factory/shim.

### API the SDK should support by default

Optional constructor/runtime knobs (no behavior change when omitted):

- `grader_middleware=...`
- `grader_context_schema=...`
- Pass outer runtime `context` into nested `grader.invoke` / `ainvoke`
- A messages/state transform hook before grade (filter internal messages), e.g. `prepare_messages_for_grader` or `prepare_grader_state`
- Built-in one-shot transport retry for mid-response read/protocol faults only (or a small retryable-failure hook if a second consumer appears)
- Re-raise `GraphBubbleUp` from nested grader HITL/interrupts before treating other exceptions as `grader_error`
[...]

PR #5874, opened 2026-08-27
feat(sdk): add rubric grader integration hooks

mdrxy:
Related: #5030

`RubricMiddleware` now lets applications configure and prepare its nested grader agent without subclassing SDK internals.

---

`RubricMiddleware` grades the main agent by creating a second, nested agent. The existing API exposes that grader's model, prompt, and tools, but real applications also need the grader to participate in app-level policy:

- apply middleware such as tool-call budgets and path restrictions;
- receive typed runtime context used by those tools and middleware;
- remove app-internal control messages before they become grading evidence; and
- carry app-specific state, such as an operation ID used to scope a budget.

Without public hooks, consumers have to override private grader construction and input-building methods. That couples applications to SDK internals and makes transcript customization especially risky because the override must preserve the SDK's payload sanitization.

### Before

An application has to subclass `RubricMiddleware` and modify the nested grader input after delegating to private methods:

```python
class AppRubricMiddleware(RubricMiddleware):
    def _grader_input(self, state, iteration, correction=None):
        state = {
            **state,
            "messages": [
                message
                for message in state["messages"]
                if not is_internal_control_message(message)
            ],
        }
        grader_input = super()._grader_input(state, iteration, correction)
        grader_input["grading_operation_id"] = f"{state['_current_grading_run_id']}:{iteration}"
        return grader_input
```

It must also override nested-agent construction to install grader middleware, a context schema, and a state schema.

### After

The same integration is expressed through constructor options:

```python
rubric = RubricMiddleware(
    model=grader_model,
    tools=verification_tools,
    grader_middleware=[verification_budget, path_policy],
    grader_context_schema=AppContext,
    grader_state_schema=AppGraderState,
    prepare_messages_for_grader=lambda messages: [
        message
        for message in messages
        if not is_internal_control_message(message)
    ],
    build_grader_state=lambda state, iteration: {
        "grading_operation_id": (
            f"{state['_current_grading_run_id']}:{iteration}"
        )
    },
)
```

This keeps the responsibilities explicit:

- `grader_middleware`, `grader_context_schema`, and `grader_state_schema` configure the nested agent through the same `create_agent` surface used elsewhere.
- `prepare_messages_for_grader` filters a copy of the transcript before the SDK builds and sanitizes the grader payload; it does not mutate the outer agent state.
- `build_grader_state` adds schema-backed fields to the nested grader input, but cannot replace the SDK-owned `messages` field and bypass sanitization.

All new options are optional, so existing `RubricMiddleware` callers keep the same behavior. Invalid callbacks fail during construction, and custom grader state requires an explicit schema.

The follow-up in #5888 uses these hooks to remove Deep Agents Code's custom grader-input override while preserving its transcript filtering and verification-budget scoping.

Made by [Open SWE](https://openswe.vercel.app/agents/983cd60f-1eb6-57fd-9146-bb0078f08d1a)

### DIFF
--- a/libs/deepagents/deepagents/middleware/rubric.py
+++ b/libs/deepagents/deepagents/middleware/rubric.py
@@ -1004,7 +1004,7 @@ class RubricMiddleware(AgentMiddleware[RubricState, ContextT, ResponseT]):
         grader_state = state
         if self._prepare_messages_for_grader:
             grader_state = cast("RubricState", dict(state))
-            grader_state["messages"] = self._prepare_messages_for_grader(state.get("messages", []))
+            grader_state["messages"] = self._prepare_messages_for_grader(list(state.get("messages", [])))
         payload = self._build_grader_payload(grader_state, iteration, correction)
         grader_input = dict(self._build_grader_state(grader_state, iteration)) if self._build_grader_state else {}
         if "messages" in grader_input:
--- a/libs/deepagents/deepagents/middleware/rubric.py
+++ b/libs/deepagents/deepagents/middleware/rubric.py
@@ -24,6 +24,7 @@ from typing import (
     Any,
     Literal,
     NotRequired,
+    cast,
 )
 
 from langchain.agents import create_agent
@@ -520,6 +521,15 @@ class RubricMiddleware(AgentMiddleware[RubricState, ContextT, ResponseT]):
             `GraderResponse`.
 
             With none, the grader reasons from the transcript alone.
+        grader_middleware: Middleware applied to the nested grader agent.
+        grader_context_schema: Runtime context schema for the nested grader.
+        grader_state_schema: State schema for the nested grader. Use this when
+            `build_grader_state` adds custom state fields.
+        prepare_messages_for_grader: Optional transform applied to the transcript
+            messages before the SDK builds the sanitized grader payload.
+        build_grader_state: Optional callback that adds custom fields to the
+            nested grader input. It cannot replace the SDK-owned `messages`
+            field.
         max_iterations: Maximum grader iterations per rubric attempt; must be a
             positive integer.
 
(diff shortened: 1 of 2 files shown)


libs/deepagents/deepagents/middleware/rubric.py
```diff
@@ -560,12 +577,27 @@ def __init__(  # noqa: D107
...
+            if callback is not None and not callable(callback):
+                msg = f"RubricMiddleware: `{name}` must be callable."
+                raise TypeError(msg)
 
         self.max_iterations = max_iterations
         self._model = model
         self._model_label = _configured_model_label(model)
         self._system_prompt = system_prompt or GRADER_SYSTEM_PROMPT
         self._tools: list[BaseTool] = list(tools) if tools else []
+        self._grader_middleware = list(grader_middleware or ())
```
mdrxy: why not a sequence?

libs/deepagents/deepagents/middleware/rubric.py
```diff
@@ -966,8 +1001,17 @@ def _grader_input(
         Returns:
             The nested grader's input state.
         """
-        payload = self._build_grader_payload(state, iteration, correction)
-        return {"messages": [HumanMessage(content=payload)]}
+        grader_state = state
+        if self._prepare_messages_for_grader:
+            grader_state = cast("RubricState", dict(state))
```
mdrxy: is cast strictly needed?

mdrxy:
I considered having the nested rubric grader inherit the parent agent’s middleware by default, but chose an explicit param instead. The grader is a separately compiled agent with its own state, context, tools, prompt, and `GraderResponse`; parent middleware may depend on incompatible schemas or inject capabilities that should not cross this trust boundary automatically. Middleware instances may also hold mutable counters or caches that should not be shared between concurrent parent and grader runs.

This matches Deep Agents' broader behavior: declarative subagents do not inherit arbitrary parent middleware. Explicit configuration keeps the grader least-privileged and makes each inherited policy intentional.

PR #5882, opened 2026-08-27
fix(code): defer the recursion limit to the LangGraph server

mdrxy:
dcode no longer imposes its own graph step budget, and the configured budget now actually reaches the graph.

---

## Dropping the product default

dcode replaced the graph's built-in limit with `2000` on every launch. The unset path now binds nothing, so the runtime's own budget stands. `--recursion-limit`, `DEEPAGENTS_CODE_RECURSION_LIMIT`, `[runtime].recursion_limit`, and managed configuration keep their existing precedence and bounds.

The fallback is the `langgraph dev` server's `10011` (`langgraph_api/utils/config.py`), not LangGraph core's `10007` and not the Deep Agents SDK's `9999` — both are shadowed on the path dcode actually runs. `dcode config` reports it rather than showing an empty default:

```
runtime.recursion_limit    10011    default
```

## Making the setting take effect

The graph executes inside the `langgraph dev` server, which stamps its own `recursion_limit` onto every run config it builds. `merge_configs` only defers to the value bound onto the compiled graph when the incoming one equals LangGraph core's default:

```python
elif key == "recursion_limit":
    if config["recursion_limit"] != DEFAULT_RECURSION_LIMIT:  # 10007
        base["recursion_limit"] = config["recursion_limit"]
```

`10011 != 10007`, so the server value overrode the binding `create_cli_agent` applied. Both the TUI and headless runs reach the graph over `RemoteGraph`, so neither honored a configured limit — the old hardcoded `2000` was discarded the same way.

`recursion_limit` is in `langgraph_api`'s `CONFIG_KEYS`, so a client-supplied value replaces the server default. `build_stream_config` now carries the resolved limit, and omits the key when nothing is configured:

```
unconfigured  ->  key absent          (server default 10011 stands)
env=3000      ->  recursion_limit=3000
```

## Hardening

- `LANGGRAPH_DEFAULT_RECURSION_LIMIT` is denied from project `.env` files. It sets the upstream default, so a cloned repo could otherwise bypass the bounded resolver.
- `RECURSION_LIMIT_FLOOR = 25` keeps its rationale in the docstring, including that rejection *raises* the effective budget: a `config.toml` value of `20` is discarded and the run falls through to `10011`. Use the CLI flag, which is exempt and accepts `>= 1`, to bound a run below the floor.

Made by [Open SWE](https://openswe.vercel.app/agents/16dfbef1-efaa-512f-9a2d-438912145ead)


openwiki/runtime-behavior.md
```diff
@@ -84,26 +84,43 @@ approach it, or never come close? A ceiling production sits far below, a limit
...
-SDK's 9,999.** The resolver clamps any override to `[RECURSION_LIMIT_FLOOR=25,
-RECURSION_LIMIT_CEILING=100_000]`
-(`config_manifest.py`, L102-L115).
-
-*So what:* if you are debugging a `GRAPH_RECURSION_LIMIT` failure, the effective
-ceiling depends on which layer built the agent. Do not assume 9,999. The check
-to run on each sample: how close do the longest baseline/outlier runs get to
-2000 graph steps? If production never approaches it, the default is comfortable;
-if outliers cluster near it, that ceiling is a real hazard for long sessions.
+### The main-agent recursion limit is set in three places, and the server wins
```
mdrxy: don't change this file in this pr

libs/code/deepagents_code/ui.py
```diff
@@ -255,7 +255,7 @@ def show_help() -> None:
     )
     console.print(
         "  --recursion-limit N        Override the agent's graph recursion_limit"
-        " (default 2000)"
+        " (default: LangGraph's 10011)"
```
mdrxy: feel like we don't need to mention the value / default here

libs/code/deepagents_code/main.py
```diff
@@ -2647,7 +2650,8 @@ def help_parent(help_fn: Callable[[], None]) -> list[argparse.ArgumentParser]:
         metavar="N",
         help="Override the main agent's LangGraph recursion_limit (graph step "
         "budget; must be >= 1). Overrides DEEPAGENTS_CODE_RECURSION_LIMIT and "
-        "[runtime].recursion_limit; defaults to 2000.",
+        "[runtime].recursion_limit; defaults to the LangGraph server budget "
+        "(10011).",
```
mdrxy: let's also not mention here

libs/code/deepagents_code/config_manifest.py
```diff
@@ -100,20 +100,20 @@
...
+`25` is the LangChain-core runnable default, the smallest budget any LangGraph
+run is expected to survive. A value below it would break otherwise-valid runs,
+so a resolved value under the floor is rejected and falls through to the next
+layer. The `--recursion-limit` flag is exempt and accepts any value `>= 1`
+(`_is_valid_cli_recursion_limit`): an operator who types a small budget on the
+command line is bounding one run on purpose.
+
+Rejection raises the effective budget rather than lowering it. A `config.toml`
+value of `20` is discarded, and the run falls through to the server default
+(`10_011`), so the user gets a far larger budget than they asked for. Use the
```
mdrxy: magic number let's not mention upstream default

libs/code/deepagents_code/config.py
```diff
@@ -2233,11 +2242,33 @@ def build_stream_config(
...
+    # server value overrides the binding `create_cli_agent` applied -- the
+    # resolved limit never took effect on this path. `recursion_limit` is in
+    # `langgraph_api`'s `CONFIG_KEYS`, so a client-supplied value replaces the
+    # server default; sending it here is what makes the setting real.
+    #
+    # The upstream environment default must be resolved too: LangGraph treats
+    # that value as its merge sentinel and would otherwise discard the API's
+    # incoming value in favor of `create_deep_agent`'s bound limit. Resolving it
+    # here also resolves it while building the agent, so both sides agree.
+    # Omitted only when neither Deep Agents nor LangGraph has an env override.
```
mdrxy: simplify

libs/code/deepagents_code/agent.py
```diff
@@ -2524,8 +2523,11 @@ def create_cli_agent(
             `auto_mode_enabled` is `True`.
         recursion_limit: Explicit LangGraph `recursion_limit` (graph step budget)
             for the main agent. When `None`, it is resolved from the
-            `DEEPAGENTS_CODE_RECURSION_LIMIT` env var, `[runtime].recursion_limit`
-            in `config.toml`, then the default via `resolve_recursion_limit`.
+            `DEEPAGENTS_CODE_RECURSION_LIMIT` env var and
+            `[runtime].recursion_limit` in `config.toml`, then from
+            `LANGGRAPH_DEFAULT_RECURSION_LIMIT`. When none is set, no
+            `recursion_limit` is bound and the LangGraph server budget (`10_011`)
```
mdrxy: don't mention magic number / upstream default

libs/code/deepagents_code/client/commands/config.py
```diff
@@ -41,6 +41,10 @@
 
 logger = logging.getLogger(__name__)
 
+_LANGGRAPH_DEFAULT_RECURSION_LIMIT_ENV = "LANGGRAPH_DEFAULT_RECURSION_LIMIT"
+_LANGGRAPH_API_DEFAULT_RECURSION_LIMIT = 10_011
```
mdrxy: feels wrong to hardcode this let's remove?

PR #5907, opened 2026-08-27
fix(sdk): clean up cancelled local shell commands

mdrxy:
Related: #5873

Local shell commands now clean up their supported process scope on timeout, interruption, and async cancellation so cancelled work does not continue consuming resources in the background.

---

`LocalShellBackend` inherits a thread-offloaded `aexecute`, so cancelling its caller cannot stop an already-running synchronous command. Descendants can also outlive timeout and interruption handling.

Execution now uses `subprocess.Popen` with one communication loop and a shared cleanup handler. POSIX cleanup kills the command's isolated process group and reaps the shell. Windows cleanup terminates only the direct shell; pipe polling allows captured handles to close even when descendants retain their write ends. Cleanup waits are bounded, and failures are logged without replacing the active interruption.

The async wrapper preserves `execute` overrides, prevents queued commands from starting after cancellation, and honors cancellation racing with output completion. Uncooperative workers are retained after the one-second cleanup grace period; application shutdown may still wait for those threads.

Partial timeout-output reporting is split into #6349, the next PR in this stack.


mdrxy:
!keep-open

PR #5922, opened 2026-08-27
feat(code): move MCP onto FastMCP and `langchain.mcp`

sydney-runkle:
`dcode` now uses FastMCP and `langchain.mcp` for MCP connection, session, and tool adaptation while preserving project trust, per-server status, OAuth, tool filtering, and provider-safe names.

---

The existing MCP layer duplicates transport construction, stderr handling, session recovery, pagination, and tool adaptation that FastMCP now owns. This replaces that machinery while keeping the behavior specific to `dcode`.

### What changes

- Mount MCP backends on one FastMCP router and retain exact backend provenance during discovery.
- Build stdio, SSE, and HTTP transports from FastMCP server models, including `cwd`, `timeout`, and `sse_read_timeout`.
- Convert tools through `langchain.mcp`, preserving optional-argument normalization, annotation metadata, disabled-tool filters, and provider-safe names.
- Report `notifications/message` diagnostics through structured logging instead of scraping subprocess stderr.
- Preserve per-server connection and discovery failures so `/mcp` can distinguish errors from authentication requirements.

### OAuth and MCP v2 compatibility

- Handle both `httpx` and `httpx2` status errors so OAuth challenges and expired credentials still offer `dcode mcp login`.
- Return MCP v2 `AuthorizationCodeResult` values from callback handlers, including RFC 9207 issuer data.
- Run OAuth login through the same FastMCP transport builder as normal connections.
- Keep FastMCP imports off the CLI startup path.

### Boundaries retained by `dcode`

- Config discovery, merging, `${VAR}` resolution, and project-server trust gating.
- Secure token persistence, refresh serialization, device flow, and provider-specific authorization parameters.
- Per-server status, actionable authentication errors, tool filters, and schema metadata for the TUI.

### Dependencies

- Add FastMCP 4 and MCP SDK 2.
- Require `langchain>=1.4.1`, which includes `langchain.mcp` and preserves free-form MCP object arguments during provider conversion.
- Remove `langchain-mcp-adapters`.

Made by [Open SWE](https://openswe.vercel.app/agents/71c6c5a6-7607-56e8-b4d0-8ed015b34a70)


libs/code/deepagents_code/mcp_tools.py
```diff
@@ -2589,318 +2161,161 @@ async def _preflight_and_connect(
...
+        contain an underscore — so `a_b_read` is ambiguous between server `a`
+        and server `a_b`. The longest configured name that matches wins, which
+        is the mount FastMCP actually resolved it against.
+        """
+        candidates = [
+            name
+            for name in backends
+            if tool_name.startswith(f"{name}_") and name not in skipped
+        ]
+        return max(candidates, key=len) if candidates else None
```
jkennedyvz, line 2203: **[P1] Preserve backend ownership when filtering tools**

agent review — blocking; high confidence.

With configured servers `a` and `a_b`, where only `a` publishes `b_read`, discovery returns `a_b_read` but `_owner` assigns it to `a_b`. Consequently, `disabledTools: ["b_read"]` on `a` never filters this tool: it is processed using `a_b`’s configuration and recorded as its `read` tool. FastMCP 4.0.2 resolves tools across providers, rather than selecting the longest namespace, so the retained tool still routes to `a` and can execute an explicitly disabled operation. The previous per-server discovery retained the actual owner. Carry backend provenance through discovery and apply filters against that owner instead of recovering it from the composed name.

libs/code/deepagents_code/mcp_tools.py
```diff
@@ -1923,100 +1476,76 @@ def _config_uses_env_interpolation(server_config: dict[str, Any]) -> bool:
...
-    """Drop empty-string values for optional MCP tool params.
-
-    Some MCP servers (e.g. Slack's `slack_search_public_and_private`) validate
-    optional ID-typed params with `value is not a channel ID` when the model
-    fills them in with `""` instead of omitting them. JSON-Schema-derived
-    Pydantic models happily accept `""` for `Optional[str]`, so the request
-    reaches the server and gets rejected with a generic `ToolException`.
+    from langchain.mcp import as_langchain_tool
+
+    tool = await as_langchain_tool(mcp_tool, client)
```
jkennedyvz, line 1503: **[P2] Wire normalization into the MCP call path**

agent review — non-blocking; high confidence.

Replacing `_build_cached_mcp_tool` removes the call to `_normalize_mcp_arguments`. The new `mcp_tool_middleware()` has no production references or registration at this head, and `langchain.mcp.as_langchain_tool` in locked LangChain 1.4.0 forwards the argument dictionary unchanged. Thus an optional string parameter supplied as `""` now reaches the backend instead of being omitted, regressing the existing normalization contract for servers that reject empty optional IDs. Register the middleware on the agents and subagents that receive these tools, or preserve normalization at the tool boundary. This is verified from the call path; no live Slack test was performed.

libs/code/deepagents_code/mcp_tools.py
```diff
@@ -1923,100 +1476,76 @@ def _config_uses_env_interpolation(server_config: dict[str, Any]) -> bool:
...
+
+    # Mounting already namespaced the tool as `server_tool`, but that name still
+    # has to satisfy the strictest provider limit, so it is recomposed through
+    # the same capping `main` applies. The bare name is recovered from the mount
+    # prefix and kept in metadata: it is what tool filters and the `/mcp` viewer
+    # show, and it is unrecoverable once the name is truncated and hashed.
+    original_tool_name = _unprefixed_tool_name(mcp_tool.name, server_name)
+    tool.name = _mcp_tool_name(server_name, original_tool_name)
+    tool.metadata = {
+        **(tool.metadata or {}),
```
jkennedyvz, line 1513: **[P2] Preserve the annotation contract for consumers**

agent review — non-blocking; high confidence.

The locked LangChain 1.4.0 adapter stores annotations under `metadata["mcp"]["tool"]["annotations"]` using snake-case keys, and this merge does not restore the old top-level keys. `mcp_tool_is_coherently_read_only` still checks top-level `readOnlyHint` and `destructiveHint`, so even a tool explicitly annotated read-only and non-destructive returns `False`. This unnecessarily gates read-only MCP calls and excludes them from the read-only context tools selected for criteria drafting/rubric grading. The previous wrapper exposed annotations at the top level. Update the consumers for the new nested schema, or preserve the existing metadata contract here.

libs/code/deepagents_code/mcp_tools.py
```diff
@@ -2589,318 +2161,161 @@ async def _preflight_and_connect(
...
+    runtime_manager.adopt(client, stack)
+    skipped.update(mount_failures)
 
-    async def _discover_server(
+    # One `list_tools` covers every mounted backend; FastMCP paginates
+    # internally and bounds itself, so a server returning a non-terminating
+    # cursor cannot hang the load.
+    try:
+        async with client:
+            mounted_tools = await client.list_tools()
```
jkennedyvz, line 2185: **[P2] Retain per-backend discovery failure outcomes**

agent review — non-blocking; high confidence.

A backend can initialize successfully and then fail `tools/list`. FastMCP 4.0.2’s `AggregateProvider` uses `provider_error_strategy="warn"` by default: `_list_tools` gathers exceptions and `_collect_list_results` logs and skips failed providers. Therefore this aggregate call succeeds and the surrounding exception handler never sees that backend failure. Since `_mount_backends` only recorded connection failures, `_build_server` receives an empty list and returns the default `status="ok"`, zero tools, and `error=None`. Previously discovery errors were caught and reported per server. Capture each backend’s discovery outcome before aggregation so the UI retains an actionable error and classification. Authentication failures during listing would also miss classification through this path; that authentication-specific scenario was not separately runtime-tested.

jkennedyvz (commented):
agent review — validated the four supplied findings against head `80bb316d` and base `bdccb82d`, including the published source for locked LangChain 1.4.0 and FastMCP 4.0.2. All four are supported; the disabled-tool filter bypass is blocking. Validation was source-based, not a fresh in-process reproduction or live Slack/authentication test. This review is scoped to the supplied findings.

jkennedyvz:
Some MCP servers provide many tools behind list, describe, call operations. Call(function, args) doesn't work properly with these dynamic servers. 

https://github.com/langchain-ai/langchain/pull/40414

jkennedyvz (approved):

PR #5946, opened 2026-08-28
feat(code): bind conversations to a recorded workspace

mdrxy:
A `dcode` server now records which directory each conversation belongs to, instead of assuming the directory the server was launched in.

---

## The problem

A server resolves its working directory once. The client passes the directory in the environment before launch; the server reads it at startup, derives the enclosing git root from it, and builds its agent, its file tools, and its offload runtime around that pair. Nothing re-reads it, so the directory is fixed for the life of the process.

The server is never told where the client actually is. It assumes. The assumption normally holds, because the client that starts the server is also the one that supplies the value. It stops holding whenever a client and its server can refer to different directories:

- A conversation is resumed whose recorded directory is not the current one.
- (Not reachable today) The client is pointed at a server it did not launch, and so cannot restart it.

The client handles the resumption case today by replacing the server process (full reboot).

The second case does not arise yet. Clients and servers are 1:1: the command-line entrypoint always launches its own server, and there is no supported way to attach to one that is already running. The app does carry guards for that mode — it refuses to switch agents or load a newly installed extra when it does not own the process — but they are reachable only by embedding the app as a library. The binding is what makes lifting the 1:1 restriction safe later, rather than a fix for a failure users can hit now.

Throughout this change, a *workspace* means a directory pair: the directory the client was started in, and the nearest enclosing git root.

> <details>
> <summary>Why the git root, and not just the working directory</summary>
> 
> The two answer different questions. The working directory is where the user is. The git root is what the project *is*, and most of the agent's local context is anchored to it, not to wherever the user happened to `cd`:
> 
> - The file tools are rooted at the git root, so the agent can reach the whole checkout from a subdirectory.
> - Project instruction files, project skills, and project subagent definitions are all discovered from the root.
> - Hook trust is recorded per project root, so approving hooks once covers the whole checkout.
> 
> Two sessions in different subdirectories of one checkout therefore share nearly all of that context, while two sessions in different checkouts share none of it. Identifying a workspace by the pair keeps both distinctions: the root separates projects, and the working directory separates positions within one.
> 
> The root is found by walking up for git metadata, so it is absent outside a repository. A workspace with no root falls back to the working directory alone, which is why the pair records it as optional rather than required.
> 
> </details>

## What changes

The client states its workspace outright, and the server records it against the conversation.

- Before it streams or offloads work, the client asks the server to bind the conversation to an absolute directory.
- The server canonicalizes the path, rejects traversal and anything that is not a directory, and writes the binding to its database. The write is atomic, and the first writer wins.
- The server returns a descriptor. Every later request on that conversation must present it. A missing, stale, or substituted descriptor is rejected rather than accepted.
- Agent and offload runtimes are selected from the validated binding and cached by workspace identity, so one process can hold runtimes for several workspaces at once.

The server stays authoritative over policy. A client proposes a directory; it never supplies permissions. The server computes the trusted policy from its own configuration and binds that. A client that sends a policy of its own gets it compared, not applied, and a mismatch is refused. Model credentials and prompt contents are never persisted. Conversation identifiers and checkpoint identity are unchanged.

## Scope

This does not remove any server restart. Every path that respawns a server still respawns it; those paths now pass the directory explicitly rather than relying on the ambient process environment. The per-workspace runtime cache makes a restart-free directory switch possible, but the interactive client does not use it yet.

Retiring the restart needs three things resolved first, none of which this change takes on:

1. The server process keeps the operating-system working directory it was launched with. Switching directories moves the client only. Any server-side code that asks the operating system for the current directory at request time, rather than reading the directory it was given, still sees the launch directory.

   

> <details>
>    <summary>A confirmed instance of this</summary>
> 
>    The shell approval prompt builds its "Working Directory:" line from the server's environment rather than from the binding. With one server hosting two workspaces, that line reads the launch directory for both, so the user would approve a command against a directory other than the one shown. It is display-only, and unreachable while clients and servers stay 1:1, but it is the exact failure this item describes and it should be fixed before that changes.
> 
>    </details>

2. The runtime cache re-reads the server's own configuration and overrides only the directory. A second workspace that needs different tool configuration or sandbox settings is refused rather than hosted.
3. Some server paths are resolved relative to the launch directory and would need to be rebased.

## Behavior

A directory mismatch is now refused, with the reason stated, instead of running against the server's launch directory:

    <conversation> has no workspace binding
    workspace context does not match <conversation>
    <conversation> is already bound to a different workspace

## Why merge this now

Read as a whole, this is the groundwork for lifting the 1:1 relationship between clients and servers, so one process can host several workspaces at once. Skipping the reboot on a directory switch is a consequence of that, not the goal.

One part is not groundwork, though. The binding and its rejection logic are live and enforced, not stubs. They simply cannot fire yet, because no current path produces a mismatch.

That is the tradeoff worth weighing: a mechanism that cannot trigger today, merged ahead of the capability it protects. The alternative is to add it alongside multi-workspace hosting, which leaves a window where conversations run unbound and a mismatch is silent rather than refused. Binding before use has to be in place first for that hosting to be safe, so it goes in first.

## References
- Plan: https://openswe.vercel.app/agents/2dd33c88-72d1-57d4-a56e-e272c1b1dde5/plan

Made by [Open SWE](https://openswe.vercel.app/agents/6ed08b3c-b007-59ca-b9ee-b24986eacc35)


mdrxy:
Note on why the binding needs its own table, since the description does not say it.

The obvious alternative is to keep the binding in LangGraph thread metadata. That store already exists, it is already per-thread and server-side, and this change already writes to it.

The reason not to: thread metadata is writable by any client over the public LangGraph API. A binding the client can rewrite is not a binding. The whole design rests on the server being the only writer, so the record has to live somewhere a client cannot reach. That is a stronger justification for a private table than anything currently in the description, and it is worth stating there.

Checkpoint metadata is weaker still, for the same reason plus one more: that table belongs to LangGraph, and write-once semantics are awkward to enforce on rows we do not own.

PR #5951, opened 2026-08-28
fix(code): reap temp artifacts left by a finished turn

mdrxy:
## What a user sees

The model creates a scratch file on one turn — say a pull-request body for `gh pr create --body-file` — and does not clean it up before that turn ends. On a later turn it tries to, and the tool refuses:

```
Denied temporary artifact cleanup: the exact path is not owned by this request.
```

Nothing else can remove the file either (`write_file`/`edit_file`/`delete` on a managed artifact path are rejected), so it sits in OS temp until the platform sweeps it — days on macOS. The model sees a failed cleanup it cannot retry, and the user sees a growing pile of `dcode-scratch-*` files. Observed with an artifact created on turn 4 and denied on turn 6.

## Why the denial is correct

`create_temp_artifact` tags each allocation with `thread_key` and `turn_id`; `delete_temp_artifact` resolves the caller's path through `_current_temp_artifacts`, which keeps only the current turn's. That narrow scope is what lets the tool act on a model-supplied `/tmp` path at all, verified by prefix and device/inode at `_delete_temp_artifact_file`. Widening it to thread scope would trade that containment away.

So this PR keeps the check and removes the leftovers instead.

## Change

`AutoModeHITLMiddleware.abefore_model` reaps, once per model step, this thread's artifacts whose allocating turn is over:

- `_stale_temp_artifacts` selects entries with a matching `thread_key` and a `turn_id` other than the latest human message's. It returns nothing when either trusted identity is missing, so an unidentifiable turn never reaps a live allocation. Within a turn the set is empty and the hook is a no-op.
- `_reap_temp_artifacts` deletes through the existing `_delete_temp_artifact_file`, so the identity checks are unchanged. An `OSError` (file already gone, identity changed) is logged at debug and the ledger entry is still cleared — the finished turn no longer owns the path either way, and a changed identity is never deleted.
- Filesystem work runs under `asyncio.to_thread`; the hook executes on the blockbuster-guarded server loop.

On the case above, the turn-4 artifact is deleted at the first model step of turn 5. A turn-6 delete attempt is still denied, and nothing is orphaned.

### DIFF
--- a/libs/code/deepagents_code/auto_mode.py
+++ b/libs/code/deepagents_code/auto_mode.py
@@ -935,6 +935,58 @@ def _current_temp_artifacts(
     }
 
 
+def _stale_temp_artifacts(
+    state: Mapping[str, object], runtime: object, messages: Sequence[object]
+) -> dict[str, AutoTempArtifact]:
+    """Return this thread's artifacts allocated by an already-finished turn.
+
+    Args:
+        state: Agent state carrying the private artifact ledger.
+        runtime: Runtime carrying the trusted thread identity.
+        messages: Messages carrying the trusted turn identity.
+
+    Returns:
+        Artifacts owned by this thread whose allocating turn is over. The result
+        is empty when either trusted identity is missing, so an unidentifiable
+        turn never reaps a live allocation.
+    """
+    thread_key = _thread_key(runtime)
+    turn_id = _latest_turn_id(messages)
+    if thread_key is None or turn_id is None:
+        return {}
+    return {
+        file_path: artifact
+        for file_path, artifact in _active_temp_artifacts(state).items()
+        if artifact["thread_key"] == thread_key and artifact["turn_id"] != turn_id
+    }
+
+
+def _reap_temp_artifacts(
+    artifacts: Mapping[str, AutoTempArtifact],
+) -> dict[str, AutoTempArtifactMutation]:
+    """Delete finished-turn artifact files and build their ledger removals.
+
+    Args:
+        artifacts: Stale artifacts keyed by their allocated path.
+
+    Returns:
+        A ledger mutation per artifact, clearing the entry. Entries are cleared
+        even when the file is already gone or its identity changed, because the
+        allocating turn no longer owns the path either way.
+    """
+    mutations: dict[str, AutoTempArtifactMutation] = {}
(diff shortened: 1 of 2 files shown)


mdrxy:
Closing this. The reaper is the right behavior, but it is the middle step of a design that is better done as one change.

## Why

`delete_temp_artifact` only accepts a path allocated by the current turn. This PR keeps that check and sweeps the leftovers on the next turn. That bounds the orphan window, but it leaves two problems.

The last turn of a session is never reaped. `abefore_model` runs at the start of the next model step. If the session ends, that step never happens and the artifact stays until the OS sweeps temp.

The model still does not know cleanup is automatic. It sees the path from an earlier turn in its own history, calls `delete_temp_artifact`, and gets `Denied temporary artifact cleanup: the exact path is not owned by this request.` That message states a refusal and gives no next action. This is the turn-4/turn-6 sequence in the description. The disk is clean, but the model still believes it leaked a file.

## What is better

Once cleanup is guaranteed at the end of a turn, `delete_temp_artifact` earns very little. It only makes a file stop existing a few minutes earlier, mid-turn. These are `mkstemp` allocations, mode 0600, same uid as the agent, in the same temp directory (`auto_mode.py:965-981`). The agent already reads the worktree and that directory. A short-lived 0600 text file is not a risk that needs a tool.

The tool costs more than it returns: a definition in every model call, a clause in the classifier policy (`auto_mode.py:1528-1529`), the denial branch at `auto_mode.py:2229-2239`, and its tests. That denial branch is the only failure mode in this area.

Removing it also gives the model a simpler contract: create a scratch file, use it, forget it. There is no lifecycle to get wrong.

## Order to do it in

1. Reap at the end of a turn, so cleanup is unconditional.
2. Remove `delete_temp_artifact`, its policy clause, and its denial branch.
3. Add a per-turn allocation cap in `_allocate_temp_artifact`. There is no cap today. In-turn deletes are what limits a pathological turn now, so the cap must replace them.

Step 1 must come first. If the tool is removed while reaping stays in `abefore_model`, the last turn of a session leaks permanently, which is worse than today.

Not urgent, so this can wait. The branch stays for whoever picks it up.

mdrxy:
Closing for now — see above. May pick this up later as the combined change.

PR #5996, opened 2026-08-30
perf(code): attach prompt-cache sessions to auto classifier calls

mdrxy:
Auto approval reviews (the background classifier calls that check whether a tool call needs user approval) now reuse a stable provider cache session per thread, instead of hitting the provider cold on every call.

## Before

Each classifier review looked like any other anonymous request to the provider. No session information was attached, so the provider couldn't associate consecutive reviews with each other and the prompt cache stayed cold — every review paid full latency and cost.

## After

Every classifier review within a thread carries the same session key, derived deterministically from the thread's identity (`sha256("auto:{thread_id}")`). Sessions remain isolated across threads, so one conversation's reviews never warm or pollute another's cache.

The middleware only strips and re-applies its own injected hints; explicit user-provided model settings remain authoritative, and the main model's session is never affected by classifier traffic.


mdrxy:
Closing because the implementation does not establish the intended performance benefit. The default inherited classifier already received the main thread’s stable provider cache key through ConfigurableModelMiddleware, so this PR primarily replaces that key with a classifier-specific deterministic key. A separately configured classifier does gain a key, but every review remains a separate request.

For OpenAI, prompt_cache_key influences routing but does not guarantee a cache hit. The classifier sends a static SystemMessage followed by changing HumanMessage content and adds no explicit cache breakpoint; current OpenAI guidance specifically warns that this shape may not make the shared static prefix reusable with implicit caching. The tests prove only that the same key is attached across reviews, not that cached_tokens increase or that latency or cost decreases.

This would be worth revisiting with a cacheable breakpoint appropriate to the provider and before/after provider measurements demonstrating improved cache hits, latency, or cost.

PR #6024, opened 2026-09-01
feat(code): default general-purpose subagents to fork mode

hntrl:
dcode's built-in `general-purpose` subagent now inherits the parent conversation by default, controlled by `DEEPAGENTS_CODE_FORKED_SUBAGENTS` (set it to `false` to opt out).

---

This enables fork mode for dcode's built-in `general-purpose` subagent while preserving a trusted user-level escape hatch. Project `.env` files cannot set the flag, so cloned repositories cannot alter whether parent conversation and private state are inherited. User-defined `general-purpose` subagents remain unchanged.

Supersedes #6021.

Made by [Open SWE](https://openswe.vercel.app/agents/047ad288-7130-5c4c-844f-89b23e80850b)


libs/code/deepagents_code/agent.py
```diff
@@ -3444,19 +3450,25 @@ def _subagent_cli_middleware(
...
-        context_schema=CLIContextSchema,
-        checkpointer=checkpointer,
-        store=store,
-        subagents=all_subagents or None,
-        name=_sanitize_agent_message_name(assistant_id),
-    )
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            message="The feature `forked subagents` is in beta",
```
ccurme, line 3456: should we filter out all langchain beta warnings

ccurme (approved):

libs/code/pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
 ]
 dependencies = [
     # Framework
-    "deepagents==0.7.11",
+    "deepagents==0.7.12",
```
mdrxy: This is stale

libs/code/deepagents_code/agent.py
mdrxy: No

PR #6103, opened 2026-09-05
fix(quickjs): clone REPL state for forked subagents

hntrl:
Forked QuickJS subagents now begin from a snapshot of the parent REPL while using a fresh interpreter slot, so child mutations cannot affect the parent.

---

This preserves forked execution context without sharing a live mutable QuickJS interpreter. The fork path falls back to a clean child slot if the parent interpreter is unavailable or snapshot cloning fails.

Tests: `make lint`; `make test COV_ARGS='' PYTEST_EXTRA='-q'` (243 passed); focused thread-affinity and snapshot tests (46 passed).

Made by [Open SWE](https://openswe.vercel.app/agents/93ae7eba-c855-5d13-af6b-2ce4cda23c4c)

### DIFF
--- a/libs/partners/quickjs/langchain_quickjs/middleware.py
+++ b/libs/partners/quickjs/langchain_quickjs/middleware.py
@@ -8,14 +8,12 @@ from collections.abc import Awaitable, Callable, Mapping
 from typing import TYPE_CHECKING, Annotated, Any, Literal, NotRequired
 
 from deepagents.middleware._utils import append_to_system_message
-from deepagents.middleware.subagents import _FORKED_CONTEXT_KEY
 from langchain.agents.middleware.types import (
     AgentMiddleware,
     AgentState,
     ContextT,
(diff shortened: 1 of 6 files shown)


libs/partners/quickjs/langchain_quickjs/middleware.py
```diff
@@ -65,7 +67,10 @@
 class REPLState(AgentState):
     """State schema for `CodeInterpreterMiddleware`."""
 
-    _quickjs_slot_id: NotRequired[Annotated[str, PrivateStateAttr]]
+    _quickjs_slot_id: NotRequired[
+        Annotated[str, OmitFromSchema(input=False, output=True)]
+        | Annotated[str, PrivateStateAttr]
```
hntrl: can we not have two annotated types? Can it be on OmitFromSchema(...) | PrivateStateAttr

PR #6108, opened 2026-09-06
feat(talon): add optional hybrid conversation history search

jkennedyvz:
Talon can search archived conversations by meaning using local, Voyage, OpenAI-compatible, or Atlas embeddings alongside keyword matching.

---

Depends on #6115; merge the storage PR first. This branch includes its latest storage changes without rewriting published commits.

Embedding profiles separate model selection from storage. Local inference remains optional; Voyage and OpenRouter need no torch installation. Profiles control dimensions, query semantics, input budgets, batching, and concurrency. Provider requests use native async clients with owned cleanup, bounded concurrency, and retry backoff. SQLite queries retain provider query/document semantics.

Credential-free fingerprints bind indexes to their embedding profile. Incompatible or unfinished indexes fail startup until an explicit `DEEPAGENTS_TALON_HISTORY_REINDEX=1` rebuild removes old vectors and backfills retained transcripts. Vector generations isolate dimensions in SQLite files, PostgreSQL schemas, or MongoDB collections. Atlas embeds server-side. Existing metadata backend plugins keep their contract; vector support has a separate optional entry point.

Opaque chat/query-scoped pagination, keyword fallback, bounded retrieval, authoritative visibility checks, and deletion before transcript release remain intact. Reset also works with semantic search disabled. New profile parameters are keyword-only and optional. The unreleased `history` extra is renamed `history-local`; remote extras are `history-voyage` and `history-openai`.

Validation: 726 Python tests and 17 bridge tests passed; four opt-in integration tests skipped in the network-free suite. A separate disposable PostgreSQL/pgvector integration test passed, including 3- and 4096-dimensional generations. Lint, formatting, types, staged secret checks, and lockfile validation passed. Live hosted embedding APIs, MongoDB/Atlas, and local Qwen inference were not run; provider and Atlas smoke tests are explicitly gated.


libs/talon/deepagents_talon/archive.py
```diff
@@ -70,36 +109,68 @@ class SQLiteConversationArchive:
...
+            from deepagents_talon.history_vectors import HistoryVectorIndex  # noqa: PLC0415
+
+            self.vectors = HistoryVectorIndex(self, store, indexing=vector_search)
 
     @classmethod
     @asynccontextmanager
-    async def from_conn_string(cls, conn_string: str) -> AsyncIterator[SQLiteConversationArchive]:
+    async def from_conn_string(
+        cls, conn_string: str, *, store: BaseStore | None = None, vector_search: bool = True
+    ) -> AsyncIterator[SQLiteConversationArchive]:
```
linted: This shouldn't be hard coded to sqlite. It needs to support all our database backends

linted (commented):
Nice work — the archive shape, opaque pagination tokens, and restart-safe indexing are the hard parts and they look solid. My feedback is all on one axis: **the embedding model is hard-wired to one specific locally-executed model**, and I'd like to loosen that before this lands, because it's much cheaper now than once there are indexes in the wild.

Concretely, I want to run this against **Voyage AI** and against **an embedding model hosted on OpenRouter**. Both are network-hosted, so this isn't only "let me name a different model" — remote inference needs to be a first-class path, and right now it isn't reachable at all.

Both are cheap to support in principle. `langchain-voyageai` ships `VoyageAIEmbeddings`, and OpenRouter's `/v1/embeddings` is OpenAI-schema-compatible, so it's `OpenAIEmbeddings(base_url="https://openrouter.ai/api/v1", model="qwen/qwen3-embedding-8b")` — no custom integration needed. `IndexConfig["embed"]` already accepts any `Embeddings`. What blocks it is the plumbing above that:

1. **The resolver hard-requires local inference.** `_embedding_index` raises unless `sentence_transformers` is importable, keyed off `enabled` rather than off which model was selected. A Voyage or OpenRouter user is forced to install torch (~2GB, transitively) to use a model that never runs on their machine.
2. **`DIMS` is a module constant.** Voyage exposes `output_dimension` (256/512/1024/2048 on voyage-3-large) and OpenAI-family models take a `dimensions` param, so dimensionality is a per-config property, not a package constant.
3. **`HistoryEmbeddings` is tuned for a CPU inference loop, not a network call.** A single-flight `asyncio.Lock` plus `batch_size=4` is a sensible memory bound for local torch and a severe throughput ceiling for an HTTP API that accepts ~96 inputs per request and benefits from concurrency.
4. **Token budgets are model-specific.** `max_seq_length = 8192` is Qwen's; against `CHUNK_SIZE = 4000` chars that's comfortable. OpenRouter's catalog spans 512-token models (Cohere embed-v3) to 8191 (text-embedding-3-small), and exceeding the limit truncates silently rather than erroring.
5. **The query instruction prefix is Qwen-specific** and applied on one backend only — see the `sqlite_history.py` comment.

There's also a third mode worth designing for even if nobody uses it yet: **the store embeds server-side**. The pinned `langgraph-store-mongodb>=0.4.0` already supports Atlas Automated Embedding — `MongoDBStore.__init__` checks `isinstance(self.embeddings, AutoEmbeddings)` and routes both the write path and the query path server-side — which is *also* Voyage, just billed through Atlas instead of directly. Its `VectorIndexConfig` requires `dims=-1` and `relevance_score_fn=None`; we hard-code `dims=DIMS` and take the `"cosine"` default, so it's unreachable. The Postgres equivalents (pgai / `pg_vectorize`) have the same shape, so "who computes the vector, us or the server?" is a general axis rather than a Mongo quirk.

Ideas, in value-for-effort order, with details inline:

1. **An `EmbeddingProfile` value object** — `(embed, dims, query_prompt, max_input_tokens, batch_size, client_side)` resolved once, replacing the constants and the duplicated index-config construction.
2. **Per-profile concurrency and batching**, so remote providers aren't throttled to the local model's single-flight lock.
3. **An embedding fingerprint persisted with the index**, so a model swap forces a reindex instead of silently mixing vector spaces.
4. **A `client_side` capability flag** covering the server-side-embedding backends.
5. **Split the local deps into their own extra**, so a Voyage/OpenRouter user installs zero ML dependencies.

One security note, since selecting a provider from config is the new surface: allow-list the *adapters* (`local` / `openai-compatible` / `voyage`) and validate at parse time in `config.py`, where `DEEPAGENTS_TALON_HISTORY_VECTOR_SEARCH` already raises `TalonConfigError` for a bad boolean — rather than letting an arbitrary env string reach `ensure_embeddings`/`init_embeddings`, where it turns into a dynamic import plus a model download. A configurable `base_url` (needed for OpenRouter) should require https and be validated, and API keys (`VOYAGE_API_KEY` / `OPENROUTER_API_KEY`) should come from env and stay out of error text — the same care `_mongodb_store` already takes with `raise ... from None` because driver errors can carry URI credentials. Worth keeping `trust_remote_code=False` on the local path for the same reason.

None of this is blocking — the feature works as scoped. It's about where the seams go before the API sets.

libs/talon/deepagents_talon/history_backends.py
```diff
@@ -55,19 +65,62 @@ async def remote_archive(config: TalonConfig) -> AsyncIterator[StoreConversation
...
+
+def _embedding_index(*, enabled: bool) -> tuple[HistoryEmbeddings, IndexConfig | None]:
+    if enabled and importlib.util.find_spec("sentence_transformers") is None:
+        msg = "Vector history requires deepagents-talon[history]: uv sync --extra history"
+        raise ImportError(msg)
+    embeddings = HistoryEmbeddings()
+    index: IndexConfig | None = (
+        {"dims": DIMS, "embed": embeddings, "fields": ["text"]} if enabled else None
+    )
+    return embeddings, index
```
linted: **Idea 1: fold the constants and this dict into one `EmbeddingProfile`, and make this the single resolver.**

These six lines are duplicated almost verbatim in `sqlite_history.py:53-59` (guard on `find_spec`, construct `HistoryEmbeddings()`, build `{"dims": DIMS, "embed": ..., "fields": ["text"]}`). Any second model has to be threaded through both, and they'll drift.

```python
@dataclass(frozen=True)
class EmbeddingProfile:
    embed: Embeddings
    dims: int
    query_prompt: str = ""
    max_input_tokens: int = 8192
    batch_size: int = 4
    client_side: bool = True  # False -> the Store embeds server-side

def resolve_profile(config: TalonConfig) -> EmbeddingProfile | None:
    if not config.history_vector_search:
        return None
    return _ADAPTERS[config.history_embed_adapter](config)
```

For the two I want to run, the adapters are thin:

```python
# voyage
VoyageAIEmbeddings(model="voyage-3-large", output_dimension=1024)  # langchain-voyageai

# openrouter (OpenAI-schema compatible)
OpenAIEmbeddings(
    model="qwen/qwen3-embedding-8b",
    base_url="https://openrouter.ai/api/v1",
    api_key=SecretStr(os.environ["OPENROUTER_API_KEY"]),
)
```

The important structural bit is that the `find_spec("sentence_transformers")` guard moves *inside* the local adapter. Today it keys off `enabled`, which is the wrong question — the right one is "does the selected profile run inference in-process?" As written, pointing Talon at Voyage still demands a torch install.

The `tuple[HistoryEmbeddings, IndexConfig | None]` annotation also stops being accurate the moment a second adapter exists; widening it to `Embeddings` is the trigger for the `aclose` issue I flagged on line 110.

libs/talon/deepagents_talon/history_embeddings.py
```diff
@@ -0,0 +1,106 @@
...
+
+        Args:
+            texts: Transcript chunks or prepared queries.
+        """
+        async with self._async_lock:
+            if self._pending is not None:
+                await asyncio.shield(asyncio.gather(self._pending, return_exceptions=True))
+            self._pending = asyncio.create_task(asyncio.to_thread(self.embed_documents, texts))
+            self._pending.add_done_callback(_consume_exception)
+            return await asyncio.shield(self._pending)
```
linted, line 103: **Idea 2: concurrency and batch size belong to the profile — this shape is right for local CPU and wrong for a network call.**

The single-flight design here is well-reasoned for what it currently does: one lock, one in-flight task, `batch_size=4`, everything serialized so a 0.6B model on CPU can't blow up memory or pile up threads after a cancelled search. The docstring says as much.

For Voyage or OpenRouter that same design becomes the bottleneck. Those are HTTP calls that are latency-bound, not memory-bound: OpenRouter accepts arrays of up to ~96 inputs per request, and both providers happily serve several concurrent requests. Serializing every batch behind one lock means background indexing of a large archive runs at roughly one round-trip per 4 chunks, and a foreground search queues behind whatever indexing batch is mid-flight.

Suggest the profile carries `batch_size` and a concurrency bound, with the local adapter keeping `(4, 1)` and remote adapters getting something like `(96, 4)` — and `asyncio.to_thread` only applying to the local path, since a remote `Embeddings` already has a native async implementation that shouldn't be wrapped in a thread.

Failure modes differ too, and it's worth being deliberate: local inference fails deterministically, whereas remote calls hit rate limits, 5xx, and per-token billing. Falling back to keyword matching on error (which the PR already does for search) is right for a *query*, but for *background indexing* a 429 should back off and retry rather than mark the chunk done — otherwise a transient rate limit leaves permanent holes in the index.

libs/talon/deepagents_talon/history_embeddings.py
```diff
@@ -0,0 +1,106 @@
...
+        *,
+        batch_size: int,
+        normalize_embeddings: bool,
+        show_progress_bar: bool,
+    ) -> _Vectors: ...
+
+
+MODEL = "Qwen/Qwen3-Embedding-0.6B"
+DIMS = 1024
+QUERY_PROMPT = "Instruct: Retrieve past conversation passages relevant to the query.\nQuery: "
```
linted: **Idea 3: these three constants are one model's identity — make them a profile, and persist a fingerprint of the active one.**

`MODEL` and `DIMS` are coupled, and the drift failure mode is the bad kind: vectors written under one model and queried under another live in different spaces, so `search_conversations` returns confident nonsense instead of erroring. That's already true for anyone who edits these constants; it becomes much likelier once the model is configurable.

The PR already persists indexing progress across restart and drops vectors before releasing transcripts on reset — storing `{"model": ..., "dims": ...}` next to that progress and refusing to serve (or forcing a reindex) on mismatch is a small addition to machinery that exists. I'd argue it's the precondition that makes the other ideas safe to ship, and worth deciding explicitly whether a mismatch is a hard error, a silent keyword-only fallback (consistent with existing degradation), or a background reindex.

`dims` specifically can't stay a constant: Voyage's `output_dimension` is selectable (256/512/1024/2048 on voyage-3-large) and OpenAI-family models take `dimensions`, so two people on the same *model* can have different vector widths.

Related, one line down — `max_seq_length = 8192` (line 64) is Qwen's ceiling, and `archive.CHUNK_SIZE = 4000` chars sits comfortably inside it. Via OpenRouter that ceiling isn't a constant either: `text-embedding-3-small` is 8191 tokens but Cohere embed-v3 is 512. Over-long input is truncated by the provider rather than rejected, so a model swap could quietly start indexing the first ~40% of each chunk with no error anywhere. `max_input_tokens` on the profile, validated against `CHUNK_SIZE` at startup, would turn that into a config error instead of a silent retrieval-quality regression.

libs/talon/deepagents_talon/history_backends.py
```diff
@@ -117,8 +174,19 @@ async def _mongodb_store(uri: str) -> AsyncIterator[BaseStore]:
                 w="majority",
             )
             stack.push_async_callback(asyncio.to_thread, client.close)
-            collection = client.get_default_database()["talon_history"]
-            store = await finish(asyncio.to_thread(driver.MongoDBStore, collection))
+            target = client.get_default_database()[collection]
+            config = (
+                driver.create_vector_index_config(
+                    embed=index["embed"], dims=index["dims"], fields=["text"]
+                )
```
linted, line 219: **Idea 4: let the profile declare whether embedding happens server-side, and pass the config the backend actually needs.**

`create_vector_index_config` also accepts `relevance_score_fn` and `embedding_key`; we take the defaults for both, which is what forecloses Atlas Automated Embedding — its required combination is `dims=-1`, `relevance_score_fn=None`, `embed=AutoEmbeddings(model=...)`, with the store keeping *text* in `embedding_key` rather than vectors.

```python
config = (
    driver.create_vector_index_config(
        embed=profile.embed,
        dims=profile.dims if profile.client_side else -1,
        fields=["text"],
        relevance_score_fn="cosine" if profile.client_side else None,
    )
    if profile is not None
    else None
)
```

Worth flagging that this is a second, cheaper route to Voyage specifically: Atlas auto-embed *is* Voyage, billed through Atlas with the key in Project Settings, with no embedding traffic from the client at all. `MongoDBStore` also takes `query_model=` for a different query-time model than index-time, which is how MongoDB suggests keeping query latency off the indexing path.

Also: `_vector_store` (line 104) constructs a `HistoryEmbeddings` unconditionally. Under any server-side mode no local model should be instantiated — it's never called, but it's a real import and a real lock.

libs/talon/deepagents_talon/sqlite_history.py
```diff
@@ -0,0 +1,70 @@
...
+class _HistorySqliteStore(AsyncSqliteStore):
+    async def abatch(self, ops: Iterable[Op]) -> list[Result]:
+        # SQLite embeds searches through aembed_documents, unlike other Stores.
+        # Add Qwen's retrieval instruction only to query operations.
+        return await super().abatch(
+            op._replace(query=QUERY_PROMPT + op.query)
+            if isinstance(op, SearchOp) and op.query
+            else op
+            for op in ops
+        )
```
linted: This subclass encodes two model-specific assumptions at once: that the query prefix is *Qwen3's* instruction format, and that it applies on this backend only (because SQLite routes queries through `aembed_documents` rather than `embed_query`).

With `query_prompt` on the profile this becomes `op._replace(query=profile.query_prompt + op.query)`, which is correct for a model with no instruction convention (empty string) as well as one with a different one.

This matters directly for both models I want to run: Voyage takes `input_type="query"`/`"document"` as an API parameter rather than a text prefix, and OpenAI-family models on OpenRouter want no prefix at all. Without the change, pointing Talon at either silently prepends `Instruct: Retrieve past conversation passages relevant to the query.` to every query string — which won't error, it'll just embed the instruction as if it were part of my search terms and quietly degrade retrieval.

The asymmetry is invisible from the other direction too: under Atlas auto-embed, `embed_query` never runs, so the instruction silently *vanishes*. Three configurations, three different behaviors, none of which fail loudly.

libs/talon/deepagents_talon/history_vectors.py
```diff
@@ -0,0 +1,233 @@
...
+
+if TYPE_CHECKING:
+    from langgraph.store.base import BaseStore
+
+    from deepagents_talon.history_index import Row, VectorArchive
+
+logger = logging.getLogger(__name__)
+_CANDIDATES = 100
+_MAX_SEARCH_PAGES = 32
+_BATCH_SIZE = 4
```
linted, line 38: `_BATCH_SIZE = 4` is sized for the local CPU model (it matches `HistoryEmbeddings`' own `batch_size=4`), and it's the batch size for the *archive row scan* that feeds embedding, so it caps remote throughput too.

If the profile grows a `batch_size`, this should read from it rather than being a module constant — 4 rows per round-trip against a provider that accepts ~96 inputs per request makes a full backfill roughly 24x more HTTP calls than it needs, each separately billed and rate-limited.

libs/talon/deepagents_talon/history_backends.py
```diff
@@ -55,19 +65,62 @@ async def remote_archive(config: TalonConfig) -> AsyncIterator[StoreConversation
...
+
+
+@asynccontextmanager
+async def _vector_store(uri: str, *, enabled: bool) -> AsyncIterator[BaseStore]:
+    embeddings, index = _embedding_index(enabled=enabled)
+    async with _remote_store(uri, index=index, vectors=True) as store:
+        try:
+            yield store
+        finally:
+            await embeddings.aclose()
```
linted: `aclose()` is specific to `HistoryEmbeddings`. `VoyageAIEmbeddings` and `OpenAIEmbeddings` have no such method, so once `embed` can be any `Embeddings` this raises `AttributeError` on teardown. Needs a `hasattr` guard, or teardown declared on the profile next to construction. Minor, but it'd only surface at shutdown on the configs nobody ran locally.

libs/talon/pyproject.toml
```diff
@@ -31,6 +31,10 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
+history = [
+    "sentence-transformers>=5.2.0,<7.0.0",
+    "transformers>=4.51.0,<6.0.0",
+]
```
linted: **Idea 5: narrow this extra once local inference is one adapter among several.**

`sentence-transformers` pulls torch transitively, making this comfortably the most expensive extra in the package. Renaming it `history-local` and adding thin siblings would mean a remote user installs no ML stack at all:

```toml
history-local = ["sentence-transformers>=5.2.0,<7.0.0", "transformers>=4.51.0,<6.0.0"]
history-voyage = ["langchain-voyageai>=0.4.1,<0.5.0"]
history-openai = ["langchain-openai>=1.0.0,<2.0.0"]  # also covers OpenRouter via base_url
```

`langchain-voyageai` 0.4.1 requires `langchain-core>=1.0.0`, worth checking against what's already resolved here. The OpenRouter case needs no dedicated package — it's `langchain-openai` with `base_url` pointed elsewhere, which is an argument for the adapter being named for the *protocol* (`openai-compatible`) rather than the vendor.

Separately, a pin inconsistency: `transformers` is `>=4.51.0,<6.0.0` here but `>=5.12.1,<6.0.0` in `media` (line 52). They resolve — intersection is `>=5.12.1,<6.0.0` — but `[history]` alone can select a 4.x that `[history,media]` never will, so the combination gets exercised while the standalone extra drifts onto an untested floor. If 4.51 is just "first release with Qwen3 architecture support," probably worth saying so in a comment and letting media's floor win.

linted (approved):

PR #6115, opened 2026-09-06
feat(talon): add configurable conversation history storage

jkennedyvz:
Talon can store conversation history in MongoDB or PostgreSQL through `DEEPAGENTS_TALON_HISTORY_URI`.

---

SQLite, MongoDB, and PostgreSQL use one archive implementation with thin storage adapters. SQLite remains the default; install the `mongodb` or `postgres` extra for remote storage. This alpha requires fresh history storage. Checkpoints use LangGraph separately.

Cursor pagination avoids rescanning prior pages. Shared code handles chat isolation, keyword search, and restartable writes/deletes, with a 500-record scan budget.

Validation: 134 focused Python tests, 17 bridge tests, lint, formatting, and types passed. Live MongoDB/PostgreSQL integration tests were not run because local servers are unavailable.


libs/talon/deepagents_talon/history_backends.py
```diff
@@ -0,0 +1,125 @@
...
+        config: Trusted configuration containing a validated remote history URI.
+
+    Yields:
+        An initialized archive closed before its metadata connection.
+    """
+    uri = config.history_uri
+    if uri is None:
+        msg = "Remote history requires DEEPAGENTS_TALON_HISTORY_URI"
+        raise TalonConfigError(msg)
+    factory = _mongodb_store if urlsplit(uri).scheme.startswith("mongodb") else _postgres_store
```
linted: This isn't easily expandable. If a user wants to use something else like mysql or a custom sqlite connection uri, this will fail.

libs/talon/deepagents_talon/config.py
```diff
@@ -176,3 +185,25 @@ def _validate_assistant_id(assistant_id: str | None) -> None:
...
     return key in _RUNTIME_ENV_KEYS or key.startswith(_RUNTIME_ENV_PREFIXES)
+
+
+def _validate_history_uri(uri: str | None) -> None:
+    if uri is None:
+        return
+    try:
+        parsed = urlsplit(uri)
+        if (
+            parsed.scheme in {"mongodb", "mongodb+srv", "postgres", "postgresql"}
```
linted: This shouldn't be hard coded here. It makes it difficult to maintain.

linted (changes requested):

linted (approved):

PR #6119, opened 2026-09-06
fix(sdk): add async methods to StoreBackend for async store support

### LINKED ISSUE #6120: StoreBackend hangs on async stores — missing async methods for ls/glob/grep/download_files
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
import asyncio
from langgraph.store.postgres.aio import AsyncPostgresStore

async def main():
    store = AsyncPostgresStore(conn)  # any async BaseStore
    backend = StoreBackend(namespace=lambda _: ("test",), store=store)

    # These hang forever — sync methods call store.get()/store.search()
    # on an async connection. BackendProtocol's default async impl
    # wraps via asyncio.to_thread(sync), which also hangs.
    await backend.adownload_files(["/AGENTS.md"])  # hangs
    await backend.als("/")                          # hangs
    await backend.aglob("*")                        # hangs

asyncio.run(main())
```

```shell

```

## Problem

StoreBackend supports async stores (e.g. AsyncPostgresStore) for aread/awrite/aedit/adelete — they use store.aget/store.aput and _asearch_store_paginated (which uses store.asearch).

However, download_files, ls, glob, and grep only have sync versions that call store.get() / store.search(). On async stores, these sync methods hang — the async connection can't be used synchronously.

The BackendProtocol default async implementations fall back to asyncio.to_thread(sync_method), which still calls the sync store methods → same hang.

## Fix

The library already has the async helper _asearch_store_paginated (used by adelete). These four methods just weren't wired up to it. Add adownload_files/als/aglob/agrep mirroring the sync bodies but using store.aget/store.asearch.

### Environment / System Info

_No response_

zhangzhiqiangcs:
Fixes #6120

`StoreBackend` no longer hangs when wrapping an async `BaseStore` (e.g. `AsyncPostgresStore`): `adownload_files`, `als`, `aglob`, and `agrep` now have proper async implementations instead of falling back to the blocked sync path.

---

`StoreBackend` already supports async stores for `aread`/`awrite`/`aedit`/`adelete` — they use `store.aget`/`store.aput` and `_asearch_store_paginated` (which uses `store.asearch`). However, `download_files`, `ls`, `glob`, and `grep` only had sync versions calling `store.get()` / `store.search()`. On an async connection these block indefinitely. The `BackendProtocol` default async fallback wraps via `asyncio.to_thread(sync_method)`, which still calls the sync store methods — same hang.

The fix reuses the existing `_asearch_store_paginated` async helper (already used by `adelete`):

- `adownload_files` → `store.aget` (mirrors sync `download_files` → `store.get`)
- `als` → `_asearch_store_paginated` (→ `store.asearch`)
- `aglob` → same helper
- `agrep` → same helper

Each method mirrors its sync counterpart exactly, only swapping the store call. Sync methods are untouched for sync stores. No API changes, no new logic — just wiring up the four methods that were missing.

This unblocks `MemoryMiddleware` + `StoreBackend` + `AsyncPostgresStore`, where `MemoryMiddleware.before_agent` calls `adownload_files` on every turn and would hang forever on an async store.


pclin1212:
Thanks for this fix — the motivation is clear and the routing through `asearch`/`aget` is the right call for async stores. A few points from my review:

**[Blocking] Missing test coverage.** The PR adds four public async methods with no tests. Since the whole point is avoiding the sync fallback that hangs on async stores, could you add a test with a store stub whose sync `search`/`get` raise (or block), proving `als`/`agrep`/`aglob`/`adownload_files` actually route through `asearch`/`aget`? A behavioral test on an in-memory async store would also help.

**[Blocking] Conflicts with #6258.** The new `size = len(file_data_to_string(fd))` sites report character counts, while #6258 is migrating `FileInfo.size` to UTF-8 byte sizes. This PR adds four new call sites of the old semantics — please align with that PR (or rebase on top of it) so the two don't land inconsistent.

**[Non-blocking] Inconsistent error handling.** `als`/`agrep`/`aglob` guard `_convert_store_item_to_file_data` with `except ValueError: continue`, but `adownload_files` calls it bare — one corrupt item would fail the whole batch. Either guard it consistently or document why download keeps sync-parity behavior.

**[Non-blocking] Code duplication.** The four async methods are near-verbatim copies of the sync ones. Consider extracting shared helpers (e.g. `_ls_from_items(items, path)`) so sync/async only differ in how items are fetched; otherwise the two copies will drift over time.

*Disclosure: this review was drafted with the help of an AI coding assistant.*

zhangzhiqiangcs:
@pclin1212 Thanks for the review — all four points addressed in the latest push.

**1. [Blocking] Missing test coverage**
Added `_SyncBlockingStore` (subclass of `InMemoryStore`) whose sync `get`/`search`/`put` raise `AssertionError`, while async `aget`/`asearch`/`aput` route through `abatch` (independent of the overridden sync methods) and count calls. Five tests prove each async method routes through the async store API and would fail loudly if it fell back to the sync path:
- `test_als_routes_through_asearch_not_sync_search`
- `test_agrep_routes_through_asearch_not_sync_search`
- `test_aglob_routes_through_asearch_not_sync_search`
- `test_adownload_files_routes_through_aget_not_sync_get`
- `test_aupload_files_routes_through_aput_not_sync_put`

Verified the negative case: temporarily removing each async override makes the corresponding test fail via the stub's `AssertionError` (rather than hanging forever).

**2. [Blocking] Conflicts with #6258**
All `size` sites now use `len(file_data_to_string(fd).encode("utf-8"))` — UTF-8 byte sizes, matching #6258's semantics. Two tests (`test_als_reports_utf8_byte_size`, `test_aglob_reports_utf8_byte_size`) use `héllo` (5 chars / 6 bytes) to assert byte counts. Since #6258 applies the same change to the same lines, it should rebase as a no-op on top of this PR with zero conflicts.

**3. [Non-blocking] Inconsistent error handling**
`adownload_files` now wraps `_convert_store_item_to_file_data` in `try/except ValueError`, returning `FileDownloadResponse(path=path, content=None, error="invalid_file")` for a corrupt item instead of failing the whole batch. This matches the `except ValueError: continue` guard already in `als`/`agrep`/`aglob`. Test: `test_adownload_files_skips_corrupt_item` plants a `content=None` item and asserts the good/missing files in the same batch still resolve correctly. (Left the sync `download_files` untouched — its bare-call behavior is pre-existing.)

**4. [Non-blocking] Code duplication**
Extracted three shared helpers so sync/async only differ in how items are fetched:
- `_ls_from_items(items, path)` — shared by `ls`/`als`
- `_files_dict_from_items(items)` — shared by `grep`/`agrep`/`glob`/`aglob` (items → `{path: FileData}` with the corrupt-item guard)
- `_glob_result_from_files(files, pattern, path)` — shared by `glob`/`aglob` (pattern match + FileInfo construction)

Net effect on `store.py`: the refactor removed 38 lines of duplicated logic. All 1062 backend tests pass.

zhangzhiqiangcs:
@ccurme ping

PR #6129, opened 2026-09-07
feat(talon): support fresh subagents and per-task tool selection

jkennedyvz:
Talon subagents use fresh context: named agents run with configured tools, while `general-purpose` defaults to no tools and accepts tools and context per launch.

---

The main agent can edit and reload named definitions, then invoke them with their configured tools. Dynamic `general-purpose` launches select tools explicitly, including `execute` for shell access, and receive context through `description`. Approval gates remain intact; invalid reloads retain the active configuration.

Compiled and remote agents must use a name other than `general-purpose`, which is reserved for dynamic launches.

Validation: 23 focused subagent/reload tests, Ruff, and type checks passed.

Fixes https://linear.app/langchain/issue/JKB-83

### DIFF
--- a/libs/talon/deepagents_talon/mcp.py
+++ b/libs/talon/deepagents_talon/mcp.py
@@ -219,12 +219,13 @@ class MCPToolProvider:
             "reload_mcp_configuration",
             description=(
                 "Reload Talon's configured MCP servers before the next agent turn. "
-                "Use after the operator changes the MCP configuration."
+                "Use after the operator changes the MCP configuration. Verify activation "
+                "with get_agent_tools; running tasks retain their original capabilities."
             ),
         )
         def reload_mcp_configuration() -> dict[str, str]:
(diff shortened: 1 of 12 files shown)


libs/talon/deepagents_talon/mcp.py
```diff
@@ -219,12 +219,13 @@ def _reload_tool(self) -> BaseTool:
             "reload_mcp_configuration",
             description=(
                 "Reload Talon's configured MCP servers before the next agent turn. "
-                "Use after the operator changes the MCP configuration."
+                "Use after the operator changes the MCP configuration. Verify activation "
```
linted, line 222: `Use after the operator changes MCP configuration` -> `Use after MCP configuration changes`

Since the agent can change the mcp config, we should make sure it runs this.

linted (approved):

PR #6153, opened 2026-09-08
fix(talon): bound archive scans, query embeddings, and deletion markers

linted:
**Stack 1 of 4 — history/archive.** Reviews cleanest bottom-up: this PR → `history-2-indexing-worker` → `history-3-drivers-adapters` → #6147.

Three High-severity defects in the conversation archive, found reviewing the ~23k lines added to `libs/talon` since `deepagents-talon==0.0.6`. All three are live in released 0.0.7.

- **Hybrid search raised on any long conversation.** `lexical()` was called unwrapped while the *optional* semantic leg beside it was defensively wrapped — backwards. With `_MAX_SCAN = 500` and `_CANDIDATES = 100`, any chat over 500 archived chunks whose query matched fewer than 100 of the newest 500 made `search_conversations` raise. `_retrieval_chain` now takes `partial=`; `entries()` and `conversations()` still raise, since a truncated page there would be a correctness lie.
- **Query embeddings were prefixed only on SQLite.** On Postgres and plugin backends the query was chunked and embedded as a *document* — no instruction prefix, no query input type, plus an extra provider round trip from inside the DB batch. `PreparedVectorStore.abatch` now publishes the vector under both cache keys.
- **The deletion-marker chain grew without bound** and was re-walked on every poll. Markers are now doubly linked and unlinked in O(1) as deletions complete.

Each test was verified to fail against pre-fix source with the exact symptom.

### DIFF
--- a/libs/talon/deepagents_talon/history_prepared_store.py
+++ b/libs/talon/deepagents_talon/history_prepared_store.py
@@ -45,7 +45,15 @@ class PreparedVectorStore(BaseStore):
         }
         for op in operations:
             if isinstance(op, SearchOp) and op.query is not None:
-                cache[True, op.query] = await self.embed.aembed_query(op.query)
+                vector = await self.embed.aembed_query(op.query)
+                cache[True, op.query] = vector
+                # Only some Stores embed a search through aembed_query; SQLite and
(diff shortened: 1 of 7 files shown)


linted:

PR #6155, opened 2026-09-08
fix(talon): share one optional-driver loader and close provider clients

linted:
**Stack 3 of 4 — history/archive.** Based on `linted/talon/history-2-indexing-worker` (#6154). Review #6153 and #6154 first.

Optional-dependency loading, provider client lifetimes, and embedding settings that were silently ignored.

- **One optional-driver loader replaces two duplicates.** `_driver` existed byte-identically in `history_adapters.py` and `history_backends.py`, differing only in a message prefix, and both did `raise ImportError(msg) from None` — so an extra that *is* installed but fails for another reason (broken `torch`, missing libpq) told the user to install what they already had, with no traceback. The shared loader in `history_drivers.py` checks `ImportError.name` and re-raises anything that isn't the extra itself, chaining the cause either way.
- **MongoDB startup was unbounded and its dispatcher never stopped**, unlike its two sibling backends. The startup budget is now explicit — see the comment on why `asyncio.timeout` alone cannot bound it, since `finish()` runs to completion by design.
- **Voyage and Atlas embedding clients were never closed**, leaking connection pools on every archive reopen. Cleanup registration moved into `open_profile` so it covers whichever adapter resolves.
- **`PreparedVectorStore.batch()` silently skipped the preparation the class exists for**, reaching the inner store with an empty cache; it now refuses. Field extraction no longer hardcodes `text`.
- **`_bridge` waited on a cross-thread future with no timeout**, so a worker thread could block forever against a stopping loop.
- **`SEND_DIMENSIONS` and `BASE_URL` were honoured by one adapter each and silently ignored elsewhere** — including by the adapter the `SEND_DIMENSIONS` error message tells you to use it with. Now honoured or refused at config time.

Verified by inspection only: Voyage client release and `output_dimension` handling — `voyageai` and `langchain-mongodb` aren't installed in CI, so the tests cover the registration path via fakes, not real SDK pool release.

### DIFF
--- a/libs/talon/deepagents_talon/archive.py
+++ b/libs/talon/deepagents_talon/archive.py
@@ -60,7 +60,7 @@ class SearchPage(TypedDict):
     pagination_status: Literal["ok", "expired"]
 
 
-def _search_page(
+def build_search_page(
     hits: list[ArchiveEntry],
     limit: int,
     status: SemanticStatus,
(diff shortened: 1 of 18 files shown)


linted:

PR #6156, opened 2026-09-08
fix(talon): make background subagents and host start/stop recoverable

linted:
**Stack 1 of 4 — background subagents & host lifecycle.** Reviews cleanest bottom-up: this PR → `lifecycle-2-subagent-orchestration` → `lifecycle-3-host-state` → #6148.

Six High-severity defects in the newest and least-exercised subsystem in the package. All are live in released 0.0.7.

- **A failing turn re-ran forever, at 1 Hz, burning tokens.** `acknowledge()` ran only on the success path, so a turn that failed while consuming a background result left the result pending and the dispatcher started a fresh turn a second later — indefinitely. Delivery attempts are now bounded with backoff, and `CancelledError` is deliberately excluded so interrupt-driven retries stay unbounded as an existing test intends.
- **The dispatcher died silently, then broke `stop()`.** No `try/except` in the loop body, and `stop()` re-raised the stored exception before tearing anything down.
- **Background workers ran in an empty `contextvars.Context()`**, so `_HISTORY_SCOPE`, `_CRON_ORIGIN` and `_AUTHORIZATION_HANDLER` all reverted to `None` and any context-scoped tool raised — surfacing only as "Subagent failed before returning a result." Now `copy_context()`, with the authorization handler deliberately withheld: a flow started after the turn ends would outlive `_clear_authorization`.
- **Every background failure was swallowed with no logging** — the module never imported `logging` at all.
- **An agent turn that raised sent the user nothing.**
- **`start()` and `stop()` leaked components on partial failure.** A channel raising during `start()` left `_running` false, so `stop()` returned early and tore down nothing — orphaning the WhatsApp Node subprocess and leaking checkpointer connections.

Every test verified to fail against pre-fix source.

### DIFF
--- a/libs/talon/README.md
+++ b/libs/talon/README.md
@@ -382,8 +382,9 @@ optional.
 ## Research defaults
 
 On startup, homes receive any missing `AGENTS.md` files for main, `internal-research`, and
-`external-research`, with defensive prompts. External research owns `fetch_url` and
-Tavily-backed `web_search`, attached at construction by default. Search is added
+`external-research`, with defensive prompts. External research declares `web: true` in its
+frontmatter, which is what attaches `fetch_url` and Tavily-backed `web_search` at
+construction; the capability follows the declaration, not the directory name. Search is added
 only when `TAVILY_API_KEY` is nonempty in the runtime environment; without it,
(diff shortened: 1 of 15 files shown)


linted:

PR #6158, opened 2026-09-08
fix(talon): stop one conversation from stalling or outliving the rest

linted:
**Stack 3 of 4 — background subagents & host lifecycle.** Based on `linted/talon/lifecycle-2-subagent-orchestration` (#6157). Review #6156 and #6157 first.

Per-conversation state and thread keying.

- **One stuck conversation stalled every other.** The background dispatcher iterated routes sequentially holding each conversation's lock, and `receive_message` holds that same lock across a cancel that can take 30s — parking the single dispatcher for the whole window. It now skips a contended route and retries next tick.
- **Per-conversation state grew without bound.** Locks, tasks, generations and block flags were keyed by conversation and never pruned. Now refcounted and dropped only when nothing holds or awaits them — and turn state only for an id with no running task, no pending approval or authorization, no background route and no interrupt block. `_generations` is treated carefully on purpose: pruning a live counter would make a superseded turn's generation match a fresh one and answer when it should stay quiet.
- **`_reset_all_history` could clear the archive but not persist its counter**, leaving the user told to retry against an archive that was already gone. The counter is now written first, so a failed write clears nothing.
- **Thread ids are now always `provider:conversation`**, removing a latent re-keying hazard where enabling history or adding a channel would silently change every thread id.

**On that last one — no user-facing change for the CLI.** `_run_host` always constructs `ConversationSaver` when a model is configured, so `history_enabled` was already true and the keying was already channel-scoped. It affects only embedders constructing `TalonHost` with their own non-archiving checkpointer on a single channel. An earlier draft of this note claimed existing conversations restart empty and that users could return to a pre-`/new` thread; both were wrong, and the second was a false privacy claim — nothing discarded with `/new` resurfaces.

`_conversation_resets` is deliberately left unpruned: its counters map a conversation to its current thread id, so dropping one would silently resurrect an older conversation.


linted:

PR #6159, opened 2026-09-08
fix(talon): harden the MCP OAuth device flow and credential paths

linted:
**Stack 1 of 4 — MCP OAuth.** Reviews cleanest bottom-up: this PR → `mcp-2-config-store` → `mcp-3-token-storage` → #6149.

Three High-severity defects in the MCP OAuth stack, all live in released 0.0.7.

- **The device-login CLI crashed after succeeding.** `login_mcp_server` caught `DeviceAuthorizationCompletedError`, but it is raised inside `ClientSession`'s anyio task group and arrives wrapped in an `ExceptionGroup` — so `deepagents-talon mcp login <github-server>` persisted credentials and then exited with a traceback and non-zero status. Now matched on leaves. The same omission made `authenticate_mcp_server` raise into the agent instead of returning `{"status": "failed"}`.
- **Device-flow HTTP clients bypassed the SSRF-safe transport** and honoured ambient `HTTPS_PROXY`/`SSL_CERT_FILE` while carrying the `device_code` and receiving tokens. All four now use the pinned client. Blocking `getaddrinfo` moved off the event loop, which also makes the 10s discovery bound real rather than nominal.
- **Credential-path placement** is checked and warned about, and the intermediate `~/.deepagents` directory no longer lands at the umask.

**What this does not do, and cannot.** MCP credentials are **not** protected from the agent's shell — tokens are cleartext under `~/.deepagents/mcp-tokens/` and the default `LocalShellBackend` reaches any absolute path. The placement check warns; it cannot enforce, and it cannot be promoted to an error later:

- `deepagents/middleware/filesystem.py:1741-1748` raises `NotImplementedError` when permissions are set on a backend where `supports_execution()` is true, unless every path routes to a non-exec backend.
- Every `_check_fs_permission` call site in that file passes `"read"` or `"write"` — **none passes `"execute"`** — so even the `CompositeBackend` escape hatch would not stop `cat`.

The real fix is architectural (OS keyring, or a directory the agent process cannot read) and is tracked separately. `update_mcp_server` provides change mediation, not confidentiality.

### DIFF
--- a/libs/talon/deepagents_talon/mcp.py
+++ b/libs/talon/deepagents_talon/mcp.py
@@ -43,10 +43,10 @@ from deepagents_talon.mcp_auth import (
     format_login_error,
     prepare_oauth_login,
 )
-from deepagents_talon.mcp_config import MCPConfigStore
+from deepagents_talon.mcp_config import MCPConfigStore, agent_workspace_root
 
 if TYPE_CHECKING:
-    from collections.abc import Awaitable, Callable, Mapping, Sequence
+    from collections.abc import Awaitable, Callable, Iterator, Mapping, Sequence
 
     from langchain_core.tools import BaseTool
     from langchain_mcp_adapters.client import Connection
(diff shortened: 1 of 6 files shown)


linted:

PR #6161, opened 2026-09-08
fix(talon): stop losing MCP refresh tokens on a refresh response

linted:
**Stack 3 of 4 — MCP OAuth.** Based on `linted/talon/mcp-2-config-store` (#6160). Review #6159 and #6160 first.

Token storage durability, plus a finding that was investigated and withdrawn.

- **`set_tokens` destroyed the stored refresh token when a response omitted one.** RFC 6749 §6 permits a refresh response to omit `refresh_token`, meaning "keep the one you have", and many servers do — so the long-lived credential was destroyed on the first such refresh, forcing a full interactive re-login and defeating the point of the persisted-refresh work in #6100 and #6090. Now merged inside the same locked read-modify-write that already preserved `client_info`.
- **A double `close()` on a file descriptor.** If the write raised, `os.fdopen`'s context manager had already closed the descriptor and the error path closed it again — with the `EBADF` hidden by `suppress(OSError)`. Because this runs under `asyncio.to_thread`, another thread could `open()` and receive that fd number in between, so the second close could silently close an unrelated file, socket or DB handle. Ownership now transfers explicitly.
- **Token writes were unsynced and unlocked**, so two concurrent refreshes silently dropped one side's tokens *and* `client_info`, and an unsynced rename could leave a partial credential file that the reader then rejected outright. Now uses the same `flock` + `fsync` discipline as the config store, extracted to one shared helper.

**Two commits, one of which reverts the other — deliberately kept.** A review finding claimed `_validate_metadata` failed to bind OAuth endpoints to the issuer origin. It was implemented, then investigated and reverted: the existing issuer check *is* RFC 8414 §3, correctly implemented in both discovery paths, because `issuer` derives from the same URL the SDK builds its candidates from. The binding closed nothing that check doesn't already close, while permanently rejecting conforming split-origin authorization servers — Google is one (`issuer: accounts.google.com`, `token_endpoint: oauth2.googleapis.com`, verified against the live discovery document). Both commits stay so the trace is answerable from `git log` rather than looking like churn.

Also noted, not fixed: the device path silently loses split-origin authorization servers — `_issuer_endpoint` enforces same-origin, the error is swallowed, and Talon falls back to auth-code with no diagnostic. Pre-existing; relaxing it is its own security decision.

### DIFF
--- a/libs/talon/deepagents_talon/mcp.py
+++ b/libs/talon/deepagents_talon/mcp.py
@@ -115,12 +115,6 @@ def _authentication_required(exc: BaseException) -> bool:
     )
 
 
-def _device_authorization_completed(exc: BaseException) -> bool:
-    return any(
-        isinstance(leaf, DeviceAuthorizationCompletedError) for leaf in _exception_leaves(exc)
-    )
-
-
 @dataclass(frozen=True, slots=True)
 class MCPToolInfo:
     """Metadata for one MCP tool."""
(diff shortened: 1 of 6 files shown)


linted:

Issue #5401, opened 2026-08-10
bug(examples): async subagent interrupt lets canceled runs overwrite thread state


abhinavkr26104:
﻿## Description

The `examples/async-subagent-server` example implements cancellation and `multitask_strategy="interrupt"` by updating the run row to `cancelled`, but it does not cancel the background `asyncio` task running `_execute_run`. When that old task eventually finishes, `_execute_run` unconditionally writes `status = 'success'` and appends its response to the thread. A canceled run can therefore overwrite the replacement run's state with stale output.

## Reproduction Steps / Example Code (Python)

1. Start `examples/async-subagent-server`.
2. Create a thread with `POST /threads`.
3. Start a run whose agent invocation takes several seconds.
4. Before it finishes, start a second run on the same thread with:

```json
{
  "input": {"messages": [{"role": "user", "content": "new task"}]},
  "multitask_strategy": "interrupt"
}
```

5. Wait for both background tasks to finish, then request `GET /threads/{thread_id}` and both run records.

## Expected behavior

The first run should remain `cancelled` and must not mutate the thread after the interrupt. The second run's result should be the only completion written to the thread.

## Actual behavior

The first run continues executing. Its completion overwrites the `cancelled` database status with `success` and appends the stale response after the replacement run has cleared/rebuilt the thread state.

## Code reference

- `examples/async-subagent-server/server.py:169-198` ? `_execute_run()` unconditionally writes success and thread messages after the agent returns.
- `examples/async-subagent-server/server.py:257-287` ? `multitask_strategy="interrupt"` only updates database rows; it does not cancel or invalidate the running task.
- `examples/async-subagent-server/server.py:287` ? runs are launched with fire-and-forget `asyncio.ensure_future()` and are not retained for cancellation or shutdown.

## Why it matters

This example is the reference implementation used by `AsyncSubAgentMiddleware`. Its cancellation and interrupt endpoints report a state that the server can later contradict, and stale results can corrupt the thread history. The example should retain task handles or use a generation/run-token check before committing results, and add a regression test that waits for the old task after an interrupt.

## Related context

`test_interrupt_strategy` currently checks the immediate `cancelled` response but does not wait for the original task to finish or assert that it cannot overwrite the replacement thread.

PR #6257, opened 2026-09-11
fix(sdk): keep final status when `cancel_async_task` hits a finished run

### LINKED ISSUE #6234: `cancel_async_task` records `cancelled` for a run that had already finished, and `list_async_tasks` then never re-fetches it
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

- #5243 — `list_async_tasks` filters by the stale cached status before fetching live statuses. Adjacent, but about the filter; this report is about `cancel_async_task` writing a status the server never had, which the terminal short-circuit then pins.
- #3008 — precise resume for `interrupted` tasks. Not about cancellation.
- #5401 — `examples/async-subagent-server`: a cancelled background task later overwrites the thread with `success`. The inverse direction; the example server's `cancel_run` is also relevant below because it flips any run to `cancelled` unconditionally.

I found no issue covering a cancel that lands after the run has already completed.

```python
# Run with: uv run --with deepagents==0.7.13 python repro.py
# Expected output on 0.7.13 is quoted in the Description below.

import dataclasses
from unittest.mock import patch

from langchain.tools import ToolRuntime
from deepagents.middleware.async_subagents import AsyncSubAgent, _build_async_subagent_tools, _tasks_reducer


class Runs:
    """Behaves like the LangGraph dev server (langgraph-runtime-inmem 0.34.0): the run finished
    before the cancel arrived, so `runs.cancel` is a silent no-op (HTTP 202) and `runs.get`
    keeps reporting `success`."""
    def __init__(self):
        self.get_calls = 0
    def cancel(self, *, thread_id, run_id):
        return None
    def get(self, *, thread_id, run_id):
        self.get_calls += 1
        return {"run_id": run_id, "thread_id": thread_id, "status": "success"}


class Threads:
    def get(self, *, thread_id):
        return {"thread_id": thread_id, "values": {"messages": [{"role": "ai", "content": "report finished"}]}}


class Client:
    def __init__(self):
        self.runs = Runs()
        self.threads = Threads()


def runtime(state, tool_call_id):
    kwargs = {}
    for f in dataclasses.fields(ToolRuntime):
        kwargs[f.name] = {"state": state, "tool_call_id": tool_call_id, "config": {}}.get(f.name)
    if "stream_writer" in kwargs:
        kwargs["stream_writer"] = lambda *_: None
    return ToolRuntime(**kwargs)


spec = AsyncSubAgent(name="worker", description="d", graph_id="worker", url="http://localhost:8123")
client = Client()
with patch("deepagents.middleware.async_subagents.get_sync_client", return_value=client):
    tools = {t.name: t for t in _build_async_subagent_tools([spec])}
    state = {"async_tasks": {"t1": {
        "task_id": "t1", "agent_name": "worker", "thread_id": "t1", "run_id": "r1",
        "status": "running", "created_at": "2026-09-10T00:00:00Z",
        "last_checked_at": "2026-09-10T00:00:00Z", "last_updated_at": "2026-09-10T00:00:00Z"}}}

    cmd = tools["cancel_async_task"].func(task_id="t1", runtime=runtime(state, "tc1"))
    print("cancel tool said:", cmd.update["messages"][0].content)
    state["async_tasks"] = _tasks_reducer(state["async_tasks"], cmd.update["async_tasks"])
    print("tracked status after cancel:", state["async_tasks"]["t1"]["status"])

    before = client.runs.get_calls
    listed = tools["list_async_tasks"].func(status_filter="all", runtime=runtime(state, "tc2"))
    text = listed if isinstance(listed, str) else listed.update["messages"][0].content
    print("list_async_tasks shows:", text.strip().splitlines()[-1])
    print("list re-fetched the run:", client.runs.get_calls > before)

    cmd = tools["check_async_task"].func(task_id="t1", runtime=runtime(state, "tc3"))
    print("check_async_task says:", cmd.update["messages"][0].content)
```

```shell

```

### Description

Output of the reproduction on 0.7.13:

```
cancel tool said: Cancelled async subagent task: t1
tracked status after cancel: cancelled
list_async_tasks shows: - task_id: t1  agent: worker  status: cancelled
list re-fetched the run: False
check_async_task says: {"status": "success", "thread_id": "t1", "result": "report finished"}
```

The fake client models what the LangGraph dev server does when a cancel arrives after the run has finished. In `langgraph-runtime-inmem` 0.34.0, `Runs.cancel` selects candidate runs by `(thread_id, run_id)` regardless of status, acts only on `pending` / `running` runs, skips the rest without raising, and the `/threads/{thread_id}/runs/{run_id}/cancel` route returns 202. So `client.runs.cancel(...)` returns normally for a run that is already `success` or `error`. The repository's own `examples/async-subagent-server/server.py` goes further: its `cancel_run` executes `UPDATE runs SET status = 'cancelled' WHERE run_id = ?` unconditionally, so there the server-side `success` is overwritten as well.

* I'm using `AsyncSubAgentMiddleware` and the agent calls `cancel_async_task` on a task whose remote run has already finished (`success` or `error`) by the time the cancel request lands.
* I expect the tracked status to reflect the run's real state, and the tool to say the task had already finished rather than that it was cancelled.
* Instead, `cancel_async_task` (sync and async) never looks at the run's state: it issues `runs.cancel`, and if that call does not raise it writes `status="cancelled"` into `async_tasks` and returns "Cancelled async subagent task: …". Three things follow:
  1. The tracker records `cancelled` for a task whose result exists on the server, and the model reports the task as cancelled to the user.
  2. `cancelled` is in `_TERMINAL_STATUSES`, so `_fetch_live_status` short-circuits and `list_async_tasks` never re-fetches it. The wrong status is permanent from the tracker's point of view; only an explicit `check_async_task` on a task the model believes is cancelled reveals the `success` result, and the tool description does not lead the model there.
  3. `list_async_tasks` and `check_async_task` disagree about the same task, as the output above shows.

A small fix seems sufficient: after `runs.cancel`, read the run back with `runs.get` and store its actual status, reporting "already finished with status X" when it is terminal, plus regression tests for cancel-after-success (result preserved, `list_async_tasks` shows `success`), cancel-after-error, and the unchanged running-task path. One related detail: the dev server marks a cancelled run `interrupted`, while the middleware writes `cancelled`, so the two vocabularies already diverge for genuine cancellations.

I am happy to open a PR with the fix and the regression tests once the issue is assigned.

### Environment / System Info

OS: macOS 26.6.2 (arm64)
Python: 3.12.13
deepagents: 0.7.13
langchain: 1.4.0
langgraph: 1.2.11
langgraph-sdk: 0.4.4

Totoro-qaq:
Fixes #6234

`cancel_async_task` no longer reports a cancellation, or records `cancelled`, for an async subagent run that had already finished; the task keeps its real `success`, `error`, or `timeout` status.

---

When a cancel reaches a run that has already finished, the LangGraph server treats it as a no-op: the dev server returns 202 and the run stays `success`. The cancel tool assumed the cancel took effect and stored `cancelled`. Because `cancelled` is terminal, `list_async_tasks` skips the live fetch for it, so the task stayed `cancelled` and the model never learned the run had finished.

After `runs.cancel`, the tool now reads the run back with `runs.get`. If the run ended on its own (`success`, `error`, `timeout`), that status is kept and the tool message says nothing was cancelled and points to `check_async_task`. Any other status, including `interrupted` (what the dev server records for a genuine cancel), is stored as `cancelled` as before, so the middleware's status vocabulary is unchanged. If the read-back fails, the tool falls back to `cancelled` and logs a warning. The sync and async tools share one `_build_cancel_command` helper.

Not addressed here: a run that is still `pending` or `running` when its status is read back can finish later while the cached task remains `cancelled`. This PR keeps cancellation non-blocking (`wait=False`) and does not wait for the run to settle. The example server's unconditional overwrite in `cancel_run` is related to #5401 and left out of this PR.

Tests cover cancel after `success`, `error`, and `timeout`; cancel of `pending`, `running`, and `interrupted` runs; a failed read-back; and the async path.

### DIFF
--- a/libs/deepagents/deepagents/middleware/async_subagents.py
+++ b/libs/deepagents/deepagents/middleware/async_subagents.py
@@ -577,6 +577,69 @@ def _build_update_tool(
     )
 
 
+_FINISHED_BEFORE_CANCEL = frozenset({"success", "error", "timeout"})
+"""Run statuses showing the run ended on its own, so a cancel request had nothing to stop."""
+
+
+def _read_status_after_cancel(client: SyncLangGraphClient, task: AsyncTask) -> str | None:
+    """Read the run's status back after a cancel request, or `None` if the read fails."""
+    try:
+        run = client.runs.get(thread_id=task["thread_id"], run_id=task["run_id"])
+    except Exception:  # noqa: BLE001  # LangGraph SDK raises untyped errors
+        logger.warning("Failed to read run status after cancelling task %s", task["task_id"], exc_info=True)
+        return None
+    return run["status"]
+
+
+async def _aread_status_after_cancel(client: LangGraphClient, task: AsyncTask) -> str | None:
+    """Async version of `_read_status_after_cancel`."""
+    try:
+        run = await client.runs.get(thread_id=task["thread_id"], run_id=task["run_id"])
+    except Exception:  # noqa: BLE001  # LangGraph SDK raises untyped errors
+        logger.warning("Failed to read run status after cancelling task %s", task["task_id"], exc_info=True)
+        return None
+    return run["status"]
+
+
+def _build_cancel_command(
+    task: AsyncTask,
+    run_status: str | None,
+    tool_call_id: str | None,
+) -> Command:
+    """Build the `Command` update after a cancel request.
+
+    A run that finished before the cancel reached it keeps its real status, so
+    its outcome stays reachable through `check_async_task`. Any other status,
+    or a failed read (`run_status` is `None`), is recorded as `'cancelled'`.
+    """
+    if run_status is not None and run_status in _FINISHED_BEFORE_CANCEL:
+        status = run_status
+        msg = (
+            f"Async subagent task {task['task_id']} had already finished with status '{run_status}' "
+            "before the cancel request, so nothing was cancelled. Use check_async_task to see its outcome."
+        )
+    else:
+        status = "cancelled"
+        msg = f"Cancelled async subagent task: {task['task_id']}"
+    now = datetime.now(UTC).strftime("%Y-%m-%dT%H:%M:%SZ")
+    updated = AsyncTask(
+        task_id=task["task_id"],
+        agent_name=task["agent_name"],
+        thread_id=task["thread_id"],
+        run_id=task["run_id"],
+        status=status,
+        created_at=task["created_at"],
+        last_checked_at=now,
+        last_updated_at=now if status != task["status"] else task["last_updated_at"],
+    )
+    return Command(
+        update={
+            "messages": [ToolMessage(msg, tool_call_id=tool_call_id)],
+            "async_tasks": {task["task_id"]: updated},
+        }
+    )
+
+
 def _build_cancel_tool(
     clients: _ClientCache,
 ) -> StructuredTool:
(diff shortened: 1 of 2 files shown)


Totoro-qaq:
Hi @colifran, this PR implements the fix for #6234 and is ready for review whenever you have a moment.

colifran:
Hey @Totoro-qaq! Thanks for putting this up. Will work on reviewing!

PR #6258, opened 2026-09-11
fix(sdk): report UTF-8 byte sizes

### LINKED ISSUE #6042: `StateBackend` and `StoreBackend` report character counts instead of UTF-8 byte sizes
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python

store = InMemoryStore()
backend = StoreBackend(
    store=store,
    namespace=lambda _runtime: ("filesystem",),
)

content = "hello 😀 €"
backend.write("/unicode.txt", content)

expected_size = len(content.encode("utf-8"))
print("Expected size:", expected_size)

ls_result = backend.ls("/")
print("ls() result:", ls_result.entries)

glob_result = backend.glob("*.txt", path="/")
print("glob() result:", glob_result.matches)

assert ls_result.entries[0]["size"] == expected_size
assert glob_result.matches[0]["size"] == expected_size
```

```shell
Expected size: 14
ls() result: [{'path': '/unicode.txt', 'is_dir': False, 'size': 9, ...}]
glob() result: [{'path': '/unicode.txt', 'is_dir': False, 'size': 9, ...}]
```

### Description

`StateBackend` and `StoreBackend` report incorrect file sizes for files containing non-ASCII UTF-8 text. `FileInfo.size` is expected to represent the file size in bytes, but `ls()` and `glob()` currently count Unicode characters instead. For example, `"hello 😀 €"` should report 14 bytes, but currently reports 9 characters.

### Environment / System Info

_No response_

eknathmali:
Fixes #6042

## Summary

Fix `FileInfo.size` reporting character counts instead of UTF-8 byte
sizes in `StateBackend` and `StoreBackend`.

## Changes

- Report file sizes in UTF-8 bytes.
- Add regression tests for non-ASCII content.
- Cover `ls()` and `glob()` behavior.

### DIFF
--- a/libs/deepagents/deepagents/backends/state.py
+++ b/libs/deepagents/deepagents/backends/state.py
@@ -156,7 +156,7 @@ class StateBackend(BackendProtocol):
                 continue
 
             # This is a file directly in the current directory
-            size = len(file_data_to_string(fd))
+            size = len(file_data_to_string(fd).encode("utf-8"))
             infos.append(
                 {
                     "path": k,
@@ -308,7 +308,7 @@ class StateBackend(BackendProtocol):
         infos: list[FileInfo] = []
         for p in paths:
             fd = files.get(p)
-            size = len(file_data_to_string(fd)) if fd else 0
+            size = len(file_data_to_string(fd).encode("utf-8")) if fd else 0
             infos.append(
                 {
                     "path": p,
--- a/libs/deepagents/deepagents/backends/store.py
+++ b/libs/deepagents/deepagents/backends/store.py
@@ -347,7 +347,7 @@ class StoreBackend(BackendProtocol):
                 fd = self._convert_store_item_to_file_data(item)
             except ValueError:
                 continue
-            size = len(file_data_to_string(fd))
+            size = len(file_data_to_string(fd).encode("utf-8"))
             infos.append(
                 {
                     "path": item.key,
@@ -641,7 +641,7 @@ class StoreBackend(BackendProtocol):
         infos: list[FileInfo] = []
         for p in paths:
             fd = files.get(p)
-            size = len(file_data_to_string(fd)) if fd else 0
+            size = len(file_data_to_string(fd).encode("utf-8")) if fd else 0
             infos.append(
                 {
                     "path": p,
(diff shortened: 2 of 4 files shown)


ErenAta16 (commented):
Checked this one out and ran it locally on Windows with CPython 3.12, since the repo's only Windows CI leg is 3.13.

**Before and after, measured.** I put identical content through `StateBackend` and `FilesystemBackend` side by side. On `main`, `"hello 😀 €"` comes back as 9 from `StateBackend.ls` and `glob`, while `FilesystemBackend` reports 14 for the same file on disk. On this branch `StateBackend` reports 14 and the two agree. `tests/unit_tests/backends/test_state_backend.py` and `test_store_backend.py` pass on the branch, 81 tests including the two added here.

**No fifth site left behind.** I grepped every `size` assignment under `deepagents/backends/`. `filesystem.py` takes `int(st.st_size)` in four places, `sandbox.py` already measures its edit payloads with `.encode("utf-8")`, `composite.py:380` is a literal `0`, and the only `len(file_data_to_string(...))` occurrences are the four this PR changes. So the character-count form is gone from the package after this, not only from the two backends named in the title.

**One case this leaves as it was, raised as a question rather than a change request.** `file_data_to_string` returns the base64 text for entries written with `encoding="base64"`, so `size` there measures the encoded string rather than the payload it stands for. A 1024-byte binary comes back as 1368 from `StateBackend`, on `main` and on this branch alike, against 1024 from `FilesystemBackend` for the same bytes. That is not a regression, since base64 is ASCII and the character count already equalled the byte count, and #6042 is scoped to text, so nothing needs to change for this PR to be correct. It is also arguably within `FileInfo.size` being documented as "File size in bytes (approximate)", and nothing inside the package reads the field, so the only consumers are external. I am naming it mainly so that leaving it stays a decision rather than an oversight, since the PR title reads as a general claim about `FileInfo.size` that a binary entry does not yet satisfy.

Environment for the above: Windows 11, CPython 3.12.10, branch at 17dc80c compared against main at 9e7d62f.

PR #6288, opened 2026-09-13
feat(sdk): reload skills when `skills_metadata` is reset to `None`

richardscarrott:
Related: #5416

`SkillsMiddleware` can now reload skills mid-thread: set `skills_metadata` to `None` through `invoke()` input, `update_state()`, or middleware, and the next run picks up added, edited, or deleted skills.

---

Skills load once per thread and are cached in state, so skills that changed mid-thread were never picked up, and a stored `None` pinned "No skills available" for the rest of the thread. `before_agent` now reloads every source when the stored value is `None`, while `[]` still means "loaded, no skills", so threads that are never reset make no extra backend calls. Every load also writes `skills_load_errors`, as `[]` when sources load cleanly, so warnings from an earlier load clear after a successful reload.

`skills_metadata` moves from `PrivateStateAttr` to `OmitFromOutput`, so it is accepted as run input but still left out of `invoke()` results. Subagents were previously kept away from it only by the private marker, so `_EXCLUDED_STATE_KEYS` now lists it by name, as in JS: non-forked subagents load their own skills, and a compiled subagent cannot overwrite the parent's list. Declarative forks still inherit it.

Docs: langchain-ai/docs#6043


libs/code/deepagents_code/plugins/adapters/skills_middleware.py
```diff
@@ -273,13 +273,15 @@ def _state_update(
...
-            update["skills_load_errors"] = errors
-        return update
+        # Always write the errors so warnings from an earlier load are cleared
+        return sdk_skills.SkillsStateUpdate(
+            skills_metadata=list(all_skills.values()),
+            skills_load_errors=errors,
+        )
 
-    def before_agent(
+    def before_model(
```
richardscarrott: Moving from before_agent -> before_model possibly makes this a breaking change in case people are subclassing it?

Also, there is a migration consideration here as if a thread is paused with SkillsMiddleware.before_agent as the next node, then I think they'll become bricked. I can't imagine it being a likely scenario though?

hntrl: for backwards compat, I think we'll need to add a passthrough before_agent hook as well. Can we do that and leave a big comment to remind us to remove in 0.8?

libs/code/pyproject.toml
```diff
@@ -27,7 +27,7 @@ classifiers = [
 ]
 dependencies = [
     # Framework
-    "deepagents==0.7.14",
+    "deepagents==0.8.0",
```
hntrl: ?

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -930,12 +948,13 @@ def modify_request(self, request: ModelRequest[ContextT]) -> ModelRequest[Contex
 
         return request.override(system_message=new_system_message)
 
-    def before_agent(self, state: SkillsState, runtime: Runtime, config: RunnableConfig) -> SkillsStateUpdate | None:  # ty: ignore[invalid-method-override]  # noqa: ARG002
-        """Load skills metadata before agent execution (synchronous).
+    def before_model(self, state: SkillsState, runtime: Runtime, config: RunnableConfig) -> SkillsStateUpdate | None:  # ty: ignore[invalid-method-override]  # noqa: ARG002
```
hntrl: for backwards compat, I think we'll need to add a passthrough before_agent hook as well. Can we do that and leave a big comment to remind us to remove in 0.8?

libs/code/tests/unit_tests/skills/test_merge.py
```diff
@@ -12,9 +12,15 @@
...
 
+    from deepagents.middleware import skills as sdk_skills
+    from langgraph.runtime import Runtime
+
 
 _MERGE_LOGGER = "deepagents_code.skills.merge"
 
+# The middleware entry points ignore `runtime`; only `state` drives the guard.
+_RUNTIME = cast("Runtime", None)
+
```
sydney-runkle, line 23: what is this

libs/code/pyproject.toml
sydney-runkle: idt this is worth minor

libs/deepagents/deepagents/middleware/skills.py
```diff
@@ -304,10 +308,10 @@ class SkillsStateUpdate(TypedDict):
     """State update for the skills middleware."""
 
     skills_metadata: list[SkillMetadata]
-    """List of loaded skill metadata to merge into state."""
+    """List of loaded skill metadata to write into state, replacing any previous list."""
 
-    skills_load_errors: NotRequired[list[str]]
-    """Skill source loading errors to merge into state."""
+    skills_load_errors: list[str]
+    """Skill source loading errors to write into state, empty when every source loaded."""
```
sydney-runkle: is this change necc

libs/deepagents/deepagents/middleware/skills.py
richardscarrott: As discussed, I've rolled back the move from before_agent -> before_model so this is now a patch change.

If implementors want mid-turn reloading, they can do this as an escape hatch:
```python
class ReloadingSkillsMiddleware(SkillsMiddleware):
    def before_model(self, state: SkillsState, runtime: Runtime, config: RunnableConfig) -> SkillsStateUpdate | None:
        return super().before_agent(state, runtime, config)
```

Once we're ready to make a breaking change, we can move before_agent -> before_model.

libs/deepagents/deepagents/middleware/skills.py
richardscarrott: I think claude felt the need to rewrite because the JS SDK went from a merge reducer to last value reducer — I'll revert.

libs/code/tests/unit_tests/skills/test_merge.py
richardscarrott: it's a stand-in for the before_agent runtime arg — impl. doesn't need it currently, hence None / just keeping ty happy.

```python
result = middleware.before_agent(
  self._state(skills_metadata=None), _RUNTIME, {}
)
```
I can inline the cast, or swap to Mock() + # ty: ignore if preferred?

libs/code/pyproject.toml
richardscarrott: reverted to patch

libs/code/tests/unit_tests/skills/test_merge.py
sydney-runkle: this is fine for tests i guess

hntrl (approved):

richardscarrott:
Superseded by #6364.

Closing in favour of the same branch pushed to `langchain-ai/deepagents` directly. This PR was opened from a fork, and fork PRs in this repo never receive CodeQL results, so the merge box was stuck on `Code scanning is waiting for results from CodeQL` with no way to clear it — `main`'s code scanning rule can't be satisfied from a fork.

#6364 is the identical branch at the same commit (`5fa3cf2`), so nothing has changed in the diff. @hntrl, sorry for the re-review ask — a PR's head repository can't be changed after the fact.

PR #6299, opened 2026-09-14
fix(sdk): reject unknown keys in the task tool arguments

### LINKED ISSUE #6286: TaskToolSchema ignores unknown keys, so task dispatches a subagent without the model's instructions
### Submission checklist

### Area (Required)

- [x] deepagents (SDK)
- [ ] dcode
- [ ] talon
- [ ] acp
- [ ] evals
- [ ] harbor
- [ ] daytona
- [ ] modal
- [ ] quickjs
- [ ] runloop
- [ ] vercel
- [ ] langsmith-sandbox

### Related Issues / PRs

_No response_

```python
# deepagents 0.7.13, Python 3.11
from deepagents.middleware.subagents import TaskToolSchema

print(
    TaskToolSchema(
        description="short label",
        subagent_type="worker",
        prompt="the real instructions, several hundred words",
    ).model_dump()
)

# printed: {'description': 'short label', 'subagent_type': 'worker'}
# the `prompt` key is gone, and a `task` call shaped like this dispatches the
# subagent with the label only
```

```shell

```

### Description

If a model calls `task` with an argument that is not in the schema, the argument is dropped without a word and the call still succeeds. `TaskToolSchema` is a plain `BaseModel`, so pydantic ignores unknown keys, and the tool binds that schema with `infer_schema=False`, so there is no second place the value could surface.

For me this is not hypothetical: Sonnet 5 keeps doing it. It leaves a short label in `description` and writes the actual instructions into an extra key, so the subagent is dispatched with the label alone.

It is not occasional. In the run where I found this, every dispatch looked like that. The subagent then comes back asking for the details that were sitting in the dropped key, so the dispatch is wasted, and the tool call is still marked as successful, which is what made it hard to spot.

What I would do: have `TaskToolSchema` reject keys it does not define, so a call like the one above comes back to the model as a tool error instead of a silent success. The model then sees which key was refused and calls `task` again with the instructions in `description`, and nothing dies in the middle of a run.

### Environment / System Info

Python 3.11
deepagents 0.7.13
Windows
model: Claude Sonnet 5

burakeyler:
## Summary

`TaskToolSchema` silently ignored keys it doesn't define. A model that puts the actual instructions under an invented key (e.g. `prompt`) instead of `description` still dispatches the subagent — with only the short `description` label as its prompt — and the tool reports success. The model never sees anything wrong.

This adds a `mode="before"` validator on `TaskToolSchema` that rejects any key outside the schema's fields (and the `runtime` parameter the tool node injects after validation). The rejection surfaces to the model as a tool error naming the offending key, instead of a silent partial dispatch.

`extra="forbid"` isn't usable directly: the tool node validates the arguments *after* injecting `runtime`, which isn't a schema field, so a plain `forbid` would break every `task` call.

Fixes #6286

## Test plan
- [x] `pytest tests/unit_tests/test_subagents.py -k unknown_key` — the two new tests (schema-level rejection + end-to-end `task` call with an extra `prompt` key) pass
- [x] `pytest tests/unit_tests/test_subagents.py` — 36 passed, 1 xfailed, 1 failed + 2 errors; verified those 3 also fail/error identically on `main` before this change (Windows-only `asyncio`/`pytest-socket` interaction, unrelated to this fix)
- [x] `ruff check` / `ruff format --diff` on the changed files — clean
- [x] `ty check deepagents` — clean (only a pre-existing, unrelated `_version.py` deprecation warning remains)


burakeyler:
Rewrote the branch history to drop the `Co-Authored-By: ... <noreply@anthropic.com>` trailers from both commits — the "validate PR has no banned trailers" check was failing against the org's `block-anthropic-coauthor` ruleset. No code changes; same diff as before, just cleaned commit messages.

Re-ran locally after the rewrite to confirm nothing regressed:
- `pytest tests/unit_tests/test_subagents.py` — 39 passed, 1 xfailed (no failures this time; the 3 Windows-only `asyncio`/`pytest-socket` failures noted in the PR description no longer reproduce)
- `ruff check` / `ruff format --diff` on the changed files — clean
- `ty check deepagents` — clean (only the pre-existing, unrelated `_version.py` deprecation warning)

The "PR trailer lint" check is green now.

burakeyler:
Merged \`upstream/main\` to resolve the merge conflict blocking this PR (branch was 59 commits behind). The only conflict was in \`test_subagents.py\`'s import block, where main had independently added \`SkillMetadata\` to the skills import — kept both that and this PR's \`TaskToolSchema\` import since both are used later in the file.

Verified after merging:
- \`pytest tests/unit_tests/test_subagents.py\` — 41 passed, 1 xfailed
- \`ruff check\` / \`ruff format --diff\` on the two changed files — clean

Pushed as a regular merge commit (fast-forward), no rebase/force-push, so the existing review history stays intact.

PR #6334, opened 2026-09-16
fix(sdk): allow colons in profile model keys

### LINKED ISSUE #4640: Improve profile lookup for `provider:model` variant specs
Deep Agents profile registries currently validate keys as either `provider` or `provider:model`, and harness profile lookup rejects specs containing more than one colon. That makes recommended provider specs with a variant suffix, such as `ollama:glm-5.2:cloud`, impossible to cover with a model-specific harness profile.

Context:

- `register_harness_profile("ollama:glm-5.2:cloud", ...)` fails validation because the key has more than one `:`.
- `_get_harness_profile("ollama:glm-5.2:cloud")` currently returns `None` before consulting the registry for the same reason.
- This means `deepagents-code` cannot apply the GLM 5.2 media-file safety prompt to the recommended Ollama Cloud selection without changing core parsing/lookup semantics.

We should decide how Deep Agents should parse provider model specs that include provider-specific suffixes or routing variants. One possible behavior is to keep registration keys canonical as `provider:model`, while allowing lookup for `provider:model:variant` to resolve a registered `provider:model` profile. That should be designed carefully so malformed specs and typos do not accidentally fall back to provider-wide profiles.

mdrxy:
Profile registrations and lookups now support model identifiers containing colons, including Ollama tags and Bedrock ARNs with empty account fields.

> [!WARNING]
> It is not possible for model specs containing multiple colons to receive registered provider defaults, including `init_kwargs` and `pre_init`, where they previously received no profile (if present).

---

Fixes #4640

Profile keys use the first colon to separate the provider from the complete model identifier. Validation requires nonempty provider/model halves with no surrounding whitespace and leaves internal model syntax to the provider. This allows `::` in accountless foundation-model ARNs; empty keys, `:model`, `openai:`, and whitespace around the first colon still raise `ValueError`.

Provider and harness lookups preserve the full identifier and merge exact model overrides with provider defaults. For pre-built models, harness resolution tries both exact candidates before provider defaults: the combined provider/identifier key, then the identifier itself when it contains a colon. A known provider remains authoritative; the identifier prefix supplies a provider fallback only when the provider cannot be inspected.


libs/deepagents/deepagents/profiles/harness/harness_profiles.py
```diff
@@ -1013,16 +1013,19 @@ def register_harness_profile(key: str, profile: HarnessProfile | HarnessProfileC
...
             - `"openai:gpt-5.4"` — specific model
+            - `"ollama:glm-5.2:cloud"` — model identifier containing a colon
         profile: The runtime harness profile or declarative config to register.
 
     Raises:
-        ValueError: If `key` is empty, contains more than one `:`, or has an
-            empty provider/model half.
+        ValueError: If `key` is malformed. See `validate_profile_key` for the
+            exact conditions: empty key, leading/trailing whitespace,
+            whitespace adjacent to the first `:`, or an empty provider/model half.
```
mdrxy: i think we can leave it at `See `validate_profile_key` for the exact conditions`

libs/deepagents/deepagents/profiles/harness/harness_profiles.py
```diff
@@ -1066,19 +1069,21 @@ def _get_harness_profile(spec: str) -> HarnessProfile | None:
...
 
-    Malformed specs (empty string, more than one `:`, or a `:` with an empty
-    provider/model half) return `None` without consulting the registry. This
-    prevents a spec like `"openai:"` from silently matching the provider-wide
-    `"openai"` registration.
+    The first colon separates the provider from the complete model identifier.
+    Providers must not contain colons, while model identifiers may. Malformed
+    specs (empty string or a `:` with an empty provider/model component) return
+    `None` without consulting the registry. This prevents a spec like
+    `"openai:"` from silently matching the provider-wide `"openai"` registration.
```
mdrxy: i feel like we jump to this kinda outta nowhere. it also feels like info most people won't need to know - if people want to see conditions of validation is this really the right place for the info to live? i'd err on side of removal probably or heavy reduction

libs/deepagents/deepagents/profiles/harness/harness_profiles.py
```diff
@@ -1252,24 +1257,43 @@ def _merge_profiles(base: HarnessProfile, override: HarnessProfile) -> HarnessPr
...
-    `provider:model` shape) and a provider-only fallback.
-
-    A *bare* identifier (no `:`) is deliberately not consulted against the
-    registry. If it were, a pre-built model whose `model_name` happened to
-    coincide with a registered provider key (e.g. an in-house proxy whose
-    identifier is `"openai"`) would silently pick up that provider's profile.
-    Registering under a bare key is supported via the `spec` path, not
-    inferred from a model's identifier.
+    Use `spec` directly when provided. Otherwise resolve the model's provider
+    and identifier, preserving any colons in the identifier, and try:
```
mdrxy:
```suggestion
    and identifier, and try:
```

libs/deepagents/deepagents/profiles/harness/harness_profiles.py
```diff
@@ -1252,24 +1257,43 @@ def _merge_profiles(base: HarnessProfile, override: HarnessProfile) -> HarnessPr
...
+
+    1. The combined `provider:identifier` key, as an exact match.
+    2. The identifier alone, as an exact match, when it contains a colon.
```
mdrxy:
```suggestion
    2. The model identifier alone, as an exact match, when it contains a colon.
```

mdrxy: i don't understand is it only when it contains a colon?

libs/deepagents/deepagents/profiles/harness/harness_profiles.py
```diff
@@ -1252,24 +1257,43 @@ def _merge_profiles(base: HarnessProfile, override: HarnessProfile) -> HarnessPr
...
+    3. Provider-wide defaults.
+
+    Exact matches inherit unset fields from the matched key's provider prefix
+    via `_get_harness_profile`. For compatibility, step 2 can therefore inherit
+    defaults from a prefix different from the model's reported provider.
+    Step 3 uses the reported provider, falling back to the identifier's prefix
+    only when the provider is unknown.
+
+    Bare identifiers are never looked up alone: a model named `"openai"` must
+    not accidentally match that provider's defaults.
```
mdrxy: ???

libs/deepagents/deepagents/profiles/provider/provider_profiles.py
```diff
@@ -231,17 +231,20 @@ def register_provider_profile(key: str, profile: ProviderProfile) -> None:
...
+            - `"ollama:glm-5.2:cloud"` — model identifier containing a colon
 
         profile: The provider profile to register.
 
     Raises:
-        ValueError: If `key` is empty, contains more than one `:`, or has an
-            empty provider/model half.
+        ValueError: If `key` is malformed. See `validate_profile_key` for the
+            exact conditions: empty key, leading/trailing whitespace,
+            whitespace adjacent to the first `:`, or an empty provider/model half.
```
mdrxy: maybe same? https://github.com/langchain-ai/deepagents/pull/6334/changes#r4030222951

libs/deepagents/deepagents/profiles/provider/provider_profiles.py
```diff
@@ -345,11 +355,13 @@ def apply_provider_profile(
     silently replaced.
 
     When no profile is registered for `spec`, returns a copy of `kwargs`
-    unchanged. This keeps the helper safe to call unconditionally.
+    unchanged and logs the miss at `DEBUG`. Profiles are optional, so this
+    helper is safe to call unconditionally.
```
mdrxy:
```suggestion
    unchanged and logs the miss at `DEBUG`.
```

PR #6342, opened 2026-09-16
fix(code): hydrate resumed history on demand

mdrxy:
Long dcode threads now resume without eagerly rendering hundreds of off-screen messages.

---

A 500-message profile showed that the initial 30-message tail rendered quickly, while background warming to the 800-message window spent tens of seconds in Textual style and layout recomputation.

This keeps the bounded initial tail and hydrates archived messages only when the user scrolls toward them.

| 500-message resume | Before | After |
| --- | ---: | ---: |
| Initial tail | ~315 ms | ~320 ms |
| Settled mounted rows | 500 | 30 |
| Background CPU work | ~57 s | none |

Made by [Open SWE](https://github.com/langchain-ai/open-swe) · [view thread](https://openswe.vercel.app/agents/a346ff9c-f31c-5f0e-b321-02c2364824da) · openai:gpt-5.6-sol (medium)


mdrxy:
Can't tell if this is actually helping

PR #6432, opened 2026-09-18
fix(code): deduplicate mixed-ID model usage

mdrxy:
The teardown Usage Stats table no longer counts one streamed subagent model invocation twice when provider and LangChain message IDs differ.

---

OpenAI Responses streaming can expose an `lc_run--…` ID on the usage-bearing chunk while the aggregated completion retains its provider `resp_…` ID. The client ledger treated those as separate requests.

- Carry the LangChain model run ID on `model_usage` completion events.
- Correlate streamed and completed usage by invocation/attempt identity while preserving legacy message-ID fallback.
- Keep completion corrections authoritative for tokens, cache details, model attribution, and cost.
- Preserve finalized message-ID aliases for interrupt/resume replay and separate genuine retries.
- Cover mixed-ID delivery through a real parent/child graph and both client accounting surfaces.

Made by [Open SWE](https://github.com/langchain-ai/open-swe) · [view thread](https://openswe.vercel.app/agents/f51cea0d-bc67-5928-93f9-2d9f50fa8082) · openai:gpt-5.6-sol (medium)


mdrxy:
One limitation to keep track of: this fix relies on LangGraph's private `langgraph.pregel._messages.StreamMessagesHandler` and its per-run `metadata` mapping. `_set_stream_invocation_id()` adds the model callback's run ID to a copy of that metadata so streamed messages and the final usage callback can be matched even when their message IDs differ. This also depends on the stream handler's start callback populating its metadata before our recorder runs.

The regression test exercises real LangGraph callbacks and checks that every chunk, including the first, carries the same invocation ID as the completed usage record. Existing coverage also checks that batched calls keep separate identities. These tests should flag upstream changes that break the bridge, but the private API dependency remains.

We should revisit this when upstream exposes the model invocation/run ID through a supported message-stream API, or changes stream callback/metadata handling. Ideally, `stream_mode="messages"` would carry the model callback's `run_id` consistently from the first chunk through completion, letting us remove the private metadata mutation. I did not find a direct upstream proposal for that exact capability; related work includes:

- https://github.com/langchain-ai/langgraph/pull/7047 — open PR preserving node metadata for custom events in `astream_events(v2)`; related metadata propagation work, but not a replacement for this bridge.
- https://github.com/langchain-ai/langgraph/issues/6394 — closed issue about usage callbacks interfering with message streaming; a different failure in the same integration area.
- https://github.com/langchain-ai/langgraph/pull/8819 — draft proposal for keyed child-task identity and persistent subgraph state. Related discussion: https://github.com/langchain-ai/deepagents/pull/6433#issuecomment-5769903148. This addresses child-task/checkpoint identity rather than model-call identity, so it would not directly replace this bridge.

PR #6433, opened 2026-09-18
fix(code): preserve JavaScript subagent costs across resume

mdrxy:
Session costs retain JavaScript subagent spending across interruptions, resumes, failures, and nested cancellation without replaying cached JavaScript task results.

---

- Keep the accounting bridge local to `deepagents-code`: only owned cost transfers reach parent state; child messages and unrelated state remain isolated.
- Persist each locally priced accounting-node delta in the existing thread checkpointer before returning its graph update. Descendants share the outer eval's accounting owner; transferred totals are excluded from receipts to avoid double-counting.
- Resume child graphs under request-fingerprint and identical-request occurrence namespaces, rather than positional `langgraph.func.task` results. Parallel completion order cannot exchange results for different requests. Direct task calls and structured output stay intact.
- Settle cancellation and receipt writes before collecting spend. Child graphs use synchronous checkpoint durability even when the parent uses exit durability, because successful nested graphs otherwise leave no resumable checkpoint.

This covers the async interpreter with a checkpointer, including reopened SQLite checkpoints. Receipts use the existing saver and thread retention, with one additional checkpoint per priced node and an eval-scoped metadata query. Arbitrary JavaScript side effects are not exactly-once; identical requests remain occurrence-ordered, and nondeterministic scripts are not made replay-safe. The adapter relies on LangGraph's private namespace/scratchpad configuration. It does not recover provider usage lost before pricing/persistence, and synchronous interpreter accounting is unchanged.

Made by [Open SWE](https://github.com/langchain-ai/open-swe) · [view thread](https://openswe.vercel.app/agents/ddfccb2f-f22b-5108-96b7-ce51b8bb57dc) · openai:gpt-6-astra (low)


mdrxy:
Related upstream work: https://github.com/langchain-ai/langgraph/pull/8819 (currently an open draft) proposes `Send(..., key=...)`, where task identity derives from an explicit key rather than packet position, and child graphs inherit a stable checkpoint namespace. Its tests cover interrupt/resume with reversed arrival order and keyed child persistence under `exit` durability. The PR description explicitly calls for LangChain/Deep Agents integration.

This looks worth coordinating with as a path to removing the namespace/scratchpad manipulation here. It is not a drop-in replacement: this JavaScript bridge directly invokes the `task` tool, whereas the proposed public API assigns keys through `Send`. We would need to route these dispatches through keyed `Send`s or have a supported API for keying direct child invocations. The key would also need to distinguish separate evals and identical-request occurrences.

The cost receipts and cancellation settlement would still be needed; stable child identity alone does not preserve spending when a child fails before returning. Separately, we could reduce private API use by passing the existing saver explicitly and setting public `durability="sync"` at the child graph invocation rather than through `__pregel_durability`.

